PR #41, opened 2016-10-26
Fix Patternfly imports

matobet:
index.js imports of patternfly js and css resources depended on non-existing
bootstrap dependency. This was fixed to align with the reference import
style given by ovirt-ui-components.


mareklibra:
Reverting.
With this change, the Patternfly JS is not loaded, so i.e. dropdown-menus are not working.

PR #43, opened 2016-10-27
Fix Patternfly imports

matobet:
index.js imports of patternfly js and css resources depended on non-existing
bootstrap dependency. This was fixed by adding an explicit npm dependency on bootstrap and aligned with the reference import style given by ovirt-ui-components.


matobet:
Verified that the tooltips are working

PR #45, opened 2016-11-01
To reduce the network usage, added Accept-Encoding: gzip

jelkosz:


jelkosz:
ok, this is automatically set by the browser, no need for this explicit setting. Closing

PR #62, opened 2016-12-01
Enable ovirt-web-ui for dependency of ovirt-engine

mareklibra:
The ovirt-web-ui is becomming a hard-dependency of ovirt-engine.

ovirt-engine.conf is dynamically updated
move ovirt-web-ui.war out from ovirt-engine.ear to same level as restapi.war or ovirt-engine.ear


mareklibra:
ovirt-engine: https://gerrit.ovirt.org/#/c/67625

PR #69, opened 2017-01-05
Implement Standalone/Development mode

matobet:
Enabled standalone run with the usage of webpack-dev-server.
Since the application now expects already authenticated user, `npm start` will
prompt the developer for credentials to the engine instance specified in the
ENGINE_URL environment variable.

The same flow is used in the accompanying docker image.

The injection of the `userData` is facilitated by extending the `env` object
(that is used to source the webpack.DefinePlugin) by the entry of
'window.userData' with the complete object containing token & username.

Note: window.userData is also used by the JSP templates to inject such data
from the engine so we maintain consistency and fetchToken() in index.js remains
the single authoritative source of obtaining pre-existing auth token on
application startup.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -2,17 +2,12 @@ import $ from 'jquery'
 
 import { setLogDebug } from 'ovirt-ui-components'
 
-// For Development
-// const CONFIG_URL = '/userportal.config'
-
-// TODO: configure path automatically
-// For Production
 const CONFIG_URL = '/ovirt-engine/web-ui/userportal.config'
 
 const AppConfiguration = {
   debug: true,
-  applicationContext: '/ovirt-engine',
-  applicationURL: '/ovirt-engine/web-ui',
+  applicationContext: '',
+  applicationURL: '',
 }
 
 export function readConfiguration () {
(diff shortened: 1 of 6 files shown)


matobet:
@mareklibra @jniederm please

src/userportal.config
```diff
@@ -1,5 +1,5 @@
 {
   "debug": true,
-  "applicationContext": "/ovirt-engine",
-  "applicationURL": "/ovirt-engine/web-ui"
+  "applicationContext": "",
```
mareklibra: Did you try in production environment (SSO, .war deployment) ?
I think this will not work there.

matobet: Ah you are right. For some reason I thought that there are 2 separate userportal.config files used for dev and production but now I notice that this will indeed end up being copied in the resulting war. 

I'm thinking that to differentiate between dev and prod we can use the presence of this config file, e.g. if we fail to load the config -> the default (development) will be used..

src/config.js
```diff
@@ -2,17 +2,12 @@ import $ from 'jquery'
...
-
-// TODO: configure path automatically
-// For Production
 const CONFIG_URL = '/ovirt-engine/web-ui/userportal.config'
 
 const AppConfiguration = {
   debug: true,
-  applicationContext: '/ovirt-engine',
-  applicationURL: '/ovirt-engine/web-ui',
+  applicationContext: '',
```
mareklibra: So for production, there MUST be the config file available.
A little bit unusual - working defaults would be better.
What about adding new (additional) config file ovirt-web-ui.devel.config which will be loaded when userportal.config (will be renamed later) is missing?

matobet:
@mareklibra verified by deploying on production ovirt-engine

PR #71, opened 2017-01-13
Start script explains why it fails

jniederm:
... when ENGINE_URL is not defined.


jniederm:
@mareklibra please review

PR #72, opened 2017-01-13
package.json not processed by autotools

jniederm:
From this patch on, package.json file is the authoritative source of
version of the project.

This allows to run npm/yarn right away without need for running `./autogen.sh` first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,5 @@ configure
 install-sh
 missing
 ovirt-web-ui.spec
-package.json
 tmp.repos/
 
--- a/configure.ac
+++ b/configure.ac
@@ -41,7 +41,6 @@ test -z "${NPM}" && AC_MSG_ERROR([NodeJS npm program not found])
 
 AC_CONFIG_FILES([
 	Makefile
-	package.json
 	ovirt-web-ui.spec
 ])
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ovirt-web-ui",
-  "version": "@PACKAGE_VERSION@",
+  "version": "0.0.0",
   "private": true,
   "devDependencies": {
     "autoprefixer": "6.4.1",


jniederm:
@mareklibra please review

mareklibra:
I do prefer as much automation as possible during release and versioning to avoid dummy mistakes.
Moreover, the wrapping makefile is the entry point for building in copr/brew. So no matter the npm/webpack can be used for development independently, the makefile is the single source of truth for building. In other words, the ./autogen.sh shall be still called after 'git clone' anyway.

From this perspective, I prefer to keep the package.json.in file.

jniederm:
I understand the the dislike towards making things more complicated. I actually believes this patch can help us make things simpler.

With this patch we can run development setup like this:

```
git clone git@github.com:oVirt/ovirt-web-ui.git
cd ovirt-web-ui
npm install
npm start
```

so it lowers bars for other contributors by using standard JS setup.

srpm and spec file for brew can also be generated using standard procedure:

```
git clone git@github.com:oVirt/ovirt-web-ui.git
cd ovirt-web-ui
./autogen.sh
make
make srpm
```

matobet:
@mareklibra I understand your concern and I also would be skeptical towards something that would break us from the makefile model or would complicate build of rpms too much. 

But in this case (maintaining the version in the `package.json`) file in sync I believe it causes more harm than good. It greately complicates maintenance of dependency versions (they always have to be manually copied to the `package.json.in` file, the .in file is not sorted, .etc). 

Considering the version number in package.json is irrelevant for packages that are not published on the npm registry I would rather go with this approach (we can set it to `"0.0.0"` for ever). You can see this decision being also made for the `ovirt-engine-nodejs-modules` on which we depend :) in a similar discussion https://gerrit.ovirt.org/#/c/69934/1/specs/ovirt-engine-nodejs-modules/package.json

jniederm:
@mareklibra please review

PR #73, opened 2017-01-13
Ovirt favicon

jniederm:
React favicon replaced by green 'O' symbol copied from webadmin.


jniederm:
@mareklibra please review

PR #86, opened 2017-02-14
Update to yarn

jniederm:
* 'yarn.lock' added containing fix versions of resolved dependencies.
* Specific dependency on "bootstrap-switch": "3.3.2" added to workaround
  problem that newest bootstrap-switch (3.3.3) introduces new peer
  dependency on "jquery": "^1.9.0" and thus collides with our (runtime)
  dependency "jquery": "^3.1.1". (Thanks Vojtech!)

This patch doesn't update automation scripts to yarn. npm should still
work for automation.

Please review @vojtechszocs @matobet @mareklibra


mareklibra:
@jniederm , Thanks for the patch!
the yarn.lock is being created even here: https://github.com/oVirt/ovirt-web-ui/pull/82

mareklibra:
Verified.
There will be followe-up patch for makefile

PR #95, opened 2017-03-03
VmsPageHeader and related components migrated

jniederm:
... from ovirt-ui-components.

@mareklibra @matobet please review

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,7 +3,9 @@ import { connect } from 'react-redux'
 
 import './App.css'
 
-import { VmsList, VmDetail, VmsPageHeader, Options } from 'ovirt-ui-components'
+import VmsPageHeader from './components/VmsPageHeader'
+
+import { VmsList, VmDetail, Options } from 'ovirt-ui-components'
 
 const App = ({ vms, visibility }) => {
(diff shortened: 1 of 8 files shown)


src/components/VmsPageHeader/style.css
```diff
@@ -0,0 +1,2 @@
+.user-name {
```
matobet: I know this was probably a blind move of code but maybe can stop and think if we need this file at all :-) What do you think @mareklibra ?

mareklibra: @jniederm , please do not hesitate and clean up similar things.

Anyway, since these changes are hard to track/review, **please be thorough in mentioning** them in commit or PR comments.

jniederm: done

mareklibra:
The migration was not properly verified, we need to be more careful next time.
Various functions are missing in the web-ui project, namely:
- helpers.js: formatTwoDigits
- actions.js: various action creators

Follow-up PR is required, web-ui build for ovirt-41 is blocked with this malfunction

matobet:
CC @jniederm

Issue #17, opened 2016-10-04
Add 'Create VM from Template'


mareklibra:
Basic Add/Edit functionality provided by ~~https://github.com/oVirt/ovirt-web-ui/pull/161~~.

For follow-ups:

 [x] update yarn.lock
 [x] The "Description" property
 [x] Render template version
 [x] Fix form update after template change
 [x] Confirm dialog Cancel if changes
 [x] Refactoring of remaining reducers
 [x] Use existing UserMessages instead of specific vmDialogReducer
 [x] Add [help](https://www.patternfly.org/pattern-library/forms-and-controls/field-level-help)

mareklibra:
https://github.com/oVirt/ovirt-web-ui/pull/98

mareklibra:
Closed by mistake, not yet finished

mareklibra:
![01](https://cloud.githubusercontent.com/assets/17194943/25433996/e0569770-2a8a-11e7-806a-916b00b19ba5.png)
![02](https://cloud.githubusercontent.com/assets/17194943/25433999/e05810be-2a8a-11e7-9212-ca7cd2008ae5.png)
![03](https://cloud.githubusercontent.com/assets/17194943/25434001/e059069a-2a8a-11e7-911d-85b2a5a36130.png)
![04](https://cloud.githubusercontent.com/assets/17194943/25433997/e056eacc-2a8a-11e7-8cc3-ea8e72cf8df4.png)
![05](https://cloud.githubusercontent.com/assets/17194943/25434000/e0580560-2a8a-11e7-97ed-190c5f6a8aff.png)
![06](https://cloud.githubusercontent.com/assets/17194943/25433998/e0575fca-2a8a-11e7-8524-c9b88e76600e.png)

mykaul:
@mareklibra  : Vm -> VM

mareklibra:
done

PR #98, opened 2017-03-08
addVM, editVM dialogs

mkrajnak:
Provides `Create VM` and `Edit VM` functionality.

In addition to review comments in the code:
- [x] move the "vm name" under the "os type"
- [x] change the memory to be in MiB (and the label to contain that it is in MiB)
- [x] change the combo boxes to be type aheads (https://www.patternfly.org/pattern-library/widgets/#bootstrap-combobox)
- [x] sort the date in combo boxes
- [x] after submit close the dialog
- [x] fail handling - reasonable user notification
- [x] Step for memory increase/decrease by Up/Down buttons should be 256MB

---
Will fix https://github.com/oVirt/ovirt-web-ui/issues/17

### DIFF
--- a/src/AddVmDialog.js
+++ b/src/AddVmDialog.js
@@ -1,0 +1,310 @@
+import React, { PropTypes } from 'react'
+import { connect } from 'react-redux'
+import $ from 'jquery'
+
+import LabeledSelect from './LabeledSelect'
+import LabeledTextField from './LabeledTextField'
+import DetailContainer from './components/DetailContainer'
+import ErrorAlert from './ErrorAlert'
+import Selectors from './selectors'
+import {
+  addNewVm,
+  editVm,
+  changeCluster,
+  changeTemplate,
+  updateOperatingSystem,
+  closeDetail,
+  updateVmName,
+  updateVmMemory,
+  updateVmCpu,
+  updateVmDialogErrorMessage,
+} from './actions'
+
+class vmDialog extends React.Component {
+  constructor (props) {
+    super(props)
+    this.createNewVm = this.createNewVm.bind(this)
+    this.changeCluster = this.changeCluster.bind(this)
+    this.changeTemplate = this.changeTemplate.bind(this)
+    this.changeOperatingSystem = this.changeOperatingSystem.bind(this)
+    this.closeDialog = this.closeDialog.bind(this)
+    this.changeVmName = this.changeVmName.bind(this)
+    this.changeVmMemory = this.changeVmMemory.bind(this)
+    this.changeVmCpu = this.changeVmCpu.bind(this)
+    this.submitHandler = this.submitHandler.bind(this)
+    this.getMemory = this.getMemory.bind(this)
+    this.clearErrorMessage = this.clearErrorMessage.bind(this)
+  }
+
+  componentDidUpdate () {
+    // without this template combobox its not rerendering
+    $(this.cluster).combobox('refresh')
+    $(this.template).combobox('refresh')
+    $(this.os).combobox('refresh')
+  }
+
+  componentDidMount () {
+    $(this.cluster).combobox()
+    $("input[type='text'].combobox").on('change', () => $(this.changeCluster))
+    $(this.template).combobox()
+    $("input[type='text'].combobox").on('change', () => $(this.changeTemplate))
+    $(this.os).combobox()
+    $("input[type='text'].combobox").on('change', () => $(this.changeOperatingSystem))
+  }
+
+  submitHandler (e) {
+    e.preventDefault()
+    this.props.type === 'edit' ? this.editVm() : this.createNewVm()
+  }
+
+  getMemory () {
+    return parseInt(this.memory.value) * 1048576
+  }
+
+  editVm () {
+    const vm = {
+      'name': this.name.value,
+      'template': { 'name': this.template.value },
+      'cluster': { 'name': this.cluster.value },
+      'memory': this.getMemory(),
+      'os': {
+        'type': this.os.value,
+      },
+      'cpu': {
+        'topology': {
+          'cores': '1',
+          'sockets': this.cpus.value,
+          'threads': '1',
+        },
+      },
+    }
+    this.props.edit(vm, this.props.vmId)
+  }
+
+  createNewVm () {
+    const vm = {
+      'vm': {
+        'name': this.name.value,
+        'template': { 'name': this.template.value },
+        'cluster': { 'name': this.cluster.value },
+        'memory': this.getMemory(),
+        'os': { 'type': this.os.value },
+        'cpu': {
+          'topology': {
+            'cores': '1',
+            'sockets': this.cpus.value,
+            'threads': '1',
+          },
+        },
+      },
+    }
+    this.props.addVm(vm)
+  }
+
+  closeDialog (e) {
+    e.preventDefault()
+    this.props.closeDialog()
+  }
+
+  changeCluster () {
+    if (!Selectors.getClusterByName(this.cluster.value)) {
+      this.props.setErrorMessage('Invalid Cluster selected')
+    } else if (this.props.cluster.get('name') === this.cluster.value) {
+      this.clearErrorMessage('Cluster')
+    } else {
+      this.props.changeCluster(Selectors.getClusterByName(this.cluster.value))
+      this.clearErrorMessage('Cluster')
+    }
+  }
+
+  changeTemplate () {
+    if (!Selectors.getTemplateByName(this.template.value)) {
+      this.props.setErrorMessage('Invalid Template selected')
+    } else if (this.props.template.get('name') === this.template.value) {
+      this.clearErrorMessage('Template')
+    } else {
+      this.props.changeTemplate(Selectors.getTemplateByName(this.template.value))
+      this.clearErrorMessage('Template')
+    }
+  }
+
+  changeOperatingSystem () {
+    if (!Selectors.getOperatingSystemByName(this.os.value)) {
+      this.props.setErrorMessage('Invalid Operating System selected')
+    } else if (this.props.os.get('name') === this.os.value) {
+      this.clearErrorMessage('Operating System')
+    } else {
+      this.props.changeOperatingSystem(Selectors.getOperatingSystemByName(this.os.value))
+      this.clearErrorMessage('Operating System')
+    }
+  }
+
+  changeVmName () {
+    this.props.changeVmName(this.name.value)
+  }
+
+  changeVmMemory () {
+    this.props.changeVmMemory(this.memory.value)
+  }
+
+  changeVmCpu () {
+    this.props.changeVmCpu(this.cpus.value)
+  }
+
+  clearErrorMessage (entity) {
+    if (this.props.errorMessage.includes(entity)) {
+      this.props.setErrorMessage('')
+    }
+  }
+
+  render () {
+    return (
+      <DetailContainer>
+        <h1>{this.props.type === 'edit' ? 'Edit Virtual Machine' : 'Create A New Virtual Machine'}</h1>
+        <hr />
+        <ErrorAlert message={this.props.errorMessage} />
+        <form className='form-horizontal'>
+          <LabeledSelect
+            id='clusterSelect'
+            label='Cluster'
+            selectClass='combobox form-control'
+            getValue={(input) => { this.cluster = input }}
+            onChange={this.changeCluster}
+            value={this.props.cluster.get('name')}
+            data={this.props.clusters.get('clusters').sort((a, b) =>
+              a.get('name').localeCompare(b.get('name'))
+            )} />
+
+          <LabeledSelect
+            id='templateSelect'
+            label='Template'
+            selectClass='combobox form-control'
+            getValue={(input) => { this.template = input }}
+            onChange={this.changeTemplate}
+            value={this.props.template.get('name')}
+            data={this.props.templates.get('templates').toList().filter(template => (
+                template.get('cluster') === this.props.cluster.get('id') || template.get('cluster') === '0')
+              ).sort((a, b) => a.get('name').localeCompare(b.get('name')))} />
+
+          <LabeledSelect
+            id='operatingSystemSelect'
+            label='Operating System'
+            selectClass='combobox form-control'
+            getValue={(input) => { this.os = input }}
+            onChange={this.changeOperatingSystem}
+            value={this.props.os.get('name')}
+            data={this.props.operatingSystems.get('operatingSystems').toList().sort((a, b) =>
+              a.get('name').localeCompare(b.get('name'))
+            )}
+            renderer={(item) => item.get('description')} />
+
+          <LabeledTextField
+            selectClass='combobox form-control'
+            getValue={(input) => { this.name = input }}
+            id='vmName'
+            label='Name'
+            placeholder='VM Name'
+            value={this.props.vmName}
+            setValue={this.changeVmName} />
+
+          <LabeledTextField
+            getValue={(input) => { this.memory = input }}
+            type='number'
+            id='vmMemory'
+            label='Memory (MB)'
+            placeholder='VM Memory'
+            value={this.props.memory}
+            setValue={this.changeVmMemory}
+            step={256} />
+
+          <LabeledTextField
+            getValue={(input) => { this.cpus = input }}
+            type='number'
+            id='vmCpu'
+            label='CPU'
+            placeholder='CPUs'
+            value={this.props.cpu}
+            setValue={this.changeVmCpu} />
+
+          <div className='form-group'>
+            <div className='col-sm-offset-2 col-sm-10'>
+              <button className='btn btn-default' type='submit' onClick={this.closeDialog}>Close</button>
+              <button className='btn btn-primary' type='submit' onClick={this.submitHandler}>Submit</button>
+            </div>
+          </div>
+        </form>
+      </DetailContainer>
+    )
+  }
+}
+
+vmDialog.propTypes = {
+  type: PropTypes.string,
+  clusters: PropTypes.object.isRequired,
+  cluster: PropTypes.object.isRequired,
+  templates: PropTypes.object.isRequired,
+  template: PropTypes.object.isRequired,
+  operatingSystems: PropTypes.object.isRequired,
+  os: PropTypes.object.isRequired,
+  cpu: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+  memory: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+  vmName: PropTypes.string.isRequired,
+  vmId: PropTypes.string.isRequired,
+  changeCluster: PropTypes.func.isRequired,
+  changeTemplate: PropTypes.func.isRequired,
+  changeOperatingSystem: PropTypes.func.isRequired,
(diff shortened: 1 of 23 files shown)


jelkosz:
@vojtechszocs FYI

jelkosz:
couple of notes:

- [ ] move the "vm name" under the "os type"
- [ ] change the memory to be in MiB (and the label to contain that it is in MiB)
- [ ] change the combo boxes to be type aheads (https://www.patternfly.org/pattern-library/widgets/#bootstrap-combobox)
- [ ] sort the date in combo boxes
- [ ] after submit close the dialog

mkrajnak:
@jelkosz All Done, except 

> change the combo boxes to be type aheads (https://www.patternfly.org/pattern-library/widgets/#bootstrap-combobox)

There is one problem which I am not sure how to solve properly. When I install bootstrap-combobox via yarn or npm installed package is this https://github.com/cchantep/bootstrap-combobox. It lacks the functionality we want.
Combobox used by patterfly is this one https://github.com/danielfarrell/bootstrap-combobox. I tried to install it via yarn from github, which should work as it is mentioned in documentation but it ends with error. So I cloned it manually to node_modules, which temporarily solved the problem. 
How should I properly include right package ?

michalskrivanek:
opened https://ovirt-jira.atlassian.net/browse/OVIRT-1249 to stop spam from lago-bot

landgraf:
ci test please

src/sagas.js
```diff
@@ -23,7 +23,41 @@ import {
 } from 'ovirt-ui-components'
 
 // import store from './store'
-import { persistState, getSingleVm } from './actions'
+import {
+  persistState,
```
matobet: I would vote for moving all *VmDialog related sagas to a separate file. This is starting to become unmanageable.

src/reducers/reducers.js
```diff
@@ -8,6 +8,13 @@ import {
   VisibilityReducer as visibility,
 } from 'ovirt-ui-components'
```
matobet: when moving from a single file `foo.js` to a directory `foo` with mutliple files I would prefer the main file to be called `index.js`. This will allow other code that imported `foo` to work as before and also enforces standardized naming across the modules.

src/sagas.js
mareklibra: Agree

landgraf:
ci test please

src/reducers/templates.js
```diff
@@ -0,0 +1,15 @@
+import Immutable from 'immutable'
+import { actionReducer } from './utils'
+
+const templates = actionReducer(Immutable.fromJS({ templates: {}, loadInProgress: true }), {
+  ADD_TEMPLATES (state, { payload: { templates } }) {
+    const updates = {}
```
matobet: Could you please wrap this peace of code into some sort of `groupById` utility function? I believe I saw this piece of code copy repeated in many places.

src/reducers/vmDialog.js
```diff
@@ -0,0 +1,53 @@
...
+  cluster: {},
+  template: {},
+  os: {},
+  name: '',
+  memory: '',
+  cpu: {},
+  errorMessage: '',
+}), {
+  UPDATE_CLUSTER (state, { payload: { cluster } }) {
+    const imUpdates = Immutable.fromJS(cluster)
```
matobet: How about inlining those trivial intermediate variables? If after inlining it would appear too long you can also consider importing the `fromJS` function using a named import: `import { fromJS } from 'immutable'`.

src/selectors.js
```diff
@@ -17,6 +17,21 @@ Selectors = {
   getVmDisks ({ vmId }) {
     return getState().vms.getIn(['vms', vmId, 'disks'])
   },
+  getFirstCluster () {
+    return getState().clusters.get('clusters').toList().first()
+  },
+  getClusterById (id) {
+    return getState().clusters.get('clusters').toList().find(cluster =>
```
matobet: If I read the reducers correctly, those entities (clusters/templates/oss) are already stored in ID-indexed hash-maps so I really not see the reason behind doing this `O(n)` lookup when a simple get by id from the immutable map should suffice.

src/templateDialog.js
```diff
@@ -0,0 +1,214 @@
...
+  updateEditTemplateDescription,
+  updateEditTemplateOS,
+  updateEditTemplateMemory,
+  updateEditTemplateCpu,
+  changeEditTemplate,
+  editTemplate,
+  closeDetail,
+} from './actions'
+
+class templateDialog extends React.Component {
```
matobet: components with `PascalCase` please - `TemplateDialog`

src/templateDialog.js
```diff
@@ -0,0 +1,214 @@
...
+    this.props.changeName(this.templateName.value)
+  }
+
+  getOperatingSystem (name) {
+    return this.props.operatingSystems.get('operatingSystems').toList().find(os =>
+      os.get('description') === name)
+  }
+
+  getTemplate (name) {
+    return this.props.templates.get('templates').toList().find(template =>
```
matobet: selector wanting to be extracted?: e.g. `getTemplateByName`

src/templateDialog.js
```diff
@@ -0,0 +1,214 @@
...
+    $(this.template).selectpicker()
+    $(this.templateOs).selectpicker()
+  }
+
+  changeName () {
+    this.props.changeName(this.templateName.value)
+  }
+
+  getOperatingSystem (name) {
+    return this.props.operatingSystems.get('operatingSystems').toList().find(os =>
```
matobet: same as below

src/templateDialog.js
```diff
@@ -0,0 +1,214 @@
...
+
+  render () {
+    return (
+      <DetailContainer>
+        <h1>Edit Template</h1>
+        <hr />
+        <ErrorAlert message={this.props.errorMessage} />
+        <form className='form-horizontal'>
+          <LabeledSelect
+            label='Templates'
```
matobet: @mareklibra please what is the current strategy regarding i18n in ovirt-web-ui? Should we already in this PR for the dialog code use some sort of i18n infra?

src/labeledTextField.js
```diff
@@ -0,0 +1,37 @@
...
+  type: PropTypes.string,
+  key: PropTypes.string,
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+  disabled: PropTypes.bool,
+}
+
+export default connect()(LabeledTextField)
```
matobet: Does this component need to be connected to Redux at all? To me it seems as a pure presentational ("dumb") component only depending on props.

src/templateDialog.js
gregsheremeta: feel free to ping @vojtechszocs and @sjd78 for their thoughts on i18n, especially if you're going to hook into Zanata like ovirt-engine-dashboard does.

src/reducers/dialogVisibility.js
```diff
@@ -0,0 +1,25 @@
+import Immutable from 'immutable'
+import { actionReducer } from './utils'
+
+const dialogVisibility = actionReducer(Immutable.fromJS({ showVmDialog: false, showVmDetail: false, showEditTemplate: false }), {
```
matobet: Now considering how this will scale for the future when we will have even more dialog types I believe a bit different store representation would be more suitable. 

One alternative would be to have a simple `dialog` field there with a string identifying the type of dialog that is to be shown (since I assume this sort of "top level" dialog will be possible to display only one at a time) with values like (`null`, `"VM_EDIT"`, `"TEMPLATE_EDIT"`, ...)

Other to have this split by 2 levels: one field identifying the general dialog and second the current "behavior" (sorry for the pun with legacy gwt UICommon model :-)) or it. But in any case I would prefer to avoid this explosion of boolean flags.....

What do you guys think: @jelkosz @mareklibra @vojtechszocs ?

src/reducers/editTemplate.js
```diff
@@ -0,0 +1,42 @@
...
+  name: '',
+  memory: '',
+  cpu: '',
+  errorMessage: '',
+}), {
+  UPDATE_EDIT_TEMPLATE (state, { payload: { template } }) {
+    return Immutable.fromJS(template)
+  },
+  UPDATE_EDIT_TEMPLATE_NAME (state, { payload: { name } }) {
+    const imUpdates = Immutable.fromJS(name)
```
matobet: same comment about removing intermediate variables

src/reducers/dialogVisibility.js
jelkosz: yes, this representation is a call for trouble because it is a question of time until due to some bug you will have {showVmDialog: true, showVmDetail: true, showEditTemplate: true} - and now what :)

The struct: {type: ..., behavior: ...} sounds good to me. Could be: {type: VM, behavior: edit}

src/templateDialog.js
mareklibra: I18n will be added to the project as a separate PR, I'm considering gettext to extract the strings.

src/index.js
```diff
@@ -13,6 +13,10 @@ window.$ = window.jQuery = require('jquery')
 require('bootstrap/dist/js/bootstrap')
 window.patternfly = {}
 window.patternfly = require('patternfly/dist/js/patternfly')
+window.selectpicker = {}
+window.selectpicker = require('bootstrap-select/js/bootstrap-select.js')
```
mareklibra: @mkrajnak , please add bootstrap-select to package.json

src/index.js
```diff
@@ -13,6 +13,10 @@ window.$ = window.jQuery = require('jquery')
 require('bootstrap/dist/js/bootstrap')
 window.patternfly = {}
 window.patternfly = require('patternfly/dist/js/patternfly')
+window.selectpicker = {}
```
mareklibra: I would need to retry, but as IIRC it's a hack to allow lazy init of those libraries.
I would not consider it as a pattern, so for the selectpicker please first try without it. If it works, let's do not use it.

src/reducers/dialogVisibility.js
mkrajnak: @mbetak @jelkosz This is basically a hack that I am able to properly handle in sagas.

>  {showVmDialog: true, showVmDetail: true, showEditTemplate: true}
This is actually necessary because the only way to "Open the Doors" is to dispatch showVmDetail as shown bellow, to open addVmDialog:
`function* showAddNewVm (action) {
  yield put(setVmDetailToShow({ vmId: '0' })) // open the gate
 .......... 
 yield put(openVmDialog())
}`
 I know it is not a best solution,  @mbetak has a good suggestion I can try it later.

mareklibra: Somewhere else, I'm using this structure:
{
topLevel: 'type as String constant  or Object with details'
// if really required later, as a hack we can add another key along the topLevel, it would be still manageble
type: {
 // type specific map of details
 // if complex, then tree structure as the topLevel
 // but oVirt web admin is a good example that there's no need for very complex 2nd or 3rd level 
}
}

The reudcer (read: at one place) handles potential complexities of dependencies.

It works well. The ImmutableJS is a big benefit.

ovirt-infra:

ovirt-infra:

ovirt-infra:

src/TemplateDialog.js
```diff
@@ -0,0 +1,203 @@
...
+        'type': this.templateOs.value,
+      },
+    }
+    this.props.editTemplate(template, this.props.template.get('id'))
+  }
+
+  render () {
+    return (
+      <DetailContainer>
+        <h1>Edit Template</h1>
```
jelkosz: This whole thing is from large part a copy/paste of the VmDialog. There are couple of issues with this, namely:
- when this becomes bigger, it becomes a maintenance nightmare
- when this will edit more entities (pools, instance types) it will become a maintenance nightmare on steroids (trust me, we had a similar issue with this dialog on the ovirt and we were drowning in in bugs like "this field is working in edit VM and Pool, but does not work in instance type" etc.)
- even now when it is small you have a bug like this here :) The memory label in VM has the changed label "Memory (MB)" and here it is still "Memory".

src/sagas.js
```diff
@@ -245,6 +282,80 @@ function* startVm (action) {
...
+}
+
+function* handleClusterChange (action) {
+  yield put(updateCluster(action.payload.cluster))
+  // After every cluster change, set template to Blank
+  const blankTemplate = Selectors.getTemplateById('00000000-0000-0000-0000-000000000000')
+  yield put(changeTemplate(blankTemplate))
+}
+
+function* handleTemplateChange (action) {
```
jelkosz: It would be maybe good to do this right now while the patch is still young and the dialog does not have too much properties. 
So, what I mean - imagine this scenario:
- you have a VM and it was a hard labor to get all it's properties tuned according to your need.
- it is working and all is fine
- than you get access to a new cluster with more powerful hosts or whatever so you want to move it there
- so, you open the edit VM dialog, change the cluster and all is gone. All your settings are reset to defaults

Here we need to be a bit more smart. Either distinguish between new and edit (what I don't like but it is what we do in oVirt) or be even more smart and be able to decide if this field has been modified by the user or did not.
For edit VM in initialization all fields would be marked as modified by user and for new none. Than if you touch on new some field, it will be marked as modified by user. Than, if you change the template for example, the particular value will be rewritten from template only if it does not make sense in the context of new template, otherwise will stay intact.

@mkrajnak @mareklibra @jniederm @vojtechszocs @matobet what you guys think?

src/TemplateDialog.js
mkrajnak: I think I will be able to wrap Operating system, memory, cpus and template to separated component.

src/sagas.js
jelkosz: OK, it may turn out to be a pretty complicated change - can be done later. This patch brings lots of value also without this proposed change.

src/TemplateDialog.js
jelkosz: yes, that sounds good

mareklibra:
Rebase is needed because of https://github.com/oVirt/ovirt-web-ui/pull/109

mkrajnak:
I pushed initial rebase

src/sagas.js
```diff
@@ -295,19 +402,106 @@ function* schedulerPerMinute (action) {
...
+}
+
+function* fetchAllOS (action) {
+  const operatingSystems = yield callExternalAction('getAllOperatingSystems', Api.getAllOperatingSystems, action)
+
+  if (operatingSystems && operatingSystems['operating_system']) {
+    const operatingSystemsInternal = operatingSystems.operating_system.map(os => Api.OSToInternal({ os }))
+    yield put(addAllOS({ os: operatingSystemsInternal }))
+    const activeOperatingSystem = operatingSystemsInternal.find(os => os.name === 'other')
+    yield put(updateOperatingSystem(activeOperatingSystem))
```
mareklibra: This flow seems weird to me, please describe what's happening here.
Do I understand correctly, it's "updating a copy" held by potentially opened dialog? If so, the flow should be redesigned. @matobet , what do you think?

src/sagas.js
```diff
@@ -295,19 +402,106 @@ function* schedulerPerMinute (action) {
...
+  }
+}
+
+function* fetchAllOS (action) {
+  const operatingSystems = yield callExternalAction('getAllOperatingSystems', Api.getAllOperatingSystems, action)
+
+  if (operatingSystems && operatingSystems['operating_system']) {
+    const operatingSystemsInternal = operatingSystems.operating_system.map(os => Api.OSToInternal({ os }))
+    yield put(addAllOS({ os: operatingSystemsInternal }))
+    const activeOperatingSystem = operatingSystemsInternal.find(os => os.name === 'other')
```
mareklibra: Why this 'other' constant?

src/actions/dialog.js
```diff
@@ -0,0 +1,304 @@
+export function updateVmDialogErrorMessage (message) {
```
mareklibra: Please rename this file to something more descriptive and related to its purpose.

src/sagas.js
```diff
@@ -245,6 +277,80 @@ function* startVm (action) {
   yield stopProgress({ vmId: action.payload.vmId, name: 'start', result })
 }
 
+function* showEditVm (action) {
+  yield put(updateDialogType('edit'))
```
mareklibra: Maybe I'm missing something, but I don't prefer saga to be used for opening a dialog.
It seems to be too procedural.

Much better approach for Edit VM Dialog can be via React itself, something like following pseudo-code:
  <EditVM vm={myVmToEdit} clusters={clusters} templates={allTemplates} ...  />

The EditVm component state consists of a VM described as references to other resources, similar to:
this.state = {
vm: {
name,
cpus,
clusterId,
templateId,
...
}
}

With this approach, we
- avoid of this procedural code
- leverage React/Redux for updates if underlying sub-resources are changed or gradually loaded
- move rendering-related code to React component

@jelkosz , @matobet , please comment.

src/sagas.js
```diff
@@ -295,19 +402,106 @@ function* schedulerPerMinute (action) {
   }
 }
 
+function* createNewVm (action) {
```
mareklibra: again, I think it's much better to wrap this logic into a React component and dispatch a "CreateVM" action once the "Submit" button is clicked.

mareklibra:
Regarding comboboxes (see review), this shall be more elegant: http://www.patternfly.org/pattern-library/widgets/#bootstrap-combobox

mkrajnak:
@mareklibra its the same bootstrap-combobox like on page you posted, X will appear when you have actual value inside
image

mareklibra:
I missed that, thanks.

No matter it's patternfly, I don't like this in our use case. 
What about preselecting a default value (means "Blank" for template, "default" for cluster) if none is selected?
We would not deviate from the patternfly behavior much and the dialog would remain in consistent state.

@jelkosz , @matobet , @mkrajnak , what do you think?

jelkosz:
@mareklibra yes, makes sense

mareklibra:
WIP: follow-up patches to resolve review issues above

src/ovirtapi.js
```diff
@@ -184,21 +250,48 @@ OvirtApi = {
...
+    return OvirtApi._httpGet({ url })
+  },
+  getAllClusters () {
+    OvirtApi._assertLogin({ methodName: 'getAllClusters' })
+    const url = '/api/clusters'
+    return OvirtApi._httpGet({ url })
+  },
+  getAllOperatingSystems () {
+    OvirtApi._assertLogin({ methodName: 'getAllOperatingSystems' })
+    const url = '/api/operatingsystems'
```
mareklibra: Done

src/actions/dialog.js
mareklibra: Done

src/AddVmDialog.js
```diff
@@ -0,0 +1,310 @@
...
+    $("input[type='text'].combobox").on('change', () => $(this.changeOperatingSystem))
+  }
+
+  submitHandler (e) {
+    e.preventDefault()
+    this.props.type === 'edit' ? this.editVm() : this.createNewVm()
+  }
+
+  getMemory () {
+    return parseInt(this.memory.value) * 1048576
```
mareklibra: Done

src/AddVmDialog.js
```diff
@@ -0,0 +1,310 @@
+import React, { PropTypes } from 'react'
```
mareklibra: Done

mareklibra:
Replaced by https://github.com/oVirt/ovirt-web-ui/pull/161

Issue #66, opened 2016-12-06
Add detection that someone is connected to the VM


jelkosz:
If someone is connected to the console of the VM, the box should show something like "Console in use".

bond95:
https://github.com/oVirt/ovirt-web-ui/pull/106
https://github.com/oVirt/ovirt-ui-components/pull/34

PR #106, opened 2017-03-17
Add feature for changing console button tooltip to "Console in use"

bond95:
Fixes: #66

### DIFF
--- a/src/actions/vm.js
+++ b/src/actions/vm.js
@@ -1,9 +1,9 @@
 import {
   ADD_NEW_VM,
   CLEAR_USER_MSGS,
+  DOWNLOAD_CONSOLE_VM,
   EDIT_VM,
   GET_ALL_VMS,
-  GET_CONSOLE_VM,
   LOGIN,
   LOGIN_SUCCESSFUL,
   LOGOUT,
@@ -15,6 +15,7 @@ import {
   SET_OVIRT_API_VERSION,
   SET_VM_CONSOLES,
   SET_VM_DISKS,
+  SET_VM_SESSIONS,
   SHUTDOWN_VM,
   START_VM,
   SUSPEND_VM,
@@ -82,9 +83,9 @@ export function startVm ({ vmId }) {
   }
 }
 
-export function getConsole ({ vmId, consoleId }) {
+export function downloadConsole ({ vmId, consoleId }) {
   return {
-    type: GET_CONSOLE_VM,
+    type: DOWNLOAD_CONSOLE_VM,
     payload: {
       vmId,
       consoleId,
@@ -266,3 +267,12 @@ export function setVmConsoles ({ vmId, consoles }) {
   }
 }
 
+export function setVmSessions ({ vmId, sessions }) {
+  return {
+    type: SET_VM_SESSIONS,
+    payload: {
+      vmId,
+      sessions,
+    },
+  }
+}
--- a/src/components/DetailContainer.js
+++ b/src/components/DetailContainer.js
@@ -27,7 +27,7 @@ class DetailContainer extends Component {
   render () {
     const { children } = this.props
     return (
-      <div className={'container-fluid ' + sharedStyle['move-left-detail']}>
+      <div className={`container-fluid ${sharedStyle['move-left-detail']} ${sharedStyle['detail-z-index']}`}>
         {children}
       </div>
     )
--- a/src/components/VmActions/actions.js
+++ b/src/components/VmActions/actions.js
@@ -1,0 +1,20 @@
+import { SET_CONSOLE_IN_USE, CONSOLE_IN_USE } from './constants'
+
+export function setConsoleInUse ({ vmId, consoleInUse }) {
+  return {
+    type: SET_CONSOLE_IN_USE,
+    payload: {
+      vmId,
+      consoleInUse,
+    },
+  }
+}
+
+export function checkConsoleInUse ({ vmId }) {
+  return {
+    type: CONSOLE_IN_USE,
+    payload: {
+      vmId,
+    },
+  }
+}
--- a/src/components/VmActions/index.js
+++ b/src/components/VmActions/index.js
@@ -7,7 +7,6 @@ import style from './style.css'
 
 import {
   Checkbox,
-
   canRestart,
   canShutdown,
   canStart,
@@ -17,13 +16,14 @@ import {
 } from 'ovirt-ui-components'
 
 import {
-  getConsole,
+  downloadConsole,
   shutdownVm,
   restartVm,
   suspendVm,
   openEditVmDialog,
   removeVm,
 } from '../../actions/index'
+import { checkConsoleInUse, setConsoleInUse } from './actions'
 
 import OnClickTopConfirmation from '../Confirmation/index'
 
@@ -102,7 +102,7 @@ const EmptyAction = ({ state, isOnCard }) => {
   return null
 }
 EmptyAction.propTypes = {
-  state: PropTypes.string.isRequired,
+  state: PropTypes.string,
   isOnCard: PropTypes.bool.isRequired,
 }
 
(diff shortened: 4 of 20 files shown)


ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

src/sagas.js
```diff
@@ -138,6 +139,7 @@ function* fetchAllVms (action) {
     yield put(removeMissingVms({ vmIdsToPreserve }))
 
     yield put(updateVms({ vms: internalVms, copySubResources: true }))
+    yield fetchVmsSessions({ vms: internalVms })
```
jelkosz: This is an additional query per each VM on every refresh cycle which is too high price for too small gain I'd say. 
I think we should do it a bit differently. Maybe something like this:
- do this if "shallowFetch == false" and than showing a msg in the VM details
- also, do this as a part of the attempt to connect to the console. So, the flow would be like this:
  - click connect to console button
  - first check the active sessions, if there is some active session, warn the user that there is an active session and ask if he wants to continue anyway
 - continue with the taking the console normally in case the user confirms

@mareklibra what you think?

mareklibra: Right, it's to costly since it's a per-VM call.

I agree with @jelkosz , let's try usability when triggered within console retrieval.

landgraf:
ci test please

mareklibra:
Rebase is needed because of https://github.com/oVirt/ovirt-web-ui/pull/109

src/components/VmActions/index.js
```diff
@@ -89,50 +92,105 @@ EmptyAction.propTypes = {
...
   }
 
-  const confirmRestart = {
-    confirmText: 'Restart?',
-    cancelText: 'Cancel',
+  onConsoleClick () {
+    this.setState({
+      openModal: true,
+    })
+    this.props.checkConsole()
```
mareklibra: change to onConsoleSessionConfirmaClose()

src/components/VmActions/index.js
```diff
@@ -142,9 +200,12 @@ VmActions.propTypes = {
 export default connect(
   (state) => ({
     icons: state.icons,
+    VmAction: state.VmAction,
   }),
   (dispatch, { vm }) => ({
+    checkConsole: () => dispatch(checkConsoleInUse({ vmId: vm.get('id') })),
```
mareklibra: Please use more self-descriptive name, instead of checkConsole use i.e. onCheckConsoleSessionInUse

src/components/VmActions/index.js
```diff
@@ -142,9 +200,12 @@ VmActions.propTypes = {
...
     onGetConsole: () => dispatch(getConsole({ vmId: vm.get('id') })),
+    modalClose: () => dispatch(setConsoleInUse({ vmId: vm.get('id'), consoleInUse: false })),
```
mareklibra: Similar: modalClose --> onConsoleSessionConfirmationClosed

src/components/VmActions/index.js
```diff
@@ -142,9 +200,12 @@ VmActions.propTypes = {
 export default connect(
   (state) => ({
     icons: state.icons,
+    VmAction: state.VmAction,
   }),
   (dispatch, { vm }) => ({
+    checkConsole: () => dispatch(checkConsoleInUse({ vmId: vm.get('id') })),
     onGetConsole: () => dispatch(getConsole({ vmId: vm.get('id') })),
```
mareklibra: With this change the overall "Get Console" flow is becoming hard to understand.
To help, please refactor throughout the application getConsole() ---> downloadConsole() and change the action type string accordingly.

src/components/VmActions/sagas.js
```diff
@@ -0,0 +1,23 @@
...
+import { setConsoleInUse } from './actions'
+
+function* getConsoleInUseModal (sagas, action) {
+  let { vmId } = action.payload
+  const sessionsInternal = yield sagas.fetchVmSessions({ vmId })
+  if (sessionsInternal &&
+    sessionsInternal.find((x) => x.consoleUser) !== undefined) {
+    yield put(setConsoleInUse({ vmId, consoleInUse: true }))
+  } else {
+    yield put(getConsole({ vmId }))
```
mareklibra: after recatoring: downloadConsole() (see different comment)

src/components/VmModals/VmYesNoModal.js
```diff
@@ -0,0 +1,45 @@
+import React, { PropTypes } from 'react'
```
mareklibra: The class is already written to be non-Vm specific. Nice!

Please rename and move this class to be more general.

<YesNoModal /> is enough.

src/components/VmModals/VmYesNoModal.js
```diff
@@ -0,0 +1,45 @@
...
+import ReactDOM from 'react-dom'
+
+class VmYesNoModal extends React.Component {
+
+  componentDidMount () {
+    const dom = ReactDOM.findDOMNode(this)
+    window.$(dom).modal('show')
```
mareklibra: please import the jquery, see config.js, i.e.

src/sagas.js
```diff
@@ -254,6 +265,15 @@ function* fetchConsoleVmMeta ({ vmId }) {
...
+function* fetchVmSessions ({ vmId }) {
+  const sessions = yield callExternalAction('sessions', Api.sessions, { payload: { vmId } })
+
+  if (sessions && sessions['session']) {
+    return Api.sessionsToInternal({ sessions })
+  }
+  return []
+}
+
 function* getConsoleVm (action) {
```
mareklibra: Please refactor (see comment above)

Use downloadVmConsole instead of getConsoleVm().

src/sagas.js
```diff
@@ -295,6 +321,12 @@ function* schedulerPerMinute (action) {
   }
 }
 
+let sagasFunctions = {
```
mareklibra: Please add comment describing purpose of this wrapping object.

Something like:
Saga functions to be called from subsagas .

Or better even more descriptive.

src/sagas.js
```diff
@@ -308,10 +340,12 @@ export function *rootSaga () {
     takeEvery('START_VM', startVm),
     takeEvery('GET_CONSOLE_VM', getConsoleVm),
```
mareklibra: refactor: GET_CONSOLE_VM --> DOWNLOAD_CONSOLE_VM

src/sagas.js
```diff
@@ -308,10 +340,12 @@ export function *rootSaga () {
...
     takeEvery('SUSPEND_VM', suspendVm),
+    takeEvery('GET_SESSIONS_VM', getSessionsVm),
```
mareklibra: Is this dispatched from somewhere? Please review whole flow.

src/sagasBuilder.js
```diff
@@ -0,0 +1,5 @@
+import { buildSagas as VmActionsSagas } from './components/VmActions/sagas'
+
+export default (sagas) => [
```
mareklibra: This is becoming one of the key parts of the overall architecture.
Please add comment describing it's purpose, try to be as much descriptive as possible.
Try to describe the general idea behind this composition and what is expected to be done for future additions.

mareklibra: It will help to keep it's use consistently and make it more understandable.

src/components/VmActions/index.js
```diff
@@ -89,50 +92,105 @@ EmptyAction.propTypes = {
...
+    let consoleProtocol = ''
+    if (!vm.get('consoles').isEmpty()) {
+      const vConsole = vm.get('consoles').find(c => c.get('protocol') === 'spice') ||
+        vm.getIn(['consoles', 0])
+      const protocol = vConsole.get('protocol').toUpperCase()
+      consoleProtocol = `Open ${protocol} Console`
+    }
+
+    if (vm.get('sessions') &&
+      !(vm.get('sessions').filter(c => c.get('consoleUser')).isEmpty())) {
```
mareklibra: With this patch, the redux store now contains two related properties:
- consoleUser
- consoleInUse

Both filled in within different flows but used for more or less similar use case. Can they be merged into one and related flows changed accordingly? This logical duplicity might cause difficulties for maintenance.

mareklibra: It's correct, the existence of potentially existing parallel console session is checked right before the console is downloaded (the new VmActions/sagas.js).

Just an idea:
Within this saga, the consoleUser might be set to some constant pessimistically enforcing the state the console is in use and once the fetchVmSessions() finishes, correct value is set

src/actions/vm.js
```diff
@@ -204,3 +204,13 @@ export function setVmConsoles ({ vmId, consoles }) {
     },
   }
 }
+
+export function setVmSessions ({ vmId, sessions }) {
+  return {
+    type: 'SET_VM_SESSIONS',
```
mareklibra: please use constants.js

src/components/VmActions/actions.js
```diff
@@ -0,0 +1,20 @@
+import { SET_CONSOLE_IN_USE, CONSOLE_IN_USE } from './constants'
```
mareklibra: I would prefer a single file to be used across whole application. 
It will help to avoid name conflicts without explicit namespacing.
Merge conflicts are easy to resolve.

mareklibra:
For user confirmation I do prefer the
   https://ethaizone.github.io/Bootstrap-Confirmation/
as written in my previous comment.

Just for a reference, if modal is supposed to be used, I would follow:
   http://www.patternfly.org/pattern-library/widgets/#modal

src/components/VmActions/actions.js
bond95: The idea of dividing is to make every module more independent, and this action use only this module, so there can not be any conflicts.

src/components/VmModals/VmYesNoModal.js
bond95: Move to ui-components?

src/components/VmActions/index.js
bond95: No, they can not be merged in to one, only if I add actions from reducer _VmAction_ to _vms_ reducer, but of whole that modularity is to make every module more independent and not to make mess in global files (like sagas.js or reducers/vms.js).
Yes, it's a good idea to add consoleInUse parameter to state

src/components/VmModals/VmYesNoModal.js
mareklibra: Please consider in context of the new popover confirmation introduced within https://github.com/oVirt/ovirt-web-ui/pull/127. I think, we can avoid this modal YesNo dialog.
Please note, the component is not yet finished (CSS polishing in progress ...)

bond95:
@mareklibra change modal window to user confirmation popup, please review

src/components/DetailContainer.js
```diff
@@ -25,7 +25,7 @@ class DetailContainer extends Component {
   render () {
     const { children } = this.props
     return (
-      <div className={'container-fluid ' + sharedStyle['move-left-detail']}>
+      <div className={'container-fluid ' + sharedStyle['move-left-detail'] + ' ' + sharedStyle['detail-z-index']}>
```
mareklibra: Can be left for a follow-up, but string composition with back-ticks is better in this case

yarn.lock
```diff
@@ -46,8 +46,8 @@ ajv-keywords@^1.0.0:
   resolved "https://registry.yarnpkg.com/ajv-keywords/-/ajv-keywords-1.5.1.tgz#314dd0a4b3368fad3dfcdc54ede6171b886daf3c"
 
 ajv@^4.7.0:
-  version "4.11.3"
-  resolved "https://registry.yarnpkg.com/ajv/-/ajv-4.11.3.tgz#ce30bdb90d1254f762c75af915fb3a63e7183d22"
+  version "4.11.5"
```
mareklibra: Since there are no changes to the package.json, please remove this file from commit.

bond95:
@mareklibra fixed. I think better if popup will look something like this,
screenshot from 2017-04-20 09-41-47

yarn.lock
```diff
@@ -1,5707 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
mareklibra: Please **do not** `git rm yarn.lock` :-)
My previous comment was just about committing changes if the package.json is untouched.

src/components/VmActions/index.js
```diff
@@ -91,76 +92,154 @@ EmptyAction.propTypes = {
...
-    id: vm.get('id'),
-    target: e.target,
-    confirmationText: 'Restart the VM?',
-    cancelLabel: 'Cancel',
-    okLabel: 'Yes',
-    onOk: onRestart,
-  })
+  onConsoleClick (e) {
+    this.setState({
+      openModal: true,
```
mareklibra: please rename to consoleConfirmationAboutToOpen

src/components/VmActions/index.js
```diff
@@ -91,76 +92,154 @@ EmptyAction.propTypes = {
...
 
-  const confirmSuspend = (e) => OnClickTopConfirmation({
-    id: vm.get('id'),
-    target: e.target,
-    confirmationText: 'Suspend the VM?',
-    cancelLabel: 'Cancel',
-    okLabel: 'Yes',
-    onOk: onSuspend,
-  })
+  onModalClose () {
```
mareklibra: please rename to onConsoleConfirmationClose()

src/components/VmActions/index.js
```diff
@@ -91,76 +92,154 @@ EmptyAction.propTypes = {
...
+      ) {
+      (() => OnClickTopConfirmation({
+        id: this.props.vm.get('id'),
+        target: this.state.notificationTarget,
+        confirmationText: 'Console in use, continue?',
+        cancelLabel: 'Cancel',
+        okLabel: 'Yes',
+        onOk: this.onConsoleDownload,
+        onCancel: this.onModalClose,
+      }))()
```
mareklibra: Can be the OnClickTopConfirmation() function executed directly without the wrapping?

src/components/VmActions/index.js
```diff
@@ -91,76 +92,154 @@ EmptyAction.propTypes = {
...
+    this.setState({
+      openModal: false,
+      notificationTarget: null,
+    })
+
+    this.props.onDownloadConsole()
+  }
+
+  componentDidUpdate (prevProps, prevState) {
+    if (
```
mareklibra: I think this code can be moved to the render() method since it's called once props/state is updated.

bond95: No, it can't cause function render must be pure and can only return component, but not change other components, or added new one inside

src/components/VmActions/index.js
bond95: I tried do it without wrapping and it throw error

src/components/VmActions/index.js
mareklibra: The flow is strange, but ok, let's keep it.
Please simplify the call on line 148, I think you can call the function directly, can't you?

bond95:
@mareklibra fixed conflicts, please review

mareklibra:
Merge is pending due to conflicts in https://github.com/oVirt/ovirt-ui-components/pull/34

bond95:
@mareklibra please review

mareklibra:
https://github.com/oVirt/ovirt-ui-components/pull/34 seems to be still needed for this PR, please fix issues there. Once merged in ovirt-ui-components, we can merge here.

bond95:
@mareklibra I closed this oVirt/ovirt-ui-components#34 PR, because all changes was already rebased to web-ui

mareklibra:
Replaced by https://github.com/oVirt/ovirt-web-ui/pull/204

PR #109, opened 2017-03-23
Migration of components from ovirt-ui-componets

jniederm:
Css handling changed: Stylesheets
* from dependencies and local stylesheets with '-nomodules.css'  are
  handled as before.
* defined locally are processed using css modules - similarly to
  ovirt-ui-components.

There are no real changes compared to previous state, just shuffling imports.


mareklibra:
Verified.
Built agains ovirt-ui-components with https://github.com/oVirt/ovirt-ui-components/pull/35 applied, UI seems to be unchanged, actions work.

To finish migration:
- merge https://github.com/oVirt/ovirt-ui-components/pull/35
- new release OUC
- increase OUC version in web-ui/package.json
- release new ovirt-engine-node-modules
- release web-ui

config/webpack.config.prod.js
```diff
@@ -109,6 +109,7 @@ module.exports = {
       // in the main CSS file.
       {
         test: /\.css$/,
+        include: /(node_modules)|(-nomodules\.css$)/,
```
matobet: nis this -nomodules solution intended as long term or are there any plans  to migrate all of the ovirt-web-ui to (arguably superior) css-modules? If yes please open an issue with this item after this PR is merged

package.json
```diff
@@ -41,30 +42,31 @@
...
     "http-proxy-middleware": "0.17.1",
+    "immutable": "^3.8.1",
     "jest": "15.1.1",
     "json-loader": "0.5.4",
     "object-assign": "4.1.0",
     "opn": "4.0.2",
     "path-exists": "3.0.0",
     "postcss-loader": "0.13.0",
     "promise": "7.1.1",
+    "react-router": "~3.0.0",
```
matobet: @mareklibra do we currently use react-router in ovirt-web-ui? If not I would avoid this until we properly introduce it (with v4 of react-router)

mareklibra: No, it's commented-out code

src/actions/visibility.js
```diff
@@ -0,0 +1,41 @@
+export function showLoginDialog () {
```
mareklibra: The code is there but effectively not executed since oVirt SSO overrides the flow.
In the future, even with remote deployment, the oVirt SSO will be still used.

The LoginForm and related code can be removed.
But to avoid bugs, I would postpone it to a followup patch.

Let's keep this PR effort as much simple as possible.

config/webpack.config.prod.js
jniederm: I've migrate all styles to css modules where it was easily possible. There is currently one sheet using this approach `src/index-nomodules.css` used to style `<body>` element.

@mareklibra, can we expect expect further styling outside the react root element? Do we want to be able to use stylesheets without  css module?

Styling of `<body>` element can be inlined into the `index.html` or link it statically using `<link>` as an alternative to current -nomodules.css suffix approach.

src/store.js
```diff
@@ -1,7 +1,7 @@
 import { createStore, applyMiddleware, compose } from 'redux'
 import createSagaMiddleware from 'redux-saga'
 
-import reducer from './reducers'
+import reducer from './reducers/index'
```
jniederm: done

package.json
jniederm: removed

config/webpack.config.prod.js
mareklibra: @jniederm , I don't expect that.

I think we can achieve all potential customization of Patternfly (or so) on module level. 
Anyway, I don't expect any major "rework" of styling in general, just local adjustments for a component or it's children. At least this is my best wish :-)

mareklibra:
Verified, still works.
@jniederm , are we ready to merge?

PR #124, opened 2017-04-07
Preparing for regular releases as docker container

jelkosz:
- handled errors in start script (especially the open browser always raised an
  error from container)
- updated node
- changed npm to yarn in dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ COPY index.html /
 COPY scripts /scripts
 COPY config /config
 COPY package.json /
-RUN npm install
+RUN yarn install
 COPY src /src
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM node:6
+FROM node:7
```
matobet: Why this version bump? I was under the impression that all ovirt projects have always the current LTS version of Node.js as their base runtime - currently v6. I believe our next bump should come when Node.js 8 enters LTS phase in october of this year https://github.com/nodejs/LTS#lts-schedule1

jelkosz: you are right, will move it back

mareklibra:
Replace by #136

Issue #16, opened 2016-10-04
Add 'Force' version of Shut down


mareklibra:

mareklibra:
Since Force Restart is currently not supported in oVirt and workaround consists of
 - Force Shut Down 
 - Run

this is skipped from implementation.

So Force Shut down is implemented only, see https://github.com/oVirt/ovirt-web-ui/pull/127.
Confirmation component is improved.

PR #127, opened 2017-04-10
"Force" shut down action added

mareklibra:
With this patch, the user can shut down a VM in either "polite" or "force" way.

Action confirmation dialog changed to bootstraps' popover.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/16

confirmation

### DIFF
--- a/UserPortal.iml
+++ b/UserPortal.iml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module type="WEB_MODULE" version="4">
-  <component name="NewModuleRootManager" inherit-compiler-output="true">
-    <exclude-output />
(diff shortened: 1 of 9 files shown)


mareklibra:
@jelkosz, @bond95, @matobet What do you think about using such a component for user confirmation?

The component can be either two-state Yes-No or three-state Yes-No-Extra.

CSS needs to be polished.

yesnopopover

mykaul:
Cancel should be in the middle to prevent accidental click on force, unless
there's confirmation.

mareklibra:
Done - "Cancel" button is rendered in the middle

mareklibra:
Text color is inherited.
New confirmation component is used for: Shutdown, Reset, Suspend

PR #161, opened 2017-04-24
Add vm dialog

mareklibra:
Replaces PR https://github.com/oVirt/ovirt-web-ui/pull/98.

On top of https://github.com/oVirt/ovirt-web-ui/pull/98 commit:
- Extensive refactoring
- Functionality fixes
- Fixed syntax
- Unused code pruned

To avoid additional merge conflicts, following is left for follow-ups:

[ ] The "Description" property
[ ] VM icons setting
[ ] Refactoring of remaining reducers
[ ] Use existing UserMessages instead of specific vmDialogReducer
[ ] update yarn.lock
[ ] Add [help](https://www.patternfly.org/pattern-library/forms-and-controls/field-level-help)
[ ] Render template version
[ ] Fix form update after template change

@jelkosz , @matobet , @jniederm , @mkrajnak, please check especially functionality to have this merged asap (today?). Any non-critical findings will be pushed within follow-ups.


mykaul:
Is there an up-to-date screenshot, so we can review it?

mykaul:
1. Shouldn't 'Name' be the first item to be filled in?
2. 'CPU' -> 'No. of CPUs' ?
3. Should some values (such as CPU and perhaps Memory) be also selectable via up/down arrows?
4. Submit and Close are too close to each other, so accidental click might cause the wrong action.
5. There is no Help whatsoever to any of the fields (or is there bubble help?). For example, does the user know what 'Cluster' mean? (does he even care?).
6. Description/comment field?
7. Do we wish to add a 'Submit and Run' button? This will launch the VM immediately after creation.

mareklibra:
@mykaul , thanks for comments.
1,2 - done
3 - not visible on the screenshot but there already are Up/Down buttons
4 - done and moved to the right
5 - good point, will be done in a follow-up
6 - follow-up (see PR description)
7 - would be nice but unreliable when implemented on FE side. Let's add support to oVirt API first.

Issue #33, opened 2016-10-20
I18n and Localization


mareklibra:
How will be the application translated?
Consider REST API error messages.

jelkosz:
The API message translations require:
- https://github.com/oVirt/ovirt-web-ui/issues/110
- https://bugzilla.redhat.com/show_bug.cgi?id=1435704

mareklibra:
Let's contemplate options for i18n/l10n.

In general, Zanata seems to be a good choice for the actual translation, so let's consider all bellow in its context.

There are plenty of choices how to i18n a JS application.
When considering the viable ones, let's focus on:

- limited amount of boilerplate code
- nothing over-sophisticated, easy to understand by newbies (just JS developers?)
- once set, low maintenance cost (scriptable)
- Zanata-ready format or just simple transformation required
- limited or no use of id-referencing, the English defaults shall be used throughout the code
- simple process of merging translations over time

Nice to have:
- format allows simple and reliable reuse of existing translations from ovirt-engine
- locale-specific data loaded on demand

So far, following approaches meet the criteria:
- "Home-made" text extraction
  - wrap all strings to something like __("My string to translate")
  - use 'sed' to extract
  - generate whatever format is needed
  - Pros:
    - full control over format and merging process
    - works fine for simple (basic) needs (do we really need more?)
    - probably lowest amount of boilerplate code
    - still supports formatted messages
    - simple to implement, simple to understand and maintain
  - Cons:
    - complicated for multi-line strings
    - a lot from l10n would not be supported (eg. time or plurals)

- gettext / Jed
  - Pros:
    - standard
    - Zanata-ready format
    - already used in cockpit-ovirt or Cockpit
  - Cons:
    - complicated to setup
    - JS support is an add-on
    - no matter gettext is standard, its use is not clear for everyone

- react-intl
  - https://medium.freecodecamp.com/internationalization-in-react-7264738274a0
  - Pros:
    - clean JS approach
    - based on JS standards
  - Cons:
    - quite a lot of boilerplate code
    - no experience so far
    - resulting format needs to be transformed to/from Zanata

I slightly prefer the react-intl. Any real-project experience?

Please comment/suggest. 
@jelkosz , @vojtechszocs , @gregsheremeta , @jniederm , @matobet , @bond95

jelkosz:
Even though the first option is tempting it would get a bit more complicated to implement things like:
___("Maximum allowed memory is {} MiB", someValue)
which we will need.
[...]

PR #163, opened 2017-04-26
All VmDialog fields enriched for help

mareklibra:
vmdialogfieldhelp

### DIFF
--- a/src/components/VmDialog/index.js
+++ b/src/components/VmDialog/index.js
@@ -5,11 +5,13 @@ import { logDebug } from '../../helpers'
 
 import style from './style.css'
 
-import LabeledSelect from './../LabeledSelect'
-import LabeledTextField from './../LabeledTextField'
-import DetailContainer from './../DetailContainer'
-import ErrorAlert from './../ErrorAlert'
-import Selectors from './../../selectors'
+import LabeledSelect from '../LabeledSelect'
+import LabeledTextField from '../LabeledTextField'
+import DetailContainer from '../DetailContainer'
+import ErrorAlert from '../ErrorAlert'
+import FieldHelp from '../FieldHelp'
+
+import Selectors from '../../selectors'
 import {
   closeDialog,
   requestCloseDialogConfirmation,
@@ -330,6 +332,7 @@ class VmDialog extends React.Component {
             label='Name'
             placeholder='Enter VM Name'
             value={this.state.name}
+            fieldHelp={<FieldHelp title='Name' content='Unique name of the virtual machine.' />}
             onChange={this.onChangeVmName} />
 
           <LabeledTextField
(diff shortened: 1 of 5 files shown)


mykaul:
When will we be taking the strings from an external resources, for easier localization?

mareklibra:
i18n is tracked by https://github.com/oVirt/ovirt-web-ui/issues/33
I would like to have a look at it soon.

I'm not sure about technical details for it yet, so far I prefer gettext as done in Cockpit but I need to look around before making that decision.
If feasible, we might share (some?) resources from ovirt-engine somehow.

src/components/VmDialog/index.js
```diff
@@ -347,6 +351,7 @@ class VmDialog extends React.Component {
             onChange={this.onChangeCluster}
             value={cluster ? cluster.get('id') : ''}
             data={sortedClusters}
+            fieldHelp={<FieldHelp title='Cluster' content='Group of servers the virtual machine can be running on.' />}
```
jelkosz: I would use the term "hosts" instead of "servers" to be aligned with the webadmin terminology

src/components/VmDialog/index.js
```diff
@@ -356,6 +361,7 @@ class VmDialog extends React.Component {
             onChange={this.onChangeTemplate}
             value={template ? template.get('id') : ''}
             data={sortedTemplates}
+            fieldHelp={<FieldHelp title='Template' content='Base configuration the virtual machine will be created from. Please customize if needed.' />}
```
jelkosz: I would add that also the disks are there. Something like:
"Contains the configuration and disks which will be used to create this virtual machine. Please customize as needed"

src/components/VmDialog/index.js
```diff
@@ -365,6 +371,7 @@ class VmDialog extends React.Component {
             onChange={this.onChangeOperatingSystem}
             value={os ? os.get('id') : ''}
             data={sortedOSs}
+            fieldHelp={<FieldHelp title='Operating System' content='Type of operating system installed in the virtual machine.' />}
```
jelkosz: I think it should be "installed on" not "in". I would anyway simplyfy it to:
Operating system installed on the virtual machine.

src/components/VmDialog/index.js
```diff
@@ -374,6 +381,7 @@ class VmDialog extends React.Component {
             placeholder='VM Memory'
             value={this.state.memory / 1024 / 1024}
             onChange={this.onChangeVmMemory}
+            fieldHelp={<FieldHelp title='Memory' content='Total memory the virtual machine will be created with. In megabytes.' />}
```
jelkosz: I'd unify with the below CPU terminology. Either both should say that "equipped with" or both "created with".

mareklibra:
Texts updated according to comments

PR #165, opened 2017-04-26
VmDialog reuses common UserMessages for error handling

mareklibra:


src/components/VmDialog/index.js
```diff
@@ -347,6 +373,7 @@ class VmDialog extends React.Component {
             onChange={this.onChangeCluster}
             value={cluster ? cluster.get('id') : ''}
             data={sortedClusters}
+            fieldHelp={<FieldHelp title='Cluster' content='Group of servers the virtual machine can be running on.' />}
```
jelkosz: Maybe use "hosts" instead of "servers" to be aligned with the webadmin's terminology

mareklibra: ok, I will change that in https://github.com/oVirt/ovirt-web-ui/pull/163

PR #187, opened 2017-05-15
Bump 0.2.1 version

mareklibra:
React and other libraries updated to the latest version


mykaul:
Where's the changelog? Should it be part of the version bump perhaps?

mareklibra:
@mykaul , The change log is updated within the .spec file and here: https://github.com/oVirt/ovirt-web-ui/releases

Issue #156, opened 2017-04-19
reconsider the door effect


jelkosz:
When a VM cube is picked, a door effect slides the VMs to the left side and the right side of the screen is filled with the VM details.

This behaviour is a more eye-candy implementation of the original basic user portal which behaves similarly without the door effect.

The biggest advantage of this UI is that it makes use of a large part of the screen (about a half of it) and scales well with the size of the screen. 

The disadvantage is that the design is quite controversial and also not in line with the patternfly which is used by other UIs of the oVirt ecosystem (like cockpit plugin or webadmin).

I would propose to use a similar approach as the trello is using when opening a card. Don't slide anything enywhere, open a large (patternfly) dialog window in the centre of the screen which will be proportional to the size of the screen (e.g. also will take about half of it and will scale with the size of the screen).

It will have all the advantages of the original design but will be more in-line with the patternfly (meaning the rest of the ecosystem) and will be less controversial.

@mareklibra @jniederm @matobet @bond95 @michalskrivanek what you guys think?

---------
Fixed by:

 - https://github.com/oVirt/ovirt-web-ui/pull/197

Depends on:

 - #200 
 - #196

matobet:
Big :+1: to a modal, patternfly-aligned dialog. The door effect is just too .... controversial.

mareklibra:
I prefer the dialog style over the sliding effect more.
Anyway, let's put this on hold for a while. Maybe we find something even better.

jniederm:
It seems that web-ui proved itself a viable product and it's gaining more and more features. I find this stage of development the right moment to create a vision of intended "final" state. Maybe we will want the project to cover more than just functionality of current basic user portal. In that case it may be handy to think through the "complete" version is supposed to look like, what kind of data we will want to show and edit. How to visually represent content of current subtabs or navigation to listings of other entities (e.g. templates or perhaps hosts).

Such vision can help us to prevent current problems of webadmin like:
* it is difficult to search for editing widget or particular name - there is no in-app search and standard "Ctrl-F" doesn't work because of side-tabs
* it is quite tedious, sometimes even impossible to find out what is value of running VM configuration and what is value of next-run configuration
* if backend validation fails, changes of whole VM dialog are lost
* some operations a represented in rather unfortunate way, for example memory hot plug
* Quota tab - the easter-egg

Regarding potential door replacements: I find following ways worth considering:
* modal dialog (similar to current webadmin)
[...]

PR #197, opened 2017-05-23
Remove door effect, replace it by react-router

bond95:
Partially fixes: https://github.com/oVirt/ovirt-web-ui/issues/156

@jelkosz @mareklibra @matobet 

------------------------
Must-Have (see https://github.com/oVirt/ovirt-web-ui/pull/197#pullrequestreview-40267230 ):

[ ] all code review comments addressed
[ ] Error message for expired/invalidated token is rendered badly - hiden by page header. Look for the TokenExpired component.
[ ] When leaving Add/Edit VM screen with "unsaved" changes, confirmation is no more shown (but should be).
[ ] When on VM Detail and Remove VM is confirmed, the flow shall lead to VmsList.
[ ] Breadcrumb: rename for Edit VM to:  Virtual Machines >> [VM Name] >> Edit
       (so no need to repeat the VM name after "Edit" text)
[ ] Change flow: Add new VM -> Create VM button -> NEW: VM detail (instead of standing on Add VM view)
[ ] Similar change for Update VM: successful submit leads to read-only VM Detail
[ ] Rearrange VM Detail to:
  - Description first and spreading over whole screen width
  - Left column:
    - State
    - Cluster
    - Template
    - Operating System
    - Memory
    - CPU
    - Address
  - Right column
    - Console
    - Disks
  - **Question ??**: Do we still need show/hide for Console and Disks? What about making "Disks" a link causing more details rendered on the right side of the page. Multiple such "subviews" can be rendered simultaneously.

[ ] ...

### DIFF
--- a/src/components/VmDialog/index.js
+++ b/src/components/VmDialog/index.js
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 
 import { connect } from 'react-redux'
+import { Link, Redirect, Prompt } from 'react-router-dom'
 
 import { logDebug, generateUnique } from '../../helpers'
 
@@ -16,10 +17,13 @@ import FieldHelp from '../FieldHelp'
 import Selectors from '../../selectors'
 import {
   closeDialog,
-  requestCloseDialogConfirmation,
-  createVm,
-  editVm,
 } from '../../actions/index'
+
+import {
+  editVm,
+  createVm,
+  setSavedVm,
+} from './actions'
 
 function sortedBy (immutableCollection, sortBy) { // TODO: move to helpers
   return immutableCollection.sort(
@@ -46,6 +50,8 @@ class VmDialog extends React.Component {
       clusterId: undefined,
       templateId: undefined,
       osId: undefined,
+      saved: false,
+      isChanged: false,
     }
 
     this.closeDialog = this.closeDialog.bind(this)
@@ -67,6 +73,10 @@ class VmDialog extends React.Component {
     this.onChangeVmDescription = this.onChangeVmDescription.bind(this)
     this.onChangeVmMemory = this.onChangeVmMemory.bind(this)
     this.onChangeVmCpu = this.onChangeVmCpu.bind(this)
+  }
+
+  componentWillMount () {
+    this.props.onInit()
   }
 
   componentDidMount () {
@@ -101,6 +111,9 @@ class VmDialog extends React.Component {
     this.props.vm
       ? this.props.updateVm(this.composeVm(), actionUniqueId)
       : this.props.addVm(this.composeVm(), actionUniqueId)
+    this.setState({
+      saved: true,
+    })
   }
 
   getLatestUserMessage () {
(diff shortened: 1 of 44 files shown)


ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

ovirt-infra:

mareklibra:
@bond95 , please
- remove yarn.lock from the commit
- rebase

Since this is quite complex change, I would like to try locally ...

bond95:
@mareklibra please review

mareklibra:
When a new VM is created (Create VM button clicked), the flow shall land on VM Detail. With recent implementation, the Create VM dialog keeps to be rendered which is confusing.

src/sagas.js
```diff
@@ -485,8 +483,8 @@ function* getConsoleVm (action) {
  * VmDetail is to be rendered.
  */
 export function* selectVmDetail (action) {
-  yield put(setVmDetailToShow({ vmId: action.payload.vmId }))
   yield fetchSingleVm(getSingleVm({ vmId: action.payload.vmId })) // async data refresh
+  yield put(loadInProgress({ value: false }))
```
bond95: No, and there is a few reasons:
1. Function fetchAllVms using for a long time, before this commit, without paired with code, because this function is using only in first data loading for showing load component (loadInProgress value has true value by default)
2. Cause of we now has more than one page we shall this use for showing loading component

mareklibra: Please consider in context of https://github.com/oVirt/ovirt-web-ui/pull/198.

The loadInProgress is can be set even later on. Recently it's set to true:
- during app initialization (full data load is expected ...)
- when full reload is requested (the Refresh button)
- might be decided later to set it whenever polling occurs (recently is the user intentionally not informed about that)

For completeness, any VM action (or so) sets Action In Progress causing the same visual "Loading ..." component to render.

@bond95 , please review the use of "components/LoadingData" within this PR, there's no need to render it on every subscreen, I think.

bond95: According to what I see code, loadInProgress set only two times, first in begining when with reducer, and second time in login sagas, that's all, and in code it present only two times, in login sagas and fetchAllVms sagas, so I don't see any reasons why it can't be used for showing Loading component when page is loading.
There is react-router, so when page is reloaded, all previous data will be lost, and web-ui need to load this data again for certain page, until it is not done I think better to show Loading component for user.

src/components/VmDialog/index.js
```diff
@@ -149,31 +162,27 @@ class VmDialog extends React.Component {
...
   }
 
   onChangeVmMemory (event) {
-    this.onIntegerChanged({ stateProp: 'memory', value: event.target.value, factor: 1024 * 1024 })
+    this.onIntegerChanged({ stateProp: 'memory', value: event.target.value, factor: 1024 * 1024, isChanged: true })
   }
 
   onChangeVmCpu (event) {
-    this.onIntegerChanged({ stateProp: 'cpus', value: event.target.value })
+    this.onIntegerChanged({ stateProp: 'cpus', value: event.target.value, isChanged: true })
```
mareklibra: I don't see this is used somewhere in the onIntergerChanged

mareklibra:
Generally looks good. Some minor UI issues remains, but this can be used.

jelkosz:
yeah! :tractor:

Issue #153, opened 2017-04-18
add a "communicating with server" or similar


jelkosz:
Sometimes the server call can take a long time to process (for example when opening a console) and it is not obvious if the server call did not finish yet or something failed or what happened.

It would be nice to have something similar than the "Loading..." in gmail to indicate that now the FE is actually waiting for a response from server - it did not fail yet and did not finish yet.

mareklibra:
When the app is starting, there's spinner with following message:

    Data are being loaded ...

Later on, when an action is issued, it's icon becomes disabled until it's finished (means response from server is received, resp.).

What about replacing the action's icon by a spinner in that case?

So the 'disabled" icon state would be reserved just for situation where "something is not met" to run the action (ie. the VM is not running to shut it down).

I was further thinking about a spinner on the page header or even the "Loading ..." patternfly allert.

mykaul:
It should be at least 'Data is being loaded'.

jelkosz:
> What about replacing the action's icon by a spinner in that case?

Im afraid it would be pretty hard to get this right in a way that it actually looks good.

> I was further thinking about a spinner on the page header or even the "Loading ..." patternfly allert.

I think the simple "Loading ..." on top of the screen is enough to start with.

PR #198, opened 2017-05-25
Render "Loading ..." message when communicating with server

mareklibra:
The user is informed about communication with server when
 - data is being initially loaded
 - waiting for an action
 - load after Refresh button
 - refreshing data when VM detail is opened

Regular polling does not lead to rendering this "Loading ..." message.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/153


mareklibra:
@jelkosz , better than screenshots :-)

docker run --rm -it -e ENGINE_URL=https://engine.local/ovirt-engine --net=host -p 3000:3000 mareklibra/ovirt-web-ui-devel:pullrequest_198

jelkosz:
@mareklibra works nice and I feel much more comfortable with the application now.

But Im not a huge fan of the visual part. Some notes there:
- it is a bit too tall - there is a lot of space above and below the text. There should be some, but this is a bit too much.
- the red border around it seems a bit too aggressive. Maybe an orange? Similar to the unknown OS icon's colour?

@jniederm could you please also check?

jniederm:
* Color: I understand that the popup is based on [patternfly alerts](https://www.patternfly.org/pattern-library/widgets/#alerts) and in general I'd also use class `alert-warning`. However it really feels a bit too red, especially compared to gmail loading info. Maybe we can try `alert-info`. 
gmail loading
* Maybe we can just put a [spinner](http://www.patternfly.org/pattern-library/widgets/#spinner) as a next block to the black top bar, optionally with a label
* If it is supposed to appear also for some long-lasting actions, we can consider label "Working ..." which is more general and covers actions better.

mareklibra:
I don't like alert-info much. I think alert-warning is closer what we are looking for ...
So I adjusted it a little bit, it looks less aggressive now.

For comparison, there's even the spinner which _replaces_ the "Refresh button" if needed.

- Let's choose one of them:
01_bothspinnerandtext


- Spinner tooltip:
02_withtooltip

- No load in progress:
03_noload

mareklibra:
After offline discussion: let's go with the text.

I'm removing the docker image referenced above.

This is the result:
result

PR #199, opened 2017-05-25
Remove ovirt standard CI

mareklibra:
Travis proved as a sufficient solution for needs of ovirt-web-ui,
so let's go with in and don't split/dupplicate maintenance effort.

Depends on:
- https://gerrit.ovirt.org/#/c/77342/1
- https://ovirt-jira.atlassian.net/browse/OVIRT-1408


mykaul:
I don't see a good reason to do it, for unit-testing kind of CI - sure. But you are missing integration tests.

michalskrivanek:
AFAICT it's not missing anything. The oVirt consumption is solved separately. There are no integration tests today, and once we do have them I still do not see any problem, they won't run on a per-patch basis but rather on a CI-verified candidate build or nightly or whatever...but that's still separate from this automation/ directory content

mykaul:
If possible, I'd be very happy with integration tests on every patch. 
What's the status of your Selenium tests to ovirt-system-tests?

michalskrivanek:
I don't think anyone is against that (though I'm yet to see a successful GUI testing project), and the "standard CI" dir here has nothing to do with it.
Travis can be configured to create the artifacts and trigger OST, or from the other side a OST for API and/or ovirt-engine can pull the latest content(even trigger a build, as the lago setup takes way longer than ovirt-web-ui build)

Issue #175, opened 2017-04-26
Max memory and min memory need to be edited when memory is edited


jniederm:
... to keep inequality: max memory >= memory => min memory (a.k.a. guaranteed memory).

![screenshot](https://cloud.githubusercontent.com/assets/5278760/25451026/168e4d8e-2ac1-11e7-8e7a-d44e78aa05c4.png)

PR #235, opened 2017-06-19
Change 'max' and 'guaranteed' memory when editing VM's memory

mareklibra:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/175

### DIFF
--- a/src/components/VmDialog/index.js
+++ b/src/components/VmDialog/index.js
@@ -27,6 +27,8 @@ import {
   setSavedVm,
 } from './actions'
 
+import { MAX_VM_MEMORY_FACTOR } from '../../constants/index'
+
 function sortedBy (immutableCollection, sortBy) { // TODO: move to helpers
   return immutableCollection.sort(
     (a, b) => a.get(sortBy).localeCompare(b.get(sortBy))
@@ -60,6 +62,7 @@ class VmDialog extends React.Component {
     this.submitHandler = this.submitHandler.bind(this)
     this.initDefaults = this.initDefaults.bind(this)
     this.onIntegerChanged = this.onIntegerChanged.bind(this)
+    this.getMemoryPolicy = this.getMemoryPolicy.bind(this)
 
     this.getCluster = this.getCluster.bind(this)
     this.getTemplate = this.getTemplate.bind(this)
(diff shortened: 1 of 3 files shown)


mareklibra:
@jelkosz , can you please have a look at the computation?

Recently user can't modify max and guaranteed memory in web-ui, it's a planed feature.
From user perspective, this seems to be advanced VM configuration and we need to be careful when/how we add it into the web-ui. So not within this PR.

src/components/VmDialog/index.js
```diff
@@ -128,6 +131,21 @@ class VmDialog extends React.Component {
     return last ? last.message : undefined
   }
 
+  getMemoryPolicy () {
+    const cluster = this.getCluster()
+    console.log('getMemoryPolocy, overCommit: ', cluster.getIn(['memoryPolicy', 'overCommitPercent']))
```
jniederm: Is this intentional? Do we want such log?

src/components/VmDialog/index.js
```diff
@@ -128,6 +131,21 @@ class VmDialog extends React.Component {
...
+    let guaranteed = (cluster && cluster.getIn(['memoryPolicy', 'overCommitPercent']))
+      ? (this.state.memory * (100 / cluster.getIn(['memoryPolicy', 'overCommitPercent']))) : this.state.memory
+
+    const memoryPolicy = {
+      'max': this.state.memory * MAX_VM_MEMORY_FACTOR,
+      'guaranteed': Math.round(guaranteed),
+    }
+    logDebug('getMemoryPolicy() resulting memory_policy: ', memoryPolicy)
```
jniederm: similar

src/components/VmDialog/index.js
mareklibra: Thanks, leftover

src/components/VmDialog/index.js
mareklibra: This was intentional for DEBUG level

src/components/VmDialog/index.js
```diff
@@ -128,6 +131,20 @@ class VmDialog extends React.Component {
     return last ? last.message : undefined
   }
 
+  getMemoryPolicy () {
+    const cluster = this.getCluster()
+    const overCommitPercent = cluster && cluster.getIn(['memoryPolicy', 'overCommitPercent'])
+    let guaranteed = overCommitPercent ? (this.state.memory * (100 / overCommitPercent)) : this.state.memory
+
+    const memoryPolicy = {
+      'max': this.state.memory * MAX_VM_MEMORY_FACTOR,
```
jelkosz: is this only for next run or also hotplug? If it is for next_run only than it is ok, if it is about hotplug than the "max" should stay as is.

src/components/VmDialog/index.js
```diff
@@ -128,6 +131,20 @@ class VmDialog extends React.Component {
     return last ? last.message : undefined
   }
 
+  getMemoryPolicy () {
+    const cluster = this.getCluster()
+    const overCommitPercent = cluster && cluster.getIn(['memoryPolicy', 'overCommitPercent'])
+    let guaranteed = overCommitPercent ? (this.state.memory * (100 / overCommitPercent)) : this.state.memory
```
jelkosz: this is ok.

src/components/VmDialog/index.js
mareklibra: Recently it's for next_run only

Issue #117, opened 2017-04-06
add support for "runs on host" for VM detail


jelkosz:
This could be an integration point between the cockpit-machines(-ovirt-provider) and ovirt-web-ui. 

The flow could be like this:
- When opening the VM detail, checking if the user can see also the host on which the VM is running on (not sure how, maybe need to check the permissions of the user, maybe could be understood from the REST response of the GET /vms/<vm id>.
- If the user does have permission to see it, add a new field "runs on host" to the VM detail
- next to the value, there should be a link or button saying "go to host" which would redirect to that host's cockpit

mareklibra:
For longer host name:
![01_longername](https://user-images.githubusercontent.com/17194943/27470169-e9cabe54-57f2-11e7-90b0-0908e66d8da8.png)

and narrow screen:
![02_narrowscreen](https://user-images.githubusercontent.com/17194943/27470170-e9cd6d70-57f2-11e7-85a6-6e08ce16361c.png)

What about rendering the `(on [HOST_NAME])` separately as a first entry of the second column (right above the `Consoles`).
Sort of `Runs on:      [LINK_TO_HOST]`

Issue #247, opened 2017-06-21
VM in pause state doesn't have any icon with tooltip


leistnerova:
also in detail is only 'paused' with no icon

![screenshot from 2017-06-21 12-46-21](https://user-images.githubusercontent.com/29575480/27380508-b30a4912-567f-11e7-9699-5d037c1d5345.png)

michalskrivanek:
can we also display the pause reason in a tooltip? IO/user/no space

mareklibra:
@michalskrivanek , within https://github.com/oVirt/ovirt-web-ui/pull/250 the `status_detail` is added, see screenshot there.

PR #250, opened 2017-06-22
Add "Runs on host" info

mareklibra:
Admin user can link-through to Cockpit running on a host with particular VM.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/117


mareklibra:
See the hyperlink in the `status` filed:

runonhost

mareklibra:
For longer host name:
01_longername

and narrow screen:
02_narrowscreen

What about rendering the `(on [HOST_NAME])` separately as a first entry of the second column (right above the `Consoles`).
Sort of `Runs on:      [LINK_TO_HOST]`

mareklibra:
Anyway, we shall avoid additional field

mareklibra:
What about this? @jelkosz , @jniederm 
How to rephrase?
hostconsole

mareklibra:
For non-admin user:
nonadmin

jelkosz:
@mareklibra 
 - maybe instead of "On" I would put "Runs on host: ..."
 - the hostname should look like you can click it so you know it you can click it

mareklibra:
runsonhost

jelkosz:
@mareklibra 👍

jniederm:
I have doubts about the tooltip solution. As a new user to webui, how am I supposed to find it? <kbd>Ctrl+F</kbd> couldn't work. I'd prefer regular additional name-value widget.

mareklibra:
Let's try it this way. We can change to name-value solution or something even better later.
The reasons for preferring the per-click context information now are:
- it's a pattern in web-ui to use blue-style of `<a href>` to route and gray-underlined-style to denote local context-aware additional information
- so the user knows the `[status]` can be clicked since the notation is used on other places as well
- I don't think the user will search (like Ctrl+F) for host name on the VmDetail page. It's for a single VM. 
- We might add new `Hosts` tab listing all VMs defined on particular host (+ more host info), if needed. Would be linked from the VmPage as well.
- This cockpit-link is rendered for admin users only (not the "typical" kind of users)
- context information of the `[status]` is used for other kind of information, like `status reason` - see later commit in this PR. So it's not single-purpose.
- the page shall provide sufficient amount of information but not too much to keep it understandable at the first glance. This is sort of "advanced" information and adding extra field for it is overwhelming in my opinion.
- Maybe we will be forced to split the recent layout of VmDetail page to "sort of subtabs or so" in the future. Anyway, I would like to avoid/postpone it as much as possible.

The web-ui is focused on standard, non-admin users. We have to keep the presented information at minimal useful level. For more details, there's webadmin.

mareklibra:
Enriched for status reason (requested in https://github.com/oVirt/ovirt-web-ui/issues/247):
statusreason

Issue #239, opened 2017-06-20
visualize the VM Type


jelkosz:
Similar to webadmin and userportal, also web-ui should show the desktop/server icon somewhere on the VM. While doing this, it would be good to take into account also the new high performance VM.

PR #253, opened 2017-06-22
Vizualize VmType

mareklibra:
Label denoting VM's type is displayed on VmDetail.
Handles `Desktop, Server, HighPerformance`.
    
Depends on not yet merged https://gerrit.ovirt.org/#/c/77961
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/239

### DIFF
--- a/src/components/VmDetail/LastMessage.js
+++ b/src/components/VmDetail/LastMessage.js
@@ -1,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Time from '../Time'
+
+const LastMessage = ({ vmId, userMessages }) => {
+  const vmMessages = userMessages.get('records')
+    .filter(msg => (msg.failedAction && msg.failedAction.payload && msg.failedAction.payload.vmId === vmId))
+    .sort((msg1, msg2) => (msg1.time - msg2.time))
+
+  const lastMessage = vmMessages.last()
+
+  if (!lastMessage) {
+    return null
+  }
+
+  return (
+    <span>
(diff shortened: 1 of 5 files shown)


mareklibra:
01_vmtype
02_vmtype

mareklibra:
I think concept of `labels` is fine for this purpose.
@jniederm , would you change styling of them? Colors?

jniederm:
I'd visualize it just like other regular properties (for example State or Description) because

* there already are two types of label in the engine (standard and SLA ones) and a time may come when we will need to visualize these
* a time may come when we will need to change or set "Optimized for" property in edit/new pages. In such situation it will be natural to visually just replace a text by a dropdown.

mareklibra:
Right, this makes sense.

mareklibra:
01_vmtype
02_vmtype

mareklibra:
@jniederm , what do you think?

mareklibra:
Cc @sgratch

sgratch:
Looks very good to me. 
Few small comments to consider:
1. desktop->Desktop (capitalize the first letter).
2. Consider locating this property under/after the 'Template' property, since changing the Template may change the 'Optimized For' value since it is inherited from it (the properties displaying order has some logic I guess..)
3. What about displaying this property also for Templates now or in the future?

mareklibra:
@sgratch , thanks
1. Done
2. Done
3. Good point, we will think about it when adding Templates view

mareklibra:
01_vmtype

mareklibra:
Merging since agreed otherwise

Issue #251, opened 2017-06-22
Handling of VM urls


jniederm:
VM list has url `/`, a vm detail `/vm/{vmId}`. When user navigates to a VM detail and then refreshes the page, VM list is loaded (`/`). Direct access to `/vm/{vmId}` is also redirected to `/`.

It would be nice if url `/vm/{vmId}` would be able to show a VM detail directly.

Version: commit 8516302

PR #254, opened 2017-06-23
Fix falsely-positive check about missing VM

mareklibra:
This gives more time to load data before decision that
particular VM is missing when `.../vm/{id}` URL is entered.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/251


mareklibra:
@jniederm , this shall be working. Can you please verify on your environment?

Issue #23, opened 2016-10-12
Test on large VM count


mareklibra:
The application lists VMs filtered on the current user.
Expected count of VMs is <100.

Anyway, the application should perform well even for 1000 VMs.
In addition, related entities (like hosts or templates) are retrieved which count might be even >1000.

Within this issue, sort of performance test should be executed and potential bottlenecks resolved.

mareklibra:
Paging will need to be added for REST API calls.

jelkosz:
requires some investigation on the ovirt side how it can be paginated and if the current support is sufficient or we need to open a bug on ovirt side.

jelkosz:
ok, after a bit of investigation the ovirt API actually does support pagination using the search query. 
Looks like this: /ovirt-engine/api/vms/;max=100?search=SORTBY NAME ASC  page 2

This is supported since 3.6 also for Filter: true, so we can use it.

PR #262, opened 2017-06-29
Add infinite scroll

bond95:
@mareklibra please review
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/23


mareklibra:
please rebase, this needs to be properly verified

bond95:
@mareklibra made rebase, please rebase

src/reducers/vms.js
```diff
@@ -6,13 +6,18 @@ const initialState = Immutable.fromJS({
   vms: {},
   pools: {},
   loadInProgress: true,
+  page: 1,
+  changed: true,
```
mareklibra: Please rename to something more descriptive, like `areAllPagesLoaded` .

src/sagas.js
```diff
@@ -304,15 +316,39 @@ function* refreshData (action) {
     console.log('refreshData(): not quiet')
     yield put(loadInProgress({ value: true }))
   }
-  yield put(getAllVms({ shallowFetch: !!action.payload.shallowFetch }))
-  yield put(getAllPools())
+  yield put(getVmsByCount({ count: action.payload.page * AppConfiguration.pageLimit, shallowFetch: !!action.payload.shallowFetch }))
```
mareklibra: How does this behave if all pages are already loaded and new VM (or multiple) are added on the server by another user?

src/index.js
```diff
@@ -102,8 +102,17 @@ function start () {
     logError('Missing SSO Token!')
   }
 
-  // start cron-jobs
-  store.dispatch(schedulerOneMinute())
+  setInterval(() => {
```
mareklibra: I would prefer to avoid `setInterval()` and keep the cron-like approach via saga.
Is there any specific need for this change?

src/sagas.js
bond95: If user add small count of VMs it just fetch VMs and add new to list, if count will be bigger it turn on infinite scroll feature and start to fetching next pages until all VMs will be loaded.

src/index.js
bond95: I gonna try to rewrite it with `delay` generator

bond95:
@mareklibra fixed app according to your comments, please review

Issue #260, opened 2017-06-29
strange behaviour with setting console Connect automatically


leistnerova:
It somehow remembers for which VMs I had set the connection and mess them up. See video.
I'm using Firefox 53.0.3 

[Screencast from 06-29-2017 02:48:28 PM.webm.zip](https://github.com/oVirt/ovirt-web-ui/files/1112083/Screencast.from.06-29-2017.02.48.28.PM.webm.zip)

mykaul:
Can you please share the issue on the oVirt users mailing list? (users@ovirt.org) ?

leistnerova:
done, waiting for approval

mareklibra:
Fixed by https://github.com/oVirt/ovirt-web-ui/issues/265

PR #265, opened 2017-06-30
Fix bug in OptionsManager, and make otions update when change VM

bond95:
@mareklibra Please review https://github.com/oVirt/ovirt-web-ui/issues/260


mareklibra:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/260

Issue #168, opened 2017-04-26
Set icon after Add/Edit VM


mareklibra:
As a follow-up for #17 , set/change VM icon after add/edit VM to correspond the chosen OS.

PR #268, opened 2017-07-07
Add check vm icon change if VM OS changed

bond95:
@mareklibra Please review https://github.com/oVirt/ovirt-web-ui/issues/168

### DIFF
--- a/src/components/VmDetail/index.js
+++ b/src/components/VmDetail/index.js
@@ -156,10 +156,15 @@ class VmDetail extends Component {
       options,
       pool,
       onRDP,
+      operatingSystems,
     } = this.props
 
     const name = isPool ? pool.get('name') : vm.get('name')
-    const iconId = vm.getIn(['icons', 'small', 'id'])
+    let iconId = vm.getIn(['icons', 'small', 'id'])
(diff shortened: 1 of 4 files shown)


src/components/VmDetail/index.js
```diff
@@ -156,10 +156,15 @@ class VmDetail extends Component {
...
       onRDP,
+      operatingSystems,
     } = this.props
 
     const name = isPool ? pool.get('name') : vm.get('name')
-    const iconId = vm.getIn(['icons', 'small', 'id'])
+    let iconId = vm.getIn(['icons', 'small', 'id'])
+    const vmOs = operatingSystems.get('operatingSystems').find((v, k) => v.get('name') === vm.getIn(['os', 'type']))
+    if (vmOs) {
+      iconId = vmOs.getIn(['icons', 'large', 'id'])
```
mareklibra: What if particular icon is not loaded?
Only the required icons are pre-fetched (means those already associated to VMs).
I suggest to keep logic as it is plus dispatch action to load icon of given ids (both small and large).

The saga performing such an action will do HTTP GET if particular iconId is missing or do nothing if already present.
Subsequent update of state enforces re-rendering.

bond95:
@mareklibra added fetching icons for OS too, now all icons will be available, please review

PR #270, opened 2017-07-10
gitignore Visual Studio Code config directory

jniederm:


mareklibra:
@jniederm , it would be very nice if you can show me Visual Studio's features related to our needs.

PR #280, opened 2017-07-12
react-infinite-scroller added to yarn.lock

jniederm:
The dependency was introduced by commit d4e373954.

From now on I suggest following handling of dependency changes:

* if dependency is changed in `package.json` then include also appropriate change in `yarn.lock` in the commit.
* after fetching a remote dependency change (in both `package.json` and `yarn.lock`), run `yarn install --check-files` which changes content of `node_modueles` directory accordingly.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4487,6 +4487,12 @@ react-dom@15.5.4:
     object-assign "^4.1.0"
     prop-types "~15.5.7"
 
+react-infinite-scroller@1.0.12:
+  version "1.0.12"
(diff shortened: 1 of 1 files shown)


mareklibra:
Thanks for this fix.

Anyway, committing `yarn.lock`together with other changes caused unnecessary merge conflicts, so considering size of this project, I think it's easier to push subsequent commit once a PR modifying `package.json` is merged. 

This way, we can be sure the resulting yarn.lock comes pairs with the package.json and is not just a product of text-file merging.

I was further thinking about splitting similar PRs into two separate ones - one for functionality, another for package.json/yarn.lock. But again, considering recent size of this project, this brings no value but makes things harder to understand (change of package.json is enforced by functionality).

PR #281, opened 2017-07-13
About dialog typo fix

jniederm:


src/components/About.js
```diff
@@ -38,6 +38,7 @@ const AboutDialog = ({ oVirtApiVersion }) => {
 
             <div className='trademark-pf'>
               Relesed under <a href='https://github.com/oVirt/ovirt-web-ui/blob/master/LICENSE' target='_blank'>Apache License 2.0</a>
```
mareklibra: Shouldn't be this line removed?

jniederm: Of course. Thanks.

PR #282, opened 2017-07-13
flow: allow unannotated arrow functions

jniederm:
I believe it's not necessary to type-annotate all arrow functions like
for example parameters of Array.prototype.{map,filter}.

### DIFF
--- a/config/eslint.js
+++ b/config/eslint.js
@@ -75,11 +75,14 @@ module.exports = {
       2,
       'comma'
     ],
-    'flowtype/require-parameter-type': 2,
+    'flowtype/require-parameter-type': [2, {
+      'excludeArrowFunctions': true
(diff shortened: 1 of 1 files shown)


mareklibra:
I'm a little big skeptical about refactoring all "bigger" arrow functions to standard ones using the `function` keyword just because of flow type - they are used heavily and I like this notation.

I think, it's fine to exclude just `expressions`. More complex map/find functions shall be checked.
If it's still an issue, we can agree on same naming-pattern for exclusion.

https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-require-parameter-type

jniederm:
Hi Marek, it was agreed that current state of the path is good without a need of additional patches.

Issue #276, opened 2017-07-11
add support for SSH public keys


jelkosz:
The ssh public key is a per user setting used for virtio-serial console. It can be set only for a particular user which is currently logged in (e.g. the admin can not set it for some user). And if the user does not have admin permissions, it can not be done in admin portal. Therefore, it has to be implemented in the user portal.

The REST endpoint is users/<user id>/sshpublickeys.

PR #290, opened 2017-07-21
Add options with ssh public key

bond95:
@mareklibra please review https://github.com/oVirt/ovirt-web-ui/issues/276


scripts/start.js
```diff
@@ -364,7 +369,8 @@ function getUserInfo () {
         resolve({
           userName: username.slice(0, username.indexOf('@')),
           ssoToken: body.access_token,
-          domain: domain
+          domain: domain,
+          userId: userId,
```
mareklibra: Please handle the case when userId is not provided by SSO.
The userId is provided by https://gerrit.ovirt.org/#/c/79579/ , so recently in oVirt 4.2 only.
Even if 79579 patch is backported, it will not be in oVirt <= 4.1.4

Please correct me, but I think we shall not increase oVirt API version dependency just because of this.

bond95:
@mareklibra fixed, please review

scripts/start.js
mareklibra: @bond95 , please describe&verify what happens if the userId is missing in the userInfo.

bond95: @mareklibra if userId is missing then `options` will be hidden, I also added this description to start.js.

mareklibra:
Recent output on DEV environment:

```
Please authenticate against oVirt running at https://engine.local/ovirt-engine/
oVirt user (admin@internal): 
oVirt password: *****
oVirt domain (internal-authz): 
oVirt user id (null), can be found there: protocol://[FQDN]/api/users/: 
```

We do not need to print the default `(null)` and the address can be composed to exact URL, so the user can "just click".

scripts/start.js
mareklibra: It would be better to always show the `Options` menu item.
In case the `userId` is missing, appropriate message shall be rendered instead of the option to see/modify the ssh-key.

Such a message should state
- what is happening
- what oVirt version is required (means with https://gerrit.ovirt.org/#/c/79579/)
- ideally even what current oVirt version is the portal connected to (see the About dialog, already there)

Reasons for that:
- the user knows there's something (s)he might do
- the dialog will be definitely reused for other config options

bond95: done

bond95:
@mareklibra done, please review

mareklibra:
Output on DEV env after last change:

```
$ node scripts/start.js 
Please authenticate against oVirt running at https://engine.local/ovirt-engine/
oVirt user (admin@internal): 
oVirt password: 
oVirt domain (internal-authz): 
oVirt user id (null), can be found there https://engine.local/ovirt-engine/api/users/ : 
```

Since recent code does not set the default user ID, please remove the `(null)` from the output above.
The  message shall be like:

`oVirt user id (from https://engine.local/ovirt-engine/api/users/): `

src/components/OptionsDialog/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+    if (!this.props.userId) {
+      let apiVersion = 'unknown'
+      if (oVirtApiVersion && oVirtApiVersion.get('major')) {
+        apiVersion = `${oVirtApiVersion.get('major')}.${oVirtApiVersion.get('minor')}`
+      }
+
+      content = (
+        <div>
+          <p>You can't manage SSH keys, because current oVirt version is to old (oVirt <strong>{apiVersion}</strong>), required: >= oVirt <strong>4.2</strong></p>
```
mareklibra: SSH keys can not be managed with recent oVirt <strong>{apiVersion}</strong> version. Please upgrade oVirt to <strong>4.2</strong> or higher.

bond95:
@mareklibra fixed, please review

scripts/start.js
```diff
@@ -347,6 +357,10 @@ function getUserInfo () {
     defaultInput: DEFAULT_DOMAIN
   });
 
+  var userId = readlineSync.question(`oVirt user id (${DEFAULT_UUID}), can be found there ${engineUrl}api/users/ : `, {
```
jniederm: I believe this is unnecessary almost every time. Provided we have an auth token (based on credentials entered above) we can access `api/users`, pick the one with matching username and domain `internal-authz` and offer it as an offered default, or just use it and let developers to edit the script in case someone wants to use  user with different domain.

AFAIK there is no clear mapping between domain entered by user (`internal`) and domain show in API (`internal-authz`).

Sample of API output:

    <user href="/ovirt-engine/api/domains/696E7465726E616C2D617574687A/users/64393165343532662D326330312D343266632D613365632D636537383930373430373862" id="64393165343532662D326330312D343266632D613365632D636537383930373430373862">
        <name></name>
        <department></department>
        <email></email>
        <last_name></last_name>
        <namespace>*</namespace>
        <principal>ovirt02</principal>
        <user_name>ovirt02@internal-authz</user_name>
        <domain href="/ovirt-engine/api/domains/696E7465726E616C2D617574687A" id="696E7465726E616C2D617574687A">
            <name>internal-authz</name>
        </domain>
        <groups/>
    </user>

Of course, it is not necessary to be part of this pull, just an idea.

mareklibra: @jniederm , you are right.
The `/api/users` call does not require admin permission. So even if `filtered`, user's id can be retrieved automatically.

bond95: @jniederm @mareklibra but we can't, I tried again on brq-dev call api/users under `ovirt02`, and still I get empty object `{}`.

jniederm: I can confirm that to `GET api/users` one needs System > Configure System > Manipulate Permissions permission which is not provided by default 'UserRole' role. Without that permission `<users />` is returned.

I guess it's still worth implementing (with fallback to user prompt) since this is for dev mode only and dev mode is often used with `admin@internal` that is compatible with guessing user id. Nevertheless I don't find it necessary to be part of this pull. For this pull we can go just with a user prompt.

mareklibra: The fallback must be there since the same script is used within docker environment.
So I would try to make GET call to get users and if it fails then recent prompt is used.

bond95:
@mareklibra add request for getting user id, if it not possible prompt

scripts/start.js
```diff
@@ -365,10 +376,29 @@ function getUserInfo () {
         return reject(err)
       }
       if (body['access_token']) {
-        resolve({
-          userName: username.slice(0, username.indexOf('@')),
-          ssoToken: body.access_token,
-          domain: domain
+        request(`${engineUrl}api/users`, { json: true, strictSSL: false, headers: { Authorization: `Bearer ${body['access_token']}` } }, (err2, response2, body2) => {
```
jniederm: Could we please use more self-explanatory names like `usersErr`, `usersResponse`, `usersBody` instead of indexing?

scripts/start.js
```diff
@@ -365,10 +376,29 @@ function getUserInfo () {
...
+                  userName: username.slice(0, username.indexOf('@')),
+                  ssoToken: body.access_token,
+                  domain: domain,
+                  userId: body2.user[i].id,
+                })
+                return
+              }
+            }
+          }
+          var userId = readlineSync.question(`oVirt user id (${DEFAULT_UUID}), can be found there ${engineUrl}api/users/ : `, {
```
jniederm: IIUIC, `(${DEFAULT_UUID})` always prints `(null)`, right?

If yes, what is the purpose of `DEFAULT_UUID` variable? Can't we just remove it, don't print 'null' for user and inline `null` at line 394?

bond95:
@mareklibra @jniederm fixed, please review

bond95:
@mareklibra @jniederm Done

src/components/OptionsDialog/index.js
```diff
@@ -0,0 +1,111 @@
...
+  onSSHKeyChange (event) {
+    this.setState({ sshKey: event.target.value })
+  }
+
+  onSaveClick () {
+    this.props.onSave({ key: this.state.sshKey, sshId: this.props.optionsDialog.get('sshId') })
+  }
+
+  render () {
+    let { oVirtApiVersion } = this.props
```
jniederm: `const`. Please prefer `const` over `let` where (reasonably) possible. It improves readability greatly since reader is not forced to scan following part of scope for other assignments.

scripts/start.js
```diff
@@ -365,10 +375,29 @@ function getUserInfo () {
         return reject(err)
       }
       if (body['access_token']) {
-        resolve({
-          userName: username.slice(0, username.indexOf('@')),
-          ssoToken: body.access_token,
-          domain: domain
+        request(`${engineUrl}api/users`, { json: true, strictSSL: false, headers: { Authorization: `Bearer ${body['access_token']}` } }, (userErr, userResponse, userBody) => {
```
jniederm: Could we have a comment here on why this request may not always contain data about current user?

src/components/OptionsDialog/index.js
```diff
@@ -0,0 +1,111 @@
...
+      <form>
+        <div className='form-group'>
+          <FieldHelp content={msg.publicSSHKey()} text={msg.SSHKey()} />
+          <textarea type='text' className='form-control' rows='7' onChange={this.onSSHKeyChange} value={this.state.sshKey || ''} />
+        </div>
+      </form>
+    )
+
+    if (!this.props.userId) {
+      let apiVersion = 'unknown'
```
jniederm: similar:

```
const apiVersion = oVirtApiVersion && oVirtApiVersion.get('major')
  ? `${oVirtApiVersion.get('major')}.${oVirtApiVersion.get('minor')}`
  : 'unknown'
```

bond95:
@jniederm done.

scripts/start.js
```diff
@@ -339,6 +339,15 @@ function getUserInfo () {
 
   var DEFAULT_USER = 'admin@internal';
   var DEFAULT_DOMAIN = 'internal-authz';
+  /**
+   * This is value of user ID, please change it according to your system.
+   * It can be found by `protocol://[FQDN]/api/users/`, where
+   * FQDN - is oVirt engine URL.
```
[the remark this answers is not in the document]
jniederm: prefix `protocol://` is still redundant since protocol is part of URL but whatever ...

Issue #295, opened 2017-07-26
Authorization lost ends in a cycle


leistnerova:
When user have lost authorization error is shown. Page should be reloaded and should go to login page, but it ends up in a cycle and the error is reloaded over and over.

![screenshot from 2017-07-26 10-18-42](https://user-images.githubusercontent.com/29575480/28632131-d0c0b740-722f-11e7-8e79-c8e2adfdad2d.png)

PR #296, opened 2017-07-27
Fix reload of subpages

mareklibra:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/295

### DIFF
--- a/src/ovirt-web-ui.config
+++ b/src/ovirt-web-ui.config
@@ -4,5 +4,5 @@
   "applicationURL": "/ovirt-engine/web-ui",
 
   "consoleClientResourcesURL": "https://www.ovirt.org/documentation/admin-guide/virt/console-client-resources/",
-  cockpitPort: "9090"
+  "cockpitPort": "9090"
 }
--- a/src/sagas.js
+++ b/src/sagas.js
@@ -237,7 +237,8 @@ function* doCheckTokenExpired (action) {
       // No matter saga is canceled for whatever reason, the reload must happen, so here comes the ugly setTimeout()
       setTimeout(() => {
         console.info('======= doCheckTokenExpired() issuing page reload')
-        window.top.location.reload(true)
+        // window.top.location.reload(true)
+        window.location.href = AppConfiguration.applicationURL
       }, 5 * 1000)
       return
     }


src/ovirt-web-ui.config
```diff
@@ -4,5 +4,5 @@
   "applicationURL": "/ovirt-engine/web-ui",
 
   "consoleClientResourcesURL": "https://www.ovirt.org/documentation/admin-guide/virt/console-client-resources/",
-  cockpitPort: "9090"
+  "cockpitPort": "9090"
```
jniederm: 👍

src/sagas.js
```diff
@@ -237,7 +237,8 @@ function* doCheckTokenExpired (action) {
       // No matter saga is canceled for whatever reason, the reload must happen, so here comes the ugly setTimeout()
       setTimeout(() => {
         console.info('======= doCheckTokenExpired() issuing page reload')
-        window.top.location.reload(true)
+        // window.top.location.reload(true)
+        window.location.href = `${AppConfiguration.applicationURL}/`
```
jniederm: Does that actually work? I suspect it might be a problem on dev setup. Snippet from Firefox console:

```
> window.location.href = '//'
[Exception... "The URI is malformed"  nsresult: "0x804b000a (NS_ERROR_MALFORMED_URI)"  location: "JS frame :: debugger eval code :: <TOP_LEVEL> :: line 1"  data: no]
```

 Values of `AppConfiguration.applicationURL`:

* dev setup: `/` (default from `src/config.js`)
* production setup: `/ovirt-engine/web-ui` (from `ovirt-web-ui.config` file)

I guess that property `AppConfiguration.applicationURL` is currently poorly defined, we can agree that it will **always** end with `/` and remove the the slash in line above.

mareklibra: Thanks for pointing that.

On dev environment (means with webpack's dev server), this will work neither.

The oVirt SSO is bypassed and sso token is injected into the application before it's start.
For that reason, "page refresh" leads to repeatedly use of the old token.
To fix that, we would need to design login in `DEV` environment differently - to follow same flow as in production. Not an easy task. Since it's for development only, I think it's low-priority.

Anyway, I removed the extra slash `/` so it at least looks better now.

Re-verified on production environment, still works.

mareklibra:
@jniederm , please comment. If agreed, I'll merge

PR #310, opened 2017-08-07
Internationalization: strings extracted

jniederm:


src/App.js
```diff
@@ -74,7 +75,7 @@ const App = ({ vms, visibility, config }) => {
   return (
     <Router getUserConfirmation={openConfirmation} basename={AppConfiguration.applicationURL}>
       <div>
-        <VmsPageHeader page={vms.get('page')} title={fixedStrings.BRAND_NAME + ' VM Portal'} />
+        <VmsPageHeader page={vms.get('page')} title={fixedStrings.BRAND_NAME + ' ' + msg.vmPortal()} />
```
mareklibra: I believe it's ok for this particular case, but in general we should avoid such string composition due to different word order in various languages.

src/intl/messages.js
```diff
@@ -4,6 +4,7 @@
  * Enter messages here.
  *
  * Please keep keys valid JavaScript identifiers.
+ * Please don't add flow type annotation
```
mareklibra: 👍 for similar comments

src/App.js
jniederm: Ok

Issue #319, opened 2017-08-15
handle smartcard


jelkosz:
There are two missing parts about smartcard handling in the new user portal comparing to the old one:
- there is no way to disable the smartcard in user portal
- there is no indication that the SPICE is configured to forward the smartcard

In old user portal it worked like this:
- if you had a smartcard enabled on the VM, it was possible to disable it on the edit console dialog
- it was not possible to enable it when it was disabled on the VM
- the console label in user portal was "SPICE" or "SPICE with Smartcard" depending if the VM had a smartcard enabled or not

The reason for all this is that the smart card is a sensitive device in terms of security and a malicious VM can brick the smartcard by trying to login 4 times, so we want a user to know when they are about to connect it to the VM.

Please note that this is a feature of SPICE so it is not relevant for RDP nor VNC.

PR #329, opened 2017-08-28
Add smartcard enabled option to VM detail

bond95:
@mareklibra Please review. https://github.com/oVirt/ovirt-web-ui/issues/319

### DIFF
--- a/src/components/VmDetail/index.js
+++ b/src/components/VmDetail/index.js
@@ -49,10 +49,6 @@ function rephraseVmType (vmType) {
   return vmType
 }
 
-function memoryToShorterFormat (memory) {
-  return memory.rounded % 1 === 0 ? memory.number.toFixed(0) + ' ' + memory.suffix : memory.str
-}
-
 class VmDetail extends Component {
   constructor (props) {
     super(props)
@@ -181,7 +177,6 @@ class VmDetail extends Component {
     return (
       <div className={style['main-container']}>
         <VmsListNavigation selectedVm={vm} expanded={this.state.vmsNavigationExpanded} toggleExpansion={this.toggleVmsNavExpansion} />
-
         <div className={style['vm-detail-main']}>
           <div className={this.state.vmsNavigationExpanded ? style['vms-nav-expanded'] : style['vms-nav-collapsed']}>
             <DetailContainer>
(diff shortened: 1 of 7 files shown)


src/components/VmDetail/index.js
```diff
@@ -198,7 +198,7 @@ class VmDetail extends Component {
                   {consoleOptionsShowHide}
                 </dt>
                 <VmConsoles vm={vm} onConsole={onConsole} onRDP={onRDP} />
-                <ConsoleOptions options={optionsJS} onSave={onConsoleOptionsSave} open={this.state.openConsoleSettings} />
+                <ConsoleOptions smartcardOptionEnable={vm.getIn(['display', 'smartcardEnabled'])} options={optionsJS} onSave={onConsoleOptionsSave} open={this.state.openConsoleSettings} />
```
mareklibra: no need to change now, but next time please split among more lines

mareklibra:
Fixes #319

mareklibra:
@bond95 , please rebase

bond95:
@mareklibra done.

src/ovirtapi.js
```diff
@@ -172,6 +172,9 @@ OvirtApi = {
       },
       cdrom: {},
       sessions: [],
+      display: {
+        smartcardEnabled: vm['display']['smartcard_enabled'] ? vm['display']['smartcard_enabled'] === 'true' : false,
```
jniederm: Is there any difference between `vm['display']['smartcard_enabled'] ? vm['display']['smartcard_enabled'] === 'true' : false` and just `vm['display']['smartcard_enabled'] === 'true'`?

src/sagas.js
```diff
@@ -697,13 +697,19 @@ function adjustVVFile ({ data, options, usbFilter }) {
   if (usbFilter) {
     data = data.replace(/^\[virt-viewer\]$/mg, `[virt-viewer]\nusb-filter=${usbFilter}`)
   }
+  if (options && isSpice) {
+    const smartcardEnabled = options.get ? options.get('smartcardEnabled') : options.smartcardEnabled
```
jniederm: AFAIK `adjustVVFile` function is only called from one place and param `options` can only originate from one place. How comes that we can't be sure whether it's a regular object or a immutable collection? Or is this "checking get method" pattern about something else?

src/sagas.js
```diff
@@ -697,13 +697,19 @@ function adjustVVFile ({ data, options, usbFilter }) {
...
+    data = data.replace(/^enable-smartcard=[01]$/mg, `enable-smartcard=${smartcardEnabled ? 1 : 0}`)
```
jniederm: Why to we need `m` regexp flag when the regexp contains both `^` and `$` and no linebreaks?

bond95: `data` is multi-line string, thus we need this modifier.

src/sagas.js
bond95: Because `options` may comes from two different places: `Selectors.getConsoleOptions` and saga worker `getConsoleOptions`.

src/ovirtapi.js
bond95: Yes, value must be `boolean` not `string`.

PR #348, opened 2017-09-21
fix #346 unecesary autoconnect

### LINKED ISSUE #346: vm console autoconnect fires unecesary requests
- fires requests every time even after the vm was deleted
- should remove the vm from local storage when deletion detected

suomiy:


src/optionsManager.js
```diff
@@ -35,6 +34,14 @@ export default {
     saveToLocalStorage(`consoleOptions.${vmId}`, JSON.stringify(options))
   },
 
+  clearConsoleOptions () {
```
jniederm: How about e.g. `clearAutoConnect`. The name seems a bit misleading for me since many console options in local storage may survive a run of this method.

suomiy:
fixed naming

PR #352, opened 2017-10-16
branding: align to new blue style oVirt branding

### LINKED ISSUE #311: ovirt has new branding colors
Ovirt ovirt-engine-webadmin-portal-4.2.0-0.0.master.20170808202742.git7c42951.el7.centos.noarch
has changed colors from green to blue and some other stuff. VM portal uses still the old colors.

screenshot from 2017-08-09 12-17-25

screenshot from 2017-08-09 12-21-23

gregsheremeta:
Change header and about box to new blue style oVirt branding.

Fixes #311


gregsheremeta:
selection_148

selection_147

vojtechszocs:
@gregsheremeta Travis fails on JSX issue, [build log](https://travis-ci.org/oVirt/ovirt-web-ui/jobs/288379577) says

```
/home/travis/build/oVirt/ovirt-web-ui/src/components/VmsPageHeader/index.js
  52:7  error  Parsing error: Expected corresponding JSX closing tag for <nav>
```

@jniederm @mareklibra is there a reason why you're building with both Node.js v4 and v6? (I guess historical reasons?)

src/components/VmUserMessages/index.js
```diff
@@ -69,12 +65,10 @@ class VmUserMessages extends React.Component {
 
     return (
       <li className='dropdown'>
-        <a href='#' onClick={hrefWithoutHistory(onToggle)}>
-          <div className={isUnread(userMessages) ? style['usermsgs-unread'] : style['usermsgs-allread']}>
-            <span className='pficon pficon-info' />&nbsp;{msg.messages()}
-          </div>
+        <a className='dropdown-toggle nav-item-iconic' href='javascript:;' title={msg.messages()} onClick={hrefWithoutHistory(onToggle)}>
```
jniederm: Please replace value of `href` with [idiomatic][1] `#` or leave the attribute empty. `javascript:;` in not necessary in combination with `hrefWithoutHistory()` function.

[1]: https://reactjs.org/docs/handling-events.html

src/components/VmsPageHeader/UserMenu.js
```diff
@@ -28,14 +22,12 @@ const UserMenu = ({ config, onLogout }) => {
   } else {
     return (
       <li className='dropdown'>
-        <a className={`dropdown-toggle`} data-toggle='dropdown' href='#'>
-          <i className='fa fa-sign-out' aria-hidden='true' />&nbsp;
-          {config.getIn(['user', 'name'])}
-          <b className='caret' />
+        <a className='dropdown-toggle nav-item-iconic' href='javascript:;' data-toggle='dropdown' title={config.getIn(['user', 'name'])}>
```
jniederm: same

src/components/VmsPageHeader/index.js
```diff
@@ -16,14 +15,19 @@ import { msg } from '../../intl'
...
+    <nav className='navbar obrand_webadminMastheadBackground navbar-pf-vertical' role='navigation'>
+
+      <div className='navbar-header'>
+        <button type='button' className='navbar-toggle' data-toggle='collapse'>
+          <span className='icon-bar' />
+          <span className='icon-bar' />
+          <span className='icon-bar' />
+        </button>
+
+        <a href='javascript:;' onClick={hrefWithoutHistory(onRefresh)} className='navbar-brand obrand_headerLogoLink'><img className='obrand_webadminMastheadLogo' src='clear.cache.gif' /></a>
```
jniederm: same

src/components/VmsPageHeader/index.js
```diff
@@ -32,10 +36,21 @@ const VmsPageHeader = ({ title, onRefresh }) => {
             </a>
           </li>
 
-          <UserMenu />
+          <li className='dropdown'>
+            <a className='dropdown-toggle nav-item-iconic' href='javascript:;' data-toggle='dropdown' title='About'>
```
jniederm: same

src/components/VmsPageHeader/index.js
```diff
@@ -32,10 +36,21 @@ const VmsPageHeader = ({ title, onRefresh }) => {
...
+              <i className='fa fa-question-circle' /><span className='caret' />
+            </a>
+            <ul className='dropdown-menu' role='menu'>
+              <li className='topLinks'><a href='javascript:;' data-toggle='modal' data-target='#about-modal'>About</a></li>
```
jniederm: same

jniederm:
> @jniederm @mareklibra is there a reason why you're building with both Node.js v4 and v6? (I guess historical reasons?)

Exactly, just historical reasons. I guess we can replace v4 by v8 now.

gregsheremeta:
I broke up the commit into smaller ones.

looks like this at bfa6b1d:

image

@jniederm please review again

branding/ovirt-common.css
```diff
@@ -0,0 +1,12 @@
...loat:left;clear:left}.obrand_ssoHeader{height:41px;position:relative;top:4px;background-color:#121212}.obrand_main_tab{height:calc(100vh - 66px);min-height:calc(100vh - 66px)}.obrand_detail_tab{min-height:265px}.obrand_landingBgTop{position:absolute;top:0;right:0;background-image:url(images/ovirt_landing_bg_top_right.png);width:561px;height:413px;background-repeat:no-repeat}.obrand_landingBgBottom{background-image:url(images/ovirt_landing_bg_bottom_left.png);position:absolute;bottom:0;left:0;width:100%;height:100%;border-top:4px solid #00659c;background-repeat:no-repeat}.obrand_mastheadLogo{background-image:url(images/ovirt_masthead_logo.png);width:272px;height:16px;position:relative;top:6px;border:0px;display:inline !important;background-size:contain;background-repeat:no-repeat}.obrand_mastheadBackground{background-image:url(images/ovirt_masthead_bg.png);background-size:cover}.obrand_aboutBackground{width:565px;height:355px;background-image:url(images/ovirt_about_bg.png)}.obrand_aboutApplicationLogo{background-image:url(images/ovirt_logo.png);width:80px;height:30px;border:0px;display:block;background-size:contain;background-repeat:no-repeat;float:right;position:relative;top:140px}
```
jniederm: Do I understand it correctly that all of these are oVirt specific rules and other brandings are required to provide similar set of rules?

I mean e.g. `.ovirt-container{font-size:13px}` looks equally applicable to all bradings and thus rather suitable to be stored in `index-nomodules.css` or similar file.

jniederm: I understand that this is probably output of a less compiler. Could use please an IDE formatting tool to make it look a bit more like source code? I can come handy in case of some web-ui specific changes.

jniederm:
Verified, looks beautiful.

branding/ovirt-common.css
gregsheremeta: You have a good point. I can slim this file down even more -- from "all common" to "just brand" since web-ui doesn't need ovirt-engine's common styles (for welcome page, sso, etc.), only the brand ones.

I've also renamed it ovirt-brand.css.

gregsheremeta: copying the scss source now (into ovirt-brand.css) instead of the css output.

gregsheremeta:
addressed @jniederm concerns, please re-review

branding/ovirt-brand.css
```diff
@@ -0,0 +1,134 @@
+/********************************************
```
jniederm: Original idea of `branding` directory was that it will contain branding specific styles. Colors, image urls and dimensions are clearly branding specific. However rules like

```
.obrand_ssoHeader {
    height: 41px;
    position: relative;
    top: 4px;
    /* background-color: #121212; */
}
```

contains declarations most probably common to all brandings - e.g. `height: 41px;`.

If this is intentionally chosen way to go since it is easy to create brandings like that by copy-pasting parts of engine code, I'm ok with that.

Otherwise please move declarations common to all brandings to `index-nomodules.css` or similar and let `branding` directory to contain only branding specific declarations - colors, image urls, image dimensions, etc. This will help us make sure that branding packages differs only intentional aspects.

gregsheremeta: I think I understand what you are saying.

This particular style 'obrand_ssoHeader' is affected by the height of the top margin, which is 4px now. If you had a d/s brand that wanted to use only 2px, you'd set that here. So it is a branding-important class and should be here. Are you wanting the class split across two different files, where 'height' and 'position' would be in nomodules, and 'top' would be in here?

jniederm: Based on IRC discussion: let's keep this format to make it easy to copy styles form engine.

Issue #308, opened 2017-08-04
autologout triggers while working with VM portal


leistnerova:
I was just clicking through VMs and its detail/edit and it logged me out. 
Set autologout to 2 minutes, start clicking on VMs and you will see.

engine-config -s UserSessionTimeOutInterval=2

in log appears

2017-08-04 08:22:01,116+02 INFO  [org.ovirt.engine.core.sso.servlets.OAuthRevokeServlet] (default task-2) [] User admin@internal successfully logged out
2017-08-04 08:22:01,188+02 INFO  [org.ovirt.engine.core.bll.aaa.TerminateSessionsForTokenCommand] (default task-15) [262f6704] Running command: TerminateSessionsForTokenCommand internal: true.
2017-08-04 08:22:16,011+02 ERROR [org.ovirt.engine.core.sso.utils.SsoUtils] (default task-10) [] OAuthException invalid_grant: The provided authorization grant for the auth code has expired.
2017-08-04 08:22:16,016+02 ERROR [org.ovirt.engine.core.aaa.filters.SsoRestApiAuthFilter] (default task-11) [] Cannot authenticate using authentication Headers: null: invalid_grant

mareklibra:
This was expected to be resolved by https://gerrit.ovirt.org/#/c/74470/ (abandoned).

Can't find any newer fixing that. @jelkosz , any luck?

leistnerova:
tested with ovirt-engine-4.2.0-0.0.master.20170802210223.git338a2bc.el7.centos.noarch

jelkosz:
@mareklibra yeah, it has been replaced by this one: https://gerrit.ovirt.org/#/c/74848/ - I think it will require some followup from us because it seems that ovirt-web-ui is one of the apps which are considered to not renew the session while only running queries.
The reasoning behind is that if you just let the app opened and don't interact with it, the periodic refresh would make the session never to expire.
In webadmin/userportal it is easy - you can set per query if it renews the session or not. So, the periodic polls dont, and everything else does. In ovirt-web-ui we don't have this luxury since the API does not support some header like that, so we need to set the scope in a way that we will always renew the sessions.

mareklibra:
This needs more effort, moving to next release

michalskrivanek:
Sorry, not without analysing impact and our options. It's too important, we cannot just break after few minutes

mareklibra:
Depends on: https://bugzilla.redhat.com/show_bug.cgi?id=1508403

[...]

PR #388, opened 2017-11-03
Use 'ovirt-app-api' scope for oVirt SSO

mareklibra:
Depends on: https://gerrit.ovirt.org/#/c/83549/
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/308

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/388)


jniederm:
https://gerrit.ovirt.org/#/c/83549/ is merged in master, unblocking.

jniederm:
Does using of 'ovirt-app-api' scope mean that web-ui will never log user out since polling of engine events?

mareklibra:
> Does using of 'ovirt-app-api' scope mean that web-ui will never log user out since polling of engine events?

yes

Unless proper support is implemented on oVirt side, this behavior is better than the recent one.

PR #428, opened 2017-11-20
Fix of sorting of VMs in left vertical panel

jniederm:
Problem: VMs were sorted according to their immutablejs Map
representation instead of according to their names. Oddly enough it
worked pretty well :D

Fix: VMs sorted according to their names. Sorted library replaced -
previous library was a bit clumsy to use in comparator function.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/428)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "recursive-readdir": "2.0.0",
     "request": "~2.79.0",
     "rimraf": "2.6.1",
+    "string-natural-compare": "2.0.2",
     "strip-ansi": "3.0.1",
     "style-loader": "0.13.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,6 +61,7 @@
     "recursive-readdir": "2.0.0",
     "request": "~2.79.0",
     "rimraf": "2.6.1",
+    "string-natural-compare": "^2.0.2",
```
bond95: Remove please ^
Otherwise LGTM.

jniederm: Done

PR #429, opened 2017-11-20
VM name validation

jniederm:
Very basic client side VM name validation. It only filter illegal character not taking into account length limits per different guest OSs.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/429)


src/components/VmDialog/index.js
```diff
@@ -201,7 +203,13 @@ class VmDialog extends React.Component {
   }
 
   onChangeVmName (event) {
-    this.setState({ name: event.target.value, isChanged: true })
+    var regexp = /^[a-zA-Z0-9._\-+]+$/
```
jniederm: There is a catch there:

Engine uses [`"^[\\p{L}0-9._-]*$"`][1] where `L` refers to [all (Unicode) letters][2] category. However javascript doesn't have any direct ability to to filter unicode categories.

Solutions I can see:
1. Being to strict, current version (probably not an option especially because of users used scripts other than latin)
2. Allow all non-wihitespace chars (probably not much useful)
3. Include a lib like [unicode][3] that contains pretty large [unicode table][4] and allows to do proper validation.
4. Develop custom regexp that would match *L* category - far more space efficient compared to option 3

I'd suggest option 3 - it's easy to do, accurate, and the lib size should be bearable.

[1]: https://github.com/oVirt/ovirt-engine/blob/2c236cae373b5634a73a0d992d523bf3aa551f27/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/utils/ValidationUtils.java#L23
[2]: https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#ucc
[3]: https://www.npmjs.com/package/unicode
[4]: https://unicode.org/Public/UNIDATA/UnicodeData.txt

michalskrivanek: what's wrong with 2?

jniederm: It catches just whitespaces. Chars like '!@#$%🚂🚙🚜...' will pass frontend validation but will fail on engine.

src/components/VmDialog/style.css
```diff
@@ -41,3 +41,7 @@
     overflow: visible !important;
     display: flex;
 }
+
+.error-text{
+    font-size: 12px;
+}
```
mareklibra: Please do enter

src/components/VmDialog/index.js
mareklibra: const

mareklibra: I like either 1 or 3

src/components/VmDialog/style.css
jniederm: done

jniederm:
screenshot

image

src/components/VmDialog/index.js
```diff
@@ -201,8 +203,15 @@ class VmDialog extends React.Component {
   }
 
   onChangeVmName (event) {
-    this.setState({ name: event.target.value, isChanged: true })
+    const nameCandidate = event.target.value
+
+    if (isVmNameValid(nameCandidate)) {
+      this.setState({ name: event.target.value, isChanged: true, vmNameErrorText: null })
```
mareklibra: event.target.value --> nameCandidate

Or we can even set just the temporary value for vmNameErrorText and call setState() just once

src/components/VmDialog/index.js
```diff
@@ -571,6 +585,20 @@ class VmDialog extends React.Component {
...
...7\u2D6F\u2D80-\u2D96\u2DA0-\u2DA6\u2DA8-\u2DAE\u2DB0-\u2DB6\u2DB8-\u2DBE\u2DC0-\u2DC6\u2DC8-\u2DCE\u2DD0-\u2DD6\u2DD8-\u2DDE\u2E2F\u3005\u3006\u3031-\u3035\u303B\u303C\u3041-\u3096\u309D-\u309F\u30A1-\u30FA\u30FC-\u30FF\u3105-\u312D\u3131-\u318E\u31A0-\u31BA\u31F0-\u31FF\u3400-\u4DB5\u4E00-\u9FCC\uA000-\uA48C\uA4D0-\uA4FD\uA500-\uA60C\uA610-\uA61F\uA62A\uA62B\uA640-\uA66E\uA67F-\uA697\uA6A0-\uA6E5\uA717-\uA71F\uA722-\uA788\uA78B-\uA78E\uA790-\uA793\uA7A0-\uA7AA\uA7F8-\uA801\uA803-\uA805\uA807-\uA80A\uA80C-\uA822\uA840-\uA873\uA882-\uA8B3\uA8F2-\uA8F7\uA8FB\uA90A-\uA925\uA930-\uA946\uA960-\uA97C\uA984-\uA9B2\uA9CF\uAA00-\uAA28\uAA40-\uAA42\uAA44-\uAA4B\uAA60-\uAA76\uAA7A\uAA80-\uAAAF\uAAB1\uAAB5\uAAB6\uAAB9-\uAABD\uAAC0\uAAC2\uAADB-\uAADD\uAAE0-\uAAEA\uAAF2-\uAAF4\uAB01-\uAB06\uAB09-\uAB0E\uAB11-\uAB16\uAB20-\uAB26\uAB28-\uAB2E\uABC0-\uABE2\uAC00-\uD7A3\uD7B0-\uD7C6\uD7CB-\uD7FB\uF900-\uFA6D\uFA70-\uFAD9\uFB00-\uFB06\uFB13-\uFB17\uFB1D\uFB1F-\uFB28\uFB2A-\uFB36\uFB38-\uFB3C\uFB3E\uFB40\uFB41\uFB43\uFB44\uFB46-\uFBB1\uFBD3-\uFD3D\uFD50-\uFD8F\uFD92-\uFDC7\uFDF0-\uFDFB\uFE70-\uFE74\uFE76-\uFEFC\uFF21-\uFF3A\uFF41-\uFF5A\uFF66-\uFFBE\uFFC2-\uFFC7\uFFCA-\uFFCF\uFFD2-\uFFD7\uFFDA-\uFFDC'
```
mareklibra: OMG :-O

mareklibra: Please describe in comment how exactly was the string generated to allow smooth fix for whatever reason

mareklibra: Please split the single line to concatenation (via '+') of multiple smaller strings

michalskrivanek: if this is a basic validation (as per PR description) please do not use such awful stuff. Just allow a bit more than engine accepts and that's it

michalskrivanek:
I also do not see a reason why pushing this to 1.3

src/components/VmDialog/index.js
jniederm: done, done

src/components/VmDialog/index.js
jniederm: This "awful stuff" is successfully used in engine since 2012.

jniederm: split to multiple variables

bond95:
@jniederm please rebase.

bond95:
@mareklibra I did rebase, please, review.

mareklibra:
Thanks @jniederm , @bond95

Issue #386, opened 2017-11-01
Add "No messages" empty state message


jniederm:
version: commit e438214

see http://www.patternfly.org/pattern-library/communication/empty-state/

![image](https://user-images.githubusercontent.com/5278760/32278182-256b2cee-bf15-11e7-9901-fe81a9cbc4ed.png)

PR #435, opened 2017-11-21
"No messages" as empty state message added

jniederm:
screenshot:
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/435)

### DIFF
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -117,7 +117,10 @@ export const messages = {
     description: 'About application',
   },
   logOut: 'Log out',
-  noMessages: 'No messages',
+  noMessages: {
+    message: 'No\u00A0messages',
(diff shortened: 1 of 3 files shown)


src/intl/messages.js
```diff
@@ -117,6 +117,10 @@ export const messages = {
     description: 'About application',
   },
   logOut: 'Log out',
+  noMessages: {
+    message: 'No\u00A0messages',
```
bond95: I think here need to add some comment about \u00A0, because it is not very clear what this sign for.

jniederm: Do you think following line (122) is not enough?

bond95: Aha, yes, sorry, I didn't saw that.

jniederm:
fixes #386

PR #445, opened 2017-11-30
Release 1.3.4

mareklibra:
- Fix for headless VMs
- Updated translations

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/445)


michalskrivanek:
I do not believe the point was to pull translations now. There's going to be more updates.

mareklibra:
Translations are pulled regularly before every release. I don't think we need to wait if there are already some new available.

Issue #448, opened 2017-12-13
entered SSH key is still in Options form even after the dialog is closed


leistnerova:
Go to user -> Options. Enter something to SSH key textarea. Close the dialog (with Close button or X). Go again to Options. Entered SSH key is still there. 
By opening Options dialog it should load value that is saved.

michalskrivanek:
is it just a display issue? which value does it actually use?

leistnerova:
Only display, the entered value is not saved.

leistnerova:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1525386

michalskrivanek:
@leistnerova, does it work for you in 1.3.6?

leistnerova:
Yes it works, in 1.3.6 entered key doesn't stay in the form.

michalskrivanek:
thanks!

PR #464, opened 2018-01-18
Fixed bug with closed SSH modal dialog and entered but not saved SSH …

bond95:
…key. https://github.com/oVirt/ovirt-web-ui/issues/448
@mareklibra please review.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/464)


src/components/OptionsDialog/index.js
```diff
@@ -65,6 +82,7 @@ class OptionsDialog extends React.Component {
       </form>
     )
 
+    console.log('wtf', this.props.userId)
```
jniederm: sure?

bond95: Done.

src/components/OptionsDialog/index.js
```diff
@@ -23,13 +23,30 @@ class OptionsDialog extends React.Component {
...
+      $(dom).on('shown.bs.modal', function (e) {
+        if (this.props.userId) {
+          this.props.getSSH()
+        }
+      }.bind(this))
+    }
+    $(dom).on('hidden.bs.modal', function (e) {
       if (this.props.userId) {
-        this.props.onOpen()
+        this.props.getSSH()
```
jniederm: I believe this request is redundant. We can save the ssh key into two separate properties of `this.state`, say `sshKey` and `originalSshKey`, in `componentWillReceiveProps` method and later restore original ssh key in `hidden.bs.modal` callback.

```
this.setState({ sshKey: this.state.originalSshKey})
```

Such caching of original ssh key would cause flickering when dialog opened in negligible number of cases.

bond95: There's no need to create one more state, because props already has original key. Thus I think we need to leave this solution as it is.

bond95: And it do, on line 49.

jniederm: Ye. So we are ok with unlikely flickering, line 48 can be removed.

bond95: But no, if someone change SSH key while modal window is hidden, then flickering still be.

bond95:
@mareklibra done.

src/components/OptionsDialog/index.js
```diff
@@ -23,19 +23,33 @@ class OptionsDialog extends React.Component {
...
     this.onSaveClick = this.onSaveClick.bind(this)
+    this.setEventHandlers = this.setEventHandlers.bind(this)
   }
 
   componentDidMount () {
+    this.setEventHandlers(true)
+  }
+
+  componentDidUpdate () {
+    this.setEventHandlers(false)
```
mareklibra: What is the purpose of this call here?

bond95:
@mareklibra removed unnecessary function `setEventHandlers`, please review.

Issue #414, opened 2017-11-15
VM boxes have different height


leistnerova:
VM boxes in grid view have different height.

![screenshot from 2017-11-15 14-31-06](https://user-images.githubusercontent.com/29575480/32838765-66a0a434-ca12-11e7-8373-934e753f0224.png)

And on iPad then can appear blank space.

![screenshot from 2017-11-15 14-31-45](https://user-images.githubusercontent.com/29575480/32838832-8ec356c8-ca12-11e7-9b0c-930559348582.png)

jniederm:
It's caused by non-standard dimensions of cake icon. We should specify set dimensions of containing element.

leistnerova:
also when there is no OS (e.g. pool)

![screenshot from 2017-12-04 16-28-32](https://user-images.githubusercontent.com/29575480/33560540-52f1008e-d910-11e7-890c-cfd4e5075767.png)

PR #475, opened 2018-02-02
Fixed different size of cards with cake icon and without.

bond95:
Fixed this issue https://github.com/oVirt/ovirt-web-ui/issues/414
@mareklibra please review.

image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/475)

### DIFF
--- a/src/components/VmsList/style.css
+++ b/src/components/VmsList/style.css
@@ -88,6 +88,10 @@
 
 .card-icon {
     text-align: center;
+    height: 134px;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
 }
 
 .operating-system-label {


mareklibra:
Please add commit message describing the issue and its fix, including link to the bug.

src/index-nomodules.css
```diff
@@ -73,3 +73,7 @@ div.toolbar-pf {
 .popover {
   min-width: 120px;
 }
+
+span.fa.fa-birthday-cake.card-pf-icon-circle {
+  margin-bottom: 6px;
```
mareklibra: Wouldn't it be better to enforce icon size in a wrapping element?
This seems like special-case hacking to me.
@jniederm , what do you think?

jniederm: Exactly. Enforcing the size of `div` wrapping the `img` element seems to be the way.

bond95:
@mareklibra @jniederm Done.

jniederm:
Snippet of `Vm.js`:

```html
            <div className={`card-pf-top-element ${style['card-icon']}`}>
              <Link to={`/vm/${vm.get('id')}`}>
                <VmIcon icon={icon} className={style['card-pf-icon']}
                  missingIconClassName='fa fa-birthday-cake card-pf-icon-circle' />
              </Link>
              <div>
                <span className={style['operating-system-label']}>{osName}</span>
              </div>
            </div>
```

Why is changed style class `card-icon` and not style of `VmIcon` component itself? The OS label seems to be there unnecessarily extra.

PR #494, opened 2018-02-19
External action failure - message fix

jniederm:
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/494)

### DIFF
--- a/src/reducers/vms.js
+++ b/src/reducers/vms.js
@@ -149,7 +149,7 @@ const vms = actionReducer(initialState, {
         if (state.getIn(['vms', vmId])) {
           return state.setIn(['vms', vmId, 'lastMessage'], payload.shortMessage ? payload.shortMessage : payload.message)
         } else { // fail, if VM not found
-          logError(`vms.updateVmIcon() reducer: vmId ${vmId} not found`)
+          logError('API reports an error associated to nonexistent VM. error=', payload, 'vmId=', vmId)
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/reducers/vms.js
```diff
@@ -149,7 +149,7 @@ const vms = actionReducer(initialState, {
         if (state.getIn(['vms', vmId])) {
           return state.setIn(['vms', vmId, 'lastMessage'], payload.shortMessage ? payload.shortMessage : payload.message)
         } else { // fail, if VM not found
-          logError(`vms.updateVmIcon() reducer: vmId ${vmId} not found`)
+          logError('API reports an error associated to nonexistent VM. error=', payload, 'vmId=', vmId)
```
jelkosz: hmmmm, looking into this I don't think this actually fixes the issue and also the issue looks much worse than just a wrong log. What I think actually happens:
1: try to create a VM based on a template which has a disk
2: sagas.js:createNewVm() fires the first rest call to create the VM
3: the VM's metadata are created, the VM gets locked and the disks of the template are started to be copied
4: the REST request returns success and keeps the VM locked and async copies the disks
5: since the REST request succeeded, the sagas.js:createNewVm() continues to call the changeCD
6: since the VM is still locked, this request fails
7: please note the newly created VM is put into the store only after this call, meaning the handling of the error above concludes that the REST returned an error associated with a non existing VM.

So, the easiest fix I can think of would be to remove the configuration of the CD from edit VM dialog and move it to the VM details the same way as we have the change icon.

jniederm: Described issue is tackled by separate pull request #496. This PR is meant to fix the error message.

PR #496, opened 2018-02-21
Cd removal from new VM page

jniederm:
Only last commit is relevant. Please try to merge #495 first.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/496)

### DIFF
--- a/src/components/VmDialog/index.js
+++ b/src/components/VmDialog/index.js
@@ -554,19 +554,23 @@ class VmDialog extends React.Component {
                   min={1}
                   step={1} />
               </dd>
-              <dt>
-                <span className='pficon pficon-storage-domain' />
-                &nbsp;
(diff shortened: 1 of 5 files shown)


src/components/VmDialog/index.js
```diff
@@ -33,6 +33,7 @@ import SelectBox from '../SelectBox'
 import { MAX_VM_MEMORY_FACTOR } from '../../constants/index'
 
 import { msg } from '../../intl'
+import { Fragment } from '../utils'
```
bond95: Can we use `React.Fragment` instead of creating new component?

jniederm: Unfortunately `(props) => props.children` only works since React 16.0. So I resorted to `<div>` wrapper.

Issue #491, opened 2018-02-19
Browser refresh on VM detail leads to VM listing


jniederm:
When browser refresh is invoked on a VM detail page and the VM is not loaded on the first page of the grid, the refresh causes redirection to a grid listing:

Version: current master, commit 62336f0

Steps to reproduce:
1. Let's have enough VMs to scroll
2. Scroll the grid listing to view VMs from page other than the first
3. View detail of a VM from page other than the first
4. Refresh the VM detail page

Actual result:
VM grid listing is shown

Expected result:
Reloaded VM detail page is shown

PR #497, opened 2018-02-21
Fixed bug with VM detail leading after refresh

bond95:
Bug: If open detail of VM that is not on first page, and refresh it, it'll leads to main page.
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/491

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/497)

### DIFF
--- a/src/components/Pages/index.js
+++ b/src/components/Pages/index.js
@@ -10,8 +10,21 @@ import { selectVmDetail, selectPoolDetail, getISOStorages } from '../../actions/
 import Selectors from '../../selectors'
 
 class VmDetailPage extends React.Component {
+  constructor (props) {
+    super(props)
+    this.requestSent = false
+  }
+
   componentWillMount () {
     if (Selectors.isFilterChecked()) {
+      this.props.getVms({ vmId: this.props.match.params.id })
+    }
+  }
+
+  componentWillUpdate () {
+    const vmInStore = this.props.vms.getIn(['vms', this.props.match.params.id])
+    if (!vmInStore && Selectors.isFilterChecked() && !this.requestSent) {
+      this.requestSent = true
       this.props.getVms({ vmId: this.props.match.params.id })
     }
   }
@@ -44,11 +57,25 @@ VmDetailPage.propTypes = {
 }
 
 class PoolDetailPage extends React.Component {
+  constructor (props) {
+    super(props)
+    this.requestSent = false
+  }
+
   componentWillMount () {
     if (Selectors.isFilterChecked()) {
       this.props.getPools({ poolId: this.props.match.params.id })
     }
   }
+
+  componentWillUpdate () {
+    const poolInStore = this.props.vms.getIn(['pools', this.props.match.params.id, 'vm'])
+    if (!poolInStore && Selectors.isFilterChecked() && !this.requestSent) {
+      this.requestSent = true
+      this.props.getPools({ poolId: this.props.match.params.id })
+    }
+  }
+
   render () {
     let { match, vms, config } = this.props
     if (vms.getIn(['pools', match.params.id, 'vm'])) {


src/actions/vm.js
```diff
@@ -1,6 +1,7 @@
 import {
   CHANGE_VM_ICON,
   CHANGE_VM_ICON_BY_ID,
+  CHECK_VM_AVIABILITY,
```
mareklibra: spelling?

src/components/Pages/index.js
```diff
@@ -110,6 +110,7 @@ const VmDetailPageConnected = connect(
   }),
   (dispatch) => ({
     getVms: ({ vmId }) => dispatch(selectVmDetail({ vmId })),
+    checkVMAviability: ({ vmId }) => dispatch(checkVMAviability({ vmId })),
```
mareklibra: I do not like this collision of names, it might be source of dummy bugs in the future or decreases readability at least.

src/actions/vm.js
```diff
@@ -345,3 +346,12 @@ export function changeVmIconById ({ vmId, iconId }) {
     },
   }
 }
+
+export function checkVMAviability ({ vmId }) {
```
jniederm: Can we follow the convention that only first letter in an abbreviation is capitalized? Like: `checkVmAvailability`.

We already use that in e.g. `VmDialog`, `VmDetail`, and others. Similarly we use `CHECK_VM_AVIABILITY`, not `CHECK_V_M_AVIABILITY`

src/sagas.js
```diff
@@ -581,6 +582,16 @@ function* fetchUSBFilter (action) {
   }
 }
 
+function* checkVmAviability (action) {
```
jniederm: Maybe this can be simplified:

```js
yield fetchSingleVm({ payload: { vmId: action.payload.vmId } })
const vm = yield select(state => state.getIn(['vms', 'vms', action.payload.vmId])
if (!vm) {
  yield put(redirectRoute({ route: '/' }))
}
```

It could save one request.

src/sagas.js
```diff
@@ -581,6 +582,16 @@ function* fetchUSBFilter (action) {
   }
 }
 
+function* checkVmAvailability (action) {
```
jniederm: *copy from previous pull request version*

Maybe this can be simplified:

```js
yield fetchSingleVm({ payload: { vmId: action.payload.vmId } })
const vm = yield select(state => state.getIn(['vms', 'vms', action.payload.vmId])
if (!vm) {
  yield put(redirectRoute({ route: '/' }))
}
```

It could save one request.

bond95: No, because if we call `fetchSingleVm` it update state `vms`, before adding fetched VM to state, thus `VmDetailPage` will update and call check again, and again, and again... Thus my version is more simplified. Otherwise it need to add more lines for take care of this recursion.

jniederm: I see. So how about (pseudo-code):

```js
const vmFromStore = yield select(state => state.getIn(['vms', 'vms', action.payload.vmId])
if (vmFromStore) {
  return
}
yield fetchSingleVm({ payload: { vmId: action.payload.vmId } })
const loadedVm = yield select(state => state.getIn(['vms', 'vms', action.payload.vmId])
if (!loadedVm) {
  yield put(redirectRoute({ route: '/' }))
}
```

I agree that originally posted version is easy to read. I find network requests expensive and requesting the same entity twice in this case not necessary.

bond95: It doing the same thing, and we already check availability of VM in current `vms` state.

bond95:
@jniederm @mareklibra Done.

bond95:
@jniederm Done.

mareklibra:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/components/Pages/index.js, line 22 at r4](https://reviewable.io:443/reviews/ovirt/ovirt-web-ui/497#-L6Qjm1c-ywfdgSIgObZ:-L6Qjm1c-ywfdgSIgOb_:b-hlsd3p) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/d972d0d2a9dfbec9909239e7cdabba678aa174ba/src/components/Pages/index.js#L22)):*
> ```JavaScript
>     const vmInStore = this.props.vms.getIn(['vms', this.props.match.params.id])
>     if (!vmInStore && Selectors.isFilterChecked()) {
>       this.props.getVms({ vmId: this.props.match.params.id })
> ```

This can cause firing of multiple same API requests at the app init.

I would move that to `componentWillMount()`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ovirt/ovirt-web-ui/497)*

bond95:
@mareklibra it already in `componentWillMount`, otherwise it wouldn't work. Because component will mount only one time, if internet speed gonna be too slow then it don't get `filterChecked` in time.

mareklibra:
I'm sorry, I don't understand following sentence:
> it already in componentWillMount, otherwise it wouldn't work.

In recent implementation, I can see the call within `componentWillUpdate()` , which is not called for first (initial) rendering but for every other, means whenever props or state change. This can cause multiple API requests until the very first one finishes.
This can be fixed in multiple ways, the simplest one seems to be just moving the call to the `componentWillMount` or similar to ensure, it is called just once.

If availability of `state.config.isFilterChecked` is an issue at the time of `componentWillUpdate()` calling, it can be fixed i.e. either by
- saga (not preferred)
- a wait function performing up to N sleeps of 50ms (or so) until `state.config.isFilterChecked === true`, then calling a callback function which is the `dispatch`  in our case

bond95:
@mareklibra 
> I'm sorry, I don't understand following sentence:

Here you can see almost the same functionality https://github.com/bond95/ovirt-web-ui/blob/d972d0d2a9dfbec9909239e7cdabba678aa174ba/src/components/Pages/index.js#L13

> This can cause multiple API requests until the very first one finishes.

It wouldn't, because if VM already in redux state, then it don't gonna send request again, you see it here in `vmInStore`:
```JavaScript
if (!vmInStore && Selectors.isFilterChecked()) {
```

mareklibra:
> It wouldn't, because if VM already in redux state,

It will. The `componentWillUpdate()` is called whenever props or state are changed. And at the app initialization, there are multiple changes occurring, so this might be called multiple times shen the VM is still not in the store (means untile the first getVm() finishes, consider scenario with overloaded server).

bond95:
@mareklibra what do you think now?

Issue #486, opened 2018-02-19
Add cloud-init configuration


michalskrivanek:
Should provide
- set ssh authorized keys
- set hostname
If Template has cloud-init it will be overwritten

michalskrivanek:
@jniederm anything else? #468 would work now?

jelkosz:
This one is done, #468 needs one more patch though: https://github.com/oVirt/ovirt-web-ui/pull/528

PR #498, opened 2018-02-23
Cloud init introduced

jniederm:
Only the last commit is relevant

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/498)

### DIFF
--- a/src/ovirtapi.js
+++ b/src/ovirtapi.js
@@ -11,6 +11,11 @@ type VmIdType = { vmId: string }
 type PoolIdType = { poolId: string }
 type InputRequestType = { url: string, input: string, contentType?: string }
 type VmType = { vm: Object }
+type CloudInitInternalType = {
+  enabled: boolean,
+  hostName: string,
+  sshAuthorizedKeys: string
+}
 
 const zeroUUID: string = '00000000-0000-0000-0000-000000000000'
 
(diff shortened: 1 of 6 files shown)


jniederm:
fixes #486 
image
image
image

jniederm:
> This does not handle the situation when the template has a cloud-init. We need to, at least, have a chance to use the cloud-init setting from the template

Done

src/ovirtapi.js
```diff
@@ -317,6 +330,7 @@ OvirtApi = {
       os: {
         type: template.os ? template.os.type : undefined,
       },
+      ...cloudInitApiToInternal(template),
```
mareklibra: For the sake of consistency, can you change to:
```
cloudInit: cloudInitApiToInternal(template)
```
and in `cloudInitApiToInternal()` return just the props of the object?

If agreed, this can be done in a follow-up (but shortly once this is merged :-) )

jniederm: done

apinnick:
I think the tooltip of the Cloud-init setting is odd: "Early initialization for Linux operating systems." Wouldn't it be better to say that it's to enable the cloud-init tool, for automating the initial setup of VMs?

jelkosz:
Or, "setup linux virtual machine using cloud init"?

apinnick:
That works, too. I'd write it, "Set up Linux virtual machine using cloud-init".

jelkosz:
okey, I'll open a separate issue on it.

jelkosz:
there you go: https://github.com/oVirt/ovirt-web-ui/issues/537

Issue #505, opened 2018-03-01
When no SSH key is stored, Options dialog shows 'Loading...'


leistnerova:
When no SSH key for logged user is saved in the engine, Options dialog shows in SSH key 'Loading...' and that doesn't change.

Tested in FF and Google Chrome on ovirt-web-ui-1.3.6-1

![screenshot from 2018-03-01 13-34-30](https://user-images.githubusercontent.com/29575480/36845052-7bf20fd0-1d55-11e8-9fcc-e6ed1d8f4e68.png)

mareklibra:
Still seen, @leistnerova

leistnerova:
Sorry, I did not see, that it's already reported.
https://github.com/oVirt/ovirt-web-ui/issues/730

PR #519, opened 2018-03-07
Fixed bug with "Loading..." message when SSH key is empty.

bond95:
Bug: If there is no SSH key saved for user, then in OptionDialog "Loading..." message doesn't disappear
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/505

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/519)

### DIFF
--- a/src/components/OptionsDialog/sagas.js
+++ b/src/components/OptionsDialog/sagas.js
@@ -10,8 +10,13 @@ function* saveSSHKey (sagas, action) {
 function* getSSHKey (sagas, action) {
   yield put(setUnloaded())
   const result = yield sagas.callExternalAction('getSSHKey', Api.getSSHKey, action)
-  if (!result.error && result.ssh_public_key && result.ssh_public_key.length > 0) {
+  if (result.error) {
+    return
+  }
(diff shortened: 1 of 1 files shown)


src/components/OptionsDialog/sagas.js
```diff
@@ -10,8 +10,12 @@ function* saveSSHKey (sagas, action) {
 function* getSSHKey (sagas, action) {
   yield put(setUnloaded())
   const result = yield sagas.callExternalAction('getSSHKey', Api.getSSHKey, action)
-  if (!result.error && result.ssh_public_key && result.ssh_public_key.length > 0) {
-    yield put(setSSHKey(Api.SSHKeyToInternal({ sshKey: result.ssh_public_key[0] })))
+  if (!result.error) {
```
jniederm: can we please use "return fast" approach.

```js
if (result.error) {
  return
}
// ...
```

I believe it improves readability, and it certainly saves my mind stack.

bond95:
@jniederm Done.

Issue #514, opened 2018-03-06
help for Console is not all visible


leistnerova:
When custom VM's icon is loaded and is smaller than default, help for Console is not all visible.

Tested in ovirt-web-ui-1.3.6-1.fc26.noarch
![screenshot from 2018-03-06 08-53-13](https://user-images.githubusercontent.com/29575480/37020464-23261a48-211c-11e8-8627-6f238d9d7f2e.png)

PR #520, opened 2018-03-07
Fixed tooltip size, now it is not overlaped by toolbar.

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/514

image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/520)

### DIFF
--- a/src/components/FieldHelp/index.js
+++ b/src/components/FieldHelp/index.js
@@ -14,9 +14,10 @@ class FieldHelp extends React.Component {
 
   render () {
     const tooltip = this.props.tooltip || msg.clickForHelp()
+    const popoverStyle = this.props.popoverWidth ? { width: this.props.popoverWidth } : null
 
     const popover = (
-      <Popover id='popover-positioned-top' className={style['field-help-min-width']} title={this.props.title}>
+      <Popover id='popover-positioned-top' style={popoverStyle} className={style['field-help-min-width']} title={this.props.title}>
         {this.props.content}
       </Popover>)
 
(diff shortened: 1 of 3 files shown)


mareklibra:
Really scary. Based on experimental feature (expected to be changed), not supported in IE or Edge.
I'm not strictly against it but what are the options to do it in a production-alligned way?

jniederm:
Here is why I believe we can stay calm and use `max-content`:
* The table has been updated (not by me).
screenshot-developer mozilla org-2018-03-14-19-35-51
* It works (tested on latest chrome and firefox)
* We use css autoprefixer so it's supposed to work even several versions back
* If the style definition is turned of in dev tools, it, the popup is shifted. It doesn't look nice, but it's fully functional

Aternatives I can see:
* Set fixed, large enough width. It would look ugly for short tooltips.
* Guestimate width based on length of tooltip string - requires computation, unreliable.

mareklibra:
Support of IE/Edge is important.
Simple setting of `z-index` did not help (tried some time ago).
The order of components should be changed to properly fix the issue.

In the meantime, since the tooltip component is local to this project and we don't have tons of tooltips, I suggest to introduce new optional parameter denoting width of the component (via appending className).

bond95:
@mareklibra please check.

src/components/FieldHelp/index.js
```diff
@@ -39,6 +40,7 @@ FieldHelp.propTypes = {
   tooltip: PropTypes.string,                                          // tooltip shown when hovering the text
   children: PropTypes.any,
   container: PropTypes.any,
+  popoverWidth: PropTypes.string,
```
mareklibra: Can you please add a comment here, that CSS's `width: max-content` is not supported in IE/Edge and this is a work-around for special cases?

mareklibra:
Will do follow-up, new release is awaiting.

Issue #511, opened 2018-03-05
bootable disk is not first in disks list


leistnerova:
In VM's detail with more disks are the disks sorted only by name. So bootable disk could be in the end, but it should be always first no matter what name it has.

Tested in ovirt-web-ui-1.3.6-1.fc26.noarch
![screenshot from 2018-03-05 16-26-22](https://user-images.githubusercontent.com/29575480/36983262-31b085ec-2092-11e8-9918-1cb1f2c21769.png)
![screenshot from 2018-03-05 16-26-08](https://user-images.githubusercontent.com/29575480/36983267-35af20e0-2092-11e8-9618-ea7a5a9a4dc7.png)

michalskrivanek:
well, that's debatable....

leistnerova:
Could be, but it worked like that few version before.

jelkosz:
> Could be, but it worked like that few version before.

probably coincidentally :) But I'd say it should at least be shown in some predictable order and I'm afraid now it is in the same order the API returns it to us which can be considered random.

sjd78:
Also note that the webadmin sorts numeric aware.

![screenshot-engine local dickerson-web com-2018 04 04-23-45-55](https://user-images.githubusercontent.com/3985964/38346357-8c37c3b2-3862-11e8-8eea-5c259f1aa11c.png)

mareklibra:
Fixed by #547

PR #547, opened 2018-03-29
Fixed the VmDisks sort so bootable disks sort first on the VmDetail page

sjd78:
Resolves issue #511

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/547)


src/components/VmDisks/index.js
```diff
@@ -26,7 +26,14 @@ class VmDisks extends Component {
...
     let classes = style['disks-list']
 
-    let disksToRender = disks.sort((a, b) => a.get('name').localeCompare(b.get('name')) - (a.get('bootable') ? 1000 : 0))
+    let disksToRender = disks.sort((a, b) => {
+      const aBoot = a.get('bootable')
+      const bBoot = b.get('bootable')
+
+      return aBoot && !bBoot ? -1
+        : !aBoot && bBoot ? 1
+        : a.get('name').localeCompare(b.get('name'))
```
jniederm: Can we please use number aware sorting:

```js
const b = ['r', 'ř10', 'ř9', 'ř']
b.sort((a, b) => a.localeCompare(b, undefined, { numeric: true }))
// b is Array [ "r", "ř", "ř9", "ř10" ]
```

sjd78: Excellent point.  I've added that and some tests so it's a bit more obvious what is intended.

sjd78:
I moved the disk sort out to its own file to keep the component simple.  Adding in explicit locales to the compare is easier to see in the **utils.js**.  The test make it explicit what we are expecting the sort to look like on screen.

**package.json** got a tweak to fix a jest complaint.

src/components/VmDisks/utils.js
```diff
@@ -0,0 +1,21 @@
+
+import { locale as appLocale, DEFAULT_LOCALE } from '../../intl'
+
+function localeCompare (a, b, locale = appLocale) {
+  return a.localeCompare(b, [ locale, DEFAULT_LOCALE ], { numeric: true })
```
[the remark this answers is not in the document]
sjd78: I did notice that _locale__ coming from intl.js will fallback to DEFAULT_LOCALE. However, the sort function accepts a locale as a parameter, primarily for testing purposes.  When the locale passed in isn't correct, and there isn't a fallback given the behavior is inconsistent across browsers.

In short, it is there to protect against edge cases if the function is ever used elsewhere.

package.json
```diff
@@ -117,7 +117,9 @@
       "^.+\\.css$": "<rootDir>/config/jest/CSSStub.js",
       "ovirtapi": "<rootDir>/src/mock/ovirtapi.mock.js"
     },
-    "scriptPreprocessor": "<rootDir>/config/jest/transform.js",
```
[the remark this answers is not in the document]
sjd78: That is exactly why I changed it.  `scriptPreprocessor` was migrated to `transform`.  When I ran the tests, jest reported the change.  It is talked about a bit here [1].

screenshot from 2018-04-05 09-59-19

[1] - https://github.com/facebook/jest/issues/1468#issuecomment-276753756

src/components/VmDisks/utils.test.js
```diff
@@ -0,0 +1,76 @@
...
+      { bootable: false, name: 'Xray' },
+      { bootable: false, name: 'Yankee' },
+      { bootable: false, name: 'Yankee 7' },
+      { bootable: false, name: 'Zulu' },
+    ]),
+  },
+
+  {
+    testTitle: 'name with number',
+    locale: 'cs',
```
jniederm: Note, action not required

---

```js
['r', 'ř10', 'ř9', 'ř'].sort((a, b) => a.localeCompare(b, 'cs', { numeric: true }))
['r', 'ř10', 'ř9', 'ř'].sort((a, b) => a.localeCompare(b, 'en', { numeric: true }))
```

both returns the same result `Array [ "r", "ř", "ř9", "ř10" ]`. So this isn't actually testing that other sorting algorithm is used. The distiguishing difference might be the fact that Czech language consider sequence 'ch' as one char that is sorted between 'h' and 'i':

```js
['cecil', 'chris', 'david', 'harry'].sort((a, b) => a.localeCompare(b, 'cs', { numeric: true }))
// Array [ "cecil", "david", "harry", "chris" ]
['cecil', 'chris', 'david', 'harry'].sort((a, b) => a.localeCompare(b, 'en', { numeric: true }))
// Array [ "cecil", "chris", "david", "harry" ]
```

src/components/VmDisks/utils.js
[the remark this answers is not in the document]
jniederm: * There is no test for invalid locale code so reproducibility of test isn't in risk.
* Current production code won't provide wrong locale code.
* Inconsistent sorting in case of potential future usage and wrong language code isn't much of an issue.

... which makes me find the second item of locales array redundant.

src/components/VmDisks/utils.test.js
sjd78: Thanks for the language lesson!  I'll adjust the tests.

src/components/VmDisks/utils.js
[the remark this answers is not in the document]
sjd78: I did a bit more reading and it is redundant and doesn't protect against bad locales passed in via function argument.  I'm ok with letting the locale functions throw the errors if the locale isn't correct.  I'll adjust accordingly.

Sorry for my confusion, it's been awhile since I dealt with the ECMA Intl APIs.

src/components/VmDisks/utils.test.js
sjd78: Oh fun.  The normal build of nodejs, at least for node 8 and newer, appears to enable the Intl API but only includes English (en).  There is a way to include the full locale support.  I'll update package.json to do that, however that will also require updating `ovirt-engine-nodejs-modules` for full CI and build support.

sjd78:
How are new libraries/version handled in ovirt-web-ui?  I added `full-icu` to enable jest to use locales other then 'en' in tests.  I see CI passed, but will rpm packaging fail since `ovirt-engine-nodejs-modules` won't have `full-icu` available yet?

mareklibra:
@sjd78 , `ovirt-engine-nodejs-modules` needs to be rebuilt to take effect.
Don't worry about it now, it's one-time action before `ovirt-web-ui` release.

Issue #541, opened 2018-03-28
Error by creating new disk - 404 Not Found


leistnerova:
When I create new disk for VM, HTTP 404 error returned. But disk is created and appears after reload in VMs detail. Error in engine.log.

tested in ovirt-web-ui-1.3.7-2.el7ev.noarch with ovirt-engine-4.2.2.4-0.1.el7.noarch

![screenshot from 2018-03-28 13-08-36](https://user-images.githubusercontent.com/29575480/38025776-e33fbae0-3289-11e8-9fa3-4ab75eae2a83.png)

leistnerova:
logs
[logs.zip](https://github.com/oVirt/ovirt-web-ui/files/1855716/logs.zip)

jniederm:
It looks like problem of engine permissions. https://bugzilla.redhat.com/show_bug.cgi?id=1554353

michalskrivanek:
hopefully no action, but let's track this for next release

bond95:
Adding disk to VM created via UserPortal, work fine without errors.

jniederm:
This happens for VMs created in webadmin with no user role assigned and admin user creating a disk. It is caused by *always* sending `Filter: true` header with POST HTTP method.

bond95:
Fixed by https://github.com/oVirt/ovirt-web-ui/pull/550.

PR #550, opened 2018-04-03
All HTTP methods uses proper 'Filter' header

jniederm:
Till now, it was set according to 'is regular/admin user' and
AlwaysFilterResultsForWebUi option only for GET method.


jniederm:
it needs further testing

fixes #541

Issue #543, opened 2018-03-28
error by deleting disk


leistnerova:
When I delete disk from VM, after a while disk is removed and error in Messages appears - Failed to retrieve disk details Not Found. No error in engine.log.

tested in ovirt-web-ui-1.3.7-2.el7ev.noarch with ovirt-engine-4.2.2.5-0.1.el7.noarch

![screenshot from 2018-03-28 13-11-00](https://user-images.githubusercontent.com/29575480/38026637-20314902-328d-11e8-9487-eeb112ac93dc.png)

console error:
XML Parsing Error: no root element found
Location: https://engine/ovirt-engine/api/disks/0e8a4376-3456-4529-a39b-276c5d20293d
Line Number 1, Column 1:
0e8a4376-3456-4529-a39b-276c5d20293d:1:1
Ajax failed: {"readyState":4,"responseText":"","status":404,"statusText":"Not Found"}
helpers.js:19:6
External action exception: {"readyState":4,"responseText":"","status":404,"statusText":"Not Found"}
helpers.js:19:6
Reducing action:  {"type":"FAILED_EXTERNAL_ACTION","payload":{"message":"Failed to retrieve disk details\nNot Found","shortMessage":"Failed to retrieve disk details","type":404,"action":{"type":"GET_DISK_DETAILS","payload":{"diskId":"0e8a4376-3456-4529-a39b-276c5d20293d"}}}}
helpers.js:19:6
External action diskattachments() starts on {"type":"GET_DISK_ATTACHMENTS","payload":{"vmId":"8a305dfc-579e-4bcb-a444-6040c0403950"}}
helpers.js:19:6
_httpGet start: url="/ovirt-engine/api/vms/8a305dfc-579e-4bcb-a444-6040c0403950/diskattachments"
helpers.js:19:6
_httpGet: url="/ovirt-engine/api/vms/8a305dfc-579e-4bcb-a444-6040c0403950/diskattachments", headers="{"Authorization":"Bearer 9ba_IsbbWN7-eRPVFwGf1iWJoApLEU1dF0mYQelAGEm9bk4eFwy3vPLDdx9uTrUadK06kWbVz6RjF_hEqcCmlw","Accept-Language":"en_US","Accept":"application/json","Filter":false}"

jelkosz:
I have a strong suspicion it is also related to #550

sjd78:
[...]

PR #551, opened 2018-04-06
Fixed 404 error when deleting disk from VM.

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/543

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/551)


bond95:
@jniederm Done.

bond95:
@mareklibra Done.

sjd78:
@bond95 Looks like your change really just tells the disk fetch to ignore fails.  I think error is probably a timing issue.  Remove the disk ... something asks for its details in future and gets a 404 since it has already been removed.

Take a look at `src/components/VmDisks/sagas.js`.  The __removeDisk()__ function kinda does that on purpose.

Issue #544, opened 2018-03-28
align Delete buttons by disks to the right


leistnerova:
Delete buttons by the disks look chaotic. Align them all to the right.

tested in ovirt-web-ui-1.3.7-2.el7ev.noarch with ovirt-engine-4.2.2.5-0.1.el7.noarch

![screenshot from 2018-03-28 16-10-16](https://user-images.githubusercontent.com/29575480/38034502-b5fb5db4-32a2-11e8-8f85-62dc99dbdec2.png)

PR #553, opened 2018-04-06
Fixed align of Delete buttons for Disks

bond95:
Bug: When some disk is bootable, arrangment is broke.
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/544
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/553)

### DIFF
--- a/src/components/VmDisks/VmDisk.js
+++ b/src/components/VmDisks/VmDisk.js
@@ -93,16 +93,17 @@ class VmDisk extends React.PureComponent {
 
     return (
       <li>
-        <span id={`${idPrefix}`}>
+        <span className={style['vmdisk-item-info']} id={`${idPrefix}`}>
           {name}
           &nbsp;
           {capacityInfo}
           {bootable}
           {inactive}
-          &nbsp;
-          {deleteButton}
-          {deletionConfirmationDialog}
         </span>
+        <span className={style['vmdisk-item-delete']}>
+          {deleteButton}
+        </span>
+        {deletionConfirmationDialog}
       </li>
     )
   }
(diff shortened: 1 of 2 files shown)


src/components/VmDisks/style.css
```diff
@@ -27,3 +28,14 @@
     font-size: small;
     cursor: pointer;
 }
+
+.vmdisk-item {
```
mareklibra: I would make these more generic, not just for `vmdisk`.
Isn't there same issue for networks?
What about potential future list of other objects on the same place?

@jniederm

jniederm: Provided we have time for this, it would surely be nice to have a component. I wouldn't go for anything complex like `AddRemoveListing` but rather use something purely visual like `AlignedListing`

src/components/VmDisks/style.css
```diff
@@ -8,6 +8,7 @@
     list-style-type: none;
     padding: 0;
     margin-bottom: 0px !important;
+    display: table;
```
jniederm: I find this difficult to read and debug. This actually relies on generation of virtual, DOM invisible, `row` and `tr` elements as described at [Anonymous table objects][1] css spec. I find it much more clear to openly admit that we are actually creating complete css table and use appropriate `display` values for `li` (`table-row`) and `span (`table-cell`) elements.

[1]: https://www.w3.org/TR/CSS22/tables.html#anonymous-boxes

bond95:
@jniederm Done.

src/components/VmDisks/VmDisk.js
```diff
@@ -93,16 +93,17 @@ class VmDisk extends React.PureComponent {
 
     return (
       <li>
-        <span id={`${idPrefix}`}>
```
mareklibra: Please do not remove `id` attributes, they are referenced from Selenium tests

bond95:
@mareklibra Done.

PR #556, opened 2018-04-09
Revert adding of full-icu package

mareklibra:
Introduced by: https://github.com/oVirt/ovirt-web-ui/pull/547
Breaks recent offline-build by using runtime-determined npm dependencies.

The npm package "full-icu" does post-installation setup consisting of downloading "icu4c-data"
_in a version corresponding to running nodejs_ which is determined at runtime.

Can be fixed by either including all potential icu4c-data versions in ovirt-engine-nodejs-modules
or using clean full-icu npm dependency mechanism.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/556)


mareklibra:
This is hot-fix to enable new build.
I'm not against ICU but the dynamic dependency needs to be resolved first.

mareklibra:
@sjd78 , merging. Please let me know if this should not be done.

sjd78:
@mareklibra, it's only used for testing.  Therefore, while annoying, it makes sense to pull the dependency and non 'en' tests for now.  I'll research other ways to get the ICU data available to node in the test phase so it can be done in an offline build.

Issue #546, opened 2018-03-29
Cloud-Init switch doesn't work properly


leistnerova:
Clicking on Cloud-Init switch in edit VM doesn't do anything. Change is rendered after clicking on Boot menu. Also writing something to textarea doesn't fill it. After clicking on Boot menu only one character from the entered text appears.
See video

tested in ovirt-web-ui-1.3.7-2.el7ev.noarch with ovirt-engine-4.2.2.5-0.1.el7.noarch
It does the same in Firefox 59.0.1 and Google Chrome 65.0.

[video.zip](https://github.com/oVirt/ovirt-web-ui/files/1859387/video.zip)

PR #559, opened 2018-04-09
Fixed bug with delayed switch in Cloud-Init enable, also Hostname and…

bond95:
… SSH Keys typing delay

Bug: When click on Cloud-Init enable switch it has very long delay, problem was in function that putting as parameter in setState function.
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/546

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/559)


bond95:
@mareklibra Done.

Issue #549, opened 2018-04-03
disable Boot menu switch in edit dialog when VM is running


leistnerova:
Boot menu switch can't be changed when VM is running. So it should be disabled. Now it seems that the value is changed, but editing VM again shows previous saved value.
In admin portal is showed flag that VM has new configuration for next run. This is other possibility how to fix this issue.

leistnerova:
Adding warning for Boot menu every time VM is running is more confusing than helpful.

I agree with the warning description in help label.
Still when I change the value, VM portal shows the old one. So the proper solution should be that the VM portal remembers the change and shows it by the VM. And also shows that something was changed with warning icon and will be applied after reboot.  The icon could be by the whole VM in detail or grid view, not in the edit dialog.

jniederm:
for future reference: this is about showing *current* x *next run* configuration difference.

PR #560, opened 2018-04-09
Add warning for Boot menu switch on running VM.

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/549

Running VM:
image

Not running VM:
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/560)

### DIFF
--- a/src/components/VmDialog/index.js
+++ b/src/components/VmDialog/index.js
@@ -435,6 +435,7 @@ class VmDialog extends React.Component {
     }
 
     const isEdit = !!vm
+    const isUp = (isEdit && vm.get('status') === 'up')
 
     const sortedClusters = sortedBy(clusters.get('clusters'), 'name')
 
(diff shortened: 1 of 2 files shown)


src/components/VmDialog/index.js
```diff
@@ -596,7 +607,14 @@ class VmDialog extends React.Component {
               )}
 
               <dt>
-                <FieldHelp content={msg.bootMenuTooltip()} text={msg.bootMenu()} />
+                {
+                  isUp &&
+                  <React.Fragment>
+                    <span className='pficon pficon-warning-triangle-o' />
+                    &nbsp;
```
mareklibra: In this case, use of padding would be better.
Considering time constraint, let's keep it for a follow-up.

src/intl/messages.js
```diff
@@ -262,6 +262,7 @@ export const messages = {
   network: 'Network',
   noNetwork: 'No network',
   vnicProfileEmpty: '<Empty>',
+  bootMenuWarning: 'All changes will take effect only after reboot!',
```
mareklibra: Please no exclamation mark `!` in similar texts.
Considering time constraint, let's keep it for a follow-up

src/components/VmDialog/index.js
```diff
@@ -435,6 +435,7 @@ class VmDialog extends React.Component {
     }
 
     const isEdit = !!vm
+    const isUp = (isEdit && vm.get('status') === 'up')
```
mareklibra: Should be `isRunning`

Issue #567, opened 2018-04-12
After click on 'Set default Icon' every additional data of VM is disappearing 


bond95:
1. Open any VM detail
2. Click on 'Set default Icon'

Before click:
![image](https://user-images.githubusercontent.com/3332176/38672646-1877670a-3e4f-11e8-885a-fa4526fdfaa3.png)
After:
![image](https://user-images.githubusercontent.com/3332176/38672666-22019b9c-3e4f-11e8-9ed9-fb2f9cec77c0.png)

PR #572, opened 2018-04-18
Fixed bug with dissapearing VM data after clicking on "Set default icon"

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/567

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/572)

### DIFF
--- a/src/sagas.js
+++ b/src/sagas.js
@@ -425,9 +425,7 @@ function* changeVmIcon (action) {
   }
 
   if (vm && vm.id) {
-    const internalVm = Api.vmToInternal({ vm })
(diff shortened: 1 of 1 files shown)


src/sagas.js
```diff
@@ -426,6 +426,13 @@ function* changeVmIcon (action) {
```
jniederm: Can we just replace the rest of this method by calling `fetchSingleVm()`?

mareklibra: Right, whole `if` is not needed, just initiate refresh of the single VM

mareklibra:
@bond95 , ping

Issue #565, opened 2018-04-12
no way to boot from a network if the VM has no disk


jelkosz:
- create a VM in ovirt-web-ui with 1 NIC but with no disk.
- try to run it

result:
- it will complain that the VM has no bootable disk so it can not be started. The bootmenu in this case will not help, because the VM will not even attempt to start

PR #575, opened 2018-04-20
Added Boot Sequence feature.

bond95:
Bug: When trying to run VM with no disks, but with one NIC, it throw error, because can't find bootable device.
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/565

@jniederm @mareklibra @leistnerova What do you think about design?
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/575)

### DIFF
--- a/src/ovirtapi.js
+++ b/src/ovirtapi.js
@@ -162,6 +162,8 @@ OvirtApi = {
 
       os: {
         type: vm['os'] ? vm.os['type'] : undefined,
+        bootDevices: vm.os && vm.os.boot && vm.os.boot.devices && vm.os.boot.devices.device
+          ? vm.os.boot.devices.device : [],
       },
 
       highAvailability: {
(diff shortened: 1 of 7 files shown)


jniederm:
It would be nice to have the numbering unified. Personally I prefer the "first device", "second device" labels option .

image

src/ovirtapi.js
```diff
@@ -267,8 +269,13 @@ OvirtApi = {
         id: vm.cluster.id,
       } : undefined,
 
-      os: vm.os && vm.os.type ? {
-        type: vm.os.type,
+      os: vm.os && (vm.os.type || vm.os.bootDevices) ? {
+        type: vm.os.type || undefined,
+        boot: {
+          devices: {
+            device: vm.os.bootDevices.filter((item) => item !== null),
```
mareklibra: weird case, but might happen: `vm.os.bootDevices === undefined`

mareklibra:
This needs rebase and addressing the comment above

src/ovirtapi.js
bond95: Actually no, because it get data from VmDialog, and VmDialog in any case will return array.

bond95:
@jniederm @mareklibra Done.
image

PR #577, opened 2018-04-24
fix: cluster can be in no datacenter

### LINKED ISSUE #576: if there is a cluster which is in no DC, web-ui crashes

jelkosz:
If the cluster was in no DC, web-ui crashed.
Since it is a valid case, made it possible.

fixes #576

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/577)

### DIFF
--- a/src/ovirtapi.js
+++ b/src/ovirtapi.js
@@ -385,7 +385,7 @@ OvirtApi = {
     return {
       id: cluster.id,
       name: cluster.name,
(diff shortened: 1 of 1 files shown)


jelkosz:
I need to test it a bit more - what happens in new VM when the cluster is in no DC and I pick it Im not sure.

PR #578, opened 2018-04-24
Storage domain in multiple data centers - fix

jniederm:
Status of storage domain that is attached to multiple data centers is
correctly merged. Previously it failed with 'Cannot convert undefined or
null to object'.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/578)


jelkosz:
I have tested it and it works as expected.

Issue #522, opened 2018-03-08
add option to poweroff a VM


jelkosz:
It would be good to have some checkbox inside of the shutdown window with label like "force" and if checked, it would do poweroff.

![shutdown](https://user-images.githubusercontent.com/701107/37150654-c8950d56-22d2-11e8-974e-4079247bf383.png)

sjd78:
Posted my solution in PR #581

PR #581, opened 2018-04-27
Add force shutdown button in shutdown confirm popup

sjd78:
Solves issue #522 

Refactored components/VmActions:
  - Removed RemoveVmAction and integrated it directly into VmActions
  - Reordered Button attributes to be consistent across all definitions
  - For buttons that use complex Confirmation text, refactored to be
    consistent and harmonize styles

Added data shape to components/Toolbar's match prop

Refactored components/Confirmation:
  - Embed string text in a `<p>` block and non-strings in a `<div>` block with
    a shared style to maintain the same styling across both renders
  - reordered Button attributes to be consistent
  - removed 'button-l' className since it isn't in use
  - moved the optional extraButton render inline to the main JSX
  - refactored styles to use flex box layout so text and buttons are always
    nicely centered

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581)


jelkosz:
@sjd please always provide a screenshots to simplify the review

Other than that, while playing with it I found 2 issues. I think both issues have been there before, but without the force you realistically did not hit them. But now with the force option you hit them immediately.

1: the shutdown and force stop should be available in all this VM states: up, powering up, powering down, reboot in progress, migrating, suspended, paused, not responding. The most important one is the "powering down" because it is a common case that you first try to shutdown without force and when you run out of patience, than you force it :)

2: the confirmation re-appears. You can get to it like this:
1: go to VM detail
2: press shutdown (pick force so it will happen quickly)
3: wait until the VM shuts down
4: start the VM
When the VM turns up, the confirmation shows up by itself again.

src/components/VmActions/index.js
```diff
@@ -166,64 +102,110 @@ class VmActions extends React.Component {
...
         <LinkButton isOnCard={isOnCard}
           shortTitle={msg.edit()}
+          tooltip={msg.editVm()} to={`/vm/${vm.get('id')}/edit`}
           button='btn btn-default'
           className={`pficon pficon-edit ${style['action-link']}`}
-          tooltip={msg.editVm()} to={`/vm/${vm.get('id')}/edit`}
           id={`action-${vm.get('name')}-edit`} />
 
-        <RemoveVmAction isOnCard={isOnCard} isPool={isPool} vm={vm} isDisks={vm.get('disks').size > 0} onRemove={onRemove} />
+        {!isOnCard && (<Button isOnCard={false} actionDisabled={isPool || !canRemove(status) || vm.getIn(['actionInProgress', 'remove'])}
```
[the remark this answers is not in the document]
sjd78: I don't have a strong feeling one way or another on that style change.  I'll leave the first attribute on the line with the tag and drop any other attributes below.  At least it'll be self-consistent in this file.

src/components/VmActions/index.js
```diff
@@ -166,64 +102,110 @@ class VmActions extends React.Component {
...
-            uniqueId={vm.get('name')} />} />
+          popover={({ close }) =>
+            <Confirmation
+              text={(
+                <div>
+                  <div>{msg.shutdownVmQuestion()}</div>
+                  <div style={{ marginTop: '8px' }}>
+                    <Checkbox
+                      checked={this.state.forceShutdown}
+                      onClick={() => { this.setState({ forceShutdown: !this.state.forceShutdown }) }}
```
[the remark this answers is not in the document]
sjd78: Ok.

sjd78:
@jelkosz  - The shutdown button's enabled/disabled logic was unchanged.  Shutdown (and force/Power Off) is currently available when:
  * the VM is not currently being shutdown
  * the VM status is:
    * up
    * migrating
    * reboot_in_progress
    * paused
    * powering_up

That leaves a gap for when a VM is powering_down, or not_responding.  In your opinion, would it be better to patch the extra states in the web-ui code or do it over in the [ovirt-ui-components project ](https://github.com/oVirt/ovirt-ui-components/blob/master/src/vm-status.js).  I vote for migrating the /can.*/ functions out of ovirt-ui-components and back to web-ui.  That would make maintenance on that particular bit easier going forward.

As for the confirm popovers not going away, it's fun to click the suspend button and then the shutdown button.  Both popovers stay.  Similar to when you use the left navigation and switch VMs - if you open the confirm popover in one VM then switch to another, the popover doesn't hide.  I consider these issues outside of the scope of this PR and Issue as that behavior wasn't introduced by adding the force shutdown option.

sjd78:
@jniederm - Honestly I don't like either style.

Having the Force checkbox or having a secondary Force button on the confirmation popover is inconsistent with the experience in both webadmin and cockpit for the force shut down.

webadmin (action labeled as "Power Off"):
screenshot-2018-4-27 ovirt open virtualization manager

cockpit (action labeled as "Force Shut Down":
screenshot-2018-4-27 virtual machines - chopper-server local dickerson-web com

for reference, this is how it looks with the PR so far:
screenshot-2018-4-27 vm portal

and this is how the remove button looks:
screenshot-2018-4-27 vm portal 1

Is there a particular style that UXD approved or do we want to stay self-consistent?

gregsheremeta:
Apologies @sjd78 I missed your last comment until today.

What @sjd78 said in the last comment is correct. Both the Force checkbox and buttons are bad / don't follow patternfly, and should be changed to secondary actions dropdown style like in Scott's screenshots:

39382705-6d56fd80-4a34-11e8-9531-5f0a2af60208

@mareklibra @jelkosz agree?

Assuming yes, @sjd78 only worry about force shutdown for this issue, and I'll open one for force remove (awaiting ack from Tomas or Marek)

mareklibra:
@sjd78 ,
>  I vote for migrating the /can.*/ functions out of ovirt-ui-components and back to web-ui. 

Sure. Historically, there was a wave migrating the code under ovirt-ui-components due to various reasons, followed by another slower one to migrate it back since the original reasons passed.
So sure - fine for me to migrate it back to this project.

mareklibra:
> What @sjd78 said in the last comment is correct. Both the Force checkbox and buttons are bad / don't follow patternfly, and should be changed to secondary actions dropdown style like in Scott's screenshots

It makes sense to get aligned. IIRC, implementation in web-ui proceeded the one in Cockpit.

Anyway, there should be extra step added to confirm this destructive action (not needed with recent design).

gregsheremeta:
I just noticed github uses the same secondary action dropdown pattern :)

selection_550

selection_552

gregsheremeta:
Update. Per Tomas, for now let's:
1. go with Jakub's suggestion so it's at least consistent
2. eventually adopt the secondary dropdown pattern
3. converting to modals is #587

sjd78:
Force shutdown is now on the extra action button in the style of the force remove extra action button.

screenshot-2018-5-3 vm portal

sjd78:
@mareklibra - I pulled and updated vm-status.js from ovirt-ui-components.  Would you like me to submit a PR on that project to remove vm-status.js?

sjd78:
@jelkosz - This is exactly what you described with the popover magically staying opened.

ezgif com-video-to-gif

mareklibra:
>  I pulled and updated vm-status.js from ovirt-ui-components. Would you like me to submit a PR on that project to remove vm-status.js?

Yes, please. Let's have it consistent ...

sjd78:
>>  I pulled and updated vm-status.js from ovirt-ui-components. Would you like me to submit a PR on that project to remove vm-status.js?
>
> Yes, please. Let's have it consistent ...

https://github.com/oVirt/ovirt-ui-components/pull/50 removes the file from ovirt-ui-components

mareklibra:
> Does it make sense to increase the dependency version for ovirt-ui-components to 0.3.0 in this PR

No, I'll take care of it independently. Thanks for the patches!

src/vm-status.js
```diff
@@ -0,0 +1,55 @@
+export function canStart (state) {
+  return state && (state === 'down' || state === 'paused' || state === 'suspended')
+}
+
+export function canShutdown (state) {
+  return state && (state === 'up' ||
```
mareklibra: nitpicking: when refactoring, we can replace this syntax by 
```
return ['up', 'migrating', ...].includes(state)
```

src/vm-status.js
```diff
@@ -0,0 +1,55 @@
...
+                   state === 'migrating' ||
+                   state === 'reboot_in_progress' ||
+                   state === 'saving_state')
+}
+
+export function canRemove (state) {
+  return state && (state === 'down')
+}
+
+/*
```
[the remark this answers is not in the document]
sjd78: Good point.  I'll take the comment out.

src/vm-status.js
sjd78: I had originally done that, but then I figured it was less efficient then the set of === || operators.  But, to satisfy my curiosity, I put together a quick benchmark across 7 different styles: http://jsben.ch/mhdZY

Net result, setting up the array outside of the function seems to be the fastest and most readable.  I'll adjust that way.

jniederm: It's nice comparison. I believe in case of UI development readability and clarity is usually more important than performance.

sjd78: I updated to indexOf() instead of includes().  They're within tens of ms of each other, but indexOf() is supported in IE.

jniederm: Please follow Marek's [suggestion][1] and use `include()`. I find it the cleanest and easiest to read solution.

---

If I understand the performance benchmark (http://jsben.ch/mhdZY) correctly, processing of predefined arrays was faster only because the time to evaluate the setup block is excluded from the test. Hence inlining of array definition shouldn't cause a performance penalty.

[1]: https://github.com/oVirt/ovirt-web-ui/pull/581#discussion_r186947553

sjd78:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/vm-status.js, line 6 at r4](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581#-LC-fSyIlgQfnmZBxhw7-r4-6:-LCC82dDBTQybS41HgWs:b-kna2vb) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/e2fd31071e2d4a82d32ff51f736515dbfed237c8/src/vm-status.js#L6)):*
<details><summary><i>Previously, jniederm wrote…</i></summary><blockquote>

Please follow Marek's [suggestion][1] and use `include()`. I find it the cleanest and easiest to read solution.

---

If I understand the performance benchmark (http://jsben.ch/mhdZY) correctly, processing of predefined arrays was faster only because the time to evaluate the setup block is excluded from the test. Hence inlining of array definition shouldn't cause a performance penalty.

[1]: https://github.com/oVirt/ovirt-web-ui/pull/581#discussion_r186947553
</blockquote></details>

In Chrome, the times come out the same for static (array defined once outside of the function) vs inline definition (array redefined on each function execution).  In Firefox, the times for static definition are significantly faster.

In this particular file, with these particular tests it really isn't an issue.

---

*[src/vm-status.js, line 37 at r4](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581#-LC-fSyIlgQfnmZBxhw7-r4-37:-LCCLv6I5zVteInBiiLX:b-896fix) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/e2fd31071e2d4a82d32ff51f736515dbfed237c8/src/vm-status.js#L37)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Good point.  I'll take the comment out.
</blockquote></details>

Done.

---

*[src/components/Confirmation/style.css, line 28 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581#-LB4R8u8ggDsd9t5kbYA-r1-28:-LCCLqtGDKYby_4QiBUE:b-896fix) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/29ea45f7d8c11aad2376bebc27c172afc0826ef0/src/components/Confirmation/style.css#L28)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Done
</blockquote></details>

Done.

---

*[src/components/VmActions/index.js, line 139 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581#-LB4R8u8ggDsd9t5kbYC-r1-139:-LCCLsLIBFkBi4oETbjx:b-896fix) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/29ea45f7d8c11aad2376bebc27c172afc0826ef0/src/components/VmActions/index.js#L139)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Ok.
</blockquote></details>

Done.

---

*[src/components/VmActions/index.js, line 182 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581#-LB4R8u8ggDsd9t5kbYC-r1-182:-LCCLtUvBL6D-GYcJEjm:b-896fix) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/29ea45f7d8c11aad2376bebc27c172afc0826ef0/src/components/VmActions/index.js#L182)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

I don't have a strong feeling one way or another on that style change.  I'll leave the first attribute on the line with the tag and drop any other attributes below.  At least it'll be self-consistent in this file.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/581)*

sjd78:
Wow - Reviewable did a few things I wasn't expecting.  Live and learn I suppose.

mareklibra:
@bond95 , @jniederm , any additional comments, please?

Issue #395, opened 2017-11-08
Error message partially overlapped


jniederm:
Error message "Authorization expired. The page is going to be reloaded to re-login." is partially overlapped.

Steps to reproduce:
1. Open web-ui
2. Restart connected engine to invalidate current authorization

![image](https://user-images.githubusercontent.com/5278760/32553620-b32b883a-c497-11e7-9168-5be0e5777625.png)

version: commit 3db3a19

michalskrivanek:
I suppose it wont' happen now when session doesnt' expire since #388?

mareklibra:
I think the issue remains for the steps described above.
To fix this issue:
- whole the message shall be visible
- the token shall be invalidated within ovirt-web-ui (its most probably already done)
- the user shall be able to reload the page to pass oVirt SSO login form once again (i.e. render `reload` within the error message

mareklibra:
Might be ralated: #452, #434

PR #586, opened 2018-05-01
Fix expired token error message to be fully visible

sjd78:
Due to the way the PageRouter is laid out, the token error message
would display beneath and only be partially visible.  This is fixed
twice by setting the z-index on .alert to be higher, and by moving
the tag below the PageRouter in the App's dom.  Both changes will
make sure the alerts are not hidden.

Refactored OvirtApiCheckFailed to use ErrorAlert, allowing future
changes to error alerts to be done in one place.

Resolves issue #395

screenshot-2018-5-1 vm portal

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/586)

### DIFF
--- a/src/components/ErrorAlert.js
+++ b/src/components/ErrorAlert.js
@@ -1,25 +1,21 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const ErrorAlert = ({ id, message, dangerouslySetInnerHTML }) => {
+const ErrorAlert = ({ id, message, children }) => {
   return (
     <div className='alert alert-danger'>
       <span className='pficon pficon-error-circle-o' />
-      {message && (
-        <strong id={id}>{message}</strong>
-      )}
-      {dangerouslySetInnerHTML && (
-        <span id={id} dangerouslySetInnerHTML={dangerouslySetInnerHTML} />
-      )}
+      <span id={id}>
+        {message && (<strong>{message}</strong>)}
+        {children}
+      </span>
     </div>
   )
 }
 ErrorAlert.propTypes = {
   id: PropTypes.string.isRequired,
   message: PropTypes.string,
-  dangerouslySetInnerHTML: PropTypes.shape({
-    __html: PropTypes.any.isRequired,
-  }),
+  children: PropTypes.node,
 }
 
 export default ErrorAlert
--- a/src/components/ErrorAlert.js
+++ b/src/components/ErrorAlert.js
@@ -1,17 +1,25 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const ErrorAlert = ({ message, id }) => {
-  return message ? (
+const ErrorAlert = ({ id, message, dangerouslySetInnerHTML }) => {
+  return (
     <div className='alert alert-danger'>
       <span className='pficon pficon-error-circle-o' />
-      <strong id={id}>{message}</strong>
+      {message && (
+        <strong id={id}>{message}</strong>
+      )}
+      {dangerouslySetInnerHTML && (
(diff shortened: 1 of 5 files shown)


gregsheremeta:
It's visible now, but not positioned correctly. I think that should be a part of this fix.

sjd78:
Just a mild centering horizontal margin:

screenshot-2018-5-2 vm portal

sjd78:
@mareklibra , @jniederm - nudge for a review
@gregsheremeta - does the latest screenshot look ok?

gregsheremeta:
Afraid not, that alert needs to have top and bottom margin. It's not supposed to float on top of stuff / have a z-index

gregsheremeta:
I think it should look like this, from top to bottom:
menu bar
line
button row
spacing
red box
spacing
gray box

all with no overlaps

But leave it as is for now, and let's wait for UXD help next week

sjd78:
To consider: 
 * http://www.patternfly.org/pattern-library/communication/inline-notifications/ specifically the "In Context" portion (although the overview image is a bit ambiguous with what the wavy pink line represents)
 * http://www.patternfly.org/pattern-library/communication/session-timeout/ (which includes an inline notification over top of the log in page)

src/components/ErrorAlert.js
```diff
@@ -1,17 +1,25 @@
...
+      )}
     </div>
-  ) : null
+  )
 }
 ErrorAlert.propTypes = {
+  id: PropTypes.string.isRequired,
   message: PropTypes.string,
-  id: PropTypes.string,
+  dangerouslySetInnerHTML: PropTypes.shape({
```
jniederm: I believe we should avoid passing html strings when possible. The translation can be passed as component child or property. It can be even merged into one `message` property. See  https://codesandbox.io/s/k25p3o94kv.

Using one property would also solve double use of `id` property.

jniederm:
> @mareklibra , @jniederm - nudge for a review

Hi @sjd78 do you have enough permissions to add others as reviewers like in https://blog.github.com/2016-12-07-introducing-review-requests/ ?

src/components/ErrorAlert.js
bond95: Completely agree, using `dangerouslySetInnerHTML` is bad programming practice in React. https://medium.com/dailyjs/exploiting-script-injection-flaws-in-reactjs-883fb1fe36c1

sjd78:
@jniederm - No, I can't add reviewers.

jniederm:
@mareklibra can you please add @sjd78? It seems I don't have enough permissions to do it by myself.

sjd78:
(rebased)

src/components/ErrorAlert.js
```diff
@@ -1,17 +1,25 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const ErrorAlert = ({ message, id }) => {
-  return message ? (
+const ErrorAlert = ({ id, message, dangerouslySetInnerHTML }) => {
+  return (
     <div className='alert alert-danger'>
```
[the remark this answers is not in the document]
sjd78: **ErrorAlert** was previously used in 2 places and now 3.  I verified its use in:
- OvirtApiCheckFailed.js
- TokenExpired.js

I did miss VmDialog.js.  I'll adjust it as well.

sjd78:
@gregsheremeta , @jelkosz - Why don't we let this issue/PR close with the UX as is and open a new issue to refactor the UX later (like was done for the popovers in PR #581)

gregsheremeta:
> Why don't we let this issue/PR close with the UX as is and open a new issue

Agree, #598

Issue #590, opened 2018-05-03
Safari browser


mareklibra:
Recent 1.3.9 and former 1.3.8 are most probably not working on Safari on macOS.

Looking for someone with Mac to verify and fix.

PR #591, opened 2018-05-03
Fix for missing pool-name

mareklibra:
Do not fail if REST API provides a pool without name.

Fixes: #590

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/591)


mareklibra:
Would be great to include Safari in our testing matrix.

PR #597, opened 2018-05-10
For 'yarn start' use a different REST endpoint to reliably pull the user's UUID

sjd78:
This works for UserRole and UserVmManager users who cannot currently list themselves via the "/api/users" endpoint.  Just pull the 'authenticated_user' id from the base "/api" endpoint.

This avoids asking the dev for the user's UUID at startup time.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/597)

### DIFF
--- a/scripts/start.js
+++ b/scripts/start.js
@@ -25,19 +25,6 @@ var DEFAULT_PORT = process.env.PORT || 3000;
 var compiler;
 var handleCompile;
 
-// You can safely remove this after ejecting.
-// We only use this block for testing of Create React App itself:
-var isSmokeTest = process.argv.some(arg => arg.indexOf('--smoke-test') > -1);
-if (isSmokeTest) {
-  handleCompile = function (err, stats) {
-    if (err || stats.hasErrors() || stats.hasWarnings()) {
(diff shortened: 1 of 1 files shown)


scripts/start.js
```diff
@@ -350,33 +338,28 @@ function getUserInfo (protocol, port) {
...
-                resolve({
-                  userName: username.slice(0, username.indexOf('@')),
-                  ssoToken: body.access_token,
-                  domain: domain,
-                  userId: userBody.user[i].id,
+      if (body.access_token) {
+        const usersApiUrl = `${engineUrl}/api/`
+        request(usersApiUrl,
+                { json: true, strictSSL: false, headers: { Authorization: `Bearer ${body.access_token}` } }, 
+                (error1, response1, body1) => {
```
jniederm: How about `apiRootError, apiRootResponse, apiRootBody` or something else more descriptive?

sjd78: Gotcha.  Amended the commit with much better var names.

mareklibra:
Works fine. Thanks!

Issue #592, opened 2018-05-03
Edit VM 'Close' without saving fails


mareklibra:
To reproduce:
- Edit a VM
- Change i.e. Name
- click Close

Behavior:
- nothing hapens, console log:
```
TypeError: Cannot read property 'contains' of undefined
    at contains.js:17
    at t.F.focus (Modal.js:557)
    at t.F.onShow (Modal.js:486)
    at t.value (Modal.js:230)
    at commitLifeCycles (react-dom.production.min.js:149)
```

Expected:
The user is navigated to previous page

Observed on 1.3.9 in Chrome

sjd78:
In summary, `VmDialog` uses `Prompt` from **react-router** to intercept link navigation off of the VM create/edit page.  Changes to fields on the page flag `Prompt` to trigger.  The current code to handle displaying a modal confirmation box in `App.js` is broken.  This code will also handle any other `Link`/`Prompt` pairs in the app.

I'm addressing the code in `App.js` now.

sjd78:
Note: The only use of `Prompt` in the app is in **src/components/VmDialog/index.js**.

PR #600, opened 2018-05-15
Refactored the react-router navigation prompt setup

sjd78:
Refactored the way **App.js** handles the react-router navigation prompt modal dialog. `App` is now a stateful component. The state controls if the `NavigationConfirmationModel` is shown and what to fire when the user responds to the modal.

Refactoring includes:
 * removing rednerModal.js and replacing the functionality with a standard patternfly-react Modal
 * renamed CloseDialogConfirmation to NavigationConfirmationModal
 * localized the messages in the new modal

Addresses issue https://github.com/oVirt/ovirt-web-ui/issues/592


sjd78:
Here is an example of the modal when you edit a field on the VM Dialog and click the 'Close' button:

screenshot-2018-5-15 vm portal

I didn't find a clear example of this interaction in patternfly, but I considered [Forms and Controls / Errors and Validations](http://www.patternfly.org/pattern-library/forms-and-controls/errors-and-validation/) and [Session Timeout](http://www.patternfly.org/pattern-library/communication/session-timeout/) when tweaking the style of the `NavigationConfirmationModal`.

src/components/NavigationConfirmationModal/index.js
```diff
@@ -0,0 +1,33 @@
...
+const NavigationConfirmationModal = ({ show, onYes, onNo }) => {
+  const idPrefix = 'close-dialog-confim'
+
+  return (
+    <Modal show={show}>
+      <Modal.Header>
+        <Modal.Title id={`${idPrefix}-title`}>{msg.unsavedChangesTitle()}</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <Alert type='warning' id={`${idPrefix}-body-text`}>{msg.unsavedChangesConfirmMessage()}</Alert>
```
sjd78: Right now the navigation prompt text is defined here.  This should be ok since **VmDialog/index.js** is currently the only place that uses `Prompt`.

If we were to add additional use of react-router `Prompt`, the specific message displayed would be better off defined with the `Prompt` itself and passed through here.  See [VmDialog/index.js:L499](https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/VmDialog/index.js#L499)

sjd78:
Right now the navigation prompt text is defined in **NavigationConfirmationModal/index.js**.  This should be ok since **VmDialog/index.js** is currently the only place that uses **react-router** `Prompt`.

If we were to add additional use of react-router `Prompt`, the specific message displayed would be better off defined with the `Prompt` itself and passed through to the modal.  See [VmDialog/index.js:L499](https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/VmDialog/index.js#L499)

mareklibra:
Works fine.

src/App.js
```diff
@@ -42,52 +41,78 @@ const NoLogin = () => {
...
+        show: false,
+        confirmYes: undefined,
+        confirmNo: undefined,
+      },
+    }
+  }
+
+  resetUserConfirmation (show = false, confirmYes = undefined, confirmNo = undefined) {
+    this.setState({
+      navigationConfirm: { show, confirmYes, confirmNo },
```
bond95: This is a really bad idea to store functions in component state, because `State it's a serializable* representation of one point in time—a snapshot.`
https://github.com/uberVU/react-guide/blob/master/props-vs-state.md

src/App.js
```diff
@@ -42,52 +41,78 @@ const NoLogin = () => {
...
+    switch (visibility.get('dialogToShow')) {
+      case 'Options':
+        detailToRender = (<Options />)
+        break
+    }
+
+    const routes = getRoutes(vms)
+    const menu = getMenu()
+
+    const getUserConfirmation = (message, confirmCallback) => {
```
bond95: I think better, this logic move to separate file.

src/App.js
gregsheremeta: What's a better way?

bond95: Define those functions like methods.

src/App.js
```diff
@@ -45,74 +40,40 @@ const NoLogin = () => {
...
+      </ContainerFluid>
+    )
   }
 
-  resetUserConfirmation (show = false, confirmYes = undefined, confirmNo = undefined) {
-    this.setState({
-      navigationConfirm: { show, confirmYes, confirmNo },
-    })
+  let detailToRender = null
+  switch (visibility.get('dialogToShow')) {
```
sjd78: I think these bits can be cleaned up later.  I don't see how it gets activated or used at all.  I can strip all the dead code out, but I'm not sure that work should be part of this PR.

gregsheremeta:
> It is a much cleaner solution

looks nice to me too

mareklibra:
To store functions in a state object is not a clean solution. It would be definitely no-go approach for the redux store.
But it will cause probably no harm in this particular case. Can't find any. 
@bond95, are you able to describe any example, please?

Otherwise LGTM.

sjd78:
Rebased and fixed the semver for the new dependency (@bond95, @mareklibra)

Issue #568, opened 2018-04-12
Pending changes tooltip overflow


bond95:
![image](https://user-images.githubusercontent.com/3332176/38673273-23f3e322-3e51-11e8-8919-760e9da17c26.png)

Issue #422, opened 2017-11-18
Popup descriptions hidden in edit page


jniederm:
version: commit 5e7eb03

Steps to reproduce:
1. Go to edit VM page
2. Click at "Cluster" label

Actual result:
Only small arrow is shown instead of whole tootlip bubble.

![image](https://user-images.githubusercontent.com/5278760/32981295-fd417d78-cc74-11e7-8cd2-6fefd7a95884.png)

mareklibra:
Fixed by #603

PR #603, opened 2018-05-16
Fix problems with FieldHelp tooltip overflow.

bond95:
Tooltips from time to time overflow container borders.
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/568 https://github.com/oVirt/ovirt-web-ui/issues/422

image
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/603)

### DIFF
--- a/src/components/DetailContainer.js
+++ b/src/components/DetailContainer.js
@@ -25,9 +25,9 @@ class DetailContainer extends Component {
   }
 
   render () {
-    const { children } = this.props
+    const { children, ...props } = this.props
     return (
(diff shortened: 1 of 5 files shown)


mareklibra:
Would LGTM if we can't find a way to avoid extensive passing of `getParentPosition`.

Would it be feasible to set an attribute on an "owning" component and then find it via `parentNode` tree hierarchy from `reactDOM`? In other words, the FieldHelp would try to find first of parent nodes having a specific attribute set (like `render-boundery` boolean or so).

bond95:
@mareklibra Good idea. I tried and it work fine.

PR #604, opened 2018-05-16
Update yarn.lock after last set of merges.

sjd78:
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/604)


mareklibra:
Curious, there is no update related to the `react-router-navigation-prompt` from #600 (after removing the `^` caret from version).

I don't see any relevant change at all, compared to recent master. 
But no harm merging that, the file will become up-to-date.

Issue #558, opened 2018-04-09
Upgrade npm dependencies


mareklibra:
Upgrade referenced npm dependencies (package.json) to the latest versions.

For safety reasons, this will be done within scope of 1.4, will require manual testing.

sjd78:
Do you plan on upgrading the webpack, eslint, babel versions to latest as well?

mareklibra:
webpack is not a priority, but if reasonably feasible, it would be great.
All other packages are planed (incl. babel and eslint)

PR #614, opened 2018-05-22
Batch yarn upgrade

mareklibra:
**This might be breaking change.**
I tried to verify but please watch behavior carefully, might be related.

---

All non-dev dependencies updated plus bunch of others affecting daily workflow.

But namely the `webpack` **stays** at `1.13.2` version.
I managed to transform webpack configuration to `v4` but patternfly is failing to load, needs farther investigation so skipping for now.

---
Fixes: #558

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/614)


src/components/FieldHelp/index.js
```diff
@@ -69,7 +69,7 @@ FieldHelp.propTypes = {
   tooltip: PropTypes.string, // tooltip shown when hovering the text
   children: PropTypes.any,
   container: PropTypes.any,
-  getParentPosition: PropTypes.func,
+  //  getParentPosition: PropTypes.func,
```
jniederm: How about just dropping these commented proptypes? Why are they left commented?

Applies to `PageRouter.js`, `Pages/index.js` as well.

mareklibra: done

src/intl/index.js
```diff
@@ -143,7 +143,7 @@ function removeMessageDescription (messages: { [MessageIdType]: MessageType }):
 
 const defaultMessages: { [MessageIdType]: string } = removeMessageDescription(messages)
 
-function createIdsMap (messages: typeof messages): { [MessageIdType]: MessageIdType } {
+function createIdsMap (messages: any): { [MessageIdType]: MessageIdType } {
```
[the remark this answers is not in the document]
mareklibra: eslint ... 
Symbol is used before declaration

mareklibra:
@jniederm , I agree, I do not like all the default rules either.
I think we should adjust/discuss the rules for this particular project in a separate PR, I'll remove this from here.

mareklibra:
@suomiy , right, they do not pop-up. The bootstrap's toggle based on data-target is happening, but the classes set seem not to be correct. 

I filed #617 for that - I think we should refactor the dialogs to patternfly-react's `Modal` so gaining two benefits at once.

mareklibra:
@jniederm , ESlint stuff separeted to #618.

There is no direct dependency, but I would like to merge this #614 before #618

package.json
```diff
@@ -4,63 +4,63 @@
...
     "path-exists": "3.0.0",
     "postcss-loader": "0.13.0",
     "promise": "7.1.1",
     "react-intl": "2.4.0",
     "readline-sync": "~1.4.5",
-    "recursive-readdir": "2.0.0",
-    "request": "~2.79.0",
-    "rimraf": "2.6.1",
+    "recursive-readdir": "2.2.2",
+    "request": "~2.87.0",
```
jniederm: same

package.json
```diff
@@ -4,63 +4,63 @@
...
     "eslint-config-standard": "~6.2.1",
     "eslint-config-standard-jsx": "~3.2.0",
     "eslint-config-standard-react": "~4.2.0",
     "eslint-loader": "1.5.0",
-    "eslint-plugin-flowtype": "2.32.1",
-    "eslint-plugin-import": "2.7.0",
-    "eslint-plugin-promise": "~3.3.0",
+    "eslint-plugin-flowtype": "2.46.3",
+    "eslint-plugin-import": "2.12.0",
+    "eslint-plugin-promise": "~3.7.0",
```
jniederm: I'm not sure: do we want to avoid version ranges completely or `~` tildas are ok?

mareklibra: sure, I missed that while concentrating on the rest. Removing semver signs

mareklibra:
Merging, let's concentrate on the eslint and especially the modal dialogs.

Issue #562, opened 2018-04-10
InfiniteScroller console warning about unique key


mareklibra:
Console log shows
```
VM27174 warning.js:33 Warning: Each child in an array or iterator should have a unique "key" prop.

Check the render method of `InfiniteScroll`. It was passed a child from Vms. See https://fb.me/react-warning-keys for more information.
    in div (created by Vms)
    in InfiniteScroll (created by Vms)
    in Vms (created by Connect(Vms))
    in Connect(Vms) (created by VmsList)
    in VmsList (created by Connect(VmsList))
```

Works fine anyway, lower priority to make code clean.

mohammad-khoddami:
Add key={0} to the loader fix it. see
https://github.com/oVirt/ovirt-web-ui/pull/616#issue-189719094

arhamymr:
hello, i add key to infinite 

```
 <InfiniteScroll
        hasMore={hasMore}
        loader={<div key={0}/>}
        loadMore={loadFunc}
        pageStart={0}
      >
```

and still have warning in console, someone can help me ?

nateg101:
Yes I too have the same problem - I am using V1.2.4

sgratch:
@nateg101 @arhamymr this is an old issue. Is it reproduced for you on latest version 1.6.4 as well?

frodolfo-duclo:
> Anybody still come out with the same issue? I use version 1.2.6 and there is still the same warning…

I also have this issue with version `1.2.6`, and it's **May 2023** now.

Ruffeng:
If someone is facing this issue, make sure you wrap your component with a div and key:
```ts
loader={
  <div key={0}>
    <MyLoaderComponent />
  </div>
}
````
I was having this warning as well because the key was inside my component, but moving it in the same place as the `loader` prop solved the issue

PR #616, opened 2018-05-22
Fix InfiniteScroller, saga deprecation and Fragment key console warnings

sjd78:
1. Fixes https://github.com/oVirt/ovirt-web-ui/issues/562 as per https://github.com/CassetteRocks/react-infinite-scroller/issues/133. Added `key={0}` to the loader and the InfiniteScroller warning went away.

2. Added an `all([])` wrapper around the rootSaga to resolve the deprecation warning.

3. Refactored `VmDetail/index.js` to add `key={0}` to the `React.Fragment` use as it was causing a warning similar to InfiniteScroller's warning.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/616)

### DIFF
--- a/src/components/VmDetail/index.js
+++ b/src/components/VmDetail/index.js
@@ -314,20 +314,16 @@ class VmDetail extends Component {
                     <FieldHelp content={msg.bootSequenceTooltip()} text={msg.bootSequence()} />
                   </dt>
                   <dd />
-                  <div>
-                    {
-                      vm.getIn(['os', 'bootDevices']).map((device, key) =>
-                        <React.Fragment>
-                          <dt className={style['field-shifted']}>
-                            <FieldHelp content={msg[`${sequence[key]}DeviceTooltip`]()} text={msg[`${sequence[key]}Device`]()} />
-                          </dt>
-                          <dd>
-                            {msg[`${device}Boot`]()}
-                          </dd>
-                        </React.Fragment>
-                      )
-                    }
-                  </div>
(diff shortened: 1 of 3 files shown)


src/components/VmDetail/index.js
```diff
@@ -315,20 +315,16 @@ class VmDetail extends Component {
...
-                          </dt>
-                          <dd>
-                            {msg[`${device}Boot`]()}
-                          </dd>
-                        </React.Fragment>
-                      )
-                    }
-                  </div>
+                  {vm.getIn(['os', 'bootDevices']).map((device, key) =>
+                    <React.Fragment key={0}>
```
bond95: Please, use unique identifier for key parameter. Also this parameter must be a string.

src/components/VmsList/Vms.js
```diff
@@ -47,7 +47,7 @@ class Vms extends React.Component {
       <InfiniteScroll
         loadMore={this.loadMore}
         hasMore={vms.get('notAllPagesLoaded')}
-        loader={<div className={style['loaderBox']}><div className={style['loader']} /></div>}
+        loader={<div key={0} className={style['loaderBox']}><div className={style['loader']} /></div>}
```
bond95: Same.

sjd78: This key doesn't matter since the loader is simply shifted on and off the scroller's children array, but I set it to a string to be type happy.

sjd78:
rebased

mareklibra:
Thanks!

PR #618, opened 2018-05-23
Upgrade ESLint rules

mareklibra:
ESlint npm dependencies are upgraded to the latest.

Some rules are adjusted along their fixes.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/618)


package.json
```diff
@@ -72,6 +72,7 @@
   "dependencies": {
     "bootstrap-select": "1.12.4",
     "bootstrap-switch": "3.3.4",
+    "eslint-plugin-node": "^6.0.1",
```
bond95: Remove please `^`

src/intl/index.js
```diff
@@ -143,7 +142,7 @@ function removeMessageDescription (messages: { [MessageIdType]: MessageType }):
 
 const defaultMessages: { [MessageIdType]: string } = removeMessageDescription(messages)
 
-function createIdsMap (messages: typeof messages): { [MessageIdType]: MessageIdType } {
+function createIdsMap (messages: { [MessageIdType]: MessageType }): { [MessageIdType]: MessageIdType } {
```
jniederm: IIUIC this change is not required by any updated dependency and we just consider this wording more readable, right?

src/intl/index.js
```diff
@@ -159,7 +158,7 @@ function createIdsMap (messages: typeof messages): { [MessageIdType]: MessageIdT
  */
 export const msgId: {[MessageIdType]: MessageIdType} = createIdsMap(messages)
 
-function createFormattingFunctionsMap (messages: typeof messages): {[MessageIdType]: ((?Object) => string)} {
+function createFormattingFunctionsMap (messages: { [MessageIdType]: MessageType }): {[MessageIdType]: ((?Object) => string)} {
```
jniederm: same

src/components/FieldHelp/index.js
```diff
@@ -63,13 +63,12 @@ class FieldHelp extends React.Component {
 }
 
 FieldHelp.propTypes = {
-  title: PropTypes.string,                                            // popover title
```
jniederm: Optional: I find this extra spaces helping with readability.

src/intl/index.js
mareklibra: Eslint fails on using symbol before it's declaration here, the `typeof messages` references the local parameter name `messages`. I saw animated discussion regarding eslint vs. flow, it does not play well together.

Semantically, this change is equal, so let's consider it as a readability improvement.

src/components/FieldHelp/index.js
mareklibra: Shorter lines looks better to me.
I think these spaces would be removed anyway within auto-reformating in most of the IDEs with default setting.

src/intl/index.js
jniederm: It's strange that `typeof messages` worked for me after es-lint update. Let's consider it a readability improvement.

Issue #573, opened 2018-04-18
VM Snapshot management


Brian-Chow:
This functionality was dropped in the VM portal between 4.1 and 4.2.  I would like to see it brought back, as we use(d) it.

Use Case:
  We have created multiple sandbox templates for Developers to use while working on enterprise line of business applications.  Let developers log into the VM portal, create their own VM's, and manage their own VM snapshots from the VM portal, keeping them out of the Administrative Portal.  

Proposed Look/Behavior: Match the snapshot functionality in the Administrative Portal

michalskrivanek:
it's not a reintroduction, this is a different project from the former GWT based User Portal.
The design needs to be done from scratch. If there are no concrete requests for functionality then we'll probably make some assumptions and simplifications.

I believe we can have a list of snapshots in details page(name plus memory status) and allow creating a new snapshot (I'd only support complete snapshots with/without memory) and "revert to" as a single action(with or without memory).

gregsheremeta:
snapshot management is part of the redesign

Brian-Chow:
@michalskrivanek Sorry for the delayed response, your listed assumptions fulfill our need for non-administrator vm users/vm portal users to be able to handle basic snapshot functionality on their own.  Thank you!

gregsheremeta:
cc @jennyhaines @lizsurette

PR #619, opened 2018-05-24
Added snapshots funcionality

bond95:
Also made some refactoring of expanded lists (like NIC, snapshots etc.) and change in VM reducer type to constants.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/573

image
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/619)


jniederm:
Snapshots font color is grey as if they were mere label.

image

src/sagas.js
```diff
@@ -646,6 +656,23 @@ function* fetchVmNics ({ vmId }) {
...
 }
 
+function* fetchVmsSnapshots ({ vms }) {
+  yield all(vms.map((vm) => call(function *() {
+    const snapshotsInternal = yield fetchVmSnapshots({ vmId: vm.id })
+    yield put(setVmSnapshots({ vmId: vm.id, snapshots: snapshotsInternal }))
+  })))
+}
+
+export function* fetchVmSnapshots ({ vmId }) {
```
jniederm: This saga is called from 3 places and each call is immediately followed by storing fetched snapshot to redux. Can we move invocation of `setVmSnapshots` into this saga?

src/actions/vm.js
```diff
@@ -288,6 +293,26 @@ export function setVmSessions ({ vmId, sessions }) {
   }
 }
 
+export function setVmSnapshots ({ vmId, snapshots }) {
```
jniederm: Why are these two action creators here and not in `VmSnapshots/actions.js`?

src/intl/messages.js
```diff
@@ -272,6 +272,12 @@ export const messages = {
   hdBoot: 'Hard Disk',
   networkBoot: 'Network (PXE)',
   cdromBoot: 'CD-ROM',
+  snapshot: 'Snapshots',
+  snapshotsTooltip: 'VM snapshots.',
+  noSnapshots: 'no snapshots',
+  newSnapshot: 'New snapshot',
+  addNewSnapshot: 'Add new snapshot',
+  areYouSureYouWantToDeleteSnapshot: 'Are you sure you want to delete snapshot {description}?',
```
jniederm: optional: Please have some sympathy for our translators and rename `description` to `snapshotName`. They usually don't have any context but the variable name.

src/components/VmSnapshots/index.jsx
```diff
@@ -0,0 +1,152 @@
...
+import style from './style.css'
+
+const MAX_DESCRIPTION_LENGTH = 100
+
+const getMinimazedString = (str, maxChar) => (
+  str.length > maxChar ? `${str.substring(0, maxChar - 3)}...` : str
+)
+
+function isSnapshotBeingDeleted (snapshotId, pendingTasks) {
+  console.log('pendingTasks', pendingTasks.toJS())
```
jniederm: Is this intentional?

jniederm:
This is a New snapshot dialog of a running VM with two disk from webadmin. It would be nice to tell users in web-ui New snapshot dialog what will happen regarding disks and memory (but not providing them any options).
image

jniederm:
I'm missing IMO key feature - "revert to" action (mentioned in assigning [comment][1]). I guess it can be a button between snapshot name and delete button.

[1]: https://github.com/oVirt/ovirt-web-ui/issues/573#issuecomment-385352405

jniederm:
Since users can create both snapshot with and without memory (based on state of the VM when snapshot) and it seems to be quite a difference to resume a VM with vs. without memory how about to render [`pficon-memory`][1] 
image
 icon after name of each snapshot with memory?

@jelkosz, @michalskrivanek , @mareklibra 

[1]: https://www.patternfly.org/styles/icons/

src/components/VmSnapshots/NewSnapshotModal.jsx
```diff
@@ -0,0 +1,93 @@
...
+  close () {
+    this.setState({ showModal: false, description: '' })
+  }
+
+  handleDescriptionChange (e) {
+    this.setState({ description: e.target.value })
+  }
+
+  handleSave () {
+    const snapshot = {
```
jniederm: It would be nice if the dialog can validate that the name is non-empty, non-blanks-only string.

src/ovirtapi.js
```diff
@@ -661,6 +678,25 @@ OvirtApi = {
       contentType: 'application/xml',
     })
   },
+  addNewSnapshot ({ vmId, snapshot }: { vmId: string, snapshot: Object }): Promise<Object> {
```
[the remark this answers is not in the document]
bond95: Actually naming consistency for sagas workers, look like that:

- `addVmSnapshot`
- `deleteVmSnapshot`
- `fetchVmSnapshots`
- `fetchVmsSnapshots`

bond95:
@jniederm Done.

image
image
image
image

src/actions/vm.js
bond95: Because `setVmSnapshots` are used not only within `VmSnapshot` component. `addVmSnapshot` I moved.

jniederm:
*[src/components/VmSnapshots/NewSnapshotModal.jsx, line 74 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/619#-LEFQ6UH2Uz5XSTsXGXT:-LEFQ6UH2Uz5XSTsXGXU:bn2v2q7) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/22b2c9da1e2a9eb5258daf25a47e613dd1fc3908/src/components/VmSnapshots/NewSnapshotModal.jsx#L74)):*
> ```JSX
>                     this.state.emptyDescription &&
>                     <HelpBlock>
>                       {msg.emptySnapshotDescription()}
> ```

It would be nice to dynamically change the message based on VM state or to expand to text of the message to explain that memory will be included only if the VM is running.

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/619)*

jniederm:
> I'm missing IMO key feature - "revert to" action (mentioned in assigning [comment][1]). I guess it can be a button between snapshot name and delete button.

I've checked with @jelkosz and "revert to" button for each existing snapshot line is considered a key part of this feature.

[1]: https://github.com/oVirt/ovirt-web-ui/issues/573#issuecomment-385352405

src/intl/messages.js
```diff
@@ -272,6 +272,15 @@ export const messages = {
...
   networkBoot: 'Network (PXE)',
   cdromBoot: 'CD-ROM',
+  snapshot: 'Snapshots',
+  snapshotsTooltip: 'VM snapshots.',
+  noSnapshots: 'no snapshots',
+  newSnapshot: 'New snapshot',
+  snapshotInfo: 'All attached disks and content of memory will be included in the snapshot.',
+  addNewSnapshot: 'Add new snapshot',
+  areYouSureYouWantToDeleteSnapshot: 'Are you sure you want to delete snapshot {snapshotName}?',
+  emptySnapshotDescription: 'Please, set snapshot description.',
```
mykaul: Rephrase. Look at what oVirt has and try to use the same phrase.
Otherwise:
'Snapshot description is missing' is probably better.

bond95:
Not running VM
image

Running VM
image

Memory tooltip
image

bond95:
@mareklibra Please review.

src/components/VmNics/index.js
```diff
@@ -84,80 +85,51 @@ VmNic.propTypes = {
...
-        </ul>
-        {moreButton}
-        {nicsModal}
-      </div>
-    )
-  }
+  })
+  const nextNicName = nextNicId !== 0 ? `nic${nextNicId}` : 'nic1'
+
+  let renderedNics = nics.sort((a, b) => naturalCompare.caseInsensitive(a.get('name'), b.get('name'))).map(nic => {
```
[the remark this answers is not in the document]
mareklibra: @bond95 , makes sense to me. WDYT? #612 is merged

src/components/VmSnapshots/sagas.js
```diff
@@ -0,0 +1,50 @@
...
+    if (apiSnapshot.error && apiSnapshot.error.status === 404) {
+      snapshotRemoved = true
+      break
+    }
+    yield delay(delaySec * 1000)
+  }
+  if (snapshotRemoved) {
+    yield fetchVmSnapshots({ vmId })
+  }
+  yield put(removeSnapshotRemovalPendingTask(snapshotId))
```
mareklibra: Do I understand correctly, that there's up to 4+4+4+30+60 seconds of no information about a pending task for the user before proper UI state changes? If so, it would be better to show a pending info immediately.

bond95: Actually no, task is added to pending tasks, so user will see marked off list item and "alert" about pending task.

bond95:
@mareklibra done.

mareklibra:
This feature is part of ongoing redesign effort.

I'm fine merging this to fixate current state, no matter there's already plan to rework it in next weeks/months. Just to decrease complexity later and potentially allow hands-on experience when UX-designing. @bond95 , @jelkosz , @gregsheremeta do you agree?

gregsheremeta:
> do you agree?

sounds good to me, esp if you think it's best :)

Issue #574, opened 2018-04-19
After reload Edit VM page, page become empty


bond95:
1. Open edit page of VM that not in first page (9th+ VMs)
2. Reload page

Result:
Empty page.

Expected:
VM edit page.

![image](https://user-images.githubusercontent.com/3332176/38991424-05ff0422-43de-11e8-81c9-92a117051a9f.png)

sjd78:
The same issue occurs for VM view and Pool view.  If the VM or Pool is in fetch page >1, the view component won't ever be shown.  The main difference is that the VM edit page will sit on a blank page while the VM and Pool view pages will redirect the user back to the card view.

Both ways are incorrect.

For all 3 pages, if the VM or Pool cannot be loaded, an error message should be displayed.  If the VM or Pool isn't in the first fetch page, it should be fetched directly.

bond95:
@sjd78 the problem is in `loadInProgress` flag of VM reducer. When you open VM detail page it try to fetch VM and VMs list, and there is the question, what will be fetched first. If VMs list, thus it set `loadInProgress` flag to `false`,  `VmDetailPage` component look at it, and will think that VM was not found, thus return `null`. Same with pools.

sjd78:
Plus, there currently is no handling for if the **vmId** on the URL is invalid or if it doesn't exists anymore.

bond95:
> VmDetailPage and PoolDetailPage both wait until login has finished (which for some reason is Selectors.isFilterChecked()) before fetching the VM/Pool directly. VmDialogPage didn't so the Filter API headers were not set correctly and the API.getVm got a 404. I updated them to all match and the VM loads (without a 404 since the API Filter headers are set correctly now) and the edit page renders.

Hmm... I never got this error before.

> Initially I thought it was a race between a single VM fetch and the initial data load and who flipped loadInProgress false first. However, this is not the core problem. And really it doesn't matter since the pages are connected to the state.vms and would get re-rendered on update regardless of which one gets merged to the store first.

For prove that problem is in `loadInProgress`, I did this trick in `VmDetailPage`:
```
    /*return <Redirect to='/' />*/
    return null
```
And everything works fine, problem gone, thus problem is in `loadInProgress` and fetch race.

[...]

PR #621, opened 2018-05-29
"Loading ..." indicator based on pending http requests

jniederm:
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/621)


src/reducers/clusters.js
```diff
@@ -3,7 +3,6 @@ import { actionReducer, removeMissingItems } from './utils'
 
 const initialState = fromJS({
   clusters: {},
-  loadInProgress: true,
```
sjd78: Yay!  Those were on my list to remove as well.

src/index.js
```diff
@@ -107,6 +108,18 @@ function start () {
...
 }
 
+function initializeApiListener (store: Object) {
+  OvirtApi.addHttpListener((requestId, eventType) => {
+    if (eventType === 'START') {
+      store.dispatch(addActiveRequest(requestId))
+      return
+    }
+    if (eventType === 'STOP') {
+      store.dispatch(delayedRemoveActiveRequest(requestId))
```
sjd78: So the purpose of this is to leave the "Loading..." indicator on screen for an extra half second?  Is that to prevent flashing if a bunch of XHR requests are done one after another?

src/reducers/index.js
```diff
@@ -49,4 +50,5 @@ export default combineReducers({
   OptionsDialog,
   NewDiskDialog,
   pendingTasks,
+  activeRequests,
```
sjd78: I like this at the core level instead of pinned in **vms**.

src/ovirtapi.js
```diff
@@ -839,6 +867,19 @@ OvirtApi = {
     OvirtApi._assertLogin({ methodName: 'getNetworks' })
     return OvirtApi._httpGet({ url: `${AppConfiguration.applicationContext}/api/networks` })
   },
+  addHttpListener (listener: ListenerType) {
+    listeners.add(listener)
+  },
+}
+
+function notifyStart (method: MethodType, url: string): Object {
+  const requestId = { method, url }
```
sjd78: ~~Will this be unique enough?~~  It should be since you're using a Set from ImmutableJS in the reducer to keep track.

If 2 requests with the same URL get fired right after each other (or concurrently), the requestId would be removed on completion of the first request.  If the second doesn't complete within 500ms the "Loading..." indicator would go away while the request is still in progress.  That's a long interval, and I can't think of a particular scenario where it would happen off the top of my head.

I guess the question is, would it be acceptable to have the "Loading..." indicator disappear in an edge case while a second request is long running (over 500ms)?

src/components/Pages/index.js
```diff
@@ -41,10 +41,10 @@ class VmDetailPage extends React.Component {
   }
 
   render () {
-    let { match, vms, config } = this.props
+    let { match, vms, config, requestActive } = this.props
     if (vms.getIn(['vms', match.params.id])) {
       return (<VmDetail vm={vms.getIn(['vms', match.params.id])} config={config} />)
-    } else if (vms.get('loadInProgress')) {
+    } else if (requestActive) {
```
sjd78: So on a VM view page, click the Run (or Stop) button.  The **requestActive** will go true for the duration of the REST request.  Will the page then render a null component and leave the user blank until the REST request is complete?

On a good note, the next PR I'm posting (to deal with #574) will remove this logic all together in this location.

src/index.js
mareklibra: yes

src/ovirtapi.js
```diff
@@ -34,6 +38,7 @@ OvirtApi = {
   },
   _httpGet ({ url, custHeaders = {} }: { url: string, custHeaders?: Object}): Promise<Object> {
     logDebug(`_httpGet start: url="${url}"`)
+    const requestId = notifyStart('GET', url)
```
mareklibra: I would add integer global counter to make this requestId unique, the `url` is not enough.

src/ovirtapi.js
mareklibra: If already refactoring, I would make it reliable. The `requestId` should be unique for each HTTP call. Please see comment above, simple global counter can ensure that.

I would not re-compose the requestId here again since it's already passed at calling (so allign it with the `notifyStop()`)

src/index.js
```diff
@@ -107,6 +108,18 @@ function start () {
     .then(onResourcesLoaded)
 }
 
+function initializeApiListener (store: Object) {
+  OvirtApi.addHttpListener((requestId, eventType) => {
+    if (eventType === 'START') {
+      store.dispatch(addActiveRequest(requestId))
```
mareklibra: I like keeping this in the store

src/index.js
jniederm: @sjd78 as mentioned in the commit message your worries came true

src/ovirtapi.js
jniederm: * The requestId properties `method` and `url` are there for debugging purposes only. The notify methods as well as immutable.js Set works with object identity (`===`) so the uniqueness is inherently ensured (Items in the Set are pure JS objects, not immutable.js instances). In another words it would work with empty objects (`{}`) as well. Please let me know in case the `requestId` serial number is required even for debugging purposes only.
* Regarding place of creation of requestId: It's created in `notifyStart()` since content of the requestId should be transparent to the calling party. The contract is supposed to be like this: Call `notifyStart()`, save the token (requestId) and whenever appropriate cancel that notification by passing the token to `notifyStop()`.
* It actually works - multiple concurrent requests in using the same method to the same url creates multiple records in the `/activeRequests` Set.

src/components/Pages/index.js
jniederm: The page will be rendered correctly (verified) since condition `vms.getIn(['vms', match.params.id])` at line 45 evaluates to `true`.

src/ovirtapi.js
mareklibra: @jniederm , you are right, I oversaw that.

src/sagas.js
```diff
@@ -700,6 +697,11 @@ function* schedulerPerMinute (action) {
   }
 }
 
+function* delayedRemoveActiveRequest ({ payload: requestId }) {
```
bond95: I think this action doesn't give any sense, because chance, that next request will be in 1 ms after `removeActiveRequest`, the same as if next request will be in 501 ms.

bond95: I think this action doesn't give any sense, because chance, that next request will be in 1 ms after `removeActiveRequest`, the same as if next request will be in 501 ms.

src/reducers/clusters.js
jniederm: It's nice to be on the same page.

src/sagas.js
jniederm: It's there to prevent most of the flickering (practically all of it). Sure there can still occur blinks. Any ideas how to improve that?

mareklibra: Maybe I miss something, but this delay improves blinking significantly.
The root cause for it is mostly in "batch" load composing of multiple calls, so executed tightly together.

My first guess was one or two seconds, but let's see - 500ms might be good enough

PR #622, opened 2018-05-29
 Templates state lifted one level up

jniederm:
Only the last commit is relevant. The rest is dependency on #621 

---

/templates/templates -> /templates because there isn't any other property
at level /templates any more. Previously there was
/templates/loadInProgress.

ADD_TEMPLATES and REMOVE_MISSING_TEMPLATES actions replaced by single
SET_TEMPLATES since they always (once) occured together.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/622)


mareklibra:
LGTM in general, let's wait for #621 merge.

@jniederm , were you "git grep" all (potential) occurrences? We need thorough testing for this kind of changes since there's probably no other reasonable way to verify unless really executed.

jniederm:
The code was grepped, I've verified that new VM page and edit VM pages can be load correctly.

sjd78:
Rebase now that #621 is merged and LGTM.

I like the SET style instead of the ADD / REMOVE, especially for this section of the store.

src/sagas.js
```diff
@@ -525,10 +524,7 @@ function* fetchAllTemplates (action) {
 
   if (templates && templates['template']) {
     const templatesInternal = templates.template.map(template => Api.templateToInternal({ template }))
-    yield put(addTemplates({ templates: templatesInternal }))
-
-    const templateIdsToPreserve = templatesInternal.map(item => item.id)
-    yield put(removeMissingTemplates({ templateIdsToPreserve }))
+    yield put(setTemplates(templatesInternal))
```
sjd78: I like SET instead of ADD/REMOVE for templates (and for the other reducers like templates that are really only set by refreshData.

src/reducers/templates.js
```diff
@@ -1,24 +1,18 @@
...
-    })
-    const imUpdates = fromJS(updates)
-    return state.mergeIn(['templates'], imUpdates)
-  },
+const initialState = fromJS({})
 
-  REMOVE_MISSING_TEMPLATES (state, { payload: { templateIdsToPreserve } }) {
-    return removeMissingItems({ state, subStateName: 'templates', idsToPreserve: templateIdsToPreserve })
+const templates = actionReducer(initialState, {
+  [SET_TEMPLATES] (state, { payload: templates }) {
```
sjd78: I like that style, using the constant's value as the key.

PR #626, opened 2018-05-31
 Operating systems state lifted one level up 

jniederm:
Only the last commit is relevant. The rest is to express dependency on #624.

---

/operatingSystems/operatingSystems -> /operatingSystems because there isn't
any other property at level /operatingSystems any more. Previously there
was /operatingSystems/loadInProgress.

ADD_ALL_OS and REMOVE_MISSING_OSS actions replaced by single
SET_OPERATING_SYSTEMS since they always (once) occured together.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/626)

PR #624, opened 2018-05-30
 Clusters state lifted one level up 

jniederm:
Only the last commit is relevant, the rest is to express dependency on #622 

---

/clusters/clusters -> /clusters because there isn't any other property
at level /clusters any more. Previously there was
/clusters/loadInProgress.

ADD_CLUSTERS and REMOVE_MISSING_CLUSTERS actions replaced by single
SET_CLUSTERS since they always (once) occured together.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/624)

PR #627, opened 2018-05-31
 Hosts state lifted one level up 

jniederm:
Only the last commit is relevant. The rest is to express dependency on #626 

---

/hosts/hosts -> /hosts because there isn't any other property at level
/hosts any more. Previously there was /hosts/loadInProgress.

ADD_HOSTS and REMOVE_MISSING_HOSTS actions replaced by single
SET_HOSTS since they always (once) occured together.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/627)


sjd78:
Rebase now that #621 is merged and LGTM.

Looks like it's unique enough to be in parallel with #622 and #624 and #626.

mareklibra:
#622 is merged, no conflicts are reported

mareklibra:
@jniederm , what about moving this #627 to #626? Can be merged together ...

jniederm:
Rebased. We can merged only this and omit #626 to get all code in. Or we can use "Rebase and merge" option (see below) on both PRs and it will work smoothly.

image

Issue #563, opened 2018-04-12
Redirect to main page after F5 on VM Detail page


bond95:
When on VM detail I click F5 it redirect me to main page. Problem is in `loadInProgress` of VM reducer. When VM has a lot of data, faster requests can set `loadInProgress` to `false`, thus it redirect to main page, because load not in progress, and VM not in reducer state.

How to reproduce:
1. Go to VM with a lot of additional data, like NICs or Disks
2. Click F5

Reproducible:
Not often, only special cases.

PR #628, opened 2018-06-01
View and Edit pages will wait for their VM/Pool to be loaded

sjd78:
Issue #574 was caused when a user would enter the app at a view or edit page with a VM or Pool that would be on a fetch page >1. The VM/Pool would be requested but often would either fail the request or would not arrive prior the initial data load flag being set true. The concurrency issues would leave the user either on a blank page or redirected to the card view page.

At a high level, a number of changes have been made to eliminate this issue:
1. The `login` saga was refactored and the `schedulerFixedDelay` is started after the initial data load

1. The `App` component will wait until the login process has completed enough activity such that subsequent REST calls can be made without failing with permission errors before it renders any routes.  Breadcrumbs are rendered with VM/Pool ids until the data is available to display.

1. The `refreshData` saga was updated such that VMs/Pools are fetched individually as needed and that no valid VM or Pool is removed.

1. The view/edit `Pages` components will monitor the VM/Pool id they're given to display and force a deep fetch of the VM/Pool when the id changes (either from nothing to something at page load time, or by a user selecting a different item to display).

---
Detailed changes:

 - `App` component will wait for login to complete before rendering routes. This will allow the route Pages to assume all relevant login data needed is available so additional API calls may be made.

 - Route titles (visible in the breadcrumbs) will show the requested VM or Pool UUID until the VM or Pool name is available from the store.

 - Each detail/edit route page will (deep) fetch the VM or Pool requested every time the requested id changes. This keeps the detail/edit information very current.

 - Each detail/edit route page will sit with only the app header and breadcrumbs rendered until the VM or Pool is available in the redux store. The "Loading..." indicator should be present while fetching data.

 - Since the refresh scheduler fetches data from the REST API, wait until the login saga is successful before starting. This prevents erroneous refreshes if the initial load fails or takes longer then the scheduler delay.

 -  The single VM fetch REST call can now __follow__ all the data paths like the fetch by page does. This reduces the number of REST calls required from 6 to 1. It now also supports both shallow and deep fetches.

 - Removal of 'stale' VMs and Pools from the store was moved to the refresh saga so it can be done once AND preserve ALL VMs and Pools regardless of how they were fetched.

 - `refreshData` now checks VM/Pool ids refreshed by page/count against the set of VM/Pool ids currently in the redux store. If there are any ids from the store not already updated, they will be fetched individually. No valid VM/Pool will be removed from the redux store.

Additionally:

 - The 'selecting VM details' unit test needed to be fixed by initializing a mock store in the test. Specifically the __selectors__ needed to be initialized or they throw an exception. The initial state of reducers used by __selectors__ are now exported to be used in the unit test.

 - Moved the loading of console options out of `VmsListNavigation` and put it in `VmDetailPage`.  The Page components are solely responsible for refreshing the data needed for their view/edit components.

 - Removed **isPool** prop from `VmDetail` since it can be derived from other props.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/628)


src/components/Pages/index.js
```diff
@@ -26,31 +24,36 @@ class VmDetailPage extends React.Component {
...
   }
 
-  componentWillMount () {
-    if (Selectors.isFilterChecked()) {
-      this.props.getVms({ vmId: this.props.match.params.id })
+  fetchTheVmIfNeeded () {
+    const { match, vms, getVms } = this.props
+    let requested = false
+
+    if (!this.requestSent && !vms.getIn(['vms', match.params.id])) {
```
mareklibra: Originally, the VM was forced to reload when the detail page is opened.

The idea is: immediately show what's in store and update once/if changes arrive. I would like to keep it.

src/App.js
```diff
@@ -53,27 +53,32 @@ const App = ({ vms, config }) => {
...
+}
+App.defaultProps = {
+  appReady: false,
 }
 
 export default connect(
   (state) => ({
     vms: state.vms,
     config: state.config,
+    appReady: state.config.get('isFilterChecked'), // When is the app ready to display data components?
```
mareklibra: Ugly imperative approach. But probably required by router.

src/components/Pages/index.js
```diff
@@ -26,31 +24,36 @@ class VmDetailPage extends React.Component {
...
     if (vms.getIn(['vms', match.params.id])) {
       return (<VmDetail vm={vms.getIn(['vms', match.params.id])} config={config} />)
-    } else if (vms.get('loadInProgress')) {
-      console.info(`VmDetailPage: VM id cannot be found: ${match.params.id}. Load is still in progress - waiting before redirect`)
-      return null
     }
 
-    console.info(`VmDetailPage: VM id cannot be found: ${match.params.id}. Redirecting to / `)
-    return <Redirect to='/' />
+    // TODO: Add handling for if the fetch runs but fails (FETCH-FAIL)
```
mareklibra: Yep, let's not forget this.

src/components/Pages/index.js
```diff
@@ -80,29 +83,36 @@ class PoolDetailPage extends React.Component {
...
   }
 
-  componentWillMount () {
-    if (Selectors.isFilterChecked()) {
-      this.props.getPools({ poolId: this.props.match.params.id })
+  fetchThePoolIfNeeded () {
+    const { match, vms, getPools } = this.props
+    let requested = false
+
+    if (!this.requestSent && !vms.getIn(['pools', match.params.id, 'vm'])) {
```
mareklibra: likewise

src/components/Pages/index.js
```diff
@@ -80,29 +83,36 @@ class PoolDetailPage extends React.Component {
...
 
     if (vms.getIn(['pools', match.params.id, 'vm'])) {
-      return (<VmDetail vm={vms.getIn(['pools', match.params.id, 'vm'])} pool={vms.getIn(['pools', match.params.id])} config={config} isPool />)
-    } else if (vms.get('loadInProgress')) {
-      return null
+      return (<VmDetail vm={vms.getIn(['pools', match.params.id, 'vm'])} pool={vms.getIn(['pools', match.params.id])} config={config} />)
     }
-    return <Redirect to='/' />
+
+    // TODO: Add handling for if the fetch runs but fails (FETCH-FAIL)
```
mareklibra: just a reminder

src/saga/login.js
```diff
@@ -179,24 +184,23 @@ function* autoConnectCheck (action) {
...
       const internalVm = Api.vmToInternal({ vm })
       yield put(updateVms({ vms: [internalVm] }))
-
       yield downloadVmConsole(downloadConsole({ vmId }))
     }
   }
 }
 
-function* initialLoad () {
+function* initialLoad () { // NOTE: This is an in-order fetch, could be made parallel?
```
mareklibra: It is: https://redux-saga.js.org/docs/api/#putaction

New action is dispatched.
And since `rootSaga` is created via [all](https://redux-saga.js.org/docs/api/#alleffects---parallel-effects), it should be executed in parallel. Please report/fix if you see otherwise.

src/ovirtapi.js
```diff
@@ -544,9 +544,12 @@ OvirtApi = {
     const url = `${AppConfiguration.applicationContext}/api/`
     return OvirtApi._httpGet({ url })
   },
-  getVm ({ vmId }: VmIdType): Promise<Object> {
+  getVm ({ vmId, additional }: { vmId: string, additional: Array<string> }): Promise<Object> {
```
mareklibra: Is the `additional` set somewhere?

src/sagas.js
```diff
@@ -164,39 +176,66 @@ function* fetchIcon ({ iconId }) {
...
+  const fetchedPoolIds = yield fetchPoolsByCount(getPoolsByCount({
+    count: action.payload.page * AppConfiguration.pageLimit,
+  }))
+
+  yield put(removeMissingPools({ poolIdsToPreserve: [ ...fetchedPoolIds, ...poolsFetchedById ] }))
+  yield put(removePoolsFetchedById({ poolIds: fetchedPoolIds }))
+
+  // update counts
+  yield put(updateVmsPoolsCount())
   yield put(loadInProgress({ value: false }))
```
mareklibra: the `loadInProgress` refactoring was just merged, please review.

src/sagas.js
```diff
@@ -286,27 +326,66 @@ function* fetchVmsByCountVLower (action) {
...
+  const { vmId, shallowFetch } = action.payload
+  yield startProgress({ vmId, name: 'refresh_single' })
+
+  const isOvirtGTE42 = compareVersionToCurrent({ major: 4, minor: 2 })
+  if (isOvirtGTE42 && !shallowFetch) {
+    action.payload.additional =
+      ['cdroms', 'sessions', 'disk_attachments.disk', 'graphics_consoles', 'nics']
+  }
+
+  yield put(addVmsFetchedById({ vmIds: [ vmId ] }))
```
mareklibra: Do I understand correctly, that this state/actions around `*FetchedById` are just because the regular refresh reads recently the paginated data only? Means the VM requested when enetering the application via URL with VM ID would not be regularly refreshed?

If so, I would like to avoid this extra level of state information.
I might be enough for the `refreshData` to check the state for list of recent VMs (those already downloaded for whatever reason) and compute difference between those already stored and reloaded with pagination.

I believe this would be much simpler and less error prone in the future since we will maintain just a single list.

src/saga/login.js
```diff
@@ -48,45 +49,45 @@ import {
   downloadVmConsole,
 } from './consoles'
 
+/**
+ * Perform login checks, and if they pass, perform initial data loading
+ */
 export function* login (action) {
+  const { payload: { token, userId, credentials: { username } } } = action
   yield put(loadInProgress({ value: true }))
```
mareklibra: loadInProgress ...

src/saga/login.js
```diff
@@ -179,24 +184,23 @@ function* autoConnectCheck (action) {
...
+function* initialLoad () { // NOTE: This is an in-order fetch, could be made parallel?
   yield put(getAllClusters()) // no shallow
   yield put(getAllHosts())
   yield put(getAllOperatingSystems())
   yield put(getAllTemplates({ shallowFetch: false }))
   yield put(getAllStorageDomains())
   yield put(getAllVnicProfiles())
 
-  yield put(getByPage({ page: 1 })) // first page of VMs list
+  yield put(getByPage({ page: 1 })) // first page of VMs list, sets _loadInProgress_ to false when done
```
mareklibra: Please remove this comment, would be confusing with the just merged patch refactoring this.

src/components/Pages/index.js
bond95: Agree, because store after initial VMs fetch doesn't have all data needed for detail page.

src/reducers/vms.js
```diff
@@ -63,23 +73,14 @@ const vms = actionReducer(initialState, {
     }
     return state
   },
-  [SET_VM_NICS] (state, { payload: { vmId, nics } }) {
```
bond95: Actually, we want to use this constants in reducers, for saving consistency.

src/sagas.js
```diff
@@ -440,27 +498,22 @@ function* startProgress ({ vmId, poolId, name }) {
   }
 }
 
-function* getSingleInstance ({ vmId, poolId }) {
```
bond95: Why do you remove this function? It fix bug with pool updating.

src/reducers/vms.js
```diff
@@ -171,3 +192,6 @@ const vms = actionReducer(initialState, {
 })
 
 export default vms
+export {
```
bond95: Do we really need this export, becauase I don't see any use of it?

sjd78: The initialState exports are for the unit test.  I needed to create a skeleton store for the selectors so the test would work.  Without the store the selectors throw an exception and the build fails.

src/reducers/vms.js
sjd78: True.  I'll follow up this PR with a refactor PR to fix those in all reducers.

src/saga/login.js
sjd78: Rebase will address this

src/saga/login.js
sjd78: Rebase will address this

src/sagas.js
sjd78: Rebase will address this

src/components/Pages/index.js
sjd78: I'll address this in a followup PR.  I will create an issue for it later today.  It'll be something like "Handle 404 conditions gracefully".

src/components/Pages/index.js
sjd78: I'll address this in a followup PR.

src/components/Pages/index.js
sjd78: I'll address this in a followup PR.

src/App.js
sjd78: What would be a better way to change App render behavior based on getting the state to a particular minimum configuration?  If something like that isn't here, the first REST call can fail because the permissions and filters aren't setup yet (by the login saga)

sjd78:
Rebase is pending.  I'll push a new rebase/merge commit so the existing comments don't go wonky.

src/sagas.js
sjd78: I don't remember doing that, maybe I missed it in a rebase conflict resolution.  I'll put it back.

src/ovirtapi.js
sjd78: It's the same `additional` as `getVmsByPage` or `getVmsByCount`.  The saga function `fetchSingleVm()` uses it to do a deep fetch of a single VM, falling back to the 6 REST calls needed in pre-4.2 versions.

src/components/Pages/index.js
sjd78: #631

sjd78: #631

src/components/Pages/index.js
sjd78: #631

src/components/Pages/index.js
sjd78: #631

mareklibra:
The [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/628) tool keeps history.

src/App.js
mareklibra: agree, I don't know

src/components/Pages/index.js
mareklibra: ok

mareklibra:
> Actually I think, this fix could be more simple. @jniederm PR was just merged, in which he monitoring requests. With some changes, we could monitoring certain single VM fetching and find out if it finished successful or if it fail. @sjd78 @mareklibra @jniederm what do you think?

I would not extend the low-level HTTP request monitoring for this kind of functionality, I'm afraid of introducing spaghetti this way.

Anyway, I agree this can be done simpler way. Copy paste of my former inscribed comment as a proposal:

> Do I understand correctly, that this state/actions around *FetchedById are just because the regular refresh reads recently the paginated data only? Means the VM requested when enetering the application via URL with VM ID would not be regularly refreshed?

> If so, I would like to avoid this extra level of state information.
It might be enough for the refreshData to check the state for list of recent VMs (those already downloaded for whatever reason) and compute difference between those already stored and reloaded with pagination.

> I believe this would be much simpler and less error prone in the future since we will maintain just a single list.

src/saga/login.js
```diff
@@ -121,13 +121,21 @@ function composeIncompatibleOVirtApiVersionMessage (oVirtMeta) {
   return `oVirt API version requested >= ${requested}, but ${found} found`
 }
 
+/**
+ * Compare the actual { major, minor } version (JS or ImmutableJS Map) to the required
+ * { major, minor } and return if the **actual** is greater then or equal to **required**.
+ * Backward compatibility of the API is assumed.
+ */
```
jniederm: IMO it's not responsibility of this function to convert data from immutable wrapper. How about using `.toJS()` at place invocation place.

src/sagas.js
```diff
@@ -163,37 +175,62 @@ function* fetchIcon ({ iconId }) {
...
 
 function* refreshData (action) {
-  console.log('refreshData(): ', action.payload)
-  if (!action.payload.quiet) {
-    logDebug('refreshData(): not quiet')
+  logDebug('refreshData(): ', action.payload)
+
+  // refresh VMs and remove any that haven't been refreshed
+  const vmsFetchedById = []
+  for (let vmId of Selectors.getVmsFetchedById().toArray()) {
```
jniederm: How about fetching by count first and then subtract fetched-by-count-vm-ids from `Selectors.getVmsFetchedById()` and fetch only the ids from subtraction result? It would prevent fetch duplicates.

src/sagas.js
```diff
@@ -163,37 +175,62 @@ function* fetchIcon ({ iconId }) {
...
+    const status = yield fetchSingleVm(getSingleVm({ vmId }))
```
jniederm: Why fetching "by id" doesn't follow `action.payload.shallowFetch` however fetch "by count" does?

src/sagas.js
jniederm: This can be parallelized by using `all()`.

src/sagas.js
```diff
@@ -282,27 +320,66 @@ function* fetchVmsByCountVLower (action) {
...
+
+  const isOvirtGTE42 = compareVersionToCurrent({ major: 4, minor: 2 })
+  if (isOvirtGTE42 && !shallowFetch) {
+    action.payload.additional =
+      ['cdroms', 'sessions', 'disk_attachments.disk', 'graphics_consoles', 'nics']
+  }
+
+  yield put(addVmsFetchedById({ vmIds: [ vmId ] }))
+  const vm = yield callExternalAction('getVm', Api.getVm, action, true)
+  let internalVm = false
```
jniederm: Can we initialize it to `null`/`undefined` or not initialize it at all? I believe it's beneficial to keep the variable type as narrow as possible.

src/sagas.js
```diff
@@ -397,7 +445,15 @@ function* fetchConsoleMetadatas ({ vms }) {
   })
 }
 
-function* fetchVmDisks ({ vmId }) {
+export function* fetchDisks ({ vms }) {
+  yield * foreach(vms, function* (vm) {
```
jniederm: This can be parallelized  by saga's `all()`.

src/sagas.js
```diff
@@ -397,7 +445,15 @@ function* fetchConsoleMetadatas ({ vms }) {
...
+    const vmId = vm.id
+    const disks = yield fetchVmDisks({ vmId })
+    yield put(setVmDisks({ vmId, disks }))
+  })
+}
+
+function* fetchVmDisks ({ vmId }) { // Would __additional__ (API param __follow_) work here as well?
```
jniederm: AFAIK `follow` parameter works for any subresource starting engine 4.2, http://ovirt.github.io/ovirt-engine-api-model/master/#documents/003_common_concepts/follow

src/sagas.js
```diff
@@ -436,27 +492,22 @@ function* startProgress ({ vmId, poolId, name }) {
...
 
-function* getSingleInstance ({ vmId, poolId }) {
-  const fetches = [ fetchSingleVm(getSingleVm({ vmId })) ]
-  if (poolId) {
-    fetches.push(fetchSinglePool(getSinglePool({ poolId })))
-  }
-  yield all(fetches)
-}
-
 function* stopProgress ({ vmId, poolId, name, result }) {
```
jniederm: obsolete

src/saga/login.js
sjd78: Ok, that makes sense.  `put` dispatches an async message to redux that will get reduced into the store sometime in future.

It's all good, I'll take the comment out.

src/sagas.js
sjd78: > Do I understand correctly, that this state/actions around `*FetchedById` are just because the regular refresh reads recently the paginated data only? Means the VM requested when enetering the application via URL with VM ID would not be regularly refreshed

No only would the VM not have been refreshed by `refreshData`, but it would have been removed completely from the list.  Refresh generally follows this path:
 - Fetch the pages the user has seen
 - Merge the fetched data to the store
 - Remove VMs from the store that were not just fetched

> If so, I would like to avoid this extra level of state information.
It might be enough for the refreshData to check the state for list of recent VMs (those already downloaded for whatever reason) and compute difference between those already stored and reloaded with pagination.

So, fetch by page, subtract the page fetched VMs from the list of VMs in the store and then fetch each of those VMs individually and either update (200 response) or remove from the store (404 response)?

Sounds like that would work.  I'll try it out.

src/App.js
```diff
@@ -51,25 +51,30 @@ const App = ({ vms, config }) => {
...
     </Router>
   )
 }
 App.propTypes = {
   vms: PropTypes.object.isRequired,
   config: PropTypes.object.isRequired,
+  appReady: PropTypes.bool,
+}
+App.defaultProps = {
+  appReady: false,
```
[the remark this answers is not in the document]
sjd78: Probably just not the right amount of coffee the day I did that.  Makes sense to make it required and leave it to the `connect()` and be done with it as you suggest.

src/components/Pages/index.js
```diff
@@ -125,28 +135,50 @@ const PoolDetailPageConnected = connect(
...
+
   componentWillMount () {
     this.props.getCDRom()
 
-    // in case the location is entered from outside, refresh data
-    if (this.props.match.params.id) {
-      this.props.getVms({ vmId: this.props.match.params.id })
-    }
+    const requested = this.fetchTheVmIfNeeded()
+    if (requested) console.info(`VmDialogPage: WillMount requesting: ${this.props.match.params.id}`)
```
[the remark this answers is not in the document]
sjd78: I don't mind the single line `if` like that for logging.  All depends on the lint style I suppose.

Could also have gone with:
```javascript
  requested && console.info(`VmDialogPage: ...`)
```
but we removed all of those from dashboard in favor of `if (condition) log` style.

Also, the logging here could be removed (or turned debug instead).  There isn't much use for it that wouldn't show up in the redux logging.

src/components/VmDetail/index.js
```diff
@@ -98,20 +98,21 @@ class VmDetail extends Component {
...
-      pool,
       onRDP,
-      operatingSystems,
-      config,
     } = this.props
 
     const sequence = [ 'first', 'second' ]
 
+    const isPool = !!pool
+    const isPoolVm = !!vm.getIn(['pool', 'id'], false)
```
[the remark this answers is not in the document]
sjd78: I had a bug with that in the VmActions component, but the `!!` will take care of ensuring `isPoolVm` is a boolean.

src/saga/login.js
sjd78: It may not be its pure responsibility but it sure is convenient.

Since it's only used in `saga.js`, I'll move it to the helper over there.

src/sagas.js
sjd78: @mareklibra has a similar (or really the same) suggestion.  I'm gonna try it out and see how it looks.

src/sagas.js
sjd78: Good question that I didn't spend a lot of time trying to unwind.

Really I think it can be reduced to saying that the fetch by page/count are intended to fetch the minimal info needed to support the card list and left navigation while the fetch by ID is intended to fetch all for a detail/edit page.  It's unfortunate that the store overlaps for these purposes.

It's something that deserves a bit of a deeper think.

src/sagas.js
sjd78: True.  I don't think it'll make much of a difference since I don't see the VMs fetched by ID list getting very long.

I'll revisit.

src/components/Pages/index.js
```diff
@@ -81,29 +84,36 @@ class PoolDetailPage extends React.Component {
...
     if (vms.getIn(['pools', match.params.id, 'vm'])) {
-      return (<VmDetail vm={vms.getIn(['pools', match.params.id, 'vm'])} pool={vms.getIn(['pools', match.params.id])} config={config} isPool />)
-    } else if (requestActive) {
-      return null
+      return (<VmDetail vm={vms.getIn(['pools', match.params.id, 'vm'])} pool={vms.getIn(['pools', match.params.id])} config={config} />)
     }
-    return <Redirect to='/' />
+
+    // TODO: Add handling for if the fetch runs but fails (FETCH-FAIL)
+    console.info(`PoolDetailPage: Pool id cannot be found: ${match.params.id}`)
```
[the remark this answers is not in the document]
sjd78: I think the logging in general should be revisited across the code base.  `logDebug()` is ok, but there are instances where I lost important call stack/location details by using it instead of `console.(log|debug|info|warn)`.

For now, I'd say it's beyond the scope of the issue this PR is addressing.

src/sagas.js
```diff
@@ -282,27 +320,66 @@ function* fetchVmsByCountVLower (action) {
...
     }
   }
 
   yield put(persistState())
+  return fetchedVmIds
+}
+
+export function* fetchSingleVm (action) {
+  const { vmId, shallowFetch } = action.payload
+  yield startProgress({ vmId, name: 'refresh_single' })
```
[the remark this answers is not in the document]
sjd78: And is a no-op if the VM doesn't exist in the store.

src/sagas.js
sjd78: I will leave `startProgress` and `stopProgress` in for now since they're used only for VM actions and the `VmActions` toolbar buttons use that state indicator as part of their enabled state calculations.

src/sagas.js
sjd78: Agreed.  I'll adjust the comment to list is as a TODO.  It can be addressed in the future since it works now but could work much better.

sjd78:
Two main points to deal with after the first set of reviews:
 - `refreshData` can probably be done without keeping a separate list of VMs fetched by id.  It wasn't obvious (to me at least) until I moved the "remove missing" bits up to the refreshData() function itself.
 - For an item view or edit Page, the item needs to be fully loaded to the redux store (a deep fetch is needed and if the VM/Pool was fetched by the page, it will only have a shallow fetch).  As @jniederm suggests, there is a better way to handle the fetch in the Pages so as not to anger react.

Additionally, and although not explicitly stated, I now understand that there is a dual use of **store.vms.vms**.  First is to drive the card view and left nav.  The store only needs minimal VM info to support that functionality.  Second is to drive the VM detail view and edit pages.  The store needs to fetch additional data to support those views.

Makes sense to me to break those 2 uses apart.  Keep the data for the lists in one place, and keep a full VM in another place.  The second place can keep multiple VMs just in case a user opens multiple tabs (or some such other silliness).  If you guys agree that change would make sense (@mareklibra , @jniederm , @bond95), I can create an issue and work on it in future.

sjd78:
These last 2 commits should address almost all of the changes requested.  The one I know I haven't addressed yet is @jniederm's suggestion:

> Hence I suggest to use fresh new shiny `getDerivedStateFromProps()` callback (which is called both on mount and props change and allows firing redux actions) that would force-fetch the VM whenever props.match.params.id is changed. (State initialization would be necessary.)

I'll work that change out next.

mareklibra:
> Makes sense to me to break those 2 uses apart. Keep the data for the lists in one place, and keep a full VM in another place. The second place can keep multiple VMs just in case a user opens multiple tabs (or some such other silliness). 

@sjd78 , not sure if I understand, so please correct me. I would like to avoid serving redux store hierarchy these purely UI needs. The React component composition can be changed very quickly, i.e. there's redesign effort in progress. The redux store hierarchy should reflect actual data set of the application and let's React derive representation for it. Duplication of data in the store does not sound good to me.

src/components/Pages/index.js
```diff
@@ -26,31 +24,36 @@ class VmDetailPage extends React.Component {
     this.requestSent = false
   }
 
-  componentWillMount () {
-    if (Selectors.isFilterChecked()) {
-      this.props.getVms({ vmId: this.props.match.params.id })
+  fetchTheVmIfNeeded () {
```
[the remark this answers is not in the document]
sjd78: This hack is to reload the VM's console options (from browser local storage) and a force a deep fetch of the VM correct?  I'd actually move that from the navigation component and put it in the page component.  That way when the page's match id changes, the page can refresh both of those things.

@jniederm, @mareklibra - Does it make sense to include that change in this PR or to do it as a followup PR?

sjd78:
>> Makes sense to me to break those 2 uses apart. Keep the data for the lists in one place, and keep a full VM in another place. The second place can keep multiple VMs just in case a user opens multiple tabs (or some such other silliness).
>
> @sjd78 , not sure if I understand, so please correct me. I would like to avoid serving redux store hierarchy these purely UI needs. The React component composition can be changed very quickly, i.e. there's redesign effort in progress. The redux store hierarchy should reflect actual data set of the application and let's React derive representation for it. Duplication of data in the store does not sound good to me.

@mareklibra - It's just finding a balance between keeping the data in the store more like the REST data (think DataAccessObject) or transforming it to be more like the view (think ViewObject).  I'm ok with leaving the store, reducers and sagas as-is.

sjd78:
> Hence I suggest to use fresh new shiny `getDerivedStateFromProps()` callback (which is called both on mount and props change and allows firing redux actions) that would force-fetch the VM whenever props.match.params.id is changed. (State initialization would be necessary.)

Done!  Works great.

Now anytime the selected VM or Pool id changes for a view or edit page, the VM or Pool will be refreshed by the Page itself.  Note: The selected id changes when the page is reloaded (nothing to something) or react-router changes the route match props (link navigation).

bond95:
Also this PR fixes https://github.com/oVirt/ovirt-web-ui/issues/563

src/components/Pages/index.js
```diff
@@ -23,52 +21,54 @@ const VmsPage = () => {
...
-    if (vms.getIn(['vms', match.params.id])) {
-      return (<VmDetail vm={vms.getIn(['vms', match.params.id])} config={config} />)
-    } else if (requestActive) {
-      console.info(`VmDetailPage: VM id cannot be found: ${match.params.id}. Load is still in progress - waiting before redirect`)
-      return null
+    const { vms, config } = this.props
+    const { vmId } = this.state
+
+    if (vms.getIn(['vms', vmId])) {
+      return (<VmDetail vm={vms.getIn(['vms', vmId])} config={config} />)
```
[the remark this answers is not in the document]
sjd78: If `vmId` is undefined (which would require getting to the page outside of react-router, or direct use of the page component elsewhere in the App), or if no VM exists in `vms`, render returns `null` and nothing will be shown.  I'll improve the check so it is clear and consistent with `VmDialog`.

src/components/Pages/index.js
```diff
@@ -78,96 +78,109 @@ const VmDetailPageConnected = connect(
...
-    let { match, vms, config, requestActive } = this.props
+    const { vms, config } = this.props
+    const { poolId } = this.state
 
-    if (vms.getIn(['pools', match.params.id, 'vm'])) {
-      return (<VmDetail vm={vms.getIn(['pools', match.params.id, 'vm'])} pool={vms.getIn(['pools', match.params.id])} config={config} isPool />)
-    } else if (requestActive) {
-      return null
+    if (vms.getIn(['pools', poolId, 'vm'])) {
+      return (<VmDetail vm={vms.getIn(['pools', poolId, 'vm'])} pool={vms.getIn(['pools', poolId])} config={config} />)
```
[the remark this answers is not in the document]
sjd78: I'll improve the check so it is clear and consistent with `VmDialog`.

src/saga/login.js
```diff
@@ -47,39 +48,41 @@ import {
...
+  yield put(loginSuccessful({ token, username, userId }))
+
+  // Verify the API (exists and is the correct version)
+  const oVirtMeta = yield callExternalAction('getOvirtApiMeta', Api.getOvirtApiMeta, action)
+  const versionOk = yield checkOvirtApiVersion(oVirtMeta)
+  if (!versionOk) {
+    console.error('oVirt API version check failed')
+    yield put(failedExternalAction({
+      message: composeIncompatibleOVirtApiVersionMessage(oVirtMeta),
+      shortMessage: 'oVirt API version check failed',
```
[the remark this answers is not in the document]
sjd78: :+1: ...followup PR

sjd78:
`git pull` merge removed :smile: 

Commits have been squashed down to 5 commits.  I can squash b2f0d1d into cce4073 if the PR needs to be fewer commits.

src/saga/login.js
```diff
@@ -47,39 +48,41 @@ import {
...
-  } else {
+  // Verify a SSO token exists
+  if (!token) {
     yield put(loginFailed({
-      errorCode: result['error_code'] ? result['error_code'] : 'no_access',
-      message: result['error'] ? (result.error['statusText'] ? result.error['statusText'] : JSON.stringify(result['error'])) : 'Login Failed',
+      errorCode: 'no_access',
+      message: 'Login Failed',
     }))
+    return
```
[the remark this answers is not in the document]
sjd78: Yes.  The check is for the failure of a part of the login process.  The `return` is needed so the process doesn't continue.

sjd78:
@bond95, @mareklibra - You guys want to take another look at this PR?  I'd like to get it merged sooner then later if possible.  Thanks.

src/components/Pages/index.js
```diff
@@ -23,52 +21,54 @@ const VmsPage = () => {
...
     }
   }
 
-  componentWillUpdate () {
-    const vmInStore = this.props.vms.getIn(['vms', this.props.match.params.id])
-    if (!vmInStore && Selectors.isFilterChecked() && !this.requestSent) {
-      this.requestSent = true
-      this.props.getVms({ vmId: this.props.match.params.id })
+  static getDerivedStateFromProps (props, state) {
+    if (state.vmId !== props.match.params.id) {
```
bond95: I think you should add one more condition to avoid sending request repeatedly.

src/components/Pages/index.js
```diff
@@ -78,96 +78,109 @@ const VmDetailPageConnected = connect(
...
     }
   }
 
-  componentWillUpdate () {
-    const poolInStore = this.props.vms.getIn(['pools', this.props.match.params.id, 'vm'])
-    if (!poolInStore && Selectors.isFilterChecked() && !this.requestSent) {
-      this.requestSent = true
-      this.props.getPools({ poolId: this.props.match.params.id })
+  static getDerivedStateFromProps (props, state) {
+    if (state.poolId !== props.match.params.id) {
```
bond95: Likewise.

src/components/Pages/index.js
```diff
@@ -78,96 +78,109 @@ const VmDetailPageConnected = connect(
...
+    this.state = {
+      vmId: undefined,
+    }
+  }
 
-    // in case the location is entered from outside, refresh data
-    if (this.props.match.params.id) {
-      this.props.getVms({ vmId: this.props.match.params.id })
+  static getDerivedStateFromProps (props, state) {
+    if (state.vmId !== props.match.params.id) {
```
bond95: Likewise.

src/components/Pages/index.js
sjd78: I'm not sure what extra condition you have in mind.  As it is, a request will only be fired if the match id (VM UUID or Pool UUID) is changed.  That is only changed if the user navigates to a new a page either by clicking a link or a left hand navigation list item.  

Even if other properties change, and this function is called repeatedly, the prop id check against state id will prevent requesting the same id multiple times.

bond95: Hmm... ok.

sjd78:
rebased

mareklibra:
LGTM
I'm about to pre-release 1.4.0 so let's see this live.

PR #634, opened 2018-06-07
fix escape key

### LINKED ISSUE #610: Typing 'escape' in the VM editor skips the navigation prompt
**Steps to reproduce**:
 - Open a VM to edit
 - Change a field
 - Type 'esc' and the VM edit page will exit without a navigation prompt

ezgif com-video-to-gif 1

**Note**: Even with the navigation prompt open, typing 'esc' will still send you back

ezgif com-video-to-gif 2

**Note 2**: When you hit 'esc' in the editor, you are always sent back to the VM Details page regardless of where you came from.  This follows the breadcrumbs.  My expectation is 'esc' will work like the 'Close' button (when the page has a Close button), otherwise it would back out to the next higher breadcrumb.

suomiy:
push new entry to history instead of redirect on escape key

- fixes #610

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)

### DIFF
--- a/src/components/PageRouter.js
+++ b/src/components/PageRouter.js
@@ -17,6 +17,8 @@ import Toolbar from './Toolbar/Toolbar'
 
 import { msg } from '../intl'
 
+const ROOT_URL = '/'
+
 const buildPath = (route) => {
   let res = []
   for (let i in route) {
@@ -55,7 +57,9 @@ Breadcrumb.propTypes = {
 }
 
 const findExactMatch = (branches) => {
-  if (branches.length === 1) {
+  if (branches.length === 0) {
+    return null
+  } else if (branches.length === 1) {
     return branches[0]
   } else {
     for (let i = 0; i < branches.length; i++) {
@@ -63,69 +67,96 @@ const findExactMatch = (branches) => {
         return branches[i]
       }
     }
+    return branches[0]
   }
-  return null
+}
+
+const propsToState = (props) => {
+  const branches = matchRoutes(props.route.routes, props.location.pathname)
+  return {
+    branches,
+    branch: findExactMatch(branches),
+  }
 }
 
 class PageRouter extends React.Component {
   constructor (props) {
     super(props)
-    this.handleKeyPress = this.handleKeyPress.bind(this)
-    this.keyPressed = null
-    this.previousPath = '/'
+    this.state = Object.assign({
+      previousPath: ROOT_URL,
+    }, propsToState(props))
+
+    this.handleKeyUp = this.handleKeyUp.bind(this)
   }
 
-  handleKeyPress (event) {
-    if (event.key === 'Escape' && this.keyPressed !== event.key) {
-      this.keyPressed = event.key
-      if (this.currentBranch.length > 1) {
-        this.props.redirectRoute(this.currentBranch[this.currentBranch.length - 2].match.url)
+  handleKeyUp (event) {
+    if (event.key === 'Escape' && this.state.branch.route.closeable) {
+      const branches = this.state.branches
+      const len = branches.length
+      let url
+
+      if (len === 1) {
+        if (branches[0].match.url !== ROOT_URL) {
+          url = ROOT_URL
+        }
+      } else if (len > 1) {
+        url = branches[len - 2].match.url
+      }
+
+      if (url) {
+        this.props.history.push(url)
       }
     }
   }
 
   componentDidUpdate () {
-    if (this.props.routeReducer.get('redirect') && this.props.routeReducer.get('redirect') !== this.props.location.pathname) {
+    if (this.props.redirect && this.props.redirect !== this.props.location.pathname) {
       this.props.onRedirect()
     }
   }
 
   componentDidMount () {
-    window.addEventListener('keydown', this.handleKeyPress)
-    window.addEventListener('keyup', (e) => { this.keyPressed = null })
+    document.addEventListener('keyup', this.handleKeyUp)
+  }
+
+  componentWillUnmount () {
+    document.removeEventListener('keyup', this.handleKeyUp)
   }
 
   componentWillReceiveProps (nextProps) {
-    if (this.previousPath !== this.props.location.pathname && nextProps.location.pathname !== this.props.location.pathname) {
-      this.previousPath = this.props.location.pathname
+    const newState = propsToState(nextProps)
+
+    if (this.state.previousPath !== this.props.location.pathname && location.pathname !== this.props.location.pathname) {
+      newState.previousPath = this.props.location.pathname
     }
+    this.setState(newState)
   }
 
   render () {
-    let { route, location, history, routeReducer } = this.props
-    if (routeReducer.get('redirect') && routeReducer.get('redirect') !== location.pathname) {
-      return (<Redirect to={routeReducer.get('redirect')} />)
-    }
-    const branches = matchRoutes(route.routes, location.pathname)
+    const { location, history, redirect } = this.props
 
-    let branch = findExactMatch(branches)
-    this.currentBranch = branches.slice()
-    this.currentBranch.unshift({ match: { url: '/' } })
-    let tools = []
-    if (branch) {
-      for (let i in branch.route.toolbars) {
-        tools.push(branch.route.toolbars[i](branch.match))
-      }
+    if (redirect && redirect !== location.pathname) {
+      return (<Redirect to={redirect} />)
+    }
+
+    const branch = this.state.branch
+    if (!branch) {
+      return null
+    }
+
+    const tools = []
+    for (let i in branch.route.toolbars) {
+      tools.push(branch.route.toolbars[i](branch.match))
     }
 
     const RenderComponent = branch.route.component
     return (<div className={style['page-router']}>
-      <Breadcrumb route={branches} root={{ title: msg.virtualMachines(), url: '/' }} />
+      <Breadcrumb route={this.state.branches} root={{ title: msg.virtualMachines(), url: '/' }} />
       <Toolbar>
         {tools}
       </Toolbar>
       <div className={style['page-router-render-component']}>
-        <RenderComponent route={branch.route} match={branch.match} location={location} history={history} previousPath={this.previousPath} />
+        <RenderComponent route={branch.route} match={branch.match} location={location} history={history} previousPath={this.state.previousPath} />
       </div>
     </div>)
   }
@@ -134,18 +165,17 @@ class PageRouter extends React.Component {
 PageRouter.propTypes = {
   location: PropTypes.object.isRequired,
   history: PropTypes.object.isRequired,
+  /* eslint-disable-next-line react/no-unused-prop-types */
   route: PropTypes.object.isRequired,
-  routeReducer: PropTypes.object.isRequired,
+  redirect: PropTypes.object, // can be null
   onRedirect: PropTypes.func.isRequired,
-  redirectRoute: PropTypes.func.isRequired,
 }
 
 export default withRouter(connect(
   (state) => ({
-    routeReducer: state.route,
+    redirect: state.route.get('redirect'),
   }),
   (dispatch) => ({
     onRedirect: () => dispatch(redirectRoute({ route: null })),
-    redirectRoute: (url) => dispatch(redirectRoute({ route: url })),
   })
 )(PageRouter))
--- a/src/routes.js
+++ b/src/routes.js
@@ -32,6 +32,7 @@ const getRoutes = (vms) => ([
       {
         path: '/vm/add',
         exact: true,
+        closeable: true, // by ESC key
         title: (match) => msg.addNewVm(),
         component: VmDialogPage,
         toolbars: [], // Recently not used. When needed, see VmDialog/style.css - .vm-dialog-buttons
@@ -44,6 +45,7 @@ const getRoutes = (vms) => ([
         routes: [
           {
             path: '/vm/:id/edit',
+            closeable: true, // by ESC key
             title: (match) => msg.edit(),
             component: VmDialogPage,
             toolbars: [], // Recently not used. When needed, see VmDialog/style.css - .vm-dialog-buttons


sjd78:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[src/components/PageRouter.js, line 79 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LER4mqrCunx6AGYaOxu:-LER4mqrCunx6AGYaOxv:bt21f8k) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L79)):*
> ```JavaScript
>   }
> 
>   handleKeyUp (event) {
> ```

Makes sense to fire on **keyup** instead of **keydown**.

In this case, are the **keydown** handlers still necessary?  If **keydown** handlers aren't need anymore, the `this.keyPressed` bits can be removed as well.

---

*[src/components/PageRouter.js, line 106 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LER49XM6sl6tQWD0UZA:-LER49XO3j0sEALegXKC:bmzo9km) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L106)):*
> ```JavaScript
> 
>   componentWillUnmount () {
>     document.removeEventListener('keydown', this.handleKeyDown)
> ```

Shouldn't this be on `window` instead of `document`?

---

*[src/components/PageRouter.js, line 124 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEROFQoAoi9ViL721zT:-LEROFQoAoi9ViL721zU:b-83hncd) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L124)):*
> ```JavaScript
> 
>     let branch = findExactMatch(branches)
>     this.currentBranch = branches.slice()
> ```

Can you define `currentBranch` in the constructor?  This is the defined in `render()` and used in `handleKeyUp()`.

I think it should really be part of component state (along with `previousPath`) instead of part of the object.

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

sjd78:
This change will fix the navigation prompt issue, but it doesn't fix the Cards -> Edit -> (hit Escape) -> Details flow.  That flow should be Cards -> Edit -> (hit Escape) -> Cards.

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-:-LERQwfvDABqIL1XtUqc:bz5e6k4)*

sjd78:
Cards → View → hit Esc → Cards .. **works**
Cards → Edit → make changes to a field → hit Esc → Navigation prompt .. **fixed**
Cards → Edit → make changes to a field → click 'Close' (or Esc) → Navigation prompt → hit Escape → _nothing_ .. **fixed**
Cards → Edit → click 'Close' → Cards .. **works**
Cards → Edit → hit Esc → Cards .. **broken** (it lands you on View)

In summary, Esc should follow what the 'Close' button does.  That doesn't exactly work out for the View pages since there isn't a 'Close' button, and you can swap VMs/Pool with the left nav list.  Having the Esc close the view page and land the user on the Cards page is probably the easiest thing to do right now.

On the "/" page, Esc probably shouldn't do anything.  I noticed if you hit Esc on the Cards view (location "/") it'll get pushed to history each time.  Using the browser back button takes awhile to back out of that.

src/components/PageRouter.js
```diff
@@ -70,29 +70,41 @@ const findExactMatch = (branches) => {
...
     this.previousPath = '/'
   }
 
-  handleKeyPress (event) {
+  handleKeyUp (event) {
     if (event.key === 'Escape' && this.keyPressed !== event.key) {
       this.keyPressed = event.key
       if (this.currentBranch.length > 1) {
-        this.props.redirectRoute(this.currentBranch[this.currentBranch.length - 2].match.url)
+        this.props.history.push(this.currentBranch[this.currentBranch.length - 2].match.url)
```
mareklibra: What is benefit of `adding` instead of removing via `history.goBack()`

https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/history.md

suomiy: I do not know what benefit is better - creating new history or going back. But anyway we are using `Link` from `react-router` for closing the dialog (using `close` button), which adds new history entry. I think these two cases should have the same behaviour.

mareklibra: Good point, it should be probably consistent.
Anyway, `removing` follows better my understanding of "going back". My understanding is sticked to the "Back button" in browser.

Does it make sense to refactor the `close` button this way?

suomiy: I would prefer to create new history. I think users are usually not expecting the page to go back in history stack while clicking the link. Also IMO I think it is a bit weird for dialog to close and then have the possibility to go forward again to the dialog. 

I would rather allow using `ESC` only for dialogs and not for all the pages.

suomiy:
Review status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @sjd78, @suomiy, and @mareklibra)

---

*[src/components/PageRouter.js, line 106 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LER49XM6sl6tQWD0UZA:-LEu-Gr_8Rr-k1IZxxdr:bb7r2f0) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L106)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Shouldn't this be on `window` instead of `document`?
</blockquote></details>

why?

---

*[src/components/PageRouter.js, line 124 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEROFQoAoi9ViL721zT:-LEu-LtKDh9TCviYIwUi:bmfp5yj) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L124)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Can you define `currentBranch` in the constructor?  This is the defined in `render()` and used in `handleKeyUp()`.

I think it should really be part of component state (along with `previousPath`) instead of part of the object.
</blockquote></details>

I put just branches, because only they will be used in the handler now.  And the matched branch is computed in the render from state 

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

suomiy:
I refactored it this time and all the navigation cases should work now.

suomiy:
Review status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @sjd78, @suomiy, and @mareklibra)

---

*[src/components/PageRouter.js, line 79 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LER4mqrCunx6AGYaOxu:-LEu0psf2BC7dNstxKwl:b-896fix) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L79)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Makes sense to fire on **keyup** instead of **keydown**.

In this case, are the **keydown** handlers still necessary?  If **keydown** handlers aren't need anymore, the `this.keyPressed` bits can be removed as well.
</blockquote></details>

Done.

---

*[src/components/PageRouter.js, line 106 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LER49XM6sl6tQWD0UZA:-LEu8K7F3ulks9oI-SZG:bjj2g84) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L106)):*
<details><summary><i>Previously, suomiy wrote…</i></summary><blockquote>

why?
</blockquote></details>

oh, I see, my mistake. Both of the calls are on document now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

sjd78:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed, 5 unresolved discussions (waiting on @mareklibra and @suomiy)

---

*[src/components/PageRouter.js, line 83 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEQ3vgmrME9kHjbvpJu-r1-83:-LEz0OBvAQKNDiShjsgy:b-tx2jgo) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L83)):*
<details><summary><i>Previously, suomiy wrote…</i></summary><blockquote>

I would prefer to create new history. I think users are usually not expecting the page to go back in history stack while clicking the link. Also IMO I think it is a bit weird for dialog to close and then have the possibility to go forward again to the dialog. 

I would rather allow using `ESC` only for dialogs and not for all the pages. 
</blockquote></details>

I chatted with UX and using `Esc` only to close dialogs was the base recommendation.

---

*[src/components/PageRouter.js, line 124 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEROFQoAoi9ViL721zT:-LEz0AUj5912NAKjej1o:bb6ywmx) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L124)):*
<details><summary><i>Previously, suomiy wrote…</i></summary><blockquote>

I put just branches, because only they will be used in the handler now.  And the matched branch is computed in the render from state 
</blockquote></details>

LGTM

---

*[src/components/PageRouter.js, line 128 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEza8Wg1mDx151H33Ab:-LEza8Wh3Cl7lsAfoh9n:b3qnikr) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L128)):*
> ```JavaScript
>     }
>     this.setState({
>       branches: matchRoutes(route.routes, location.pathname),
> ```

Could do this in a single `setState` to ensure the state changes are done together?

---

*[src/components/PageRouter.js, line 133 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEzbASj5UfSoEtgFPin:-LEzbASj5UfSoEtgFPio:b-jlvy4p) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L133)):*
> ```JavaScript
> 
>   render () {
>     let { location, history, redirect } = this.props
> ```

nit pick (since you're in the file already): s/let/const/

---

*[src/components/PageRouter.js, line 142 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LF0aiUA6O5cFUxXAT_J:-LF0aiUB4qrblEdQ39hJ:bqex98g) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L142)):*
> ```JavaScript
> 
>     if (!branch) {
>       return null
> ```

This check could be avoided if you init state in the constructor with `matchRoutes` instead of `[]`

---

*[src/components/PageRouter.js, line 145 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEzb7kQ9eDFA4IKQPyh:-LEzb7kQ9eDFA4IKQPyi:b-jlvy4p) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L145)):*
> ```JavaScript
>     }
> 
>     let tools = []
> ```

nit pick (since you're in the file already): s/let/const/

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

suomiy:
@sjd78 fixed

suomiy:
Review status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @sjd78, @mareklibra, and @suomiy)

---

*[src/components/PageRouter.js, line 83 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEQ3vgmrME9kHjbvpJu-r1-83:-LF805IN7Vh0BOTobgoB:b-l5y2w3) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L83)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

I chatted with UX and using `Esc` only to close dialogs was the base recommendation.
</blockquote></details>

agreed, works only for dialogs now and can be optionally set in routes

---

*[src/components/PageRouter.js, line 142 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LF0aiUA6O5cFUxXAT_J:-LF80V_g2J-s5I2aGH8_:bouubju) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L142)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

This check could be avoided if you init state in the constructor with `matchRoutes` instead of `[]`
</blockquote></details>

it is initialized in the constructor now, but I would prefer to put this check here anyway. For example if we get empty routes in props.

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

sjd78:
LGTM. I like the `closable` flag in routes.

---

Reviewed 2 of 2 files at r3.
Review status: all files reviewed, 1 unresolved discussion (waiting on @mareklibra)

---

*[src/components/PageRouter.js, line 83 at r1](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEQ3vgmrME9kHjbvpJu-r1-83:-LFAvz0x9fAnHdWQkp6R:bb6ywmx) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/87511821eb764218955ae11d19cff122f44a4225/src/components/PageRouter.js#L83)):*
<details><summary><i>Previously, suomiy wrote…</i></summary><blockquote>

agreed, works only for dialogs now and can be optionally set in routes
</blockquote></details>

LGTM

---

*[src/components/PageRouter.js, line 128 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LEza8Wg1mDx151H33Ab:-LFAy0sP-d-8J8T7_b6a:bb6ywmx) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L128)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

Could do this in a single `setState` to ensure the state changes are done together?
</blockquote></details>

LGTM

---

*[src/components/PageRouter.js, line 142 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LF0aiUA6O5cFUxXAT_J:-LFAyxSW3vPQBD4pP-u_:b-o8xeeo) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L142)):*
<details><summary><i>Previously, suomiy wrote…</i></summary><blockquote>

it is initialized in the constructor now, but I would prefer to put this check here anyway. For example if we get empty routes in props.
</blockquote></details>

The component only gets pulled in through react-router so the route wouldn't be empty, but leaving the check in there is ok too.

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-:-LFB1OqvEDLrfwcGiNyn:b48vdcc)*

suomiy:
Review status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @sjd78 and @mareklibra)

---

*[src/components/PageRouter.js, line 142 at r2](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634#-LF0aiUA6O5cFUxXAT_J:-LFJR86UBIZY6Z-KsEJP:b-jo3nz0) ([raw file](https://github.com/ovirt/ovirt-web-ui/blob/6e48b81631816f143dc5185e6f1a9c4957e6b646/src/components/PageRouter.js#L142)):*
<details><summary><i>Previously, sjd78 (Scott Dickerson) wrote…</i></summary><blockquote>

The component only gets pulled in through react-router so the route wouldn't be empty, but leaving the check in there is ok too.
</blockquote></details>

ok, I put it back there for completeness

---


*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

sjd78:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed, 1 unresolved discussion (waiting on @mareklibra)

---



*Comments from [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/634)*

sjd78:
I adopted a bunch of the changes from the PR in PR #648 and #653.  More to the point, the Esc key handling has already been merged.

@suomiy - Can you review the current merged changes vs this PR?  If the merged changes look good, I think this PR can be closed.

suomiy:
@sjd78 ok, looks fine

PR #637, opened 2018-06-11
build: Add workaround for failing `yarn check` 

mareklibra:
The `yarn check` is recently failing on
- old webpack (v1, at least v2 is "needed")
- bootstrap-switch has peerDependency on bootstrap "^3.1.1" (but "bootstrap@4.1.1" used)
- redux-devtools-extension has peerDep on redux@^3.1.0" (but "redux@4.0.0" used)
- "patternfly-react#table-resolver#redux@>= 3.0.0 < 4.0.0" doesn't satisfy found match of "redux@4.0.0"
- error "patternfly-react#eonasdan-bootstrap-datetimepicker#jquery" not installed
- error "patternfly#eonasdan-bootstrap-datetimepicker#jquery" not installed

It's manually verified, that these errors have no effect on recent ovirt-web-ui code base.
But they need to be watched carefully, since they might in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/637)


mareklibra:
Related to #636, the new `yarn v1.7.0` is more strict in case of `yarn check`

sjd78:
Is there any particular reason `bootstrap` >3.3.7 is being used?

`patternfly-react` and `react-bootstrap` are both built on `bootstrap` 3.

mareklibra:
No, introduced by #614 .
Looking for comments - I'm fine with downgrading if it makes sense.

gregsheremeta:
@vojtechszocs see above

gregsheremeta:
Why do we need yarn check in builds?

gregsheremeta:
bootstrap should be locked at 3.3.7 for everyone using patternfly

sjd78:
So some of the errors are valuable.  For example, this shows web-ui really needs to upgrade `webpack` to use those new loaders "properly":
```
error "babel-loader#webpack@2 || 3 || 4" doesn't satisfy found match of "webpack@1.13.2"
error "eslint-loader#webpack@>=2.0.0 <5.0.0" doesn't satisfy found match of "webpack@1.13.2"
```

But some are not valuable.  For example:
```
warning "patternfly-react#patternfly#eonasdan-bootstrap-datetimepicker#jquery@^1.8.3 || ^2.0 || ^3.0" could be deduped from "3.3.1" to "jquery@3.3.1"
error "patternfly-react#eonasdan-bootstrap-datetimepicker#jquery" not installed
```

I suggest using `yarn check --integrity` in the build.  At least in yarn 1.7, that will check **package.json** against **yarn.lock**.  That would keep the check shallow and only error out if someone changed **package.json** without also committing a new **yarn.lock**.  It won't look deep in **node_modules/**, but it won't error on problems deep in patternfly dependencies...

sjd78:
Just noticed - the `bootstrap` dependency really should be in the __dependencies__ block, not __devDependencies__.

mareklibra:
> Why do we need yarn check in builds?

> I suggest using yarn check --integrity in the build.

Still makes sense to verify that since failure would point to potential issue with merged PRs. So to double-check.

Recently, the `yarn check` is called from `ovirt-engine-nodejs-modules/setup-env.sh`. I think it makes sense to fix it there. @vojtechszocs ?

mareklibra:
> this shows web-ui really needs to upgrade webpack to use those new loaders "properly":

There's no doubt about that. I had already more-or-less working version of this upgrade, there's bunch of changes needed, it's not straightforward as one would think. I used recent ejected `react-create-app`as a frame to subsequently re-introduce changes specific for ovirt-web-ui - this approach helped a lot. 

Unfortunately, still not everything was finished and I have not enough time to continue on this.
The `v1` is not a long-term way for us ...
Anyway, recent webpack v1 is not perfect but I don't observe any real issue with it.

mareklibra:
Added commit downgrades bootstrap to `3.3.7` and moves some packages from `devDependencies` to `dependencies`

vojtechszocs:
> Why do we need yarn check in builds?

Until now, using `yarn check` was a recommended practice - its purpose was to validate `package.json` (version ranges of all dependencies) against `yarn.lock` (version resolutions of all dependencies). In other words, `yarn check` was supposed to ensure a fully deterministic dependency resolution via the lock file.

It seems that Yarn 1.7.x made `yarn check` command cover a wider area of potential problems. I'll check this and get back to you.

vojtechszocs:
> Recently, the `yarn check` is called from `ovirt-engine-nodejs-modules/setup-env.sh`. I think it makes sense to fix it there. @vojtechszocs ?

Normally, for the purpose of RPM build (e.g. when using [oVirt Standard CI](http://ovirt-infra-docs.readthedocs.io/en/latest/CI/Build_and_test_standards/)), `ovirt-engine-nodejs-modules` *should* be declared as build-time dependency and the `setup-env.sh` script *should* be called during the `%prep` phase, check [this example spec file](https://github.com/oVirt/ovirt-engine-dashboard/blob/ovirt-engine-dashboard-1.2.3-2/packaging/spec.in#L24).

We added `yarn check` to `setup-env.sh` script to ensure that dependency resolution check is always performed.

> I suggest using `yarn check --integrity` in the build.

[CLI docs](https://yarnpkg.com/lang/en/docs/cli/check/) on using `--integrity` flag:

> Verifies that versions and hashed value of the package contents in the project’s package.json matches that of yarn’s lock file. This helps to verify that the package dependencies have not been altered.

Notice the "versions **and** hashed value" part. The `--integrity` flag should do what `yarn check` does, plus it should verify checksums (content hashes) of all packages. Therefore, I don't see how using the `--integrity` flag would solve the problem at hand.

vojtechszocs:
@gregsheremeta @sjd78 I think we have the same issue in current `ovirt-engine-dashboard` master.

When using Yarn 1.5.1 (reasonably recent):
```bash
$ rm -rf ./node_modules && yarn install && yarn check ; echo $?
..
error "patternfly#eonasdan-bootstrap-datetimepicker#jquery" not installed
..
error An unexpected error occurred: "Found 1 errors.".
..
1
```

When using Yarn 1.7.0 (latest stable):
```bash
$ rm -rf ./node_modules && yarn install && yarn check ; echo $?
..
error "patternfly#eonasdan-bootstrap-datetimepicker#jquery" not installed
..
error Found 1 errors.
..
1
```

Note that in `ovirt-engine-yarn` package (used by oVirt CI), we upgraded Yarn version from 0.24.4 straight into 1.7.0.

gregsheremeta:
My take:

until patternfly is adjusted, just add this to package.json
  "resolutions": {
    "bootstrap": "3.3.7",
    "jquery": "3.2.1"
  }

We didn't have this in our old yarn.

Now the check will succeed and you don't need the "yarn check || /bin/true"

vojtechszocs:
@sjd78 see [this comment](https://github.com/yarnpkg/yarn/issues/2287#issuecomment-359147179)

> —integrity does a light check that verifies that no one touched node_modules after yarn installed it.

> Regular check runs Yarn’s resolution mechanism that checks that  the installation tree matches yarn.lock.

> The warnings indicate that something is not optimal but installation is
still correct

sjd78:
This diff will fix all the `yarn check` errors except the webpack one:
```diff
diff --git a/package.json b/package.json
index 4b5eb59..c67f4a3 100644
--- a/package.json
+++ b/package.json
@@ -92,9 +92,12 @@
     "react-router-navigation-prompt": "1.6.3",
     "react-test-renderer": "16.4.0",
     "redux": "4.0.0",
-    "redux-devtools-extension": "2.13.2",
     "redux-saga": "0.16.0"
   },
+  "resolutions": {
+    "patternfly/**/jquery": "~3.3.1",
+    "redux": "4.0.0"
+  },
   "scripts": {
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",

```

sjd78:
@mareklibra - I feel your pain on the webpack upgrade.  With applying optimizations etc on dashboard, took me a few weeks to get webpack v3 all sorted.

mareklibra:
@sjd78 , I would keep the `redux-devtools-extension` in the dependencies. I just briefly verified, it's still working even with the `redux 4.0.0` .

mareklibra:
Big thank all of you for your help on this.

sjd78:
> I would keep the `redux-devtools-extension` in the dependencies. I just briefly verified, it's still working even with the `redux 4.0.0`.

I took a look at the extension and I couldn't find where it was used in the code base, but I don't have any particular issue with leaving it there.

mareklibra:
Any objection merging this?
@sjd78 , @gregsheremeta , @vojtechszocs , @bond95 ?

sjd78:
@gregsheremeta  @mareklibra - 

Keep the jquery version at **3.3.1** and the resolution block is required. 
Dropping the version to **3.2.1** allows dropping the jquery entry from the resolution block.

Either way LGTM

mareklibra:
@sjd78 , both ways makes sense to me as well.

Anyway, since this hack is potential subject to change in the future, I'm aligning it with the `dashboard` project.

PR #639, opened 2018-06-15
Refactor the scheduled data refresh to be able to stop and start

sjd78:
- The scheduler can be started via `startSchedulerFixedDelay` with
   any delay (in seconds).

 - The scheduler can be stopped via `stopSchedulerFixedDelay` action.
   This will cancel the current `delay` and skip the task.

 - Starting a scheduler while another is already running will cause
   the existing scheduler to be stopped.

This will be useful if/when the refresh is enhanced to allow the user to pick the refresh interval.

Also useful for devs to control the refresh using the redux tools browser extension. Dispatch `{ type: 'STOP_SCHEDULER_FIXED_DELAY', payload: {} }` to completely stop the refresh.  Dispatch `{ type: 'START_SCHEDULER_FIXED_DELAY', payload: { delayInSeconds: 600 } }` to (re)start the scheduler with a 10 minute delay.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/639)


sjd78:
**NOTE** - This PR touches some of the same files as #628.  If that PR is merged first, this one will need some changes.  If this PR is merged first, #628 will need some changes.

src/sagas.js
```diff
@@ -664,31 +669,54 @@ function* fetchAllNetworks () {
...
+  yield put(removeActiveRequest(requestId))
+}
 
-  // TODO: do we need to stop the loop? Consider takeLatest in the rootSaga 'restarts' the loop if needed
-  while (true) {
-    yield delay(AppConfiguration.schedulerFixedDelayInSeconds * 1000)
-    logDebug(`schedulerWithFixedDelay() event after delay of ${AppConfiguration.schedulerFixedDelayInSeconds} seconds`)
+function* startSchedulerWithFixedDelay (action) {
+  // if a scheduler is already running, stop it
+  if (Selectors.isSchedulerEnabled()) {
```
bond95: Here must be `if (!Selectors.isSchedulerEnabled()) {`, no?

src/sagas.js
```diff
@@ -664,31 +669,54 @@ function* fetchAllNetworks () {
...
+    yield put(enableScheduler({ enabled: false }))
```
bond95: Why do you have here `yield put(enableScheduler({ enabled: false }))` if `stopSchedulerFixedDelay()` doing the same thing?

src/reducers/config.js
```diff
@@ -50,6 +51,9 @@ const config = actionReducer(initialState, {
   SET_USB_FILTER (state, { payload: { usbFilter } }) {
     return state.set('usbFilter', usbFilter)
   },
+  ENABLE_SCHEDULER (state, { payload: { enabled } }) {
```
bond95: Likewise

src/sagas.js
sjd78: No.

It is setup to have only 1 scheduler running at a time.  When starting, if the scheduler is already enabled, stop it.

src/sagas.js
sjd78: Ah - that is confusing.  I'll streamline that a bit.

The `start` can call `stop` directly, and the `stop` can use a different action string so it doesn't share with the `take` in the `race` in `schedulerWithFixedDelay`

src/actions/index.js
```diff
@@ -27,18 +29,31 @@ export * from './pendingTasks'
...
+export function stopSchedulerFixedDelay () {
+  return {
+    type: STOP_SCHEDULER_FIXED_DELAY,
     payload: {},
   }
 }
 
+export function enableScheduler ({ enabled = true }) {
+  return {
+    type: 'ENABLE_SCHEDULER',
```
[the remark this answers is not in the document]
sjd78: I'll do the entire file.

src/reducers/config.js
sjd78: I'll do the entire file.

src/sagas.js
```diff
@@ -658,33 +662,53 @@ function* fetchAllNetworks () {
...
-function* schedulerWithFixedDelay (action) {
-  logDebug('Starting schedulerWithFixedDelay() scheduler')
+function* delayedRemoveActiveRequest ({ payload: requestId }) {
+  yield delay(500)
+  yield put(removeActiveRequest(requestId))
+}
+
+function* startSchedulerWithFixedDelay (action) {
+  // if a scheduler is already running, stop it
+  yield put(stopSchedulerFixedDelay())
```
sjd78: Just ask the running scheduler to stop.  If there isn't one, it become a no-op.

src/actions/index.js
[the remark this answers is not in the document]
sjd78: I ended up removing `ENABLE_SCHEDULER`, but I still fixed up constant use.

src/sagas.js
sjd78: I was able to take out the `isSchedulerEnabled` selector and reducer etc.  It wasn't needed.

src/reducers/config.js
sjd78: `ENABLE_SCHEDULER` was removed, but I fixed up the constant use in the file anyway.

sjd78:
There was a simpler way that didn't require `enabledScheduler` in the config reducer.

sjd78:
#643 included the commit in this PR, so I'm closing this one out since it is no longer relevant.

PR #643, opened 2018-06-19
Use constants for all actions types and reducer function names

sjd78:
- All actions now use constants instead of inline strings.
 - All reducers now use constants for function names instead of inline strings.
 - constants JS files have been sorted alphabetically
 - constants imports have been sorted alphabetically

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/643)


sjd78:
The first commit in this PR is from #639.  Once that PR is merged, this PR will need to be rebased.

PR #648, opened 2018-06-26
ux-redesign: Adopt use of connected-react-router

sjd78:
Removed existing redirect / redirectRoute handling:
 - route reducer
 - route/redirect actions and constants
 - redirect handling in `PageRouter`

Refactored `PageRouter` further to:
 - have Esc key use connected-react-router for navigation
 - use component state to track what is needed for Esc handling

Refactored App init to incorporate __connected-react-router__
 - redux store creation
 - app init passes store around instead of using an imported var
 - `App` react-router BrowserRouter to use the ConnectedRouter

Now navigation can be done via redux actions from the __connected-react-router__ package.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/648)


sjd78:
NOTE: PR #653 builds on (therefore includes the changes of) this PR

sjd78:
Closing since 0e1b521 (cherry-picked as 3a83f61) was merged along with #653.

Issue #99, opened 2017-03-13
Add CODING STYLE guide


matobet:
It would be useful to have a single document explaining the guidelines how to name components, files, structure package directories and in general develop code in this repository.

As the basis I would probably use the practices developed in https://github.com/oVirt/ovirt-engine-dashboard e.g. naming multi-word files files with `"-"` (`main-tab.js`, `vm-dialog.js`, .etc...), and standard react best practices from http://reactpatterns.com/

gregsheremeta:
test comment - delete me

PR #651, opened 2018-06-27
ux-redesign: Refactor console components and redux code

sjd78:
This decouples the console related redux code from the existing components and pushes the use of redux state and dispatches to the actual button components.  Consumers of the buttons don't need to lookup the data held in the redux store, the buttons can lookup for itself.

Specific changes:
 - Elevate constants, actions, sagas and reducers up to base from component specific definitions. __consoleInUse__ now lives in __store.vms.vms[vmid]__

 - Console related sagas are held in 'src/saga/vm-console'

 - Pushed config and dispatches to be close to the console buttons

 - Refactor `ConsoleButton`'s connected component to pull props directly from redux state instead of relying on the data being passed from the parent component

 - Created a `WindowsRdpButton` specifically for RDP

 - Removed __config__, __usbFilter__, and __userId__ use from components that used `ConsoleButton` in favor of `ConsoleButton`'s connected component pulling the data directly (`VmActions`, `VmDetail` and `VmConsoles`)

 - Simplified `VmDetail/VmConsoles`

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/651)


src/saga/vm-console/index.js
```diff
@@ -85,28 +55,22 @@ export function* downloadVmConsole (action) {
       }
 
       data = adjustVVFile({ data, options, usbFilter, isSpice })
-      fileDownload({ data, fileName: 'console.vv', mimeType: 'application/x-virt-viewer' })
+      fileDownload({ data, fileName: `console-${consoleId}.vv`, mimeType: 'application/x-virt-viewer' })
```
mareklibra: If this is needed (I doubt, the file can be automatically removed by the desktop viewer app), then let's be consistent with the RDP bellow

src/saga/vm-console/vv-file-utils.js
```diff
@@ -0,0 +1,39 @@
...
+  if (options && isSpice) {
+    const smartcardEnabled = options.get ? options.get('smartcardEnabled') : options.smartcardEnabled
+    data = data.replace(/^enable-smartcard=[01]$/mg, `enable-smartcard=${smartcardEnabled ? 1 : 0}`)
+  }
+
+  logDebug('adjustVVFile data after adjustment:', data)
+  return data
+}
+
+export {
```
mareklibra: nit: rest of the code uses (usually) `export` close to the function itself

src/components/VmActions/WindowsRdpButton.js
```diff
@@ -0,0 +1,47 @@
...
+import { connect } from 'react-redux'
+
+import { getRDP } from '../../actions'
+import { isWindows, hrefWithoutHistory } from '../../helpers'
+import { canConsole } from '../../vm-status'
+
+/**
+ * Button to send a RDP connection file to the user for a Windows VM.
+ */
+const WindowsRdpButton = ({ vm, className, onRDP }) => {
```
mareklibra: +1 for new component

gregsheremeta:
Is "redux bits" a known term? If not I'd rather not use the word "bit" :)

s/redux bits/redux code/g

src/components/VmActions/ConsoleButton.js
```diff
@@ -96,7 +99,7 @@ class ConsoleButton extends React.Component {
 
     return (
       <span className={style['full-button']}>
-        <a href='#' onClick={this.consoleConfirmationAboutToOpen} className={`${button} ${style['link']}`} id={shortTitle}>
```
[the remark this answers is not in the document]
sjd78: I'll move things around a bit.  I think I can do it better.  Good point!

src/components/VmActions/index.js
```diff
@@ -154,8 +153,7 @@ class VmActions extends React.Component {
           button='btn btn-default'
           className='pficon pficon-screen'
           vm={vm}
-          usbFilter={config.get('usbFilter')}
-          userId={config.getIn(['user', 'id'])} />
```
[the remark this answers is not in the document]
sjd78: Because now the `ConsoleButton` pulls it from redux store directly.

src/components/VmDetail/VmConsoles.js
```diff
@@ -21,48 +29,46 @@ const VmConsoles = ({ vm, onConsole, onRDP, usbFilter }) => {
...
 
-        {
-          isWindows(vm.getIn(['os', 'type']))
-            ? (<a href='#' key={vm.get('id')} id={`${idPrefix}-rdp`} onClick={hrefWithoutHistory(onRDP)} className={style['left-delimiter']}>RDP</a>) : null
-        }
+        {isWindowsVM && <WindowsRdpButton vm={vm} className={style['left-delimiter']} />}
       </dd>
     )
-  }
+  } else {
```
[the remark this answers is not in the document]
sjd78: I have a better form for that. I just like having the 2 render blocks at the same indentation level.  Makes my OCD happy.

src/saga/vm-console/index.js
sjd78: It was useful, while I was testing it, to see the different ids in the filename between VNC and SPICE on the same VM.  Also I'm not sure what id I'd add to the RDP file name since it doesn't have a console id.

Oh, I just thought of this ... what happens if a user has >1 VM set to auto-open its console?  I guess the browser would rename the second file or ask the user where to save it.  Adding the console id would help step around that scenario a bit more gracefully.

src/saga/vm-console/index.js
```diff
@@ -85,28 +55,22 @@ export function* downloadVmConsole (action) {
...
-  yield getConsoleOptions(getConsoleOptionsAction({ vmId: action.payload.vmId }))
-}
-
+/**
+ * Push a RDP connection file (__console.rdp__) to connect a user to the Windows VM's RDP session
+ */
 export function* getRDPVm (action) {
   const rdpBuilder = new RDPBuilder(action.payload)
   const data = rdpBuilder.buildRDP()
   fileDownload({ data, fileName: 'console.rdp', mimeType: 'application/rdp' })
```
sjd78: Any suggestions for an id to add to this filename?

src/components/VmActions/index.js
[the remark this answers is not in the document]
bond95: This is not working like that, second parameter of `mapDispatchToProps` function is `ownProps`, that mean - not props that was got from store through `connect` function. https://github.com/reduxjs/react-redux/blob/master/docs/api.md

src/saga/vm-console/index.js
mareklibra: I think it's same approach as Web Admin does. There's presumption, the user does not click so fast :-)

The expected flow is: download file, open associated application which removes the file. Might take a few seconds.

So if you think this is an issue, you can add unique appendix to the filename. IMO, there must not be reference to the VM. I think, value of a counter is good enough. Same for the RDP.
But as I wrote - just if you see issue here.

src/saga/vm-console/index.js
mareklibra: console id just denotes VNC vs SPICE.
So right, no need here. Just the uniqie value, if you feel its needed - as discussed above

src/components/VmActions/index.js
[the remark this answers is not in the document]
sjd78: Right.  `mapDispatchToProps` second argument is ownProps which is the props sent by the parent component plus the props created by the `mapStateToProps` function.  So as long as the prop is sent by the parent or redux `connect`ed, it'll get through to the dispatch functions.  This is done in other web-ui components as well (which is where I got the idea).

It makes sense to me to put props that can be pulled from redux state as close to the component that uses it as possible.  In this case, the only use for **config**, **usbFilter** and **userId** in all of `VmDetails/index.js` is to pass data to `ConsoleButton`.

https://github.com/sjd78/ovirt-web-ui/blob/f99b93af30597ac9deb8468a99d3561ca12ca5e2/src/components/VmActions/ConsoleButton.js#L133-L143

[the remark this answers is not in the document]
bond95: No, not `plus the props created by the mapStateToProps function`. As I said it not work like this, `ownProps` means not props + props created by the mapStateToProps function, it means just props. Thus, please, move it back to `VmActions`.

Here is example and explanation: https://stackoverflow.com/questions/41198842/what-is-the-use-of-the-ownprops-arg-in-mapstatetoprops-and-mapdispatchtoprops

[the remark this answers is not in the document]
sjd78: Bohdan - [1] is a much better explanation of what you're describing.  Thank you for catching it as I wasn't aware `connect()` worked in that matter and why.

But, that doesn't mean the solution is to put those props in the containing component who doesn't need to know anything about how the `ConsoleButton` needs to map data to actions.  Since the main problem with how I tried to do things is repeated function rebinding on state change causing performance issues, it's easier to provide that action's data via function arguments.  I'm making that change now.  That way the `ownProps` isn't an issue, AND the parent component doesn't need to know so much about the `ConsoleButton`.

[1] - https://github.com/reduxjs/redux/blob/master/docs/faq/DesignDecisions.md#no-asynch-in-mapDispatchToProps

src/saga/vm-console/index.js
sjd78: Yeah, no change.

src/saga/vm-console/index.js
sjd78: Backed the change out, I don't think it is really needed.  User's browser can handle multiple file downloads if they have auto-connect turned on for >1 VM.

src/components/VmActions/ConsoleButton.js
```diff
@@ -109,28 +121,31 @@ class ConsoleButton extends React.Component {
...
   (state) => ({
-    VmAction: state.VmAction,
+    usbFilter: state.config.get('usbFilter'),
+    userId: state.config.getIn(['user', 'id']),
   }),
-  (dispatch, { vm, consoleId, usbFilter, userId }) => ({
-    onCheckConsoleSessionInUse: () => dispatch(checkConsoleInUse({ vmId: vm.get('id'), usbFilter, userId })),
-    onConsoleSessionConfirmaClose: () => dispatch(setConsoleInUse({ vmId: vm.get('id'), consoleInUse: false })),
-    onDownloadConsole: () => dispatch(downloadConsole({ vmId: vm.get('id'), consoleId, usbFilter })),
+  (dispatch) => ({
```
bond95: Why did you remove `{ vm, consoleId }`, if in any case this is `ownProp`?

src/components/VmActions/index.js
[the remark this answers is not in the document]
bond95: Right, good point.

src/saga/login.js
```diff
@@ -46,7 +46,7 @@ import {
 
 import {
   downloadVmConsole,
-} from './consoles'
+} from './vm-console'
```
bond95: Please use for names `camelCase`.

src/saga/vm-console/index.js
sjd78: :man_facepalming: I didn't realize that the "connect automatically" switch only holds on to 1 VM id in browser local storage, so it's really a moot point now.

src/components/VmActions/ConsoleButton.js
sjd78: I removed them to be consistent with all of the variables going to the dispatch actions.

src/saga/login.js
sjd78: I'll rename `src/saga/vm-console` to `src/saga/console` to be consistent with the names under `src/actions`

sjd78: Fixed the file names as well.

src/components/VmActions/ConsoleButton.js
bond95: But it's not needed, because if you have some component, that represent Console, it has by default some props that define this component, in this case it is `vmId` and `consoleId` (not in every case). Thus you don't need to put them as argument to those function directly. The code will be more readable and it will be more consistent with the rest of code then now.

sjd78: I'll agree to disagree on this.  Either way the info is getting to the actions in the dispatch.

Perhaps you can write up a few of these web-ui coding best practices and put them in README or similar so other devs don't have to go through the same comment stream in the future.

bond95: > Perhaps you can write up a few of these web-ui coding best practices and put them in README or similar so other devs don't have to go through the same comment stream in the future.

Actually, yes, I had thoughts about some documentation and we already have issue for this https://github.com/oVirt/ovirt-web-ui/issues/99. Soon I get to this.

gregsheremeta: I'll need this too as I'm trying to catch my skills up

sjd78:
@bond95 - Can you take another quick look at the current changes to see if I've covered everything you saw before?

src/components/VmActions/WindowsRdpButton.js
```diff
@@ -0,0 +1,52 @@
...
+  config: PropTypes.object.isRequired,
+  onRDP: PropTypes.func.isRequired,
+}
+
+export default connect(
+  (state) => ({
+    config: state.config,
+  }),
+  (dispatch) => ({
+    onRDP: (vmName, fqdn, domain, username) => dispatch(getRDP({ vmName, fqdn, domain, username })),
```
bond95: I think we can get `vmName` and `fqdn` from `vm` as `ownProps`.

sjd78: Done!

gregsheremeta:
@sjd78 good to merge?

sjd78:
@gregsheremeta - Yes! The PR is all buttoned up.

PR #656, opened 2018-07-03
ux-redesign: remove `VerticalMenu` and update `VmsPageHeader`

sjd78:
- `VerticalMenu` was already hidden and unused, isn't included in the redesign work so can be removed

 - `VmsPageHeader` pulls it's __page__ from redux so it's dispatch action data is self-contained

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/656)


src/routes.js
```diff
@@ -63,16 +64,4 @@ const getRoutes = (vms) => ([
...
-const getMenu = () => ([
-  {
-    icon: 'pficon pficon-virtual-machine',
-    title: msg.virtualMachines(),
-    to: '/',
-  },
-])
-
-export { getRoutes, getMenu }
+export { getRoutes }
```
bond95: I think it would enough to export getRoutes as default.

sjd78: Done.

PR #657, opened 2018-07-03
ux-redesign: fix PageRouter previousPath bug when reloading edit page

sjd78:
If the user reloads/refreshes the VM edit page, the `PageRouter` component had a bug that would push a __null__ to __previousPage__ and prevent the edit component from rendering.  A truthy check on the previous path fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/657)

### DIFF
--- a/src/components/PageRouter.js
+++ b/src/components/PageRouter.js
@@ -53,7 +53,7 @@ class PageRouter extends React.Component {
       updates.closeable = !!updates.branch.route.closeable
 
       updates.currentPath = newPath
-      if (previousPath !== currentPath) {
+      if (currentPath && (previousPath !== currentPath)) {
         updates.previousPath = currentPath
       }
     }


src/components/PageRouter.js
```diff
@@ -53,7 +53,7 @@ class PageRouter extends React.Component {
       updates.closeable = !!updates.branch.route.closeable
 
       updates.currentPath = newPath
-      if (previousPath !== currentPath) {
+      if (currentPath && previousPath !== currentPath) {
```
mareklibra: Can you please add braces to improve readability and avoid future dummy bugs?

mareklibra: @sjd78 , thanks, I meant these parenthesis :-)
I understand they are not needed, but why not to help a human when parsing that in the future ...

PR #659, opened 2018-07-13
ux-redesign: refactor getRoutes() route title functions

sjd78:
A route's title function is now passed __vms__ as the second argument.

With the `ConnectedRouter` in place, passing __vms__ to `getRoutes()` was rendering the route but not changing the render when __vms__ is updated. The route doesn't change even if the __vms__ does, so titles would not get updated.

The only need for __vms__ in `getRoutes()` is to pull a VM's name for the route's title used by `Breadcrumb`. On a browser page reload for any route that uses __vms__ (VM/Pool Details and Edit), the Breadcrumb will render the VM/Pool ID initially. However, it didn't get updated to the VM/Pool name since getRoutes() was holding on to an old copy of __vms__. Pushing the __vms__ to the `title` functions ensures the correct copy of __vms__ is used every time.

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/659)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -36,7 +36,7 @@ const NoLogin = () => {
  * Main App component. Wrap the main react-router components together with
  * the various dialogs and error messages that may be needed.
  */
-const App = ({ history, vms, config, appReady }) => {
+const App = ({ history, config, appReady }) => {
   if (!config.get('loginToken')) { // login is missing
     return (
       <Grid fluid>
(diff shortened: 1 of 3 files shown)


sjd78:
@gregsheremeta , @bond95 - Can you guys take a look at the PR, it fixes a problem with the the VM's name in breadcrumbs?  It would be great to get it out in by the next release.

Open a VM's details page and reload the browser page.  You'll see the VM's ID in the breadcrumbs instead of its name.

gregsheremeta:
we'll have a 1.4.1 today / tomorrow that's what's currently merged. This one can go in 1.4.2.

gregsheremeta:
Don't merge yet :) I need to release 1.4.1 today first.

sjd78:
@gregsheremeta - This PR is reviewed and ready to merge.  There are no overlaps with #654 or #660.

Issue #647, opened 2018-06-26
Redesign notifications


bond95:
![image](https://user-images.githubusercontent.com/3332176/41908599-cd8a3a70-7944-11e8-9a9c-0486538589aa.png)

mareklibra:
Let's not group the events for now, just list them chronologically

mareklibra:
Part of https://github.com/oVirt/ovirt-web-ui/issues/596

gregsheremeta:
I'm building a notification drawer for admin portal that has 'dismiss all' and 'snooze for ...' buttons.
https://gerrit.ovirt.org/#/c/92443/

lizsurette:
@bond95 
Minor nitpick from the demo today:
The Clear All button and footer area should not be shown when there are no notifications.

PR #660, opened 2018-07-13
ux-redesign: UserMessages redesign according to Patternfly

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/647

image
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/660)


lizsurette:
@bond95 Sorry for my delayed review! Back from PTO :)

This is looking great! Just one comment:
1) Can you add an empty state when there are no notifications? I think it will help make it more clear. In this case the Clear All and Mark all Read actions can go away too. Check out an example here by clicking "Clear All": http://www.patternfly.org/pattern-library/communication/notification-drawer/#code

bond95:
@lizsurette Done.
image
image

bond95:
@gregsheremeta @sjd78 Please code review.

src/components/VmUserMessages/index.js
```diff
@@ -3,120 +3,103 @@ import PropTypes from 'prop-types'
...
   }
 
   render () {
-    const { userMessages, onClearMessages } = this.props
+    const { userMessages, onClearMessages, onDisplayMessages, onDismissMessage } = this.props
 
     const idPrefix = `usermsgs`
-    const onToggle = () => {
-      this.setState({ show: !this.state.show })
+    const handleToggle = () => {
```
sjd78: If you move this to the class level (and rebind against this in the constructor), the arrow function will not need to be recreated at each render.

src/components/VmUserMessages/index.js
```diff
@@ -3,120 +3,103 @@ import PropTypes from 'prop-types'
...
+      this.setState((prevState) => ({ show: !prevState.show }))
     }
 
-    let show = ''
-    if (this.state.show) {
-      show = 'show'
+    const handleExpand = () => {
```
sjd78: Same as `handleToggle`

src/components/VmUserMessages/index.js
```diff
@@ -3,120 +3,103 @@ import PropTypes from 'prop-types'
...
-        <div className={`blank-slate-pf ${style['no-messages']}`}>
-          <div className='blank-slate-pf-icon'>
-            <span className='pficon pficon pficon-info' />
-          </div>
-          <h1>
-            {spaceToNonbreakableSpace(msg.noMessages())}
-          </h1>
-        </div>)
+    const filteredMessages = userMessages.get('records').filter((r) => r.get('show'))
+    const messagesList = filteredMessages.size ? filteredMessages.map(r => {
```
sjd78: Could be simplified:
````javascript
    let idCounter = 0
    const messagesList = filteredMessages.size
      ? filteredMessages.map(r =>
        <UserMessage
          key={`msg-${r.get('time')}`}
          record={r}
          id={`${idPrefix}-msg-${r.get('time')}-dropdown`}
          onDismissMessage={() => onDismissMessage(r.get('time'))}
        />
      )
      })
      : <NotificationDrawer.EmptyState title={msg.noMessages()} />
````

File: src/components/VmUserMessages/index.js [hunk 4]
```diff
@@ -3,120 +3,103 @@ import PropTypes from 'prop-types'
...
+        </NotificationDrawer>
       </li>
     )
   }
 }
 VmUserMessages.propTypes = {
   userMessages: PropTypes.object.isRequired,
   onClearMessages: PropTypes.func.isRequired,
+  onDisplayMessages: PropTypes.func.isRequired,
+  onDismissMessage: PropTypes.func.isRequired,
```
sjd78: What is the difference between "Clear" and "Dismiss"?

src/components/VmUserMessages/index.js
```diff
@@ -3,120 +3,103 @@ import PropTypes from 'prop-types'
...
-        <pre className={style['message-box']} id={id}>
-          <Time time={record.time} cssClass={style['usermsg-time']} />
-          {record.message}
-        </pre>
-      </span>
-    </li>
-  )
+const UserMessage = ({ record, id, onDismissMessage }) => {
+  const time = getFormatedDateTime(record.get('time'))
+  return (<Notification seen>
```
sjd78: Where is `seen` defined?

sjd78:
screenshot-localhost-3000-2018 08 05-15-52-54

In empty state, 3 questions.
1. Should the 'Clear all' button be disabled since it is a no-op anyway.
2. Should the 'Display all' button should also be disabled if there are no hidden messages? 
3. Why the extra space below the buttons?  I assume the buttons should be at the bottom of the min-height of the notification pane, or the pane should be shorter.

lizsurette:
@sjd78 Adding my thoughts on some of the design-y questions :)

> If I "Clear All", should "Display All" bring them all back? (based on the code, yes)
I would even wonder if we need the concept of "Display All".

> If I hit "Clear All" now and then 20 minutes from now open the user messages and hit "Display All", should ALL of the previously generated messages come back? (based on the code, yes)
Again, what's the use case for needing to bring them back?

> Is "Dismiss" on a single message the same function as "Clear All"? (based on the code, yes)
It should just clear that one message.

gregsheremeta:
> Again, what's the use case for needing to bring them back?

+1. Seems useless to me.

src/helpers.js
```diff
@@ -217,3 +217,11 @@ export function arrayToMap (array, keySelector) {
     return accum
   }, {})
 }
+
+export function getFormatedDateTime (timestamp) {
```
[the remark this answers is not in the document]
gregsheremeta: +1. We can do this in a follow up PR

src/index-nomodules.css
```diff
@@ -75,3 +75,13 @@ div.toolbar-pf {
 .popover {
   min-width: 120px;
 }
+
+.navbar-nav > .dropdown {
+  position: inherit;
+}
+
+.drawer-pf .blank-slate-pf .blank-slate-pf-title {
+    font-size: 1em;
```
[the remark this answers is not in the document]
gregsheremeta: sometimes that's good practice for accessibility reasons

gregsheremeta:
> Why the extra space below the buttons?

indeed, that should be shortened up

sjd78:
@lizsurette , @gregsheremeta , @bond95  - the tray buttons match what is currently in webadmin (not that it makes sense either).

&lt;off-the-top-of-my-head-before-I-looked-at-the-patternfly-website-again&gt;
Dismiss (single), Clear all, and Display all could make more sense in terms of unread messages vs read messages (seen) and deleting the messages.  So the badge shows the unread count, the tray shows unread + read messages, and the bottom buttons could be "Hide all read" and "Show all read" or "Delete all read" and "Mark all unread" or something similar.  Read/unread/delete -- the operations that can be taken on messages.
&lt;/off-the-top-of-my-head&gt;

[PF Notification Drawer](https://www.patternfly.org/pattern-library/communication/notification-drawer/) has "Unread count", "Mark All Read", and "Clear All".  It also says "this option may be used differently across products", so there is just mild guidance.

src/index-nomodules.css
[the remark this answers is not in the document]
sjd78: I read all about font-size and relative units etc etc.  Percents and ems seem to work very similar as they both do an adjustment to the inherited height.  At this point, I have no real concerns, just curious.

src/helpers.js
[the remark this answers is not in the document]
sjd78: People will complain.  People don't like their dates in the "wrong" format.  In my i18n experience, it always comes up surprisingly quickly.

src/actions/userMessages.js
```diff
@@ -0,0 +1,28 @@
+import {
```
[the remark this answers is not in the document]
bond95: VmPortal have only one type of notifications, thats why notification drawer have only one category. Thus doing universal action like `addUserMessage` better in follow PR, for example here https://github.com/oVirt/ovirt-web-ui/issues/695

src/components/VmUserMessages/index.js
bond95: There is not need to define `seen`, enough just add it as props.
P.S. Best practice for set bool props.

bond95:
@sjd78 

> Why the extra space below the buttons? I assume the buttons should be at the bottom of the min-height of the notification pane, or the pane should be shorter.

Actually this is default behavior of notification drawer. Like here for example https://www.patternfly.org/pattern-library/communication/notification-drawer/#code
or in WebAdmin. So for consistent better to leave it as it is.

> I would even wonder if we need the concept of "Display All".
@sjd78 @lizsurette @gregsheremeta

lizsurette:
> Actually this is default behavior of notification drawer. Like here for example https://www.patternfly.org/pattern-library/communication/notification-drawer/#code
> or in WebAdmin. So for consistent better to leave it as it is.

This could be a bug in PF-core. Taking a look at the PF-React example looks better to me without the extra spacing... https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=patternfly-react%2FCommunication%2FNotification%20Drawer&selectedStory=Notification%20Drawer&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs



> @sjd78 @lizsurette @gregsheremeta Okay so you think we need to delete "Display All"?

Yes, I'm in favor of removing.

src/components/VmUserMessages/index.js
sjd78: Ok, I got confused. You're just rendering all Notifications as seen.

At first glance, I assumed you intended to keep seen vs unseen (read vs unread).

bond95:
I removed 'Display All' button. And made rebase.

image
image

gregsheremeta:
> We could do better than Date.now() for a unique id for the messages

+1

> Get rid of `FAILED_EXTERNAL_ACTION` and `LOGIN_FAILED` and just replace them with a new `ADD_USER_MSG`

I don't know enough yet to get this one :)

> would make more sense from the data perspective to be `HIDE_ALL_USER_MSGS`, `HIDE_USER_MSG`, `SHOW_ALL_USER_MSGS`

I don't follow this one -- are we lining those up with something?

> Either they should be removed, or they should be updated on every operation.

+1

> @gregsheremeta - Should we add an issue for this so we don't forget?

Yep!

gregsheremeta:
@sjd78  opened #712 #713

Can you elaborate on points 2 and 3 above?

Issue #666, opened 2018-07-16
snapshots: order of snapshots is alphabetical


jelkosz:
it should be ordered by creation date, no alphabetically.

PR #667, opened 2018-07-16
snapshots: Sort snapshots by creation time

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/666

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/667)

### DIFF
--- a/src/components/VmSnapshots/index.jsx
+++ b/src/components/VmSnapshots/index.jsx
@@ -7,7 +7,6 @@ import DeleteConfirmationModal from '../VmModals/DeleteConfirmationModal'
 import { msg } from '../../intl'
 import { isRunning } from '../utils'
 import FieldHelp from '../FieldHelp'
-import naturalCompare from 'string-natural-compare'
 import { PendingTaskTypes } from '../../reducers/pendingTasks'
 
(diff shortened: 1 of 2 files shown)


src/components/VmSnapshots/index.jsx
```diff
@@ -141,8 +140,20 @@ const VmSnapshotConnected = connect(
   (dispatch) => ({})
 )(VmSnapshot)
 
+function sortSnaphots (a, b) {
```
sjd78: This would be a good function to test in a unit test.

src/ovirtapi.js
```diff
@@ -339,6 +339,8 @@ OvirtApi = {
     return {
       id: snapshot.id,
       description: snapshot.description,
+      type: snapshot.snapshot_type,
+      date: snapshot.date,
       persistMemoryState: snapshot.persist_memorystate === 'true',
```
sjd78: I'd add
````javascript
  isActive: snapshot.snapshot_type === 'active',
````
here since the test will always be done at least once in sorting and probably again when rendering the snapshot's component item.

PR #670, opened 2018-07-16
snapshots: Snapshot restore confirmation

bond95:
Fix restore confirmation modal dialog. Before it shows same as delete confirmation dialog.
Only last commit!

Fixes: #662

image

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/670)

### DIFF
--- a/src/components/VmSnapshots/RestoreConfirmationModal.js
+++ b/src/components/VmSnapshots/RestoreConfirmationModal.js
@@ -1,0 +1,37 @@
+import React from 'react'
+import PropsTypes from 'prop-types'
+import { Modal } from 'patternfly-react'
+import { msg } from '../../intl'
+
+const RestoreConfirmationModal = ({ show, onRestore, description, onClose }) => {
+  return (
+    <Modal onHide={onClose} show={show}>
+      <Modal.Header>
+        <button
+          className='close'
+          onClick={onClose}
+        >
+          <span className='pficon pficon-close' title='Close' />
+        </button>
+        <Modal.Title>{msg.confirmRestore()}</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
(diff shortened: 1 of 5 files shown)


src/components/VmSnapshots/RestoreConfirmationModal.js
```diff
@@ -0,0 +1,37 @@
...
+        <button className='btn btn-default' onClick={onClose}>{msg.cancel()}</button>
+        <button className='btn btn-primary' onClick={onRestore}>{msg.restore()}</button>
+      </Modal.Footer>
+    </Modal>
+  )
+}
+
+RestoreConfirmationModal.propTypes = {
+  show: PropsTypes.bool.isRequired,
+  description: PropsTypes.string.isRequired,
```
sjd78: **s/description/snapshotName/** since the value is put directly in an externalized string with a placeholder of that same name.

src/components/VmSnapshots/index.jsx
```diff
@@ -99,14 +100,14 @@ class VmSnapshot extends React.Component {
           </span>
           {snapshotMemoryInfo}
           {
-            !isUp
+            !isUp && !isActive
```
mareklibra: unnecessary ternary operator

src/components/VmSnapshots/RestoreConfirmationModal.js
mareklibra: Agree, `description` denotes longer text with an explanation. This prop is used just as a parameter of that.

mareklibra:
@sjd78 , I would normally not merge this unless the comments are addressed but there's pressure on having new release shortly.

@bond95 , please don't forget to address them in a follow-up.

PR #674, opened 2018-07-20
Developer tooling convenience changes

sjd78:
- Use **redux-devtools** hook (without the npm package) to hook the devtools and blacklist the ***_ACTIVE_REQUEST** actions from the redux browser extension. Now devs can see the actions they care about.

 - Reduce action logging on the browser console by ignoring ***_ACTIVE_REQUEST** actions.

 - Remove `redux-devtools-extension` since it isn't needed.

 - Upgrade `chalk` so string tagged template literals can be used (in the dev server and build scripts).

 - Dev server start script now prints out the user / user id and sso token used to connect to the REST api.

 - `.eslintignore` so vscode's eslint plugin won't lint the build, config and script files.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,6 @@
+#
+# eslint to ignore config, script and build output
+#
+build/*
+config/*
+scripts/*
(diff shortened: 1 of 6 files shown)


gregsheremeta:
lgtm. @bond95 ?

PR #678, opened 2018-07-26
ux-redesign: refactor the main layout to use flex box, simplify CSS use

sjd78:
Using flex-box in the layout allows better control of the height of components to fill space when the content is not longer then the window.

There should be no (major) visible difference between the old layout and the new layout.


sjd78:
nudge for a review ... @bond95, @gregsheremeta

gregsheremeta:
> send, please, screenshot.

i'm guessing there isn't one because

> There should be no (major) visible difference between the old layout and the new layout.

src/components/sharedStyle.css
```diff
@@ -1,78 +1,42 @@
...
-    cursor: pointer;
+  cursor: pointer;
+}
+
+/* For DetailContainer (in VmDetail and VmDialog) */
+.move-left-detail { /* VM detail on the right */
+  height: 100%;
+  display: block;
+  max-width: 95%;
+  width: 100%;
```
bond95: Why you set width 100% if max-width is 95%?

bond95:
Hmm... Okay, I was just curious how it looks like, because I see changes in CSS.

src/components/sharedStyle.css
sjd78: This rule is unchanged (just moved from the top to the bottom of the file), applies to `src/components/DetailContainer.js`, and will eventually go away when the `VmDetail` and `VmDialog` pages are dropped.

But I did look at it again and it can be simplified to `width: 95%` and look the same.

bond95: As I see previous `.move-left-detail` look like this (and it without any `width:100%`):
```
.move-left-detail { /* VM detail on the right */	
    height: 100%;	
    display: block;	
    max-width: 95%;	
    opacity:1;	
}
```

sjd78:
> Hmm... Okay, I was just curious how it looks like, because I see changes in CSS.

You've actually been running the patch in your snapshot PR work (#691) for awhile now!

sjd78:
Here is current master's list and details pages cross faded from master(before flex-box) to PR(after flex-box) screen caps:

ezgif com-gif-maker

ezgif com-gif-maker 1

PR #683, opened 2018-07-30
reattach VmDialog's sagas to the root saga so VM edits get saved

### LINKED ISSUE #684: VM edits don't save
In PR #651, refactoring missed VmDialog's sagas the same way as VmDisks, NewDiskDialog and VmSnapshots. As a result, the VM Edit page will not currently save any changes.

sjd78:
In PR #651, I missed refactoring `VmDialog`'s sagas the same way as `VmDisks`, `NewDiskDialog` and `VmSnapshots`.  As a result, the VM Edit page will not currently save any changes.  This PR re-links the `VmDialog` sagas back to the root saga so they can be executed and edits committed.

Fixes #684

---
This change is [Reviewable](https://reviewable.io/reviews/ovirt/ovirt-web-ui/683)

### DIFF
--- a/src/components/VmDialog/sagas.js
+++ b/src/components/VmDialog/sagas.js
@@ -1,21 +1,24 @@
 import { put, takeEvery, takeLatest } from 'redux-saga/effects'
-import { refresh, getSingleVm } from '../../actions/index'
-import { ADD_NEW_VM, EDIT_VM } from './constants'
-import { setSavedVm } from './actions'
-import Api from '../../ovirtapi'
 
-function* createNewVm (sagas, action) {
-  const result = yield sagas.callExternalAction('addNewVm', Api.addNewVm, action)
+import { ADD_NEW_VM, EDIT_VM } from './constants'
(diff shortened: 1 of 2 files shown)


src/components/VmDialog/sagas.js
```diff
@@ -1,21 +1,24 @@
...
     yield put(setSavedVm({ vm: Api.vmToInternal({ vm: result }) }))
   }
 }
 
-function* editVm (sagas, action) {
-  const result = yield sagas.callExternalAction('editVm', Api.editVm, action)
+function* editVm (action) {
+  const result = yield callExternalAction('editVm', Api.editVm, action)
   if (!result.error) {
-    const result2 = yield sagas.callExternalAction('changeCD', Api.changeCD, {
```
bond95: I think it could be named more specific.

gregsheremeta: I was gonna say the same, but it was already named that, so we're good

Issue #686, opened 2018-07-31
RDP button sends 'console.rdp' to the user at render time instead of onClick event time


sjd78:
When a Windows VM is opened in the "legacy" VM Detail screen, the `WindowsRdpButton` is rendered to allow the user to connect to the RDP session.  Unfortunately, the `onClick` event handler is being fired immediately at render time and not when the user clicks the button.   This causes web-ui to send the user a **console.rdp** connection file.

Also, since the background refresh triggers a re-render and the refresh happens every minute, the user is sent a new **console.rdp** file every minute.

Caused by PR #651

gregsheremeta:
is this a recent regression?

what's the symptom?

gregsheremeta:
@michalskrivanek do you want to release 1.4.1 asap (without this) or wait for it?

sjd78:
Closing since the issue is resolved with the merge of #687.

PR #687, opened 2018-07-31
fix RDP button event handler

sjd78:
Addresses #686 

- The event handler was being executed at render time instead of at `onClick` event time. Correctly wrapping the event handler in an arrow function solves the problem.

### DIFF
--- a/src/components/VmActions/WindowsRdpButton.js
+++ b/src/components/VmActions/WindowsRdpButton.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 
 import { getRDP } from '../../actions'
-import { isWindows, hrefWithoutHistory } from '../../helpers'
+import { isWindows } from '../../helpers'
 import { canConsole } from '../../vm-status'
 
 /**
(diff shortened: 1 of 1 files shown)


gregsheremeta:
please split into 2 separate PRs -- one for the CSS / view, one for the rdp file

src/components/VmActions/WindowsRdpButton.js
```diff
@@ -23,7 +23,7 @@ const WindowsRdpButton = ({ vm, className, config, onRDP }) => {
         href='#'
         key={vm.get('id')} id={`${idPrefix}-rdp`}
         className={className}
-        onClick={hrefWithoutHistory(onRDP(domain, username))}
+        onClick={(e) => { e.preventDefault(); onRDP(domain, username) }}
```
bond95: Why do you use here `preventDefault`?

sjd78: Same exact thing as `hrefWithoutHistory` does, but without that function call.

Equivalent change with maintaining the helper function...
````javascript

  onClick={hrefWithoutHistory(e => onRDP(domain, username))}

````

I removed the helper function use because it tripped me up.

gregsheremeta: `hrefWithoutHistory` is weird if we can just use `preventDefault`. +1.

gregsheremeta:
lgtm. Testing, then I'll merge.

PR #689, opened 2018-08-01
fix the render of the console button list on the VmDetail component

sjd78:
On the VM detail page, the button are now rendered with separators
both when the VM `canConsole()` and it can't `canConsole()`.


gregsheremeta:
this isn't critical and Bohdan wants to go first ;) so I'll hold off merging this until 1.4.2.

@sjd78 can you insert before and after screens (and on all such things)? ty

sjd78:
`VmDetail` is being replaced in #699 and #654 reworks the console button completely.

This PR is no longer needed as it just cleans up the console button list on VmDetail prior to merging #699.

PR #679, opened 2018-07-26
refactor `ovirtapi` to separate concerns

sjd78:
`ovirtapi` has been split to have transport, data transform and API interactions in separate files:

 - HTTP transport functions are in `ovirtapi/transport.js`
 - data transform functions (from API to internal structures and back again) are in `ovirtapi/transform.js`
   - convertInt() in API transform for some values that
     transport as strings
 - API interaction functions are in `ovirtapi/index.js`
 - `ovirtapi/index.js` presents the save facade as the previous `ovirtapi.js`
 - create `ovirtapi` types, initially applied to the transform
   functions
   - each transform object has 2 corresponding types:
      - Api_Type to define the type in and out of the REST API
      - _Type to define the type within the application/redux
   - Most of the types are stubbed to `Object` so the names can be used and the actual contents can be defined from the REST API definitions and transform functions at a later time.

PR #690, opened 2018-08-01
 Adapt sub-resource/deep fetch handling for VM resources

sjd78:
This PR builds on PR #679. ~~Until #679 is merged, its commit is included in this PR.~~

VM sub-resources are the part of the VM in the REST API that require either a follow-up fetch request or a `?follow=linkItem1,linkItem2,linkItem3` deep fetch.

Updated **sagas.js**:
 - the "deep fetch" additional list is shared across all uses
 - `fetchSingleVm` will have the `updateVms` action copy the VM sub resources as appropriate (this allows deep fetch details to be copied forward when the VM is updated by a shallow fetch from the background data refresh task)

Updated the base VM transform (**ovirtapi/transform.js**) to handle all sub-resources consistently. This allows the VM reducer to properly carry forward the sub-resources after a shallow fetch.

Updated **reducers/vms.js** to handle the deep fetch data (the resources retained when _copySubResources_ is _true_) consistently.

### DIFF
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -123,21 +123,9 @@ const VM = {
       },
       bootMenuEnabled: vm.bios && vm.bios.boot_menu && convertBool(vm.bios.boot_menu.enabled),
       cloudInit: CloudInit.toInternal({ vm }),
-
-      // TODO: Should snapshots be part of 'subResources'?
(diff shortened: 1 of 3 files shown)


sjd78:
nudge for a review ... @bond95, @gregsheremeta

PR #691, opened 2018-08-01
ux-redesign: Redesign snapshot card

bond95:
If you have an idea what add to Snapshot detail or to card or if I forgot something, please write it to comments.

ToDo list of features:
- [ ] Snapshot status
- [ ] Memory included

Only last commit!

image
image
image
image


lizsurette:
@bond95 This is looking great. I have a few comments:
- The "Create Snapshot" link should be at the top of the list to allow for scaling and it will never be hidden from the user.
- Creation time can be changed to "Created".
- I think Defined Memory should just be "Memory".
- How about adding any disks information to the snapshot info tooltip?
- Restore button should be left aligned to the panel.
- Titles in the confirmation modals should Camel case. So "Confirm Restore" rather than "Confirm snapshot".
- The Add New Snapshot modal should have a title and button that match in terminology. I'd suggest "Create Snapshot" for both modal title and button label.

Let me know if you have any questions on these suggestions!

bond95:
New version.

image
image

gregsheremeta:
@sjd78 let's try to get this one in today too

sjd78:
@bond95 - rebase to upstream/master or #699 please

package.json
```diff
@@ -75,9 +75,9 @@
     "intl-messageformat": "2.2.0",
     "jquery": "=3.2.1",
     "object-assign": "4.1.1",
-    "patternfly": "3.48.1",
+    "patternfly": "3.53.1",
```
gregsheremeta: the build is failing on node6. Looks like it's from this upgrade.

gregsheremeta:
@bond95 please rebase on master.

What's left on this? Is it ready for final review?

bond95:
@gregsheremeta @lizsurette @sjd78 @lwrigh So what do you think?

image
image
image
image
image
image

lizsurette:
@bond95 Looks awesome! A couple of questions:

1) What are the icons referring to about the snapshots? Looks like there is a "locked" state and "in preview" state. Do you know what these mean for the user? Does the green check icon mean everything is okay? Just curious and I can make any recommendations based on this.

2) Do we know how many more disks we will be showing in the "Show More" tooltip? If yes, I'd suggest we change that link text accordingly to "1 More Disk..." just to describe it a little more to the user. If we can't compute it ahead of time, I think it looks great the way it is.

gregsheremeta:
> What are the icons referring to about the snapshots? Looks like there is a "locked" state and "in preview" state

+1. Can we add more text to the tooltip? "Locked -- this snapshot is locked because ...", "Preview -- this snapshot is in preview run, so ..."

I don't completely know what locked and preview mean, so adjust my explanations as necessary

bond95:
@lizsurette @gregsheremeta 'Locked' state means that now snapshot can't be used, it be because of snapshot still creating, or snapshot deleting. 'In preview' means that we temporary restore snapshot, and because of that we can't create any snapshot, or delete some, or even restore..

gregsheremeta:
@lizsurette @lwrigh can you provide tooltip text based on https://github.com/oVirt/ovirt-web-ui/pull/691#issuecomment-413828714

gregsheremeta:
works for me

lizsurette:
@bond95 @gregsheremeta @lwrigh I didn't actually give any recommendations on tool-tip text, but I'm thinking we try to keep it simple so for "Locked" something like "This snapshot is locked because it is still being created. You can not delete it at this time." I'm thinking the same with "In Preview". "Creating a snapshot is currently unavailable while <snapshot x> is being restored."

lwrigh:
@bond95 @gregsheremeta @lizsurette Those all sound like good directions to me for the tooltips. It totally makes sense to me to disable the actions if they are not currently available at a time and give the reason why they are unavailable.

src/sagas.js
```diff
@@ -354,7 +354,12 @@ export function* fetchSingleVm (action) {
       internalVm.sessions = yield fetchVmSessions({ vmId: internalVm.id })
       // TODO: Support <4.2 for snapshots?
     }
-
+    for (let i in internalVm.snapshots) {
```
sjd78: Can this be done with `'snapshots.disks'` or similar on the `vmFetchAdditionalList` list (which translates to `?follow=...` on the API request)?  Then add the disks transform to the `Snapshot.toInternal` transform...

As is, if a VM has a lot of snapshots, it'll make a lot of requests.

If the disks cannot be loaded by a follow, I recommend you open a BZ on the rest api to get it added!  The fewer requests the better.  Leave a TODO note in the saga explaining this.

src/components/VmDetails/style.css
```diff
@@ -108,3 +138,61 @@
...
+}
+
+div.popover {
+  min-width: 200px;
+  max-width: 1000px;
+  z-index: 20;
+}
+
+.green {
+  color: #24d011;
```
sjd78: Wrong green. See #709

src/components/VmModals/DeleteConfirmationModal.js
```diff
@@ -1,36 +1,54 @@
...
+          primaryContent={<p className='lead'>{ children }</p>}
+        />
+      </React.Fragment>
+    )
+  }
 }
 
 DeleteConfirmationModal.propTypes = {
-  show: PropsTypes.bool.isRequired,
+  trigger: PropsTypes.node.isRequired,
```
sjd78: This style (also on `RestoreConfirmationModal`) seems a bit awkward, with a trigger component like this.  But I'm not sure how you could change it around to be more like `OverlayTrigger`.

I'm ok with how it is, but it would be worth looking at how it could look a bit more like `OverlayTrigger` (https://react-bootstrap.github.io/components/popovers/#popovers-overlay-trigger)

src/components/VmDetails/BaseCard.js
```diff
@@ -72,6 +72,8 @@ class BaseCard extends React.Component {
     const hasBadge = itemCount !== undefined
     const hasIcon = icon && icon.type && icon.name
 
+    const renderedChildren = React.isValidElement(children) ? React.cloneElement(children, { isEditing: editing }) : children({ isEditing: editing })
```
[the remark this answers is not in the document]
bond95: > `<this.renderChildren isEditing={editing} />`

It doesn't work like this. But yeah, it makes sense.

src/components/VmDetails/cards/SnapshotsCard/NewSnapshotModal.jsx
```diff
@@ -1,7 +1,11 @@
...
 import PropTypes from 'prop-types'
+
+import { connect } from 'react-redux'
+import { addVmSnapshot } from './actions'
+
 import { Button, Modal, Icon, FormControl, Alert, FormGroup, HelpBlock } from 'patternfly-react'
-import { msg } from '../../intl'
+import { msg } from '../../../../intl'
 
 class NewSnapshotModal extends Component {
```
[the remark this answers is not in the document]
bond95: Oh yeah right, thank you. No, there is no any reason to use `jsx` instead of `js`.

src/components/VmDetails/cards/SnapshotsCard/sagas.js
```diff
@@ -23,7 +23,7 @@ function* deleteVmSnapshot (action) {
   }
   yield put(addSnapshotRemovalPendingTask(snapshotId))
   let snapshotRemoved = false
-  for (let delaySec of [ 4, 4, 4, 30, 60 ]) {
+  for (let delaySec of [ 4, 4, 4, 30, 30, 60 ]) {
```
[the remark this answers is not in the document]
bond95: Using sync requests for this type action is not very good idea, because deleting can last to long, thus even can be 'TIMEOUT'. It means that async way is more safe way to then sync. Also we can do some manipulation in delete action, while snapshot still deleting.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
```diff
@@ -0,0 +1,96 @@
...
+import RestoreConfirmationModal from './RestoreConfirmationModal'
+import DeleteConfirmationModal from '../../../VmModals/DeleteConfirmationModal'
+import SnapshotDetail from './SnapshotDetail'
+import { deleteVmSnapshot } from './actions'
+import { formatDateFromNow } from '../../../../helpers'
+import { PendingTaskTypes } from '../../../../reducers/pendingTasks'
+
+const StatusTooltip = ({ icon, text, id }) => {
+  return <OverlayTrigger overlay={<Tooltip id={id}>{text}</Tooltip>} placement='left' trigger={['hover', 'focus']}>
+    <a>{icon}</a>
```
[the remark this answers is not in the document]
bond95: In `SnapshotItem` I use `a` as indicator that there's some action. So if you click on it or hover it, action will do something. And it is easy to style, not needed a lot of classNames.

src/sagas.js
bond95: Unfortunately `'snapshots.disks'` doesn't work.

sjd78: Are there other `follow` options that do work?

If the snapshot's disks cannot be loaded by a follow, I recommend you open a BZ on the rest api to get it added!

src/components/VmDetails/BaseCard.js
```diff
@@ -57,6 +57,11 @@ class BaseCard extends React.Component {
     }
   }
 
+  renderChildren = (childProps) => {
```
sjd78: I just posted #735 to combine the `renderChildren` technique while converting the event handler style to match the rest of the code base.

src/components/VmDetails/cards/SnapshotsCard/sagas.js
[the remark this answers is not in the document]
bond95: The problem is not on client. Because after 14 minutes Engine drop connection by itself and return empty result.

sjd78:
rebase on master to fix the CI build

src/components/VmDetails/cards/SnapshotsCard/sagas.js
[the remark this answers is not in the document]
sjd78: > The problem is not on client. Because after 14 minutes Engine drop connection by itself and return empty result.

That makes more sense.  The request would timeout on the server side.

Another question then - does the snapshot get put into a "delete in progress" state or anything useful like that?  If it did, you could at least filter out or distinctly display that snapshot...

src/ovirtapi/index.js
```diff
@@ -203,7 +207,7 @@ const OvirtApi = {
       contentType: 'application/xml',
     })
   },
-  addNewSnapshot ({ vmId, snapshot }: { vmId: string, snapshot: Object }): Promise<Object> {
+  addNewSnapshot ({ vmId, snapshot }: { vmId: string, snapshot: SnapshotType }): Promise<Object> {
     assertLogin({ methodName: 'addNewSnapshot' })
     const input = JSON.stringify(OvirtApi.internalSnapshotToOvirt({ snapshot }))
```
sjd78: Looking at the REST docs, looks like you only need the description to create a snapshot.  Using a full `SnapshotType` here is a bit much and requires the transform default values to prevent errors with undefined values.

If you need to support fine grained details (`persist_memorystate` or `disk_attachments`) making those parameters helps explicitly define how the API function works.  In other words, don't require a developer to know exactly how the rest call works when they look at it.

src/ovirtapi/transform.js
```diff
@@ -288,10 +289,12 @@ const Pool = {
 const Snapshot = {
   toInternal ({ snapshot }: { snapshot: ApiSnapshotType }): SnapshotType {
     return {
-      id: snapshot.id,
+      id: snapshot.id || '',
```
sjd78: ~~Why add default values?~~  Ok, I see you're using in the in the API function and need to avoid undefined values in the object causing problems.

src/ovirtapi/index.js
bond95: Yeah, you right, here must be `ApiSnapshotType`.

bond95:
Rebased and replace `SnapshotType` by `ApiSnapshotType`. @sjd78 Please review.

src/components/VmDetails/cards/SnapshotsCard/SnapshotDetail.js
```diff
@@ -0,0 +1,154 @@
...
+        <dd>
+          {statusMap[snapshot.get('status')]}
+        </dd>
+      </dl>
+      <dl className={style['snapshot-properties']}>
+        <dt>
+          {msg.template()}
+        </dt>
+        <dd>
+          {template ? templateNameRenderer(template) : ''}
```
sjd78: Doesn't need to be a ternary when the else is `''`.

src/components/VmDetails/cards/SnapshotsCard/SnapshotDetail.js
```diff
@@ -0,0 +1,154 @@
...
+                key='info'
+              >
+                <a>Show more</a>
+              </OverlayTrigger>
+            }
+          </div>
+        </dd>
+      </dl>
+    </div>
+    <div style={{ textAlign: 'left' }}><RestoreConfirmationModal snapshot={snapshot} vmId={vmId}><Button bsStyle='default'>{ msg.restoreSnapshot() }</Button></RestoreConfirmationModal></div>
```
sjd78: nit-pick: Can you wrap this a bit to cut down on horizontal scrolling when reading the code?  Thanks.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
```diff
@@ -0,0 +1,100 @@
...
+  let statusIcon = null
+  if (!snapshot.get('isActive')) {
+    buttons.push(<OverlayTrigger
+      overlay={<SnapshotDetail key='detail' snapshot={snapshot} vmId={vmId} />}
+      placement='left'
+      trigger='click'
+      rootClose
+      key='info'
+    >
+      <a><Icon type='pf' name='info' /></a>
```
sjd78: If the `a` is just a tag to style, it can be any other tag (or no tag at all).  Do browsers still complain about `a` tags without an `href`?

src/components/VmDetails/cards/SnapshotsCard/index.js
```diff
@@ -0,0 +1,53 @@
...
+    <React.Fragment>
+      {
+        !isVmInPreview &&
+        <div className={style['snapshot-create']}>
+          <NewSnapshotModal vmId={vmId} />
+        </div>
+      }
+      {snapshots.sort((a, b) => b.get('date') - a.get('date')).map((snapshot) => (
+        <SnapshotItem key={snapshot.get('id')} snapshot={snapshot} vmId={vmId} isEditing={!isVmInPreview} />
+      ))}
```
sjd78: Can you double check that the new snapshot left icon is the same size as the left status/info icons on the snapshots?  I remember them off by just a bit.

src/components/VmDetails/cards/SnapshotsCard/sagas.js
```diff
@@ -1,10 +1,10 @@
 import { takeEvery, put } from 'redux-saga/effects'
 
 import { ADD_VM_SNAPSHOT, DELETE_VM_SNAPSHOT, RESTORE_VM_SNAPSHOT } from './constants'
-import Api from '../../ovirtapi'
-import { callExternalAction, delay } from '../../saga/utils'
-import { fetchVmSnapshots, startProgress, stopProgress } from '../../sagas'
-import { addSnapshotRemovalPendingTask, removeSnapshotRemovalPendingTask } from '../../actions'
+import Api from 'ovirtapi'
```
sjd78: Oops on my part, see #740.  Rebase and the build should work again.

src/components/VmDetails/cards/SnapshotsCard/sagas.js
sjd78: You have something on the snapshot card that looks at the `pendingTasks` to try to show a snapshot as "delete in progress" correct?

src/helpers.js
```diff
@@ -225,3 +225,22 @@ export function getFormatedDateTime (timestamp) {
     date: `${t.getDate()}/${t.getMonth()}/${t.getFullYear()}`,
   }
 }
+
+export function formatDateFromNow (d) {
```
sjd78: Would be great to have a few unit tests for this function so:
  1. its output can be stable, and
  2. we can see, at a glance, what the output should be based on various inputs (sometimes it is easier to figure out what a function does by looking at its tests)

src/ovirtapi/index.js
```diff
@@ -203,7 +207,7 @@ const OvirtApi = {
       contentType: 'application/xml',
     })
   },
-  addNewSnapshot ({ vmId, snapshot }: { vmId: string, snapshot: Object }): Promise<Object> {
+  addNewSnapshot ({ vmId, snapshot }: { vmId: string, snapshot: ApiSnapshotType }): Promise<Object> {
```
sjd78: I think you need `SnapshotType` here since 2 lines down there is a `OvirtApi.internalSnapshotToOvirt({ snapshot })`

src/sagas.js
sjd78: Do you plan on checking other `follow` options or opening a BZ on the rest api to get that functionality (potentially in the future)?

src/components/VmDetails/cards/SnapshotsCard/sagas.js
bond95: According to new design, it just hide `delete` and `restore` button for snapshot that is being deleted.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
bond95: Not any, but yes, and it can be `a` as well. Browsers don't complain about that.

src/ovirtapi/index.js
```diff
@@ -232,6 +236,10 @@ const OvirtApi = {
       contentType: 'application/xml',
     })
   },
+  snapshotDisks ({ vmId, snapshotId }: { vmId: string, snapshotId: string }): Promise<Object> {
+    assertLogin({ methodName: 'snapshotDisks' })
+    return httpGet({ url: `${AppConfiguration.applicationContext}/api/vms/${vmId}/snapshots/${snapshotId}/disks` })
+  },
   snapshot ({ vmId, snapshotId }: { vmId: string, snapshotId: string }): Promise<Object> {
```
sjd78: You may be able to `Promise<Object>` to `Promise<ApiSnapshotType>` here

src/components/VmDetails/BaseCard.js
```diff
@@ -76,9 +76,7 @@ class BaseCard extends React.Component {
 
   renderChildren (childProps) {
     const children = this.props.children || noop
-    return typeof children === 'function' ? children(childProps)
-      : React.isValidElement(children) ? React.cloneElement(children, childProps)
-        : children
+    return React.isValidElement(children) ? React.cloneElement(children, childProps) : children(childProps)
```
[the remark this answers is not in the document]
bond95: React component can be function as well, but not every function is React component. Thus we need to check first React component, because React component can include some props.

package.json
gregsheremeta: +1.

@bond95 please extract this change into a separate PR.
cc @michalskrivanek

bond95:
@gregsheremeta @sjd78 created separate PR for Patternfly update. https://github.com/oVirt/ovirt-web-ui/pull/746

gregsheremeta:
>> New updates.

> Did you change anything other than rebasing and #746?

Indeed, please be more specific :)

src/ovirtapi/index.js
sjd78: inputs are internal types (or at least partials) while the Promise returned would be the Api type (or an error).

eventually it would be nice to have the Apis typed properly, although keeping it updated by hand is less then ideal

Issue #416, opened 2017-11-17
ScrollPositionHistory not working


jniederm:
When getting back from VM page to grid listing, scroll position is not restored.

At this occasion I'd like to suggest to use memory instead of session storage to store scroll offset since
* there is on need to preserve scroll position during whole browser session. It's ok if scroll is shifted to the top after page reload.
* doesn't play well with multiple tabs.

version: commit 9dccbfd

jniederm:
Relates to #417

mykaul:
@jniederm  - there's no severity or target milestone label on this.

mareklibra:
@jniederm , can you please check if this is already fixed?

jniederm:
it's still there in commit d1c3a43, current master

gregsheremeta:
@bond95 Bohdan, can you check this after you rebase the new main page? If it's still an issue, it sounds easy and I targeted to 1.4.2 (August 10)

bond95:
Yes, this is still issue.

gregsheremeta:
@bond95 ok, thanks for verifying. Does this sound like something you can fix in a day or two?

bond95:
@gregsheremeta yes, of course.

PR #696, opened 2018-08-02
Fix bug in ScrollPositionHistory.

bond95:
Now it save position right.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/416


gregsheremeta:
lgtm

PR #700, opened 2018-08-03
ux-redesign: VM Details > Utilization Card

sjd78:
Includes the commits from #699 (the first 6).


sjd78:
For a stopped VM:
screenshot-localhost-3000-2018 08 08-19-03-05

For a running VM:
screenshot-localhost-3000-2018 08 08-19-14-07

With a bit of history:
screenshot-localhost-3000-2018 08 08-21-27-15

sjd78:
FYI @lizsurette, @lwrigh 

Also, if the systems doesn't have any historical data available, does it make sense to
 1. not render anything and keep the space,
 2. reclaim the space,
 3. put a "no history available" message where the chart should be,
 4. something else?

bond95:
> Also, if the systems doesn't have any historical data available, does it make sense to
> 
> 1. not render anything and keep the space,
> 2. reclaim the space,
> 3. put a "no history available" message where the chart should be,
> 4. something else?

I think put 'no history available' instead of chart is more consistent variant.

lizsurette:
> I think put 'no history available' instead of chart is more consistent variant.

+1. The Angular version of the Sparkline PatternFly component supports this where there is no data available. It's a small play on the empty state:
https://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfSparklineChart

gregsheremeta:
>  Angular version of the Sparkline PatternFly component supports this where there is no data available. It's a small play on the empty state

I don't see a screenshot there

For now, let's do regular empty state and spin out a low-priority task to add something more if we need to (unless I'm missing something...)

sjd78:
Ok, at this point, the PR is set except for two caveats:

 - Disk information is disk based allocated/provisioned data (filesystem info from guest isn't available via REST yet).

 - If historic data isn't available, the sparkline chart is just whitespace.  I'll need to adopt the style from pf-angular and add a "No Data Available" empty state.

sjd78:
[rebased]

gregsheremeta:
@sjd78 please rebase on master

sjd78:
[rebased]

I have not yet addressed the sparkline chart empty state.

sjd78:
@lizsurette , @lwrigh - Added "no historic data" placeholder if the sparkline isn't shown.

Running VM with history:
screenshot from 2018-08-20 21-52-15


Running VM without history (and no network summary data):
screenshot from 2018-08-20 21-52-25


Stopped VM:
screenshot from 2018-08-20 21-52-49

sjd78:
@bond95, @gregsheremeta - nudge for a code review (and visual review if you feel so inclined)

lwrigh:
For the info tip, would hover text read along the lines of,"There is currently no historic data available right now so we are not able to show any historic visualizations at this time."?

sjd78:
> For the info tip, would hover text...

I didn't add any tooltip or hover text.  I double checked the angular implementation and they don't have a tooltip on empty state either.  I'm happy to add one if we think it could reduce confusion.

lizsurette:
@sjd78 @lwrigh Yeah I'm not sure how much extra info we'd really be adding in a tooltip, so I'm fine without it. It does look like the color needs to be updated for contrast sake. If it's helpful, both the icon and the text are #363636 in the angular implementation.

sjd78:
(rebased on master, builds from and includes #735 and #724)

@lizsurette, @lwrigh - Updated the colors for both "no data" and "no historic data" placeholders.  Icon is slightly lighter than the text.  Icon is 30% black, text is 65% black (rgba makes a lot more sense to me for gray colors).

screenshot from 2018-08-21 13-29-08

src/utils/round.js
```diff
@@ -1,10 +1,25 @@
...
 export function round (number, precision = 0) {
-  const factor = Math.pow(10, precision)
-  const temp = number * factor
-  const roundedTemp = Math.round(temp)
-  return roundedTemp / factor
+  let rounded
+
+  if (Array.isArray(number)) {
+    rounded = number.map(n => round(n, precision))
+  } else if (number === Object(number)) {
```
bond95: Nice solution. But it is a little bit hard to understand this condition, if you don't know what `Object` function do. So please use `typeof`, it would be more clear.

src/components/VmDetails/cards/UtilizationCard.js
```diff
@@ -20,160 +23,333 @@ import {
   SparklineChart,
 } from 'patternfly-react'
 
+const CpuCharts = ({ stats, isRunning }) => {
```
bond95: Could you, please, split this file. Because it has to much components. And also it would be nice if every card would have own directory.

src/utils/round.js
sjd78: `null === Object(null)` is false
`typeof null === 'object'` is true
but at the same time, `function a() {}; a === Object(a)` is also true

I'll tweak the checks

sjd78:
(rebased after merging #735 and #724)

src/components/VmDetails/cards/UtilizationCard.js
sjd78: > Could you, please, split this file. Because it has to much components.

The entire file with 7 component is <400 lines so it doesn't really both me.  There are other places where there is a collection of components in a single file or smaller components in 1 file with bigger components in another file.  So there isn't a consistent standard across the app.

> And also it would be nice if every card would have own directory.

That does make sense.  With your snapshots in a directory and the `UtilizationCard` in a directory (broken apart maybe), that would be self-consistent.  I'll split up `UtilizationCard` and if it still looks ok, I'll post an update.

sjd78:
@bond95 - I ended up pushing the split files.

src/components/VmDetails/cards/UtilizationCard/style.css
```diff
@@ -0,0 +1,66 @@
+.utilization-card .chart-card {
+  box-shadow: none;
+  margin-left: -20px;
+  margin-right: -20px;
+  margin-bottom: 0;
+
+  padding: 0 10px;
+
+  min-height: fit-content;
```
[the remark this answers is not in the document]
bond95: Oh, okay. As I see fist public release version support it, thus everything must be fine. 
https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content
https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history

src/components/VmDetails/cards/UtilizationCard/index.js
```diff
@@ -0,0 +1,49 @@
...
+import CpuCharts from './CpuCharts'
+import MemoryCharts from './MemoryCharts'
+import NetworkingCharts from './NetworkingCharts'
+import DiskCharts from './DiskCharts'
+import NoLiveData from './NoLiveData'
+
+/**
+ * VM dashboard style Utilization charts (CPU, Memory, Network, Disk)
+ */
+const UtilizationCard = ({ vm, onEditChange }) => {
```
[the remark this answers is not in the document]
sjd78: Good point. I'll remove the old boiler plate code.

src/components/VmDetails/cards/UtilizationCard/style.css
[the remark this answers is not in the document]
sjd78: https://developer.mozilla.org/en-US/docs/Web/CSS/min-height#Browser_compatibility

I'll look to see if it is even necessary.  Probably will just drop the rule, but I'll have to try some things out.

sjd78:
@bond95, please review the latest changes in 0a2bdd8.

src/components/VmDetails/cards/UtilizationCard/NetworkingCharts.js
```diff
@@ -82,7 +90,7 @@ const NetworkingCharts = ({ stats, isRunning }) => {
   )
 }
 NetworkingCharts.propTypes = {
-  stats: PropTypes.object.isRequired,
+  netStats: PropTypes.object.isRequired,
```
bond95: If there is can be a case when netStats is `null` or `undefined`, then `isRequired` need to be removed.

sjd78: There cannot be that case.  The chart components require at least the current set of stat data.  They can handle the history not being there, but they cannot handle no stats.  The base card handles if the stats have not been loaded or are missing after a load attempt.

(See **UtilizationCard/index.js**)

bond95: Then why do you check `netStats` on line 31?
`const haveNetworkStats = netStats && netStats['current.total']`

sjd78: Because I missed cleaning it up in the last commit.

The Networking statistics isn't always available/present in the api call, but there is always a placeholder for it in the internal `VmStatistics` object.  So the object the index component checks is good but the network component needs to see if the stats it needs are present.

Not doing this check for CPU and Memory since those stats have been present for a long time.  Network stats is new in 4.3.

bond95: So if I understand you right this part `netStats &&` is not needed.

PR #702, opened 2018-08-03
ux-redesign: VM Details > Overview Card

sjd78:
Includes the commits from #699 (the first 6).

This is for the card with no title, top row, far left.


sjd78:
On a stopped VM, normal edit flow:
ezgif com-video-to-gif

sjd78:
FYI @lizsurette , @lwrigh

lwrigh:
@sjd78 It's looking good to me.

lizsurette:
@sjd78 Looking great! A few nits:
1) Do we have control over the error message? I'm hoping we can clean it up a little bit and get rid of some of the code jargon. I think it would be nice if it read "Virtual Machine Name contains special characters. Only lower-case and upper-case letters, numbers, '-', '_', and '.' are allowed."
2) There should be more contrast between the color of the text used in the tag, in this case "Linux", and the background color for good readability. Either the text should switch over to white or the background color should be updated to #EDEDED.

sjd78:
> 1. Do we have control over the error message? I'm hoping we can clean it up a little bit and get rid of some of the code jargon. I think it would be nice if it read "Virtual Machine Name contains special characters. Only lower-case and upper-case letters, numbers, '-', '_', and '.' are allowed."

There is always _something_ that can be done with enough time and effort.

Right now there is no frontend validation that block the REST call.  We're just sending what was input and displaying any and all error messages that are returned.  Not ideal.  Worse yet, if >1 field is not valid, you only get the error on the first field instead of getting all errors.  (@gregsheremeta maybe that can be part of the REST gaps)

**TL;DR** - Front end field validation is currently the easiest way out.

> 2. There should be more contrast between the color of the text used in the tag, in this case "Linux", and the background color for good readability. Either the text should switch over to white or the background color should be updated to #EDEDED.

Cool.  I plan on using the same CSS class that Bohdan is using on his main page redesign so the colors, margins, font-size, etc all match.

bond95:
@sjd78 rebase, please.

gregsheremeta:
@sjd78 please rebase on master

src/components/VmDetails/cards/OverviewCard.js
```diff
@@ -1,47 +1,204 @@
...
+    return null
+  }
+
+  componentDidUpdate (prevProps, prevState) {
+    if (prevState.isEditing && !this.state.isEditing) {
+      this.props.onEditChange(false)
+    }
+  }
+
+  handleCardOnStartEdit = () => {
```
bond95: I would like to avoid declaring methods like this, because:
1. It is still experimental and don't part ECMA specification
2. It is not consistent with rest of code
3. It is at odds with class-based style of OOP

src/components/VmDetails/cards/OverviewCard.js
```diff
@@ -1,47 +1,204 @@
...
+    this.setState({ isEditing: true, isDirty: false, actionCorrelationId: null, correlatedMessages: null })
+    this.props.onEditChange(true)
+  }
+
+  handleCardOnCancel = () => {
```
bond95: Same.

src/components/VmDetails/cards/OverviewCard.js
```diff
@@ -1,47 +1,204 @@
...
+    this.setState({ isEditing: false, isDirty: false, actionCorrelationId: null, correlatedMessages: null })
+    this.props.onEditChange(false)
+  }
+
+  handleCardOnSave = () => {
```
bond95: Same.

src/components/VmDetails/cards/OverviewCard.js
```diff
@@ -1,47 +1,204 @@
...
+    this.setState({ actionCorrelationId })
+    this.props.saveChanges(vmUpdates, actionCorrelationId)
+
+    // saveChanges will add the result of the operation to the vm under the given actionCorrelationId
+    // so when the vm prop changes, the result can be checked and close the edit as needed
+
+    return false // componentDidUpdate will swap the BaseCard out of edit mode as appropriate
+  }
+
+  handleChange = (fieldName, { target: { value } }) => {
```
bond95: Same.

src/components/VmDetails/cards/OverviewCard.js
```diff
@@ -1,47 +1,204 @@
...
+  }
+
+  render () {
+    const { vm, icons, operatingSystems } = this.props
+    const { isEditing, correlatedMessages } = this.state
+
+    const icon = getVmIcon(icons, operatingSystems, vm)
+
+    const changesRequireReboot = false // vm.get('nextRunExists')
+    console.log('correlatedMessages', correlatedMessages && correlatedMessages.toJS())
```
bond95: Did you forget to remove this?

src/actions/vm.js
```diff
@@ -137,6 +140,31 @@ export function suspendVm ({ vmId }) {
...
 
+export function createVm ({ vm, transformInput = false, pushToDetailsOnSuccess = false }, { actionCorrelationId, ...actionPassThrough }) {
+  return {
+    type: CREATE_VM,
+    payload: {
+      vm,
+      transformInput,
+      pushToDetailsOnSuccess,
+    },
+    actionCorrelationId,
```
bond95: I think for this purpose better create something like `metaPayload` in action.

src/components/VmDetails/cards/OverviewCard.js
```diff
@@ -1,47 +1,204 @@
...
-import style from '../style.css'
+import VmIcon from '../../VmIcon'
+import VmStatusIcon from '../../VmStatusIcon'
+import style from './OverviewCard.css'
+
+function getParts (vm, ...parts) {
+  const subset = {}
+  for (const part of parts) {
+    subset[part] = vm.get(part)
+    if (subset[part].toJS && typeof subset[part].toJS === 'function') {
```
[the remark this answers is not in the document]
sjd78: Sometimes the `part` is a single value, sometimes a piece of Object data, sometimes an ImmutableJS construct.  The `typeof` check is needed to make sure the call doesn't error, no matter what it contains.  I prefer to keep the check at this point.

I planned on moving this to a higher level helper to reuse in at least the details card.  I'll move it now.

src/components/VmDetails/cards/OverviewCard.js
sjd78: > 1. It is still experimental and don't part ECMA specification

It's stage 3 and that's why we have babel.

> 2. It is not consistent with rest of code

True.  Let's tag this one to go into the yet-to-be-created style guide.  Perhaps there is an es-lint rule that we can turn on to prohibit this style.

I'll change the style to match the rest of the code base.

> 3. It is at odds with class-based style of OOP

I'd never expect to extend one of our react components, nor have I seen that in patternfly-react.  So this isn't really a relevant point.

sjd78:
@lizsurette, @lwrigh - Nit picking a detail here...  what about new lines in the VM's description?  The user can enter as much white space (well spaces and line breaks) as they want.

I can easily display the text a few different ways:
  - collapse all the white space (like now)
  - preserve the line breaks
  - preserve all the white spaces

Example:
ezgif com-video-to-gif 2

Thoughts?

src/actions/vm.js
sjd78: The `actionCorrelationId` is for the action itself, not for the payload.  When I read `metaPayload`, I read it as information about the payload itself.

AFAIK, there is no standard for redux action data beyond what we do in the app with putting the data to be worked on in the `payload` field.

src/components/VmDetails/cards/OverviewCard.js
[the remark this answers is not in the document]
bond95: I still don't get it. Why you use `typeof` condition, if you already check `toJS` function to existence. Can really be case when `toJS` is not a function?

src/actions/vm.js
bond95: Okay. It can be `metaAction`, the point is to group data to make them more understandable.

bond95:
@sjd78 I think better to check new lines and white spaces, if there gonna be more then one `break lines` then replace it by single `break line`, and with white spaces do the same.

@lizsurette @gregsheremeta @lwrigh What do you think?

lwrigh:
@sjd78 I think in a situation like that it would make sense to preserve the line breaks but I think preserving the white space could get a little crazy. 

@lizsurette Do you know if there's a PatternFly standard for line breaks or white space within text boxes? 

cc @bond95 @gregsheremeta

src/components/VmDetails/cards/OverviewCard.css
```diff
@@ -0,0 +1,34 @@
...
+
+.vm-status .vm-status-icon {
+  margin: 2px;
+}
+
+.vm-status .vm-status-text {
+  margin-left: 5px;
+}
+
+.vm-description {
```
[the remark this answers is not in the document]
sjd78: No need to remove, it is now in use!

sjd78:
(rebased) and addressed review comments other then how to add the correlation id to the save action

src/actions/vm.js
sjd78: I did some research and it is more common to keep the action types clean and use the form:
````javascript
{
   type: CONSTANT,
   payload: { someData: true },
   meta: { correlationId: 123 },
}
````

As you mentioned, it does make the intention more explicit.  Additionally, since the technique being used to trace back the success/fail of an action via a `correlationId` can be used in a number of other places as well.  I'm updating the code to follow the `meta` field style in the action type.

sjd78: Moved to #724

sjd78:
> Cool. I plan on using the same CSS class that Bohdan is using on his main page redesign so the colors, margins, font-size, etc all match.

@lizsurette - apart from exact placement rules, the OS tags are sharing styles :smile: 

----

@lizsurette, @lwrigh, @gregsheremeta, @bond95 - The description text is currently `white-space: pre-wrap` so all spaces and newlines are rendered and text is wrapped as necessary.  It is trivial to collapse multiple spaces or newlines to a single, I just need to know if that should be done.

----

Current screen caps:
screenshot from 2018-08-16 15-05-06
screenshot from 2018-08-16 15-05-16
screenshot from 2018-08-16 15-05-31

sjd78:

sjd78:
(rebased after merging #735 and #724)

PR #709, opened 2018-08-10
Fix the green for 'Up' on VmStatusIcon to be PF green

sjd78:
We had the wrong color green for the Running/Up status icon.

Still waiting to hear if the Off/Down status icon should be red (the specific pf-red) instead of black.

### DIFF
--- a/src/components/VmStatusIcon/style.css
+++ b/src/components/VmStatusIcon/style.css
@@ -1,3 +1,3 @@
 .green {
-    color: #24d011;
+    color: #3B9D2E;
 }


gregsheremeta:
lgtm

sjd78:
@lizsurette, @lwrigh - Thoughts on the red for a stopped VM?

lwrigh:
@sjd78 Red can be tricky sometimes since it can be associated with something going wrong or there being an error so I would err on the side of caution with using it. I would suggest just using a black icon to represent the VM being stopped instead. @lizsurette What do you think?

gregsheremeta:
>  I would suggest just using a black icon to represent the VM being stopped instead

+1

gregsheremeta:
in the spirit of smaller PRs, stopped icon changes can be another PR

Issue #708, opened 2018-08-10
InfiniteScroll is broken


bond95:
InfiniteScroll start to fetch data in begining and continue even if no data is returned.

gregsheremeta:
Was this introduced in main page updates, or has it always been a problem?

bond95:
Not exactly, in main page branch everything worked fine. But after merge with the latest master it broke.

bond95:
@gregsheremeta Also it can freeze whole application. Thus I think it is not `medium` task.

PR #715, opened 2018-08-13
Fixed InfiniteScroll infinite fetch.

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/708


sjd78:
What was the problem and why did the change fix it?

src/components/VmsList/style.css
```diff
@@ -61,10 +61,6 @@
     z-index: 20;
 }
 
-.full-window {
```
sjd78: Did this change impact the fix or is it just cleanup?

Is the class used in Vms.js?  Does it need to be anymore?  Perhaps remove it from there as well.

src/index-nomodules.css
```diff
@@ -43,18 +43,14 @@ body {
   flex-grow: 1;
   display: flex;
   flex-direction: column;
-
+  background: #f5f5f5;
```
sjd78: And that will force every page in the app to have a gray background, including the current Edit page.  That is why `.route-page-component` exists.

src/components/VmsList/index.js
```diff
@@ -51,7 +51,7 @@ const VmsList = ({ vms, requestActive }) => {
     el = <NoVmAvailable />
   }
 
-  return <div className='route-page-component'>{el}</div>
+  return el
```
sjd78: How does removing a wrapper div with a class fix the problem?

src/components/VmsList/style.css
bond95: It is just cleanup.

src/index-nomodules.css
bond95: Main background color of new design is gray, thus other cases are just exceptions, that can be fixes by adding additional style.
And thank you for your comment I forgot to add this style to Dialog page.

src/components/VmsList/index.js
bond95: Because infinite scroller use parent node for tracking scrolling. Thus with this div wrapper it doesn't work, because div has `overflow: visible` by default.

src/index-nomodules.css
bond95: Gray is main color of background in new design. Thus other cases is just exceptions, that can be fixed by adding additional styles.
Thank you for your comment, I forgot to add additional styles to Dialog page.

src/components/VmsList/index.js
sjd78: So the `div` that contains the `InfiniteScroll` component needs `overflow: auto`.  That makes sense.  I think we can do better so that a change to `#page-router-render-component` won't cause problems in the future.

bond95: Actually I can't imagine what problems can it cause. Because `#page-router-render-component` has only one goal it must be be scrollable.

src/components/DetailContainer.js
```diff
@@ -5,8 +5,10 @@ import sharedStyle from './sharedStyle.css'
 
 const DetailContainer = ({ children, ...props }) => {
   return (
-    <div className={`container-fluid ${sharedStyle['move-left-detail']} ${sharedStyle['detail-z-index']}`} {...props}>
-      {children}
+    <div className='detail-container'>
```
sjd78: Don't need a wrapper div, just add 'detail-container' to the containing div as the first style.

src/components/VmsList/index.js
sjd78: This change is ok.

src/index-nomodules.css
```diff
@@ -152,3 +148,7 @@ body {
     margin-bottom: 0;
     margin-top: 15px;
 }
+
+.detail-container {
+  background: #fff;
```
sjd78: +1   (although I would have just used `white` :wink:)

src/index-nomodules.css
sjd78: I'm ok with that.

src/components/VmsList/index.js
sjd78: > Actually I can't imagine what problems can it cause.

Which is why we take simple precautions now...especially since a simple change to add a wrapper div already caused so much havoc.

src/components/VmsList/style.css
gregsheremeta: our github standards are complete yet, but single-purpose commits are definitely in there :) so next time let's cut cleanups into separate commits and PRs

src/components/VmsList/index.js
gregsheremeta: Thanks! I spun this discussion into #716

src/components/DetailContainer.js
bond95: It will not work as expected, because `<div className={`container-fluid` has different non-zero margin.

sjd78: Ok

sjd78:
> Add a wrapper `<div style={{ overflow: 'auto' }}>...</div>` in **src/components/VmsList/Vms.js to make the Vms component resilient.**

#716

> Remove `route-page-component` class from **src/components/VmDetails/index.js** (line 53)

The removed `.full-window` style is still referred to in **src/components/VmsList/Vms.js** (line 55)

Those to be cleaned up sometime in the future.

gregsheremeta:
Thank you @sjd78 for adding a proper PR message

PR #717, opened 2018-08-14
Enable access to the pre-redesign VM detail page

sjd78:
- With incremental releases planned (after each agile sprint), not all
   of the previous functionality will be available

 - Enable navigation to the pre-redesign / legacy VM detail page from
   the current VM details page

 - This allows access to the old Console, Disks, Nics and Snapshots
   functionality

Solves #718

### DIFF
--- a/src/components/Pages/index.js
+++ b/src/components/Pages/index.js
@@ -6,6 +6,7 @@ import { RouterPropTypeShapes } from '../../propTypeShapes'
 import VmDialog from '../VmDialog'
 import VmsList from '../VmsList'
 import VmDetails from '../VmDetails'
+import { default as LegacyVmDetails } from '../VmDetail'
 
 import { selectVmDetail, selectPoolDetail, getIsoStorageDomains, getConsoleOptions } from '../../actions'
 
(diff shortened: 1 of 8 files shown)


src/components/Pages/index.js
```diff
@@ -6,6 +6,7 @@ import { RouterPropTypeShapes } from '../../propTypeShapes'
 import VmDialog from '../VmDialog'
 import VmsList from '../VmsList'
 import VmDetails from '../VmDetails'
+import { default as LegacyVmDetails } from '../VmDetail'
```
bond95: Here you can do more simpler: `import LegacyVmDetails from '../VmDetail'`

bond95:
@sjd78 Did you tested it? Because I got this errors:
image
image

sjd78:
@bond95 - I did test it.  I saw those errors, but they are in code that was untouched and the page still works so I didn't care.  They may not even be new errors (these come from the PropType check which is hidden in production code).

I really don't want to take any more time working on this.

sjd78:
@bond95 , @gregsheremeta - I just tested the details page from tags/1.4.0-3 and the same errors get reported on the dev server.

bond95:
@sjd78 the problem not in console errors by itself, but in that they cause blank page.

gregsheremeta:
> the problem not in console errors by itself, but in that they cause blank page.

for posterity, we determined this wasn't a real issue

PR #723, opened 2018-08-16
enhanced browser console logging

sjd78:
- may be used as a module import

 - captures `console.(log|info|warn|error)` and routes it through the same
   logging functions as the module import

 - allows disabling both `logger` and `console` logging via configuration

 - retains proper file and line number reporting

 - adds styles to visually indicate the log level on the browser console

 - retire helpers.js `logDebug` and `logError` in favor of logger.js

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,15 +1,16 @@
 // @flow
-/* eslint-disable import/first */
-
 /**
-  Flow agreement:
-  For simple types, like number, boolean, string and etc.: use lower-case,
-  For complex types, like Object, Array and etc.: use first letter in upper-case
-**/
+ Flow agreement:
+ For simple types, like number, boolean, string and etc.: use lower-case,
+ For complex types, like Object, Array and etc.: use first letter in upper-case
+ **/
+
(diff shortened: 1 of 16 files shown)


sjd78:

src/index.js
```diff
@@ -6,6 +6,9 @@
   For simple types, like number, boolean, string and etc.: use lower-case,
   For complex types, like Object, Array and etc.: use first letter in upper-case
 **/
+
+import logger from './logger'
```
bond95: Is this necessary have `logger` import on the top? Because for consistency first must be imported `React`.

sjd78: This is the app entry point file, so if we want to be sure to capture ALL potential console logging, the logger needs to initialize (starting the capture of `console.log` etc) before all other imports.

However, the risk of moving that import below the React imports is quite minimal.  A few `console` logs may get to the browser console (from the modules imported before the logger) before **logger.js** can setup the capture and be configured by **readConfiguration**.

I'll move it down to maintain "import React first" consistency.

PR #701, opened 2018-08-03
ux-redesign: VM Details > Details Card 

sjd78:
Includes the commits from #699 (the first 6).

PR #724, opened 2018-08-16
Elevate create and edit action/saga/reducers from VmDialog to core app

sjd78:
Work to support #702 and all of the rest of the cards that will need to create or update a VM

 - Moved the `createVm` and `editVm` actions/sagas to global actions/sagas

 - `createVm` saga has an option to push the user directly to the
   details page of the newly created VM

 - redux actions should now keep the action objects clean and only
   include `type`, `payload` and `meta` field (flow-type to follow to
   enforce this style)

 - `editVm` saga will update the effected VM with the success of the
   action if the action contains a `correlationId` in its `meta` field
   using the new `SET_VM_ACTION_RESULT` action

 - `addNewVm` and `editVm` ovirtapi functions can optionally transform
   the VM from internal to API format

 - Added flow-types to `src/actions/errors.js`'s `failedExternalAction`

 - Added action reducer `SET_VM_ACTION_RESULT` to **vms** reduce to
   store the result of VM actions that have a `correlationId` in its
   `meta` field. This allows a component to react to the completion of
   an action before changing render state.

 - `VmDialog` updated:
   - uses the new global actions/sagas
   - remove its reducer in favor of checking action correlated userMessages
   - Create and Edit modes work as appropriate


sjd78:
@bond95, @gregsheremeta - Overview #702 and Details #701 are both dependent on this functionality. Expediting review and merge would be appreciated.

sjd78:

PR #725, opened 2018-08-16
update TravisCI to run yarn in node 8 (drop node 6)

sjd78:
Current version of patternfly don't like node6, and since ovirt builds with node8, it doesn't make sense to retain node6 CI builds in travis any longer.

Also swapped from npm to yarn for the build to match what we're doing locally and in packaging.


sjd78:

PR #735, opened 2018-08-20
refine BaseCard event handlers and children render

sjd78:
Based on work in #691 and #700, refine the `BaseCard`:

  - better description of the component in the class comment

  - event handlers are defined functions bound to `this` in the
    constructor, matching the style in the rest of the app

  - children are rendered as a function, element or any other
    node type via `renderChildren`

  - update components in **GridComponent.js** so they don't blindly pass props to a `div` tag and cause React errors

### DIFF
--- a/src/components/VmDetails/BaseCard.js
+++ b/src/components/VmDetails/BaseCard.js
@@ -19,16 +19,28 @@ import CardEditButton from './CardEditButton'
 /**
  * Base VM details card.  Support common layouts and view vs edit modes.
  *
- * Specifying __editMode__ allows the containing component to control the card's
- * edit state.  Leave __editMode__ undefined to allow the card to control itself.
+ * Content of the component may vary based on the edit state of the card. Children
+ * may be a function of the form -
+ *   ({ isEditing }) => ...
  *
- * When the user interacts with the card, the card will fire event handler:
- *  - When in non-edit state, user clicks the __Edit__ icon -> onStartEdit() is called
- *  - When in edit state, user clicks the __Cancel__ button -> onCancel() is called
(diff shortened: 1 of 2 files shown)


src/components/VmDetails/BaseCard.js
```diff
@@ -90,7 +113,7 @@ class BaseCard extends React.Component {
             <CardEditButton tooltip={editTooltip} editEnabled={editing} onClick={this.clickEdit} />
           )}
 
-          {children({ isEditing: editing })}
+          <this.renderChildren isEditing={editing} />
```
bond95: This is not right calling react component. Please, assign it to capitalize variable before use.
https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized

src/components/VmDetails/BaseCard.js
```diff
@@ -118,7 +141,7 @@ BaseCard.propTypes = {
   onStartEdit: PropTypes.func,
   onCancel: PropTypes.func,
   onSave: PropTypes.func,
-  children: PropTypes.func,
+  children: PropTypes.oneOfType([ PropTypes.func, PropTypes.ndoe ]),
```
bond95: Not `ndoe`, but `node`. Actually I think it can be set as `isRequired`.

sjd78: :man_facepalming:  I'll fix both.

src/components/VmDetails/BaseCard.js
sjd78: Makes sense.  I'll make that change.

Issue #698, opened 2018-08-03
A VM's CD can be changed, but the currently loaded CD is never displayed


sjd78:
On entering the VM details or edit screen, the VM's full data is fetched.  To reduce the number of calls needed to get a complete picture of the VM, the _follow_ parameter is used.  Unfortunately, in the case of __cdroms__, an extra matrix parameter is required on its fetch (current=true) to get the correct information and it is not currently possible to specify the matrix parameter within the __folllow__ parameter.

The "Change CD" function on webadmin and the CD select box on web-ui both change the current configuration and not the saved configuration.

As a result, on web-ui, a VM's CD can be changed, but it currently never displays correctly.

Either a second fetch with the extra matrix param needs to be done, or the REST api needs an update to allow sending matrix params along with the __follow__ param.

![screenshot from 2018-08-03 08-42-19](https://user-images.githubusercontent.com/3985964/43643419-3aaca7a6-96f9-11e8-9cde-39ad5601c2f7.png)

gregsheremeta:
> As a result, on web-ui, a VM's CD can be changed, but it currently never displays correctly.

Seems like we need a place to hold intermediate states

sjd78:
> Either a second fetch with the extra matrix param needs to be done, or the REST api needs an update to allow sending matrix params along with the follow param.

#742 adds a second fetch for the cdrom info when the VM is "up"

sjd78:
@gregsheremeta ...
> Seems like we need a place to hold intermediate states

The cd-rom API has it and makes the distinction based on the `current` parameter.  Just need to adjust the code so when the VM is running, the `current` cdrom info is grabbed (and pushed to), and when the VM is not running, the `not current` cdrom info is grabbed (and pushed to).

With the latest updates to #742, that is how it'll work.

PR #742, opened 2018-08-24
refactor VM cd-rom handling to work on the current/running VM

sjd78:
Fixes: #698 
Needed for: #701

Change details:
 - When the VM is stopped, changing the CD will change the next-run ("current=false") cd-rom settings

 - When the VM is running, the next-run cd-rom will be initially loaded. Changing the CD will change the current ("current=true") cd-rom settings.  The CD will be replaced with the next-run cd-rom at VM start up.

 - A VM's cd-rom settings are only needed in redux for the VM detail pages and will be loaded as appropriate by `fetchSingleVm`

 - For a VM fetch, `fetchSingleVm` will pull both the "current=false" (via the additional/follow params) and "current=true" (via second query if the VM is running) cd-rom info

 - The list of iso files available from each storage domain is loaded at login time

 - Modified the internal cd-rom object to make it flatter

 - Adjusted legacy view and edit components to use the new internal structure for cd-rom

 - Added actions/constants for `getVmCdRom` and `changeVmCdRom`

 - Added saga `changeVmCdRom` so the CD can be changed without editing the entire VM

 - Fixed up the camelCase on the functions


src/ovirtapi/index.js
```diff
@@ -303,17 +306,16 @@ const OvirtApi = {
     return httpGet({ url: `${AppConfiguration.applicationContext}/api/storagedomains/${storageId}/files` })
   },
 
-  getCDRom ({ vmId, running }: { vmId: string, running: boolean }): Promise<Object> {
+  getCDRom ({ vmId, current = true }: { vmId: string, current?: boolean }): Promise<ApiCdRomType> {
```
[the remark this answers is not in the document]
sjd78: The API still supports it, so it could be used, it just isn't right now.  If someone in the future does want to be able to fetch/change the offline CD, they could.  I don't see any harm in leaving the parameter available but defaulted.

src/ovirtapi/index.js
```diff
@@ -303,17 +306,16 @@ const OvirtApi = {
...
     assertLogin({ methodName: 'getCDRom' })
-    return httpGet({ url: `${AppConfiguration.applicationContext}/api/vms/${vmId}/cdroms/${zeroUUID}?current=${running ? 'true' : 'false'}` })
+    return httpGet({ url: `${AppConfiguration.applicationContext}/api/vms/${vmId}/cdroms/${zeroUUID}?current=${current ? 'true' : 'false'}` })
   },
-
-  changeCD ({ cdrom, vmId, running }: { cdrom: Object, vmId: string, running: boolean }): Promise<Object> {
+  changeCD ({ cdrom, vmId, current = true }: { cdrom: CdRomType, vmId: string, current?: boolean }): Promise<ApiCdRomType> {
```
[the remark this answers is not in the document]
sjd78: I'll fix the saga.

src/sagas.js
```diff
@@ -424,8 +441,8 @@ function* fetchVmsSessions ({ vms }) {
   })
 }
 
-function* fetchVmCDRom ({ vmId, running }) {
-  const cdrom = yield callExternalAction('getCDRom', Api.getCDRom, { type: 'GET_VM_CDROM', payload: { vmId, running } })
+function* fetchVmCDRom ({ vmId }) {
```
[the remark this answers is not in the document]
sjd78: Just an inconsequential oversight  (yay destructuring doesn't break when you send in more object keys then it's looking for).  I'll clean it up.

src/sagas.js
```diff
@@ -435,8 +452,10 @@ function* fetchVmCDRom ({ vmId, running }) {
 
 function* fetchVmsCDRom ({ vms }) {
   yield * foreach(vms, function* (vm) {
-    const cdromInternal = yield fetchVmCDRom({ vmId: vm.id, running: vm.status === 'up' })
-    yield put(setVmCDRom({ vmId: vm.id, cdrom: cdromInternal }))
+    if (vm.status === 'up') {
```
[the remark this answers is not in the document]
sjd78: Humm.  This actually made me rethink how the cdrom changes are handled.  I went through and tweaked things a bunch.  I'll re-document things and post a change this morning.

src/ovirtapi/index.js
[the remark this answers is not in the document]
sjd78: My rethink makes use of the current option.  :-)

src/ovirtapi/transform.js
```diff
@@ -374,16 +374,15 @@ const CdRom = {
   toInternal ({ cdrom }: { cdrom: ApiCdRomType }): CdRomType {
     return {
       id: cdrom.id,
-      file: {
-        id: cdrom.file ? cdrom.file.id : '',
-      },
+      fileId: cdrom && cdrom.file && cdrom.file.id,
```
bond95: If cdrom can be `null` or `undefined`, then, please, add same check to `id` and set type as `{ cdrom?: ApiCdRomType }`

src/ovirtapi/transform.js
```diff
@@ -374,16 +374,15 @@ const CdRom = {
...
     }
   },
 
   toApi ({ cdrom }: { cdrom: CdRomType }): ApiCdRomType {
     return {
+      id: cdrom.id,
       file: {
-        id: cdrom.file.id,
+        id: (cdrom && cdrom.fileId) || '', // no fileId == Eject == ''
```
bond95: Likewise.

src/sagas.js
```diff
@@ -265,7 +274,7 @@ function* fetchVmsByPageVLower (action) {
     if (!shallowFetch) {
       yield fetchConsoleMetadatas({ vms: internalVms })
       yield fetchDisks({ vms: internalVms })
-      yield fetchVmsCDRom({ vms: internalVms })
+      yield fetchVmsCdRom({ vms: internalVms, current: false })
```
bond95: As I see `fetchVmsCdRom` has parameters `{ vms }`, thus remove, please, `current`.

src/sagas.js
```diff
@@ -323,7 +335,7 @@ function* fetchVmsByCountVLower (action) {
...
```
bond95: Likewise.

src/sagas.js
```diff
@@ -424,19 +443,29 @@ function* fetchVmsSessions ({ vms }) {
...
-function* fetchVmsCDRom ({ vms }) {
+/*
+ * For each given VM, first fetch the appropriate cdrom info based on the status of the VM,
+ * then push the info to VM reducers for the redux store.
+ */
+function* fetchVmsCdRom ({ vms }) {
   yield * foreach(vms, function* (vm) {
-    const cdromInternal = yield fetchVmCDRom({ vmId: vm.id, running: vm.status === 'up' })
-    yield put(setVmCDRom({ vmId: vm.id, cdrom: cdromInternal }))
+    const cdrom = yield fetchVmCdRom({ vmId: vm.id, isVmUp: vm.status === 'up' })
```
bond95: `fetchVmCdRom` has parameter `current`, but not `isVmUp`

src/sagas.js
sjd78: `current: false` for all VMs to make sure that the statically configured ISO is loaded as a base.  

`current: true` is then used when a VM's full details are loaded when entering a details page.

src/sagas.js
sjd78: good catch, thanks

src/sagas.js
sjd78: `fetchVmsCdRom` was wrong, `current` should be passed here.

`fetchVmsByPageVLower` and `fetchVmsByCountVLower` both need to fetch the `current: false` cd-rom info to match what the >=4.2 fetches pull.

src/sagas.js
sjd78: This is correct, `current: false` should be passed.  `fetchVmsCdRom` was wrong.

src/ovirtapi/transform.js
sjd78: Updated the check assuming `cdrom` is passed (as per the flow type definition)

src/components/VmDialog/index.js
```diff
@@ -277,7 +270,7 @@ class VmDialog extends React.Component {
   }
 
   onChangeCD (fileId) {
-    this.setState({ cdrom: { file: { id: fileId } }, isChanged: true })
+    this.setState({ cdrom: { fileId: fileId }, isChanged: true })
```
bond95: It would be better to write like this `{ fileId }`

gregsheremeta:
doesn't need QE / will get QA'd during the Details card QA

PR #748, opened 2018-08-28
added global react exception handler

kozloffsky:


bond95:
@kozloffsky Please, provide more info about PR.

gregsheremeta:
> provide more info about PR
+1

Also:
 - I don't see anything about ErrorBoundary -- why? We need a fallback UI to tell the user there was an error
 - what happens when there is an API call that returns bad data?
 - what happens when a saga returns an error?
 - what happens when there is UI corruption and React goes to unmount the app? (... needs ErrorBoundary)
 - what about de-obfuscating in production mode? Printing a stack to console in production is useless

bond95:
@kozloffsky @gregsheremeta I tried to throw some error within this PR in `login` sagas and it is not what I expected (maybe because I don't know what must be expected result). I got not white screen of death, but just empty main page:
image

I think there must at least some component with text "Something goes wrong, please, reload page".

kozloffsky:
Please check now. Here is correct code. Due to ErrorBoundary is just react name for class that has componentDidCatch method I`ve put this method into App component so it will handle all child components exceptions that are not handled by try catch blocks. Also it has only white screen with info about error due to I have no UI guides for this screen. Error also pushed into console and if source maps configured correctly then deobfuscation should be ok.

gregsheremeta:
> Due to ErrorBoundary is just react name for class that has componentDidCatch method I`ve put this method into App component so it will handle all child components exceptions that are not handled by try catch blocks

ok, makes sense. Probably worth a code comment

> Also it has only white screen with info about error due to I have no UI guides for this screen

Please proactively reach out to our UX people (Laura). I've also mentioned you can put a simple message like "Sorry, there was an error, try to refresh etc etc". But replacing a white screen with another white screen doesn't make sense.

> Error also pushed into console and if source maps configured correctly then deobfuscation should be ok.

Did you test it? (I don't think source maps are configured correctly. Can you check?)

sjd78:
@kozloffsky - You can update old commits until the PR is ready for review and a review is posted.

Squashing the commits you have so far would de-clutter history.

After a review, we're trying to push commits that implement changes from the review.  That allows the old reviews to stick with the old commits and prevent that "I coldn't find what you're looking for because you force pushed" kind of error.

If you need to force push `git push -f origin ...` to avoid merges in the PR that is also preferred.

Really the only time that we should NEED to rebase (as opposed to merge from master) is if merge conflicts come up.  If they do, resolving with a rebase and a force push is preferred (and more inline with what gerrit does).

src/App.js
```diff
@@ -37,28 +37,60 @@ const NoLogin = () => {
  * Main App component. Wrap the main react-router components together with
  * the various dialogs and error messages that may be needed.
  */
-const App = ({ history, config, appReady }) => {
-  if (!config.get('loginToken')) { // login is missing
+/* const App = ({ history, config, appReady }) => {
+
+} */
```
sjd78: git remembers what is was before, no need to keep that comment

src/App.js
```diff
@@ -37,28 +37,60 @@ const NoLogin = () => {
...
+      )
+    }
+
+    if (hasError) {
+      return (
+        <div>
+          <h1>Error</h1>
+          <p>{error.message}</p>
+          <pre>{error.stack}</pre>
+        </div>
```
sjd78: Pull this `div` out to another component like `NoLogin`.  Then it should be easier to quickly find the place to make any changes to the presentation and those changes won't touch the `App` component.

It'll need a bunch of (static component) polish in the near future.

sjd78:
> - what happens when there is an API call that returns bad data?

API calls are wrapped in the saga util function `callExternalAction` and tends to just dump the error message to toast and notification tray then cause the calling saga to fail.

> - what happens when a saga returns an error?

Redux basically ignores it just like any other reducer that fails.  I really don't know what best practice is for monitoring reducer fails etc.  If a reducer fails, redux store isn't updated.  If the store isn't updated, nothing in the UI should change.  So...silent fail?

> - what happens when there is UI corruption and React goes to unmount the app? (... needs ErrorBoundary)

Putting some extra comments on the `App` component referencing that it doubles as an `ErrorBoundary`, or creating a separate component that wraps `App` (maybe in **src/index.js**'s `renderApp`) would make it more pronounced.

Either way, whatever component is used to render the error state needs to be bullet proof.

> - what about de-obfuscating in production mode? Printing a stack to console in production is useless

+1  ...  I'd like to see what the stack trace looks like by default (install ovirt-web-ui to your dev engine and access it from there), and what it looks like when source maps are enabled (they might already be, but I'm not 100% sure on that).

kozloffsky:
@sjd78 @gregsheremeta 

> what about de-obfuscating in production mode? Printing a stack to console in production is useless
I`ve check this one, we have `devtool: 'source-map'` so source maps are generated in production already.

src/App.js
kozloffsky: done

src/App.js
kozloffsky: done

src/App.js
```diff
@@ -33,32 +33,69 @@ const NoLogin = () => {
...
+  }
+
+  render () {
+    const { history, config, appReady } = this.props
+    const { hasError, error } = this.state
+
+    if (!config.get('loginToken')) { // login is missing
+      return (
+        <Grid fluid>
+          <NoLogin error={error} />
```
sjd78: I don't think the `error` prop is needed (or used) here.

src/App.js
```diff
@@ -33,32 +33,69 @@ const NoLogin = () => {
...
+        </Grid>
+      )
+    }
+
+    if (hasError) {
+      return (
+        <Grid fluid>
+          <Error />
```
sjd78: You need the `error` prop here.

What about the `errorInfo` from `componentDidCatch`?  That should probably be given to the error boundary report component as well.

src/App.js
```diff
@@ -33,32 +33,69 @@ const NoLogin = () => {
   )
 }
 
+const Error = (error) => {
```
sjd78: I'd prefer a different component name.  Maybe simply "ErrorBoundaryReport" 

This should be `const Error = ({ error }) => {` to conform to the normal structure of a function component.

You'll also need to add 
````js
Error.propTypes = {
  error: PropTypes.object,
}
````
below the component function to keep eslint happy.

**Comment by sjd78** [comment]
Looking at https://reactjs.org/docs/error-boundaries.html is fairly interesting.

May be cleaner to create a single `GlobalErrorBoundary` component (in src/GlobalErrorBoundary.js) mirroring the example component on the react docs, and make it the root element for the entire app.  So instead of:
https://github.com/oVirt/ovirt-web-ui/blob/a49fc4f3043cd18842ec953621565b0e5e5ae5d0/src/index.js#L48-L57

It could look like:
````js
function renderApp (store: Object) {
  ReactDOM.render(
    <GlobalErrorBoundary>
      <Provider store={store}>
        <IntlProvider locale={locale} messages={getSelectedMessages()}>
          <App history={store.history} />
        </IntlProvider>
      </Provider>
    </GlobalErrorBoundary>,
    (document.getElementById('root'): any)
  )
}
````

That would allow it to capture render errors everywhere.

@kozloffsky, @bond95, @gregsheremeta - Thoughts?

**Comment by kozloffsky** [comment]
@sjd78 I`ve implement GlobalErrorBoundary ant it work as expected so all error handling code was removed from App.js

**Comment by bond95** [comment]
> > what happens when a saga returns an error?

> Redux basically ignores it just like any other reducer that fails. I really don't know what best practice is for monitoring reducer fails etc. If a reducer fails, redux store isn't updated. If the store isn't updated, nothing in the UI should change. So...silent fail?

Actually this is more complicated then it looks like.

I did some tests. I threw errors in different places in sagas. Like login saga, fetchByPage and startVm. And after every throw sagas dropped completely (redux still work). So in this case VMPortal must show the same component as in case with React fail.

Also we must handle fails in reducers (errors doesn't drop redux), but in some cases (because of error) functionality can be limited, or even cause more complex errors.

File: src/GlobalErrorBoundary.js [hunk 6]
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import * as branding from './branding'
+
+class ErrorMessages {
+  static TITLE = 'We are currently having some issues with VM portal.'
+  static DESCRIPTION = 'Please refresh page or log out and log back in. If the issue persists, please report a bug on GitHub: '
```
sjd78: These messages can go in **src/intl/messages.js**.

Then you can access it like:
````js
import { msg } from './intl'

const title = msg.globalErrorBoundaryTitle()
const desc = msg.globalErrorBoundaryDescription()
````

File: src/App.js [hunk 7]
```diff
@@ -37,28 +37,34 @@ const NoLogin = () => {
  * Main App component. Wrap the main react-router components together with
  * the various dialogs and error messages that may be needed.
  */
-const App = ({ history, config, appReady }) => {
-  if (!config.get('loginToken')) { // login is missing
+
+class App extends React.Component {
+  render () {
+    const { history, config, appReady } = this.props
```
sjd78: No real `render()` change to `App` right?  Maybe just drop these changes since `App` doesn't need to be a React.Component class.

(I'd use `git gui` to ammend the last commit and unstage the change to the file)

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+class GlobalErrorBoundary extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = { hasError: false }
+  }
+
+  componentDidCatch (error, info) {
```
sjd78: Have you looked at `info` and what is available in dev and prod compiles?

It is supposed to be "component stack" and could be used to locate the exact component in the App where the error came from.  That may be useful.

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,54 @@
...
+    this.setState({ hasError: true })
+    console.error(error)
+  }
+
+  render () {
+    if (this.state.hasError) {
+      // You can render any custom fallback UI
```
gregsheremeta: you can delete this comment -- it's from the example :)

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,54 @@
...
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
+              </a>
+            </div>
+          </nav>
+          <div className='container text-center' style={containerStyle}>
+            <img src={branding.resourcesUrls.errorImg} />
+            <h1 className='bolder'>{ErrorMessages.TITLE}</h1>
+            <p className='h4'>{ErrorMessages.DESCRIPTION}<a href='https://github.com/oVirt/ovirt-web-ui/issues'>https://github.com/oVirt/ovirt-web-ui/issues</a></p>
+            <div>
+              <a href='/' className='btn' style={{ margin: '5%' }}>Refresh</a>
```
gregsheremeta: better to use stylesheets and `className` instead of `style`

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,54 @@
...
+              <a className='btn-primary btn' style={{ margin: '5%' }}>Logout</a>
```
gregsheremeta: same

package.json
```diff
@@ -133,7 +134,8 @@
       "<rootDir>/(build|docs|node_modules|__tests__/helper|config)/"
     ],
     "testEnvironment": "jsdom",
-    "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(js|jsx)$"
+    "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(js|jsx)$",
+    "testURL": "http://localhost/"
```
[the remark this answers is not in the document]
kozloffsky: this one fix test in localhost env.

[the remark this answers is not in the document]
kozloffsky: @gregsheremeta I`ve got this issue https://github.com/jsdom/jsdom/issues/2304 and solve it with this line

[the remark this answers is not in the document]
sjd78: @kozloffsky  Where did you get the issue?  Did the unit tests fail to run?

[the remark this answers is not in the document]
gregsheremeta: > Where did you get the issue?
+1, I don't get such issue, nor does travis

gregsheremeta:
@kozloffsky please reply to the review comments as you resolve them. (In particular, the comment "// You can render any custom fallback UI" is still in there")

src/intl/messages.js
```diff
@@ -304,6 +304,8 @@ export const messages = {
     message: 'Are you sure you want to drop your changes?',
     description: 'Message in the modal dialog opened when a user tried to navigate off an editor page after changes have been made.',
   },
+  globalErrorBoundaryTitle: 'We are currently having some issues with VM portal.',
```
gregsheremeta: change to: "Sorry, VM Portal is currently having some issues."

src/intl/messages.js
```diff
@@ -304,6 +304,8 @@ export const messages = {
...
+  globalErrorBoundaryDescription: 'Please refresh page or log out and log back in. If the issue persists, please report a bug on GitHub: ',
```
gregsheremeta: change to: "Please refresh the page or log out and log back in. If the issue persists, please report a bug on [link to issue tracker] github."  (missing "the" and "GitHub" should be the link)

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,51 @@
...
+            <div className='navbar-header'>
+              <a href='/' className='navbar-brand obrand_headerLogoLink' id={`pageheader-logo`}>
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
+              </a>
+            </div>
+          </nav>
+          <div className='container text-center obrand_errorContainer'>
+            <img src={branding.resourcesUrls.errorImg} />
+            <h1 className='bolder'>{title}</h1>
+            <p className='h4'>{descr}<a href='https://github.com/oVirt/ovirt-web-ui/issues'>https://github.com/oVirt/ovirt-web-ui/issues</a></p>
```
gregsheremeta: change the link to:
```<a href='https://github.com/oVirt/ovirt-web-ui/issues'>GitHub</a>```

File: branding/brand.css [hunk 16]
```diff
@@ -132,3 +132,12 @@ over) by other installed brands.
     bottom: 25px;
     right: 25px;
 }
+
```
gregsheremeta: this file is actually reserved for things that are copied from the admin portal branding file. (We can't put things in here because they'll be overwritten eventually.)

you should be able to create a new style.css in /src and import it:
import style from './style.css'

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,51 @@
...
+      return (
+        <div>
+          <nav className='navbar obrand_mastheadBackground obrand_topBorder navbar-pf-vertical'>
+            <div className='navbar-header'>
+              <a href='/' className='navbar-brand obrand_headerLogoLink' id={`pageheader-logo`}>
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
+              </a>
+            </div>
+          </nav>
+          <div className='container text-center obrand_errorContainer'>
```
gregsheremeta: once you have
`import style from './style.css'`
you can then do
```className={`${style['errorContainer']}```

(don't use the `obrand` prefix, since those are reserved for things that change between brands)

File: src/GlobalErrorBoundary.js [hunk 18]
```diff
@@ -0,0 +1,53 @@
...
+    this.setState({ hasError: true })
+    console.error(error)
+    console.error(info.componentStack)
+  }
+
+  render () {
+    const title = msg.globalErrorBoundaryTitle()
+    const descr = msg.globalErrorBoundaryDescription()
+    if (this.state.hasError) {
+      // You can render any custom fallback UI
```
bond95: Is this comment needed?

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+      return (
+        <div>
+          <nav className='navbar obrand_mastheadBackground obrand_topBorder navbar-pf-vertical'>
+            <div className='navbar-header'>
+              <a href='/' className='navbar-brand obrand_headerLogoLink' id={`pageheader-logo`}>
```
bond95: Use, please, `'pageheader-logo'`, instead of ```{`pageheader-logo`}```

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
+              </a>
+            </div>
+          </nav>
+          <div className='container text-center globalErrorContainer' >
```
bond95: For consistent in the end of tag must be no spaces, thus it must look like this:
`<div className='container text-center globalErrorContainer'>`

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+            <img src={branding.resourcesUrls.errorImg} />
+            <h1 className='bolder'>{title}</h1>
+            <p className='h4'>{descr}<a href='https://github.com/oVirt/ovirt-web-ui/issues'>GitHub</a></p>
+            <div>
+              <a href='/' className='btn'>Refresh</a>
```
bond95: Use, please, `msg` localization.

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,53 @@
...
+              <a className='btn-primary btn'>Logout</a>
```
bond95: Likewise.

src/GlobalErrorBoundary.js
bond95: > It is supposed to be "component stack" and could be used to locate the exact component in the App where the error came from. That may be useful.
+1 Agree, it would be nice to return not full stack, but error, and name of component that cause error.

gregsheremeta:
@kozloffsky There are a lot of unresolved code review comments above. Also the PR body is still missing a description.

...

In addition, the error page should render when there is 1. a saga failure, 2. bad API data, 3. react component corruption [1].

Please post some test code (can use a branch or gists) that demonstrate how the error handling works in all 3 of these cases.

[1] https://medium.com/@Tam_mo/react-16-error-boundaries-117262e2d7d4

src/index.js
```diff
@@ -121,19 +125,33 @@ function initializeApiListener (store: Object) {
   })
 }
 
+class ErrorThrower {
+  errorHandler = null
+  setErrorHandler (errorHandler: Function) {
+    this.errorHandler = errorHandler
+  }
+  throw (error: Object) {
```
bond95: This is not very good idea to use key words as name of methods or variables.

src/App.js
```diff
@@ -46,6 +46,11 @@ const App = ({ history, config, appReady }) => {
     )
   }
 
+  const errorHandler = (error) => {
```
bond95: Is there any reason to define this function insinde of component?

src/index.js
```diff
@@ -46,13 +47,16 @@ window.patternfly = require('patternfly/dist/js/patternfly')
...
   ReactDOM.render(
-    <Provider store={store}>
-      <IntlProvider locale={locale} messages={getSelectedMessages()}>
-        <App history={store.history} />
-      </IntlProvider>
-    </Provider>,
+    <GlobalErrorBoundary>
+      <Provider store={store}>
+        <IntlProvider locale={locale} messages={getSelectedMessages()}>
+          <App history={store.history} errorThrower={errorThrower} />
```
sjd78: I see that you are bridging the saga error handler through to the react error stack with this.  I'm not 100% sold on the idea as it may end up trapping errors that may not be real errors -- I haven't read up on error handling in saga and redux in general.

But ... the react way is to create a new component for this instead of adding another purpose to an existing component.

So .... rename `errorThrower` to `sagaErrorBridge` and then ...
````jsx
<GlobalErrorBoundary>
  <React.Fragment>
    <RethrowSagaErrorAsReactError errorBridge={sagaErrorBridge} />
    <Provider store={store}>
    ......
    </Provider>
  </React.Fragment>
<GlobalErrorBoundary>
````

The `Fragment` may not be necessary based on how `GlobalErrorBoundary` handles its children.

`App` should remain untouched by any of the error handling code.

Create `RethrowSagaErrorAsReactError` as a component that has the bits you currently have in `App`.

`ErrorThrower` is too generic of a name.  Name is something specific because its job will be to:
  - collect components that will bridge errors into react (not forgetting to also release the component when the component is unmounted)
  - listen for error from saga
  - when an error occurs in saga, broadcast it out to all of its collect components

**Comment by gregsheremeta** [comment]
> In addition, the error page should render when there is 1. a saga failure, 2. bad API data, 3. react component corruption [1].

> Please post some test code (can use a branch or gists) that demonstrate how the error handling works in all 3 of these cases.

Case 1, saga failure. I hacked in an error in `updateVms`, which is called from `function* fetchVmsByPageV42 (action)`

nx1qhq_vm js vm js - sourcegear diffmerge_036

works as expected

selection_027

^ the first entry in the stack gives us a line number. Clicking that gives us the source mapped `updateVms`.

selection_028


Case 2, bad API data. I hacked the VM data to be missing everything but name, id, and description for the first VM returned.

r2etml_sagas js sagas js - sourcegear diffmerge_037

works as expected

selection_024

again, clicking the line number gets us the source mapped function, where it's clear to see that `vms.icons` is indeed missing.

selection_026


Case 3, corrupt component. I hacked in a corrupt component.
selection_030

works as expected

chzumw_index js index js - sourcegear diffmerge_038

clicking gets us the source mapped code

selection_031

**Comment by gregsheremeta** [comment]
@kozloffsky the only visual fix we need is to increase the min-height of the header. Please add another commit with this:

r8rech_index-nomodules css index-nomodules css - sourcegear diffmerge_034

You'll also need to fix up the messages file, as it currently conflicts because we sorted it. Like so:

t7cn6c_messages js messages js - sourcegear diffmerge_035

File: package.json (hunk 12 above)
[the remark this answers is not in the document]
Comment (gregsheremeta) [reply package.json]: seems like a valid issue and doesn't harm anything, +1

File: src/GlobalErrorBoundary.js [hunk 26]
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import * as branding from './branding'
+import { msg } from './intl'
+import './error_styles-nomodules.css'
```
sjd78: When importing a css, there is no need to use the `nomodules` tag.  The form to use when importing a CSS in the tag code is like this:
````jsx
import style from './error.css'

<div className={style['globalErrorContainer']}> ... </div>

<div className={`other-classes ${style['my-error-css-style']}`}> ... </div>
````

File: src/GlobalErrorBoundary.js [hunk 27]
```diff
@@ -0,0 +1,63 @@
...
+      return (
+        <div>
+          <nav className='navbar obrand_mastheadBackground obrand_topBorder navbar-pf-vertical'>
+            <div className='navbar-header'>
+              <a href='/' className='navbar-brand obrand_headerLogoLink' id='pageheader-logo'>
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
+              </a>
+            </div>
+          </nav>
+          <div className='container text-center globalErrorContainer'>
```
sjd78: Use the `style['globalErrorContainer']` form as I mentioned above.  This will maintain consistency with the rest of the code and prevent CSS class name collisions.

src/error_styles-nomodules.css
```diff
@@ -0,0 +1,8 @@
+.globalErrorContainer {
+    width: 40%;
+    margin-top: 10%;
+}
+
+.globalErrorContainer a.btn {
```
sjd78: If you remove the '-nomodules.css' extension, and if `a.btn` is to reference a bootstrap/patternfly button, then you would need to rewrite the css as:
````css
.globalErrorContainer a:global(.btn) {
 ...
}
````

`:global()` is a directive to the CSS module loader to leave that classname untouched.

File: src/index.js [hunk 29]
```diff
@@ -121,19 +125,34 @@ function initializeApiListener (store: Object) {
...
+
 function onResourcesLoaded () {
   console.log(`Current configuration: ${JSON.stringify(AppConfiguration)}`)
 
   addBrandedResources()
 
   const store = configureStore()
-  store.runSaga(rootSaga)
+  const rootTask = store.runSaga(rootSaga)
+  const sagaErrorBrigde = new SagaErrorBridge()
```
sjd78: s/sagaErrorBrigde/sagaErrorBridge/

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,63 @@
...
+import * as branding from './branding'
+import { msg } from './intl'
+import './error_styles-nomodules.css'
+import AppConfiguration from './config'
+
+class GlobalErrorBoundary extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = { hasError: false }
+    this.props.errorBridge.setErrorHandler((error) => {
```
sjd78: That is nice and clean!

Only suggestion is to add a blank line and a comment above just explaining that the component is registering itself as a generic error handler.

I know the component is only used once in the App so it isn't necessary, but in any other component, you'd want to register in `componentDidMount` and unregister `componentWillUnmount` to avoid any memory leaks.  Otherwise the error bridge could hang on to every instance of the component (and its children).  That would be bad.

sjd78:
@kozloffsky, you need to rebase to resolve the conflict on `src/intl/messages.js`.  As @gregsheremeta mentioned, we sorted the messages file so we need to maintain the sort.

src/error_styles-nomodules.css
kozloffsky: thanks

kozloffsky:
@bond95 when saga thows error redux stops all actions and no dispatching is going. i`ve tried this way of sending.

src/GlobalErrorBoundary.js
```diff
@@ -0,0 +1,63 @@
...
+            <div className='navbar-header'>
+              <a href='/' className='navbar-brand obrand_headerLogoLink' id='pageheader-logo'>
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
+              </a>
+            </div>
+          </nav>
+          <div className={`container text-center ${styles['globalErrorContainer']}`}>
+            <img src={branding.resourcesUrls.errorImg} />
+            <h1 className='bolder'>{title}</h1>
+            <p className='h4'>{descr}<a href='https://github.com/oVirt/ovirt-web-ui/issues'>{gitHub}</a></p>
```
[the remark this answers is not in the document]
gregsheremeta: @sjd78  (for my own learning) why call `dangerouslySetInnerHTML` when you don't have to / it has "dangerous" in the name implying don't call it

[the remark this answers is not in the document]
sjd78: @gregsheremeta - You need to use `dangerouslySetInnerHTML` if your string variable has markup that you would like rendered as markup.  Any other way to include the string and React auto escapes it.

For example:
````jsx
const s = 'This <span style="color: red">red text</span> is red.'

return <React.Fragment>
  <div>{s}</div>
  <div dangerouslySetInnerHTML={{ __html: s }} />
</React.Fragment>
````
Will render like:
````html
<div>This &lt;span style="color: red"&gt;red text&lt;/span&gt; is red.</div>
<div>This <span style="color: red">red text</span> is red.</div>
````
React Docs: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

File: src/App.js [hunk 32]
```diff
@@ -35,7 +35,7 @@ const NoLogin = () => {
 
 /**
  * Main App component. Wrap the main react-router components together with
- * the various dialogs and error messages that may be needed.
+ * the various dialogs and error messages that may be neded.
```
bond95: `neded`, maybe `needed`?

sjd78:
@kozloffsky - Are you going to fix the typo/spelling error and remove App.js from the PR?

src/error_styles-nomodules.css
```diff
@@ -0,0 +1,8 @@
+.globalErrorContainer {
```
bond95: Is there any reason to not remove this file?

bond95:
@kozloffsky Also, rebase, please.

src/App.js
gregsheremeta: done

kozloffsky:
> @kozloffsky please reply to the review comments as you resolve them. (In particular, the comment "// You can render any custom fallback UI" is still in there")

it is already resolved

ovirt-infra:

2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically.

gregsheremeta:
waiting on QA

gregsheremeta:
ci add to whitelist

gregsheremeta:
closing in favor of squahed #820

PR #756, opened 2018-09-05
ux-redesign: VM Details > Nics Card (view only)

sjd78:
View only portion of #757

API changes:
  - Add API &rarr; internal transforms to capture more info about NICs
  - Add better api typing, mostly Nic based

Nic Card changes:
  - each Nic is listed on the card
  - the IPv4 and IPv6 addresses are shown below each Nic

Note: There may be a render issue when there are >4 Nics. The nic list may overflow the card. Not quite sure why yet.

### DIFF
--- a/src/components/VmDetails/cards/NicsCard.js
+++ b/src/components/VmDetails/cards/NicsCard.js
@@ -1,13 +1,35 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
 
 import BaseCard from '../BaseCard'
-import style from '../style.css'
+import { Grid, Row, Col } from '../GridComponents'
+
+import style from './NicsCard.css'
 
 /**
  * List of NICs connected to a VM
  */
-const NicsCard = ({ vm, onEditChange }) => {
+const NicsCard = ({ vm, vnicProfiles, onEditChange }) => {
+  const nicList = vm.get('nics')
+    .sort((a, b) => a.get('name').localeCompare(b.get('name')))
+    .map(nic => ({
+      id: nic.get('id'),
+      name: nic.get('name'),
+      plugged: nic.get('plugged'),
+      ipv4: nic.get('ipv4').toJS(),
+      ipv6: nic.get('ipv6').toJS(),
+      vnicProfile: {
+        id: nic.getIn(['vnicProfile', 'id']),
+        name: vnicProfiles.getIn([nic.getIn(['vnicProfile', 'id']), 'name']),
+        network: vnicProfiles.getIn([nic.getIn(['vnicProfile', 'id']), 'network', 'name']),
+      },
+    }))
+    .toJS()
+
+  const ip4Label = 'IPv4'
+  const ip6Label = 'IPv6'
+
   return (
     <BaseCard
       icon={{ type: 'pf', name: 'network' }}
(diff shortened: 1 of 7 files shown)


bond95:
@sjd78 could you please add screenshots?

sjd78:
> could you please add screenshots?

When the PR docs are done, the issue is created, and it's ready to take a look at, I'll let you know.

bond95:
@sjd78 great, thank you.

lizsurette:
@sjd78 This looks awesome! What do you think about adding a small label to the IP and MAC addresses? Maybe just as simple as: 
IP:x.x.x.x        MAC:aaaa:aaaa:aaaa:aaaa:aaaa

sjd78:
@lizsurette ...IPv6 does _kinda_ look like a MAC address... :grin:

I put the IPv4 and the IPv6 addresses there.  We're ignoring MAC addresses in web-ui.

I could put labels in so it looks like:
````
    IPv4: 123.123.123.1232    IPv6: fe80::21a:4aff:fe16:123
````
I don't have any strong feelings either way.

**Comment by lizsurette** [comment]
> IPv6 does kinda look like a MAC address... 😁

LOL. Yep, whoops!

> I could put labels in so it looks like:
> 
>     IPv4: 123.123.123.1232    IPv6: fe80::21a:4aff:fe16:123

I think that would look nice and just give a bit more detail. Although I'm sure the folks using these VMs will be smarter than I am :)

**Comment by sjd78** [comment]
@lizsurette -- with labels:
screenshot-localhost-3000-2018 09 07-07-45-11

**Comment by sjd78** [comment]
@gregsheremeta, @bond95 - nudge for review
@lizsurette - how does the card look with the labels?

File: src/components/VmDetails/cards/NicsCard.js [hunk 1]
```diff
@@ -18,25 +40,49 @@ const NicsCard = ({ vm, onEditChange }) => {
...
-        )
-      }}
+      {({ isEditing }) =>
+        <Grid className={style['nics-container']}>
+          {nicList.map(nic =>
+            <Row key={nic.id}>
+              <Col style={{ display: 'block' }}>
+                <div>
+                  <span>{nic.name}</span>
+                  <span>&nbsp;({nic.vnicProfile.name}/{nic.vnicProfile.network})</span>
```
bond95: Is this really necessary use `&nbsp;` instead of `padding-left` for example?

src/ovirtapi/index.js
```diff
@@ -164,7 +165,7 @@ const OvirtApi = {
     return httpGet({ url })
   },
 
-  addNewVm ({ vm, transformInput = false }: VmType): Promise<Object> {
+  addNewVm ({ vm, transformInput = false }: { vm: Object, transformInput?: boolean }): Promise<Object> {
```
bond95: I think here must `ApiVmType` instead of `Object`.

src/ovirtapi/index.js
```diff
@@ -174,7 +175,7 @@ const OvirtApi = {
       input,
     })
   },
-  editVm ({ vm, transformInput = false }: VmType): Promise<Object> {
+  editVm ({ vm, transformInput = false }: { vm: Object, transformInput?: boolean }): Promise<Object> {
```
bond95: Likewise

src/ovirtapi/index.js
sjd78: I think this got out-of-sync with the same change on the Details PR #701.

At this point, and in general, I think the inputs should by the internal type or `Object`.  There are plenty of times a rest action only needs a partial object.  The returns we should leave as `Object` for now as the return typing needs a bit more thought.

All that being said, I'll update the function's type to:
````js
addNewVm ({ vm, transformInput = false }: { vm: VmType | Object, transformInput: boolean }): Promise<Object> {
````

File: src/ovirtapi/transform.js [hunk 4]
```diff
@@ -472,9 +472,21 @@ const Cluster = {
 //
 const Nic = {
   toInternal ({ nic }: { nic: ApiNicType }): NicType {
+    const ips = (
+      nic.reported_devices &&
+      nic.reported_devices.reported_device &&
```
bond95: In this case, better use ternary operator `?:`

sjd78: Ok.  Using a ternary looks like it would be more consistent to the rest of the file.

PR #760, opened 2018-09-06
intl: sort the messages alpha by key, adjust flow typing

sjd78:
Initial sort of the message keys in **src/intl/messages.js** to make adding keys a bit more straightforward.


src/intl/index.js
```diff
@@ -16,7 +16,7 @@ import csLocalData from 'react-intl/locale-data/cs'
 import huLocalData from 'react-intl/locale-data/hu'
 import heLocalData from 'react-intl/locale-data/he'
 
-import { messages, type MessageIdType } from './messages'
+import { messages, type MessageIdType, type MessageType } from './messages'
```
gregsheremeta: what does this do? (doesn't seem related to sorting)

sjd78: It's the "adjust flow typing" bit of the PR.

I made a slight adjustment to the flow typing shared between **intl/index.js** and **intl/messages.js**. Now both types used by the messages are in **intl/messages.js**.  Also the `messages` object now has a type of `{ [messageId: string]: MessageType }`.  Flow type checking should fail if a key is added that doesn't match the `MessageType` pattern.

gregsheremeta: thanks!

PR #762, opened 2018-09-11
fix multiple types import in ovirtapi/index.js

gregsheremeta:
fix multiple types import in ovirtapi/index.js

to fix the travis build https://travis-ci.org/oVirt/ovirt-web-ui/builds/427162428

### DIFF
--- a/src/ovirtapi/index.js
+++ b/src/ovirtapi/index.js
@@ -1,6 +1,9 @@
 // @flow
 import type {
-  CdRomType, ApiCdRomType,
+  CdRomType,
+  ApiCdRomType,
+  ApiSnapshotType,
+  SnapshotType,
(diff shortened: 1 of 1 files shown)


src/ovirtapi/index.js
```diff
@@ -1,6 +1,9 @@
 // @flow
 import type {
-  CdRomType, ApiCdRomType,
+  CdRomType,
```
bond95: Can you, please, pair types, like Scott had it, I think it would be easier to understand?
```
import type {
  CdRomType, ApiCdRomType,
  SnapshotType, ApiSnapshotType,
} from './types'
```

gregsheremeta: done

PR #774, opened 2018-09-14
ux-redesign: Made snapshot improvements in new detail dashboard.

bond95:
List of improvements:

 - When snapshot is deleting, buttons are disabled
 - When restoring, buttons are disabled for all snapshots
 - While creating new snapshot, button `Create snapshot` is disabled 
 - Changed `description` to `name`
 - Removed cloud-init and replace it by NICs
 - Fixed bug with missing disks
 - Fixed bug with locked icon when creating snapshot
 - Added close button to popover 
 - When click enter in new snapshot modal it create new snapshot

Restoring:
image

Deleting:
image

Popover:
image


src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
```diff
@@ -22,6 +22,19 @@ import { PendingTaskTypes } from '../../../../reducers/pendingTasks'
 
 const MAX_DESCRIPTION_SIZE = 50
 
+const SnapshotAction = ({ children, className, disabled, onClick }) => {
+  return <a onClick={!disabled ? onClick : null} className={`${className} ${disabled && 'disabled'}`}>
```
sjd78: You used `null` here for the onClick but `undefined` in another file.  It would be nice to stay consistent.  Could still do `{!disabled && onClick}` and let it default to undefined by itself.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
```diff
@@ -34,31 +47,62 @@ StatusTooltip.propTypes = {
...
+      <a>
+        <Icon type='pf' name='info' />
+      </a>
     </OverlayTrigger>)
-    if (isEditing && !snapshot.get('isActive') && !beingDeleted) {
-      buttons.push(<RestoreConfirmationModal key='restore' snapshot={snapshot} vmId={vmId} />)
-      buttons.push(<DeleteConfirmationModal key='delete' trigger={<a key='delete'><Icon type='pf' name='delete' /></a>} onDelete={onSnapshotDelete}>
+
+    // Delete and restore actions
+    if (!snapshot.get('isActive')) {
```
sjd78: Looks like this if condition is repeating the if condition above.  If so, this `isActive` check isn't necessary.

src/components/VmDetails/cards/SnapshotsCard/index.js
```diff
@@ -1,34 +1,46 @@
...
+  beingCreated: PropTypes.bool,
+  beingRestored: PropTypes.bool,
 }
 
+const ConnectedSnapshots = connect(
+  (state) => ({
+    beingCreated: !!state.pendingTasks.find(t => t.type === PendingTaskTypes.SNAPSHOT_ADD),
+    beingRestored: !!state.pendingTasks.find(task => task.type === PendingTaskTypes.SNAPSHOT_RESTORE),
+  }),
+  (dispatch) => ({})
```
sjd78: You can leave the `dispatch` section out completely.  `connect` can be called with a single argument.

src/components/VmDetails/cards/SnapshotsCard/sagas.js
```diff
@@ -38,9 +55,11 @@ function* deleteVmSnapshot (action) {
 }
 
 function* restoreVmSnapshot (action) {
+  yield put(addSnapshotRestorePendingTask())
   yield startProgress({ vmId: action.payload.vmId, name: 'restoreSnapshot' })
   const result = yield callExternalAction('restoreSnapshot', Api.restoreSnapshot, action)
   yield stopProgress({ vmId: action.payload.vmId, name: 'restoreSnapshot', result })
+  yield put(removeSnapshotRestorePendingTask())
```
sjd78: Is restore sync so once the call returns it is done?  If not, the `delaySec` dance would need to be done here as well...

src/sagas.js
```diff
@@ -809,7 +813,20 @@ export function* fetchVmSnapshots ({ vmId }) {
   let snapshotsInternal = []
 
   if (snapshots && snapshots['snapshot']) {
-    snapshotsInternal = snapshots.snapshot.map(snapshot => Api.snapshotToInternal({ snapshot }))
+    snapshotsInternal = yield all(snapshots.snapshot.map(snapshot =>
+      call(function* () {
```
sjd78: I suggest putting a `NOTE` comment here saying something like "Snapshot Disks and Nics are not currently available via additional/follow param on the snapshot fetch so we need to fetch them directly".

I also recommend filing a BZ RFE to add that capability, and an issue to track it.  There are a bunch of places I've seen like that where it would be great to not have to make additional calls.  Might as well start logging RFE gaps now right?  Or are these the types of things Mike is going to fix?  Either way, a BZ would help track the change in engine project...

cc: @gregsheremeta

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
sjd78: Er, nope.  Still onClick still needs a function.  Will use `noop`.

src/sagas.js
sjd78: I teased it apart a bit to a few different functions so it is easier to read.  Applied the same work to `fetchSingleVm`.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
sjd78: I think the check was moved up to `isActionsDisabled` and this `if` isn't needed anymore.  The icons will show but be disabled.

gregsheremeta:
thanks! followed up and thus replaced by #783

PR #784, opened 2018-09-24
Add permission checks/filtering for Create VM, Edit VM and Cluster Lists

### LINKED ISSUE #737: Clusters should be filtered by user permission on creating and editing VM pages
Clusters must be filtered by user permission on creating and editing VM pages. Also if user has no permissions for creating VM on Cluster, then hide `Create Virtual Machine` button. For creating VM user must have `VmCreator` or `PowerUserRole` permissions for cluster.

Related issues: https://github.com/oVirt/ovirt-web-ui/issues/258 https://github.com/oVirt/ovirt-web-ui/issues/286 https://github.com/oVirt/ovirt-web-ui/issues/507

### LINKED ISSUE #745: "Create VM" button should not be available when user is granted only with UserRole
Description of problem:
"Create VM" should not be available when user is granted only with UserRole

Version-Release number of selected component (if applicable):
4.2.3

How reproducible:
always

Steps to Reproduce:
1. create user and grant him UserRole permission to system or just some vm
2. after login with this user - "create vm" button is visible and when you attempt to create vm you are not able to close the vm creation window and you have to go to rhv manager home page and navgate from there

Actual results:
"create vm" button is visible

Expected results:
"create vm" button is not visible

sjd78:
Disable Create VM and Edit VM by checking editing permissions on clusters.

Fixes #737 (and therefore [BZ1599617](https://bugzilla.redhat.com/show_bug.cgi?id=1599617))
Fixes #745 (and therefore [BZ1597705](https://bugzilla.redhat.com/show_bug.cgi?id=1597705))
Replaces #754 (rebase to fix merge conflicts initially)
Includes the commit from #786 (arbitrary userMessages)

Summary:
  - If user has no clusters with permissions that allow creating a VM, then Create Virtual Machine button is hidden
  - If user without permissions to create a new VM tries to get directly to new VM page via URL, redirect them back to main page.
  - If user cannot edit a VM, hide the Edit buttons (for the legacy edit page) and hide the edit pencils on the VM Details page.

Follow-up changes:
  - permission helpers are now in **src/utils/permissions.js**

  - `checkPermissions` uses constants instead of using Selectors for each iteration of the checks

  - moved `canUserUseAnyClusters` from **Pages/index.js** to **permissions.js**

  - **Pages/index.js** `VmCreatePage`:
    - Added a permission check to `render` so in theory the page won't flash content if a user with insufficient permission tries to open it directly
    - Added a toast/tray notification to explain why the user landed on the VM list page instead of 'Create VM'

  - **Pages/index.js** `VmEditPage`:
    - cluster permission check added like in `VmCreatePage`
    - added a direct check on the VM to ensure the user can edit the requested VM
    - both checks have toast/tray notification

  - New toast/tray notifications have new intl messages


sjd78:
(rebased to resolve merge conflicts with master)

src/components/Pages/index.js
```diff
@@ -269,23 +298,25 @@ VmEditPage.propTypes = {
   vms: PropTypes.object.isRequired,
   previousPath: PropTypes.string.isRequired,
   match: RouterPropTypeShapes.match.isRequired,
-  isFilteredClusters: PropTypes.bool, // deep immutable, {[id: string]: Cluster}
+  isAnyUsableCluster: PropTypes.bool, // deep immutable, {[id: string]: Cluster}
```
bond95: Could you, please, remove comment, because this is not `deep immutable`? =)

src/components/Pages/index.js
```diff
@@ -192,37 +200,43 @@ class VmCreatePage extends React.Component {
     props.getAvailableCDImages()
   }
 
-  componentDidUpdate () {
-    if (!this.props.isFilteredClusters && this.props.clusterSize) {
+  render () {
+    if (!this.props.isAnyUsableCluster && this.props.clusterSize) {
```
bond95: Is there any reason to move it from `componentDidUpdate`? Bacause `render` method must be pure, and dispatching actions this is not very pure solution. 
https://reactjs.org/docs/react-component.html#render

src/components/Pages/index.js
```diff
@@ -253,11 +261,32 @@ class VmEditPage extends React.Component {
...
+      previousPath,
+      isAnyUsableCluster,
+      clusterSize,
+      redirectToMainPage,
+      addUserMessage,
+    } = this.props
     const { vmId } = this.state
 
+    // Verify the user will be able to edit any VMs (not necessarily the one that is requested)
+    if (!isAnyUsableCluster && clusterSize) {
```
bond95: Likewise.

src/components/Pages/index.js
sjd78: I did it specifically this way since `componentDidUpdate` does not run during the first render cycle and the page was flashing before kicking the user back to the main page.  Humm...I'm not 100% sure which way to go to keep `render` pure.

What do you think about having `render` do the test and just returning null and then putting the dispatch back in `componentDidUpdate`?

sjd78: I'm updating following that direction...

bond95: > What do you think about having render do the test and just returning null and then putting the dispatch back in componentDidUpdate?

Yes, it sounds fine.

PR #786, opened 2018-09-25
Allow adding an arbitrary user message to toast and tray

sjd78:
Extends the `userMessages` framework so any arbitrary toast/tray message can be added for display.

Previously only an API error or a failed login could generate a message.

This will be useful to push toast notification when async events start (e.g. restarting a VM) or for other short term messages (e.g. trying to jump to a page you don't have permissions for).


sjd78:
No screenshots since the toats and tray notification are unchanged.

#784 uses this to send toast notification when a user gets booted to the main list pages when they try to directly access VM create or VM edit.

PR #791, opened 2018-09-27
Combine cluster permissions and VM Details, Details card PRs

sjd78:
Base on #784 (cluster and VM permissions), add #701 (details card) and apply the VM edit permissions and cluster 'canUse' filtering to the details card itself.

Only the most recent 3 commits are unique to this PR.


src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -736,7 +741,7 @@ DetailsCard.propTypes = {
 }
 
 const DetailsCardConnected = connect(
-  (state) => ({
+  (state, { vm }) => ({
```
bond95: Why do you have `{ vm }` here?

sjd78: It's a hold over from merging.  This is what it looked like in just the permissions #784:
https://github.com/oVirt/ovirt-web-ui/blob/b05b015d1c1cd1bfb6ab3df7b03742f73fdc1cd1/src/components/VmDetails/cards/DetailsCard.js#L42-L47

No need for it now, I'll take it out.

PR #799, opened 2018-10-03
implement the ExpandCollapseSection patternfly pattern in react

sjd78:
To support the intended design of Nic Edit and Disk Edit modals, an advanced options section is needed.  Patternfly has the pattern as 'Expand/Collapse Section' [1], but there is not currently a patternfly-react implementation.  This commit adds a react implementation of that pattern.

[1] - https://www.patternfly.org/pattern-library/forms-and-controls/expand-collapse-section/


sjd78:
Example of the section closed:
closed


Example of the section opened:
opened

src/components/ExpandCollapseSection.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+ExpandCollapseSection.propTypes = {
+  id: PropTypes.string.isRequired,
+  opened: PropTypes.bool,
+  sectionHeader: PropTypes.string,
+  children: PropTypes.node.isRequired,
+}
+ExpandCollapseSection.defaultProps = {
+  opened: false,
```
bond95: As I see, default value for prop `opened` is not needed, because:
1. By default bool props a `false`
2.  On line 24 `props.opened || false` explicitly set `false` if `props.opened` is `false`

src/components/ExpandCollapseSection.js
```diff
@@ -0,0 +1,65 @@
...
+  sectionHeader: 'Advanced Options',
```
bond95: Translate, please.

src/components/ExpandCollapseSection.js
```diff
@@ -0,0 +1,65 @@
...
+
+    this.state = {
+      opened: props.opened || false,
+    }
+
+    this.handleToggleExpand = this.handleToggleExpand.bind(this)
+  }
+
+  handleToggleExpand (e) {
+    if (e && e.preventDefault) e.preventDefault()
```
bond95: And as I already asked in previous PR, is this `if (e && e.preventDefault)` check really need?

src/components/ExpandCollapseSection.js
sjd78: Yes, with `defaultProps` setting opened to false, the `|| false` is not needed.

src/components/ExpandCollapseSection.js
sjd78: No.  

This follows the style with other patternfly-react components where text is defaulted to English and it is up to the consumer of the component to provide the translated text.

src/components/ExpandCollapseSection.js
sjd78: Here the check isn't needed, but the `e.preventDefault()` is.

Issue #811, opened 2018-10-09
Filter clusters according to permits, not roles


bond95:
User can create own roles, thus need to filter clusters according to permits, not roles, because permits are static and can't be changed or added new one by user.

sjd78:
**Note**: The merged solution, PR #784, to issues #745 / [BZ1597705](https://bugzilla.redhat.com/show_bug.cgi?id=1597705) and #737 / [BZ1599617](https://bugzilla.redhat.com/show_bug.cgi?id=1599617) relied on Permissions and not on Permits.  This issue changes the base unit of authorization checking from Permissions to Permits themselves.

The two BZs will need to be re-tested once this issue is resolved by PR #813.  I'll update them to reflect the current situation.

sjd78:
Reopening since the merge of #836, which depends on #813, should have waited for #813.  My mistake.

PR #813, opened 2018-10-10
Change cluster filtering by permits instead of roles

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/811

### DIFF
--- a/src/utils/permissions.js
+++ b/src/utils/permissions.js
@@ -2,44 +2,16 @@
 import Selectors from '../selectors'
 import type { ClusterType, PermissionType } from '../ovirtapi/types'
 
-function checkPermissions (allowedPermissions: Array<string>, permissions: Array<PermissionType>): boolean {
-  const userFilter = Selectors.getFilter()
-  const userGroups = Selectors.getUserGroups()
-  const userId = Selectors.getUserId()
-
-  return permissions.find((role) => (
-    (userFilter ||
-      (
-        (role.groupId && userGroups.includes(role.groupId)) ||
-        (role.userId && role.userId === userId)
-      )
-    ) &&
-    allowedPermissions.includes(role.name)
-  )) !== undefined
+function checkUserPermit (permit: string, permits: Set<string>): boolean {
+  return permits.has(permit)
 }
 
-export function canUserUseCluster (permissions: Array<PermissionType>): boolean {
-  const allowedPermissions = [
-    'VmCreator',
-    'PowerUserRole',
-    'ClusterAdmin',
-    'DataCenterAdmin',
-    'UserVmManager',
-    'UserVmRunTimeManager',
-    'SuperUser',
-  ]
-  return checkPermissions(allowedPermissions, permissions)
+export function canUserUseCluster (permits: Set<string>): boolean {
+  return checkUserPermit('create_vm', permits)
 }
 
-export function canUserEditVm (permissions: Array<PermissionType>): boolean {
-  const allowedPermissions = [
-    'ClusterAdmin',
-    'DataCenterAdmin',
-    'UserVmManager',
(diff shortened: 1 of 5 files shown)


src/utils/permissions.js
```diff
@@ -2,7 +2,7 @@
 import Selectors from '../selectors'
 import type { ClusterType, PermissionType } from '../ovirtapi/types'
 
-function checkPermissions (allowedPermissions: Array<string>, permissions: Array<PermissionType>): boolean {
+function checkPermissions (permission: string, permissions: Array<PermissionType>): boolean {
```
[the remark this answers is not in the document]
bond95: No, because purpose of `checkPermissions` is to check that user have permissions to do something. Previous it was array, because roles can contain same permits, but now we don't need to check bunch of permissions, we can check permit directly.

src/utils/permissions.js
```diff
@@ -14,32 +14,16 @@ function checkPermissions (allowedPermissions: Array<string>, permissions: Array
         (role.userId && role.userId === userId)
       )
     ) &&
-    allowedPermissions.includes(role.name)
+    role.permits.find((permit) => permit.name === permission) !== undefined
```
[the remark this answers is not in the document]
bond95: True.

src/utils/permissions.js
```diff
@@ -33,3 +21,20 @@ export function canUserEditVm (permissions: Array<PermissionType>): boolean {
...
+  const userGroups = Selectors.getUserGroups()
+  const userId = Selectors.getUserId()
+  let permits = []
+  permissions.forEach((role) => {
+    if (userFilter ||
+      (
+        (role.groupId && userGroups.includes(role.groupId)) ||
+        (role.userId && role.userId === userId)
+      )) {
+      permits = [...permits, ...role.permits.map((permit) => permit.name)]
```
sjd78: Nit-pick: This will allow duplicate permits in the `permits` array if the user has multiple user/group/role mappings.  You could use a [ES6 Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) and then convert it to an array upon return.

bond95: Good point.

bond95:
@sjd78 Done. Please, review.

bond95:
@sjd78 @gregsheremeta Added external permissions fetch as temporary solution, because of this bug https://bugzilla.redhat.com/show_bug.cgi?id=1639784

src/utils/permissions.js
bond95: In this case I prefer to use Array instead of Set, because Set is too slow. I know that Set not so much slow, but still better to think about performance now, then when it will be to late.

gregsheremeta:
ci build please

sjd78:
ci build please

gregsheremeta:
[ovirt-web-ui-1.5.0-0.20181026git9b9bd99.el7.noarch.rpm](https://jenkins.ovirt.org/job/oVirt_ovirt-web-ui_standard-check-pr/150/artifact/build-artifacts.el7.x86_64/ovirt-web-ui-1.5.0-0.20181026git9b9bd99.el7.noarch.rpm)

[ovirt-web-ui-1.5.0-0.20181026git9b9bd99.fc28.noarch.rpm](https://jenkins.ovirt.org/job/oVirt_ovirt-web-ui_standard-check-pr/150/artifact/build-artifacts.fc28.x86_64/ovirt-web-ui-1.5.0-0.20181026git9b9bd99.fc28.noarch.rpm)

bond95:
ci build please

bond95:
ci build please

bond95:
@leistnerova rebased.

leistnerova:
LGTM

Issue #450, opened 2017-12-20
Make saga error handling more robust


mareklibra:
If any saga fails for whatever reason (including an exception), the application must stay in working state.

To verify, e.g. within a saga, use code like
```
   const dummy = undefined
   dummy.failHere()
```

Such a flow must be properly logged but the rootSaga must continue to live and process next actions.

gregsheremeta:
@sjd78 is this still an issue?

bond95:
@gregsheremeta GlobalErrorBoundary already implemented, can we close this issue?

sjd78:
@bond95, @gregsheremeta - The error boundary is a good kernel panic trap, and it prevents the app to run with inconsistent state and prevent the app from being unloaded.  However, it isn't quite what the issue here is asking for.  I'm ok with closing this issue with the caveat that the App just fails to an error state, and I'm ok with leaving it open so we can circle back to provide a more self-healing approach.

gregsheremeta:
> closing this issue with the caveat that the App just fails to an error state

Indeed.

Issue #449, opened 2017-12-20
improve handling of unexpected API responses


mareklibra:
If a pool has no `vm` assigned, the `poolToInternal()` method fails on `undefined`.

Data retrieval must be more robust, no matter how corrupted data are retrieved from server.

Subsequent flows in React must be revised to verify even for the case of no assigned VM to the pool.

mareklibra:
Since the data from engine seems to be corrupted due to nonstandard flow of ovirt installation, this issue can wait till 1.4

michalskrivanek:
@mareklibra, what do you mean by "data seems to be corrupted"? that's not a reassuring sentence to defer with;-)

mareklibra:
ah :-) the `pool` resource is expected to be associated with a `vm`. 

The original reporter of this bug was facing multiple issues during his oVirt setup and subsequent test data load. Anyway, the environment became up and running. When connected to VM portal, the data provided by API was inconsistent, like no `vm` within a `pool`.

Severity is low in this case, but it's good to improve handling of unexpected API response.

mareklibra:
I think this should go along with https://github.com/oVirt/ovirt-web-ui/issues/450

mareklibra:
In general, the `vmToInternal` and similar translation functions should be made more robust to
- expect certain structure from API (as they do)
- do not assume the structure is really returned (as in any other distributed system, it's good to assume whatever crap as a response)

bond95:
@gregsheremeta GlobalErrorBoundary already implemented, can we close this issue?

gregsheremeta:
closing  with the caveat that the app just fails to an error state now

PR #820, opened 2018-10-11
implement GlobalErrorBoundary with fallback UI

### LINKED ISSUE #729: implement ErrorBoundary to prevent white screen crash
implement ErrorBoundary to prevent white screen crash
https://reactjs.org/docs/error-boundaries.html
https://hackernoon.com/error-boundaries-in-react-16-32fb8e185a3?gi=8b779e0eeac5

see also #449 and #450

When we have an exception in a saga, API call, or anywhere else, and the UI gets unmounted, we should never show a plain white screen with no feedback. At the very least there should be a "fallback UI" that indicates something went wrong and the user should try refreshing the page.

gregsheremeta:
implement GlobalErrorBoundary with fallback UI to prevent white screen crash

https://reactjs.org/docs/error-boundaries.html
https://hackernoon.com/error-boundaries-in-react-16-32fb8e185a3?gi=8b779e0eeac5

see also #449 and #450

When we have an exception in a saga, API call, or anywhere else, and the UI gets unmounted, we should never show a plain white screen with no feedback. At the very least there should be a "fallback UI" that indicates something went wrong and the user should try refreshing the page.

Fixes #729


gregsheremeta:
verified:
selection_045

gregsheremeta:
(see conversation at ##748)

Issue #843, opened 2018-10-30
Aliases for the most usable files and folders


bond95:
I don't like when need to use a lot of '../', after 2 or 3 it is very hard to use it. So I suggest to use aliases for the most usable files and folders. Like for `utils.js` or `actions` or `helpers.js`.

Example:
```JavaScript
import { escapeHtml } from '../../../utils'
```

PR #845, opened 2018-11-01
Added aliases for the most used files, to avoid long paths.

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/843

- Main utils - 'app-utils'
- Intl - 'app-intl'
- Reducers folder - 'app-reducers'
- Actions - 'app-actions'
- Helpers in root - 'app-helpers'
- Main constants - 'app-constants'


config/webpack.config.dev.js
```diff
@@ -74,6 +74,12 @@ module.exports = {
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
       'react-native': 'react-native-web',
       'ovirtapi': `${paths.appSrc}/ovirtapi/index.js`,
+      'utils': `${paths.appSrc}/utils/index.js`,
```
sjd78: I suggest moving our aliases to `webpack.config.commons.js` as a function and spreading it here.  That way we can have the alias defs in 1 files instead of 2.  And we can drop the `react-native` alias, I don't think we'll be using that anytime soon.

````js
  alias: {
    ...commonAliases()
  }
````

and in `webpack.config.commons.js`:
````js
exports.commonAliases = function () {
  return [
    'app-intl': `${paths.appSrc}/intl/index.js`,
    'reducers': `${paths.appSrc}/reducers`,
    'actions': `${paths.appSrc}/actions/index.js`,
    'helpers': `${paths.appSrc}/helpers.js`,
    'app-constants': `${paths.appSrc}/constants/index.js`
  ]
}
````

File: config/webpack.config.prod.js [hunk 2]
```diff
@@ -69,7 +69,13 @@ module.exports = {
       // Support React Native Web
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
       'react-native': 'react-native-web',
-      'ovirtapi': `${paths.appSrc}/ovirtapi/index.js`
+      'ovirtapi': `${paths.appSrc}/ovirtapi/index.js`,
```
sjd78: Use common aliases as in the dev config file.

bond95:
> so the aliases are added for webpack and for jest. Now we just need to also add them for vscode as per this article and jsconfig.json reference on using webpack aliases.

I think it must do each user by itself, because IDE configuration file is unique and must be ignored by git.

bond95:
Also added to README paragraph about VSCode configuration.

sjd78:
> > so the aliases are added for webpack and for jest. Now we just need to also add them for vscode as per this article and jsconfig.json reference on using webpack aliases.
> 
> I think it must do each user by itself, because IDE configuration file is unique and must be ignored by git.

That file is a project configuration file, and other project do include it.  Actual vscode config files on a per user basis are stored in .vsconfig/.  I don't see a problem with including it and enhancing the developer experience for vscode users.

@gregsheremeta - Any opinion one way or another?

gregsheremeta:
> @gregsheremeta - Any opinion one way or another?

yeah, I'll be annoyed if I have to set something up manually after this is merged :)
Instead of a readme, I just want it to work out of the box. So if that's possible...

bond95:
> That file is a project configuration file, and other project do include it. 
 Actually I didn't saw that other pro

bond95:
> That file is a project configuration file, and other project do include it.

Actually, I never saw that other projects add IDE config files, further more adding files that not depend to project is bad practice. It can make conflicts and etc.
For example we can look at ovirt-engine, in .gitignore there is next lines:
```
# IDE files
###########################
.settings
.project
.classpath
.factorypath
.externalToolBuilders
.checkstyle
maven-eclipse.xml
.pmd
*.ipr
*.iml
.idea
.pydevproject
```

Thus I think it would be better not to add IDE depend files to project.

sjd78:
While there is decent comparison, ovirt-engine isn't a javascript react project.  Ovirt-engine also configures the project with maven so a lot of the IDE files get rebuilt from there.  Web-ui is built by yarn, but that isn't as rich of a project definition as the maven config.

**.gitignore** in **patternfly-react**:
````
node_modules
dist
.out
/.changelog
.DS_Store
coverage
public/
.cache
.tmp
packages/react-icons/src/icons/
packages/react-icons/src/index.js
packages/react-icons/src/index.d.ts

# package managers
yarn-error.log
lerna-debug.log

# IDEs and editors
/.idea
.project
.classpath
.c9/
*.launch
.settings/
*.sublime-workspace

# IDE - VSCode
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
````

Also, patternfly-react includes a **tsconfig.json** which is essentially the same as **jsconfig.json**: https://github.com/patternfly/patternfly-react/blob/master/tsconfig.json

Thus, I think that adding **jsconfig.json** is acceptable and useful.

**Comment by sjd78** [comment]
Also, @bond95, any particular reason why you want to have discrete aliases instead of just a `"_/*"` or `"App/*"` pointed at `${paths.appSrc}`?

**Comment by bond95** [comment]
> While there is decent comparison, ovirt-engine isn't a javascript react project. Ovirt-engine also configures the project with maven so a lot of the IDE files get rebuilt from there. Web-ui is built by yarn, but that isn't as rich of a project definition as the maven config.

@sjd78 It doesn't matter that ovirt-engine is not javascript project. I trying to say, that using IDE files in project it is bad decision. Because different developers using different IDE, and sometimes it can make conflicts. Also for alias configuration can be used `package.json`.

Thus I still don't like this. But ok, I added `jsconfig.json`.

> Also, @bond95, any particular reason why you want to have discrete aliases instead of just a "_/*" or "App/*" pointed at ${paths.appSrc}?

Nothing, I think this is good idea. Thus I rewrite aliases to `_`.

File: jsconfig.json [hunk 3]
```diff
@@ -0,0 +1,16 @@
+{
```
sjd78: I tried it out, and found the following works well for `jsconfig.json`:

````json
{
  "compilerOptions": {
    "baseUrl": "./",
    "paths": {
      "_/*": [ "src/*" ]
    }
  },
  "exclude": [
    "node_modules",
    "build",
    "extra"
  ]
}

````

PR #838, opened 2018-10-18
ux-redesign: VM Details > Disks Card

sjd78:
Closes: #775
~~Fixes: #851~~ (PR #863 fixed Issue #851), Fixes: #852

**What was added?**
In View:
 - Handle "no disks" scenario
 - `DiskListItem` renedered per disk attached to the VM.  It includes:
   - Disk state icon (active, inactive, locked)
   - Alias/Name
   - bootable label if the disk is bootable
   - provisioned size in GiB
   - Edit mode can be activated with the pencil icon based on the user's authorization to edit the VM

In Edit:
 - Create New Disk link available
 - Each existing disk is given an edit and delete icon
   - A disk may be deleted if the VM is down, it isn't in process of being deleted and it isn't locked
   - A disk may be edited as long as it isn't in process of being deleted and it isn't locked

Create Disk:
 - To create a disk, a new modal dialog is opened
 - Creation requires:
   - Disk alias (one will be suggested based on VM name and existing disks)
   - Disk size (defaults to 1GiB)
   - Storage domain (if the VM already has attached Disks, the most commonly used storage domain will be selected by default)
   - Disk format (raw && !sparse or cow && sparse)

Edit Disk:
 - To edit a disk, a new modal dialog is opened
 - The current size of the disk is always displayed
 - Edits are only allowed on:
   - Disk alias
   - If the disk is an **image** type, Disk size (only to extend/expand the disk image)
 - Once change to a disk have been submitted (asynchronously) the card is kept up to date with the disk's state (it'll show as locked while the disk image is being expanded)

Delete Disk:
 - The VM must be stopped/down to allow deletion of a VM disk
 - A delete confirmation modal verifies the action
 - While the deletion is in progress, the locked state of the disk is kept current on the card itself

**Legacy Disk Updates**
 - Adopted the redux/ovirtapi updates
 - Fixed the way disks are created so they are attached properly.  See [BZ1639630](https://bugzilla.redhat.com/show_bug.cgi?id=1639630).

**ovirtapi updates**
 - In transforms, created new helper function `cleanUndefined` to remove keys from Objects whose values are `undefined`.  It makes it a bit easier to work with ImmutableJS and doing `mergeDeepIn` operations of transformed objects.
 - All redux create and delete related things moved to the top level
 - Added the ability to EDIT/UPDATE a disk attachment/disk (new actions, sagas, VMS reducers, and API calls)

[...]

PR #847, opened 2018-11-03
[ovirt-web-ui-1.4] adapt the disk create chages from #838 to 1.4

sjd78:
Disks created on 1.4 branch behave differently if the VM is up or down at the time. This PR makes the active/inactive disk attachment state an explicitly checked thing.  This way a disk can be added to a VM that is "up" or "down".

This solves [BZ1639630](https://bugzilla.redhat.com/show_bug.cgi?id=1639630) in the 1.4 branch.

An update was also made to the disk delete button.  The button will now be hidden if the VM is not "down".  Only down VMs may have their disks removed with the technique being used.

For a running VM:
screenshot-localhost-3000-2018 11 03-01-26-30

The same VM but down:
screenshot-localhost-3000-2018 11 03-01-27-14


sjd78:
ci build please

sjd78:
ci build please

sjd78:
@bond95 , @gregsheremeta - please review so the BZ can be moved forward

src/ovirtapi/index.js
```diff
@@ -244,6 +244,10 @@ const OvirtApi = {
     assertLogin({ methodName: 'icon' })
     return httpGet({ url: `${AppConfiguration.applicationContext}/api/icons/${id}` })
   },
+  diskattachment ({ vmId, attachmentId }: { vmId: string, attachmentId: string}): Promise<Object> {
+    assertLogin({ methodName: 'diskattachments' })
```
bond95: `diskattachment`

PR #860, opened 2018-11-13
console: fix the 'in use' check

sjd78:
The existing logic to check if the current web-ui user is a current console user on a VM was inverted and causing the "console in use" dialog to be presented at the wrong time.

Fixes: #861

### DIFF
--- a/src/saga/console/index.js
+++ b/src/saga/console/index.js
@@ -84,16 +84,20 @@ export function* fetchConsoleVmMeta ({ vmId }) {
 }
 
 /**
- * Check the consoleUser sessions on a VM and
+ * Check the sessions on a VM and if someone other then the given user has an open
+ * console, flag the console as "in use" requiring manual confirmation to open the
+ * console (which will disconnect the other user's existing session).
  */
 export function* getConsoleInUse (action) {
   let { vmId, usbFilter, userId } = action.payload
   yield put(setConsoleIsValid({ vmId, isValid: false }))
 
   const sessionsInternal = yield fetchVmSessions({ vmId })
-  logger.log(`vmId: ${vmId}, sessions: ${JSON.stringify(sessionsInternal)}`)
+  const consoleUsers = sessionsInternal && sessionsInternal
+    .filter(session => session.consoleUser)
+    .map(session => session.user)
 
-  if (sessionsInternal && sessionsInternal.find((x) => x.consoleUser && (!userId || x.user.id === userId)) !== undefined) {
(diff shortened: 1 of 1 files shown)


src/saga/console/index.js
```diff
@@ -84,16 +84,20 @@ export function* fetchConsoleVmMeta ({ vmId }) {
...
+ * console (which will disconnect the other user's existing session).
  */
 export function* getConsoleInUse (action) {
   let { vmId, usbFilter, userId } = action.payload
   yield put(setConsoleIsValid({ vmId, isValid: false }))
 
   const sessionsInternal = yield fetchVmSessions({ vmId })
-  logger.log(`vmId: ${vmId}, sessions: ${JSON.stringify(sessionsInternal)}`)
+  const consoleUsers = sessionsInternal && sessionsInternal
+    .filter(session => session.consoleUser)
```
bond95: I think this is strange solution, because I don't see any reason why not to add here one more condition. For example:
```JavaScript
const consoleUsers = sessionsInternal && sessionsInternal
    .filter(session => session.consoleUser && session.user.id === userId)
```
This solution will be faster, because it contains less cycles.

gregsheremeta: I didn't follow you

bond95: At line 96 @sjd78 filtering sessions, at line 100 Scott using function `find` to find out if current user use console. And I don't understand why to split this two actions, if it can be done more easier with just adding one more condition. And as result, we can avoid of using one unnecessary cycle.

sjd78: @bond95 , @gregsheremeta -- The technique I used comes from the reason the bug existed in the first place...packing way too much semantics in a single line of code.  By unpacking what is being checked to something VERY explicitly derived, the code become much easier (IMHO at least) to quickly read.

In terms of complexity, the new check uses the same complexity as the old.  Slightly more memory space may be used, but I think the trade-off is worth the added space used.

If we tried to reduce everything down to the most efficient check, a lot of things would look very different.  As such, I do not feel compelled to change this solution just to reduce down to a single line of code.  While both would work, I prefer my solution clearly checking the current user against a list of of VM console users.

samuelmacko:
ci build please

gregsheremeta:
@samuelmacko it looks like the bot is ignoring you. Let me look into that. In the meantime, ci please build

gregsheremeta:
ci please build

gregsheremeta:
@samuelmacko it's ready now

Issue #777, opened 2018-09-18
ux-redesign: VM Details > Details Card (advanced options)


sjd78:
![image3](https://user-images.githubusercontent.com/3985964/45703176-89903d80-bb41-11e8-96ca-d3d8de2bdf44.png)

The "Advanced Options" fold on the details card in edit mode.

See the [Edit VM- Basic and Advanced Options](https://docs.google.com/document/d/1PuaaeEsPbUlFaFoh9lCvxkw3gOh8-wQ7vHhdYBFFVfI/edit#heading=h.zh2twut5rkij) design doc.

samuelmacko:
ci build please

PR #867, opened 2018-11-15
Added advanced options to VM Details Card

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/777

image

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -110,6 +110,26 @@ function createTemplateList (templates) {
   return templateList
 }
 
+/*
+ * Return a normalized and sorted list of OS ready for use in a __SelectBox__ from
+ * the Map of provided templates.
+ */
+function createOsList (architecture) {
+  const osList =
+    Selectors.getOperatingSystemsByArchitecture(architecture)
+      .toList()
+      .map(os => (
+        {
+          id: os.get('id'),
+          value: os.get('description'),
+        }
+      ))
+      .sort((a, b) => localeCompare(a.value, b.value))
+      .toJS()
+
+  return osList
+}
+
 function rephraseVmType (vmType) {
   const types = {
     'desktop': msg.vmType_desktop(),
@@ -163,7 +183,7 @@ NotAvailable.propTypes = {
   id: PropTypes.string.isRequired,
 }
 
-const defaultBootDevices = List(['hd', null])
+const DEFAULT_BOOT_DEVICES = List(['hd', null])
 
 /*
  * Specific information and details of the VM (status/up-time, data center, cluster,
@@ -173,6 +193,10 @@ const defaultBootDevices = List(['hd', null])
 class DetailsCard extends React.Component {
   constructor (props) {
     super(props)
+
+    const clusterId = props.vm.getIn(['cluster', 'id'])
+    const cluster = props.clusters && props.clusters.get(clusterId)
+    const architecture = cluster && cluster.get('architecture')
 
     this.state = {
       vm: props.vm, // ImmutableJS Map
@@ -188,6 +212,7 @@ class DetailsCard extends React.Component {
       isoList: createIsoList(props.storageDomains),
       clusterList: createClusterList(props.clusters),
       templateList: createTemplateList(props.templates),
+      osList: createOsList(architecture),
     }
     this.trackUpdates = {}
     this.hotPlugUpdates = {}
@@ -252,6 +277,13 @@ class DetailsCard extends React.Component {
 
     if (prevProps.templates !== this.props.templates) {
       this.setState({ templateList: createTemplateList(this.props.templates) }) // eslint-disable-line react/no-did-update-set-state
+    }
+
+    if (prevProps.operatingSystems !== this.props.operatingSystems) {
+      const clusterId = this.props.vm.getIn(['cluster', 'id'])
+      const cluster = this.props.clusters && this.props.clusters.get(clusterId)
+      const architecture = cluster && cluster.get('architecture')
+      this.setState({ osList: createOsList(architecture) }) // eslint-disable-line react/no-did-update-set-state
     }
   }
 
@@ -328,22 +360,19 @@ class DetailsCard extends React.Component {
           {
             const template = this.props.templates.get(value)
             if (template) {
+              const templateOsType = template.getIn(['os', 'type'], 'other')
+              const templateOs = this.props.operatingSystems.find(os => os.get('type') === templateOsType)
+
               // fields that are editable on the card
               changeQueue.push(
                 { fieldName: 'memory', value: template.get('memory') / (1024 ** 3) }, // input assumed to be GiB
                 { fieldName: 'cpu', value: template.getIn(['cpu', 'vCPUs']) },
                 { fieldName: 'bootMenuEnabled', value: template.get('bootMenuEnabled') },
+                { fieldName: 'os', value: templateOs && templateOs.get('id') },
+                { fieldName: 'cloudInitEnabled', value: template && template.getIn(['cloudInit', 'enabled']) },
+                { fieldName: 'cloudInitHostName', value: template && template.getIn(['cloudInit', 'hostName']) },
+                { fieldName: 'cloudInitSshAuthorizedKeys', value: template && template.getIn(['cloudInit', 'sshAuthorizedKeys']) },
               )
-
-              // fields that need to be changed but are not on the card
-              const osType = template.getIn(['os', 'type'], 'other')
-              const operatingSystems = this.props.operatingSystems
-              const operationSystem = operatingSystems.find(os => os.get('type') === osType)
-              updates = this.updateOs(updates, operationSystem)
-
-              // TODO: When could-init editing is added, make this look like the other followup changes
-              const cloudInit = template.get('cloudInit').toJS()
-              updates = updates.set('cloudInit', cloudInit)
             }
           }
           break
@@ -384,7 +413,7 @@ class DetailsCard extends React.Component {
           break
 
         case 'bootDevices':
-          const copiedDevices = updates.getIn(['os', 'bootDevices'], defaultBootDevices).toJS()
+          const copiedDevices = updates.getIn(['os', 'bootDevices'], DEFAULT_BOOT_DEVICES).toJS()
           copiedDevices[additionalArgs.device] = value
           for (let i = additionalArgs.device + 1; i < copiedDevices.length; i++) {
             copiedDevices[i] = copiedDevices[i] === value ? null : copiedDevices[i]
@@ -492,7 +521,7 @@ class DetailsCard extends React.Component {
 
     if (this.trackUpdates['bootDevices']) {
       vmUpdates['os'] = {
-        bootDevices: stateVm.getIn(['os', 'bootDevices']),
+        bootDevices: stateVm.getIn(['os', 'bootDevices']).toJS(),
       }
     }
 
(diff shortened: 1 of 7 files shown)


sjd78:
@bond95, @gregsheremeta, @lwrigh - I know it isn't on #777, but what about the Console switches as in the design document?

image18

bond95:
> I know it isn't on #777, but what about the Console switches as in the design document?

@sjd78 I think it better to do in separate PR. Also would be nice to discuss about what else can be added to `Advanced options`.

lwrigh:
@bond95 That looks like a good option to me. Are you imagining the console link would take the user to the console view and open up in another window or tab? Or would the user be transported to a different page that features the console?

gregsheremeta:
@lwrigh right now those links send the console.vv or console.rdp file that opens the external application, probably remote-viewer. When we implement the console feature, it will go to your new design (a new page).

lwrigh:
@gregsheremeta Cool that sounds good to me!

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -436,12 +486,29 @@ class DetailsCard extends React.Component {
       vmUpdates['bootMenuEnabled'] = stateVm.get('bootMenuEnabled')
     }
 
+    if (this.trackUpdates['cloudInit']) {
+      vmUpdates['cloudInit'] = stateVm.get('cloudInit').toJS()
+    }
+
+    if (this.trackUpdates['bootDevices']) {
+      vmUpdates['os'] = {
+        bootDevices: stateVm.getIn(['os', 'bootDevices']),
```
sjd78: Needs a `.toJS()`

Could even slice the array to make sure only 2 boot devices are set.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -508,40 +575,71 @@ class DetailsCard extends React.Component {
...
+    // VCPU
     const vCpuCount = vm.getIn(['cpu', 'vCPUs'])
 
+    // Boot devices
+    const allowedBootDevices = ['hd', 'network', 'cdrom']
+    const FIRST_DEVICE = 0
+    const SECOND_DEVICE = 1
+    const bootDevices = []
+
+    for (let i = 0; i < defaultBootDevices.size; i++) {
```
sjd78: Why even bother with an array if we know we only support exactly 2 devices? For example:
````js
    const bootDevices = [
      vm.getIn(['os', 'bootDevices', 0], defaultBootDevices.get(0)),
      vm.getIn(['os', 'bootDevices', 1], defaultBootDevices.get(1)),
    ]
````

File: src/components/VmDetails/cards/DetailsCard/index.js [hunk 3]
```diff
@@ -344,6 +361,39 @@ class DetailsCard extends React.Component {
...
+          break
+
+        case 'os':
+          fieldUpdated = 'os'
+          const operatingSystems = this.props.operatingSystems
+          const os = operatingSystems.find(os => os.get('id') === value)
+          updates = this.updateOs(updates, os)
+          break
+
+        case 'bootDevices':
```
sjd78: Suggestion -- have 2 `case` blocks.  One 'firstBootDevice', and the other 'secondBootDevice'.  The `additionalArgs` wouldn't be necessary anymore.  It would also simplify the update logic.  For example:

````
case firstBootDevice:
  update 'os', 'bootDevice', 0 to the new value
  if the second boot device === new first, use the changeQueue to set the second to null
  flag 'bootDevices' as updated

case secondBootDevice:
  if the new value equals the first boot device, do nothing
  update 'os', 'bootDevice', 1 to the new value
  flag 'bootDevices' as updated

File: src/components/VmDetails/cards/DetailsCard/index.js [hunk 4]
```diff
@@ -724,6 +813,113 @@ class DetailsCard extends React.Component {
...
+                                item !== bootDevices[FIRST_DEVICE]
+                              )).map(item => (
+                                { id: item, value: msg[`${item}Boot`]() }
+                              ))]}
+                              selected={bootDevices[SECOND_DEVICE]}
+                              onChange={(selectedId) => { this.handleChange('bootDevices', selectedId, { device: SECOND_DEVICE }) }}
+                            />
+                          </FormGroup>
+                        </Col>
+                      </div>
```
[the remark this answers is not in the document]
gregsheremeta: @bond95 I'm pretty sure we agreed in the last design meeting or demo to have labels next to fields  like @sjd78 suggests. But ultimately it's up to @lwrigh , Laura what do you think about labels on top vs left of field for boot order? See screens above

lwrigh:
@bond95 I would recommend keeping the field sizes all the same size in the right hand column for first and second boot device for consistency. Since the boot device names are typically shorter than the SSH key, I do not see the need to extend the field. 

@sjd78 @gregsheremeta What do you think?

bond95:
@lwrigh But first and second boot devices fields are subitems for `Boot Order` it means it must be consistent to `Hostname` and `SSH Key`, but not to regular field.

gregsheremeta:
> But first and second boot devices fields are subitems for Boot Order it means it must be consistent to Hostname and SSH Key, but not to regular field.

I think you're over-thinking it -- it just looks better lined up :)

bond95:
> I think you're over-thinking it -- it just looks better lined up :)

@gregsheremeta but no, user will be confused, because there is empty `Boot Order` and right after are devices without any indentation, like regular fields. So user will think that this is regular fields, not subfields, thus why there is `Boot Order` in this case?

sjd78:
> > I think you're over-thinking it -- it just looks better lined up :)
> 
> @gregsheremeta but no, user will be confused, because there is empty `Boot Order` and right after are devices without any indentation, like regular fields. So user will think that this is regular fields, not subfields, thus why there is `Boot Order` in this case?

@bond95 - We discussed the advanced options for the details card quite a few times in redesign meetings and what we ultimately agreed on is what is in the design document.  Labels to the left, select box to the right.  If anything, adding a style to the label to indent it or un-bold the font would add a bit of distinction, but still makes sense to keep the label and field on the same line.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -344,6 +361,39 @@ class DetailsCard extends React.Component {
...
+        case 'cloudInitSshAuthorizedKeys':
+          updates = updates.setIn(['cloudInit', 'sshAuthorizedKeys'], value)
+          fieldUpdated = 'cloudInit'
+          break
+
+        case 'os':
+          fieldUpdated = 'os'
+          const operatingSystems = this.props.operatingSystems
+          const os = operatingSystems.find(os => os.get('id') === value)
+          updates = this.updateOs(updates, os)
```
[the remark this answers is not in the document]
bond95: I prefer to leave `updateOs`, to make code more readable.

src/components/VmDetails/cards/DetailsCard/index.js
bond95: This is unnecessary split code into separate cases, thus I prefer to leave as it is.

bond95:
@sjd78 @gregsheremeta @lwrigh please, review.
image

lwrigh:
@bond95 The updates look good to me, thank you for updating it! +1

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -253,6 +278,13 @@ class DetailsCard extends React.Component {
     if (prevProps.templates !== this.props.templates) {
       this.setState({ templateList: createTemplateList(this.props.templates) }) // eslint-disable-line react/no-did-update-set-state
     }
+
+    if (prevProps.operatingSystems !== this.props.operatingSystems) {
```
sjd78: Is this enough of a check for when the `osList` needs to be rebuilt?  Since `osList` is dependent on the VM (to get the cluster) and Clusters (to get the architecture), may be worth adding a check for new props on those as well.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -253,6 +278,13 @@ class DetailsCard extends React.Component {
...
+      const clusterId = this.props.vm.getIn(['cluster', 'id'])
+      const cluster = this.props.clusters && this.props.clusters.get(clusterId)
+      const architecture = cluster && cluster.get('architecture')
+      this.setState({ osList: createOsList(architecture) }) // eslint-disable-line react/no-did-update-set-state
```
sjd78: I suggest pushung the `clusterId`, `cluster` and `architecture` lookups to `createOsList()` and just pass in the VM and Clusters.  That way those lookups only happen in `createOsList()` instead of being repeated here and in the constructor.

src/components/VmDetails/cards/DetailsCard/index.js
sjd78: Note: So technically having an ImmutableJS object attached in normal object will convert ok with `JSON.stringify(obj)`.  Adding the `toJS()` is nice for devs so we can verify contents of `vmUpdates` without clicking through the object structure in the debugger.

src/components/VmDetails/cards/DetailsCard/index.js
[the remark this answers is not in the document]
sjd78: That is reasonable.

src/components/VmDetails/cards/DetailsCard/index.js
sjd78: Ok.

To split hairs a bit... I still find the list processing difficult to read.  We already know that we are only dealing with 2 boot devices, why not simplify assuming that?  Plus, your loop assumes that the value won't overlap with previous values (in other words, it assume that the second device will never be changed to the first device's value).  Thoughts?

bond95: Actually I don't see any problem with loop. Because instead of using if's, it use small pretty loop. I added comment and space to make it more clear.

sjd78: > Actually I don't see any problem with loop. Because instead of using if's, it use small pretty loop. I added comment and space to make it more clear.

I'm ok with that.  Thanks for taking another look.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -110,6 +110,26 @@ function createTemplateList (templates) {
   return templateList
 }
 
+/*
+ * Return a normalized and sorted list of OS ready for use in a __SelectBox__ from
+ * the Map of provided templates.
+ */
+function createOsList (architecture) {
```
sjd78: I suggest pushing the `clusterId`, `cluster` and `architecture` lookups to this function and just pass in the VM and Clusters. That way those lookups only happen here instead of being repeated in the constructor and `componentDidUpdate`.

samuelmacko:
ci build please

gregsheremeta:
ci build please

gregsheremeta:
wow, the rpm build fails. It kinda just occurred to me that travis tests using yarn / node, while stdci uses the make rpm.

@sjd78 culprit is almost certainly c79146219f07381a24ef7b5a6a0dc2224863e86f. Can you fix?

gregsheremeta:
ci build please

gregsheremeta:
[rebased]

leistnerova:
I have only one thing, RHEL doesn't fit into OS select box. Otherwise it looks fine.

screenshot from 2019-01-04 08-56-18

bond95:
@leistnerova @lwrigh @gregsheremeta 
image

lwrigh:
In a case like this we would probably want to truncate the RHEL name so it becomes 'Red Hat Enterprise Linux 3.x...'. The full name would appear in the dropdown however.

bond95:
@lwrigh 
image

lwrigh:
@bond95 That looks great to me!

Issue #869, opened 2018-11-16
[BZ#1644970] web-ui does not load on IE11


bond95:
Description of problem:

Internet Explorer is a Tier 2 browser[1]. But the User Portal (ovirt-web-ui) does not load on it. After entering the credentials and proceeding it crashes.

The error on the console is:
SCRIPT483: Object doesn't support property or method 'includes'
main.353f9a25.js (1,9689)

See the second screenshot for more details, it points to a function calling this method.

Version-Release number of selected component (if applicable):
ovirt-web-ui-1.4.2-1.el7.noarch
ovirt-engine-4.2.6.4-1.el7.noarch

How reproducible:
100%

Steps to Reproduce:
1. Open the user portal using IE11

Actual results:
Blank screen

Expected results:
Load

Additional info:
[1] https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.2/html/planning_and_prerequisites_guide/rhv_requirements#client-requirements_RHV_planning

PR #870, opened 2018-11-16
Fixed IE11 support bug

bond95:
Added `babel-polyfill` and removed `promise` and `object-assign` packages.
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/869


sjd78:
@bond95 - Can you look into my questions and let us know what you find?

bond95:
@sjd78 

> Does this add a bunch of other polyfils that we don't need that may cause other issues? (I guess not much, but it is worth a look anyway)

`babel-polyfill` include generator support and a lot of ES6 features (Promises, Array methods and etc), more info you can find here: https://github.com/zloirock/core-js

> What does it do to size of the chunk? In other words, how much bigger is the js download after putting this polyfill in place? Does it make sense to do some app-init time checking and only require the polyfill dynamically, as needed, and keep it in another chunk so current browsers don't have to deal with it?

Size of built JS file without `babel-polyfill` ~ 2.5 MB, with polyfill ~ 2.6 MB. So no, it is not needed to do some app-init time checking.

leistnerova:
ci build please

PR #871, opened 2018-11-19
Move the core error messages above PageRouter

### LINKED ISSUE #513: hidden error message when user is logged out
When user is logged out error message appears but is not visible.

Steps to reproduce:
1. open two VM potals and log in with one user
2. in one portal log out
3. in the other one click on Reload button in top menu

Tested in ovirt-web-ui-1.3.6-1.fc26.noarch
screenshot from 2018-03-06 08-20-22

sjd78:
After the redesign layout upgrade, the error message (oVirt version
check and SSO token expiry) alerts were on the bottom of the page
and not always visible.  Moved the alerts above the page render so
they are always visible.

Fixes #513

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -50,10 +50,10 @@ const App = ({ history, config, appReady }) => {
     <ConnectedRouter history={history}>
       <div id='app-container'>
         <VmsPageHeader title={fixedStrings.BRAND_NAME + ' ' + msg.vmPortal()} />
-        { appReady && renderRoutes(getRoutes()) }
-        <LoadingData />
         <OvirtApiCheckFailed />
         <TokenExpired />
+        { appReady && renderRoutes(getRoutes()) }
+        <LoadingData />
         <ToastNotifications />
       </div>
     </ConnectedRouter>


sjd78:
What it looks like with this change:
screenshot-localhost-3000-2018 11 20-13-36-56

screenshot-localhost-3000-2018 11 20-13-37-42

lwrigh:
@bond95 Are you saying the error notification should be located in the grey box of the dashboard?

lwrigh:
@bond95 I see, I do think that's a cleaner layout option for the error message.

sjd78:
@bond95 , @lwrigh - But these are hard errors that should prevent web-ui from functioning.  And they can occur regardless (at least the timeout one) of what page you're on.  I'd leave them as errors for now.  Going forward, there a few issues to adopt better timeout handling and login errors.  For now, I'll settle for visible error messages...

gregsheremeta:
what does this patch look like on the dasbboard page now when there is an error? (we're talking about dashboard but looking at the listing page :) )

bond95:
> we're talking about dashboard but looking at the listing page

Actually I talk about more consistent with dashboard style errors on main page.

sjd78:
> > we're talking about dashboard but looking at the listing page
> 
> Actually I talk about more consistent with dashboard style errors on main page.

I think it is important to keep in mind that these errors are the "halt and catch fire" types of errors when the api endpoint is either too old, went away, or SSO timed out.  I don't think they should look like the existing "oh this VM will need to be restarted to pickup the new configurations" info message.

Really I think those three cases should be handled differently all together, but for right now, I'm just making sure the messages are visible and prominent whereas they were not before.

sjd78:
For comparison, here is the authorization expired error (generated by using the Redux DevTools and submitting the dispatch `{ type: 'SHOW_TOKEN_EXPIRED_MSG', payload: {} }`) across all of the rest of the pages.

**Create**
screenshot-localhost-3000-2018 11 27-12-44-03

**Legacy View**
screenshot-localhost-3000-2018 11 27-12-44-43

**Dashboard Details**
screenshot-localhost-3000-2018 11 27-12-45-56

gregsheremeta:
On the legacy pages let's not care, so we are only thinking about the dashboard view:

image

vs

screenshot-localhost-3000-2018 11 27-12-45-56

I prefer Scott's version, the second one (red above the buttons, not integrated).

@lwrigh to send it to a tie or make it 3-1 ?

lwrigh:
@gregsheremeta I think I'm going to have to go with Scott's version if the goal of the error is to be a  "halt and catch fire" sort of error since it's so prominent.

pnovotny:
ci build please

gregsheremeta:
ci build please

pnovotny:
Verified in ovirt-web-ui-1.5.0-0.20181129git8e1b8d3.el7.noarch

The "Authorization expired..." alerts look exactly like the ones from [Scott's comment](#issuecomment-442153219).

Issue #778, opened 2018-09-18
ux-redesign: VM Details > Utilization Card > Disk Details Charts


sjd78:
Update the Disk details chart to use horizontal bar charts to indicate a per attached disk allocation v provisioned space used.  See the [VM Portal Dashboard- Disk Details](https://docs.google.com/document/d/1z74VD--M75fxklLJOQQhDO1Wc-PfA0W55fvwM3a9NwM/edit?ts=5b75d71d#heading=h.mrj4pzxmy7tw) design doc.

For VMs with guest agents reporting file system data, bar charts for the mounted file systems should be displayed instead of raw disk allocated vs provisioned sizes.

With disk charts:
![image14](https://user-images.githubusercontent.com/3985964/45703556-ad07b800-bb42-11e8-936e-fc2604fa19b7.png)

With filesystem charts:
<img width="300" alt="image5" src="https://user-images.githubusercontent.com/3985964/47512318-ee7a3a00-d849-11e8-8dec-1eb80ec212d0.png">
<img width="300" alt="image7" src="https://user-images.githubusercontent.com/3985964/47512320-f0dc9400-d849-11e8-8af8-07045e78e17d.png">

PR #873, opened 2018-11-19
ux-redesign: Redesign utilization disk card

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/778

Running VM with Guest Agent:
image
image

Running VM without Guest Agent or without OS at all
image

Stopped VM:
image

Running Windows:
image

### DIFF
--- a/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
@@ -9,7 +9,7 @@ import {
   UtilizationCardDetailsDesc,
   UtilizationCardDetailsLine1,
   UtilizationCardDetailsLine2,
-  DonutChart,
+  UtilizationBar,
 } from 'patternfly-react'
 
 import { round } from '_/utils'
@@ -24,6 +24,36 @@ const EmptyBlock = () => (
   <div className={style['no-history-chart']} />
 )
 
+const DiskBar = ({ path, total, used }) => {
+  const { unit, value } =
+    convertValueMap(
+      'B',
+      {
+        total: total,
+        used: used,
+      })
+  const thresholdError = 90
+  const thresholdWarning = 70
+  const usedInPercent = round(used / total * 100, 0)
+  return <div className={style['disk-fs-box']}>
+    <div className={style['disk-fs-name']}>{path}</div>
+    <div className={style['disk-fs-bar']}>
+      <UtilizationBar
+        now={usedInPercent}
+        thresholdWarning={thresholdWarning}
+        thresholdError={thresholdError}
+      />
+    </div>
+    <div className={style['disk-fs-used']}><strong>{round(value.used, 0)} of {round(value.total, 0)} {unit}</strong> Used</div>
+  </div>
+}
+
+DiskBar.propTypes = {
+  path: PropTypes.string.isRequired,
+  total: PropTypes.number.isRequired,
+  used: PropTypes.number.isRequired,
+}
+
 /*
  * Disks, but intended to be in terms of guest agent reported data (file system viewpoint),
  * not in terms of storage allocation (infrastructure viewpoint - like dashboard/webadmin).
@@ -31,8 +61,8 @@ const EmptyBlock = () => (
  * NOTE: File system usage data from the guest agent is not currently (Aug, 2018) available
  *       via REST. Storage allocation is being used instead.
  */
-const DiskCharts = ({ vm, isRunning, id, ...props }) => {
-  const hasDisks = vm.get('disks').size > 0
+const DiskCharts = ({ vm, diskStats, isRunning, id, ...props }) => {
+  const diskDetails = diskStats['usage'].datum
 
   let actualSize = 0
   let provisionedSize = 0
@@ -47,7 +77,6 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
     }
   })
 
-  const usedFormated = userFormatOfBytes(actualSize)
   const availableFormated = userFormatOfBytes(provisionedSize - actualSize)
   const totalFormated = userFormatOfBytes(provisionedSize)
 
@@ -55,10 +84,13 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
     <UtilizationCard className={style['chart-card']} id={id}>
       <CardTitle>Disk <span style={{ fontSize: '55%', verticalAlign: 'super' }}>(storage allocations)</span></CardTitle>
       <CardBody>
-        { !hasDisks &&
-          <NoLiveData id={`${id}-no-live-data`} message='This VM has no attached disks.' />
+        { !diskDetails && isRunning &&
+          <NoLiveData id={`${id}-no-live-data`} message='It seems that no guest agent is configurated on VM.' />
         }
-        { hasDisks &&
+        { !diskDetails && !isRunning &&
+          <NoLiveData id={`${id}-no-live-data`} />
+        }
+        { diskDetails &&
         <React.Fragment>
           <UtilizationCardDetails>
             <UtilizationCardDetailsCount id={`${id}-available`}>
@@ -69,26 +101,13 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
               <UtilizationCardDetailsLine2 id={`${id}-total`}>of {round(totalFormated.number, 1)} {totalFormated.suffix} Provisioned</UtilizationCardDetailsLine2>
             </UtilizationCardDetailsDesc>
           </UtilizationCardDetails>
-
-          <DonutChart
-            id={`${id}-donut-chart`}
-            data={{
-              columns: [
-                [`allocated`, actualSize],
-                [`unallocated`, provisionedSize - actualSize],
-              ],
-              order: null,
-            }}
-            title={{
-              primary: `${round(usedFormated.number, 0)}`,
-              secondary: `${usedFormated.suffix} Allocated`,
-            }}
-            tooltip={{
-              show: true,
-              contents: donutMemoryTooltipContents,
-            }}
-          />
-
+          <div className={style['disk-fs-list']}>
+            {
+              diskDetails.map((disk) =>
+                <DiskBar path={disk.path} total={parseInt(disk.total)} used={parseInt(disk.used)} />
+              )
+            }
+          </div>
           {/* Disks don't have historic data but stub the space so the card stretches like the others */}
           <EmptyBlock />
         </React.Fragment>
@@ -100,6 +119,7 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
 DiskCharts.propTypes = {
   id: PropTypes.string.isRequired,
   vm: PropTypes.object.isRequired,
+  diskStats: PropTypes.object.isRequired,
   isRunning: PropTypes.bool,
 }
 
--- a/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
@@ -10,9 +10,10 @@ import {
   UtilizationCardDetailsLine1,
   UtilizationCardDetailsLine2,
   UtilizationBar,
+  DonutChart,
 } from 'patternfly-react'
 
-import { round } from '_/utils'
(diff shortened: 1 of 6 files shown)


lwrigh:
This is looking great @bond95! Overall everything looks good to me, I have a few comments for the messaging though. 

For the no guest agent message, I would suggest having it say, "It looks like no guest agent is configured on the VM.". 

I would also suggest changing the no data available message to, "Utilization data is only available when the VM is running.".

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -19,15 +18,45 @@ import style from './style.css'
...
+  return <div className={style['disk-fs-box']}>
+    <div className={style['disk-fs-name']}>{path}</div>
+    <div className={style['disk-fs-bar']}>
+      <UtilizationBar
+        now={usedInPercent}
+        thresholdWarning={thresholdWarning}
+        thresholdError={thresholdError}
+      />
+    </div>
+    <div className={style['disk-fs-used']}><strong>{round(value.used, 0)} of {round(value.total, 0)} {unit}</strong> Used</div>
```
sjd78: The `strong` should be in the style instead ... `font-weight: bold;`

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -43,18 +72,20 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
...
   const total = round(disks.provisionedSize, 1)
 
   return (
     <UtilizationCard className={style['chart-card']} id={id}>
       <CardTitle>Disk <span style={{ fontSize: '55%', verticalAlign: 'super' }}>(storage allocations)</span></CardTitle>
       <CardBody>
-        { !hasDisks &&
-          <NoLiveData id={`${id}-no-live-data`} message='This VM has no attached disks.' />
+        { !diskDetails && isRunning &&
+          <NoLiveData id={`${id}-no-live-data`} message='It seems that no guest agent is configurated on VM.' />
```
sjd78: This reminds me that Utilization Card needs to be localized.  Perhaps you can start with this message.

src/ovirtapi/transform.js
```diff
@@ -260,20 +260,24 @@ const VmStatistics = {
...
       // no values -> undefined, 1 value -> value.datum, >1 values -> [...values.datum]
+      // ?disks.usage -> {detail...}
       const datum =
         stat.values &&
         stat.values.value &&
-        (stat.values.value.length === 1
-          ? stat.values.value[0].datum
-          : stat.values.value.map(value => value.datum))
+        (stat.name === 'disks.usage'
+          ? JSON.parse(stat.values.value[0].detail || 'null')
```
sjd78: I recommend moving the JSON parsing down lower and have it `convertInt()` values here so the `parseInt` doesn't need to be done in the components.

````js
if (stat.name === 'disks.usage') {
  datum = JSON.parse(datum)
  datum.total = convertInt(datum.total)
  datum.used = convertInt(datum.used)
}
````

File: src/components/VmDetails/cards/UtilizationCard/DiskCharts.js [hunk 4]
```diff
@@ -43,18 +72,20 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
...
   const available = round(disks.provisionedSize - disks.actualSize, 1)
   const total = round(disks.provisionedSize, 1)
 
   return (
     <UtilizationCard className={style['chart-card']} id={id}>
       <CardTitle>Disk <span style={{ fontSize: '55%', verticalAlign: 'super' }}>(storage allocations)</span></CardTitle>
       <CardBody>
-        { !hasDisks &&
-          <NoLiveData id={`${id}-no-live-data`} message='This VM has no attached disks.' />
+        { !diskDetails && isRunning &&
```
sjd78: Humm.... If the guest agent isn't reporting file system use, the disk allocated vs provisioned is still available to render.  I think it makes sense to put the `NoLiveData` messages below the `UtilizationCardDetails`.  The disk allocations, assuming the VM has attached disks, will always be available regardless of the state of the VM.

TL;DR - Put the "no attached disks" `NoLiveData` message check back here and move the 2 new `NoLiveData` messages to the same spot as the `DiskBar`s.

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -43,18 +72,20 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
   })
 
   const { unit, value: disks } = convertValueMap('B', { actualSize, provisionedSize })
-  const used = round(disks.actualSize, 1)
   const available = round(disks.provisionedSize - disks.actualSize, 1)
   const total = round(disks.provisionedSize, 1)
 
   return (
     <UtilizationCard className={style['chart-card']} id={id}>
       <CardTitle>Disk <span style={{ fontSize: '55%', verticalAlign: 'super' }}>(storage allocations)</span></CardTitle>
```
sjd78: We can probably take the super-scripted "storage allocations" off the `CardTitle` now.

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
bond95: It is better to do in separate PR, all at once: https://github.com/oVirt/ovirt-web-ui/issues/879

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
bond95: In this case I would prefer `strong`, because it will bold only some part of text, it is more readable, then adding a lot of `span`s and classes. Also this style would work for localization.

bond95:
@lwrigh @sjd78 @gregsheremeta What do you think about this solution

With guest agent
image

Without guest agent
image

Without disks
image

VM is off
image

lwrigh:
@bond95 It looks good to me!👍

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
sjd78: Good point.

sjd78:
@bond95, @lwrigh, @gregsheremeta - When the VM is off, all of the attached disks and allocations are still known.  The card detail section can still be displayed (X unallocated of Y provisioned).

The chart below I think should only be displayed when guest agent data is available.

So, I see the variations as:
  1. No disks attached to the VM
  2. Disks attached, VM stopped (X unallocated of Y provisioned, no chart data)
  3. Disks attached, VM running, no guest agent (X of Y, no chart data)
  4. Disks attached, VM running, guest agent (X of Y, bar charts for file systems)

Looks like 1 and 4 are good.  2 and 3 are close, just need the X of Y headers.

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -43,48 +74,60 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
...
 
   return (
     <UtilizationCard className={style['chart-card']} id={id}>
-      <CardTitle>Disk <span style={{ fontSize: '55%', verticalAlign: 'super' }}>(storage allocations)</span></CardTitle>
+      <CardTitle>Disk</CardTitle>
       <CardBody>
-        { !hasDisks &&
-          <NoLiveData id={`${id}-no-live-data`} message='This VM has no attached disks.' />
+        { !diskDetails && !isRunning &&
+          <NoLiveData id={`${id}-no-live-data`} />
```
sjd78: I think this `NoLiveData` needs to go below the `UtilizationCardDetails`.

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -43,48 +74,60 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
...
-            }}
-            tooltip={{
-              show: true,
-              contents: patternfly.pfDonutTooltipContents,
-            }}
-          />
-
-          {/* Disks don't have historic data */}
-        </React.Fragment>
+        { isRunning && vm.get('disks').size === 0 &&
```
sjd78: `isRunning` is not necessary, doesn't matter if the VM is up or down, the "no disk is attached" message should still be displayed.

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -43,48 +74,60 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
...
+          <NoLiveData id={`${id}-no-live-data`} message='It looks like no disk is attached to VM.' />
+        }
+        { isRunning && vm.get('disks').size > 0 &&
```
sjd78: Here again, I don't think `isRunning` is necessary.  If disks are attached, the `UtilizationCardDetails` showing the allocated/provisioned disk total isn't dependent on the VM's state.

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -43,48 +74,60 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
...
+          <React.Fragment>
+            <UtilizationCardDetails>
+              <UtilizationCardDetailsCount id={`${id}-available`}>{available}</UtilizationCardDetailsCount>
+              <UtilizationCardDetailsDesc>
+                <UtilizationCardDetailsLine1>Unallocated</UtilizationCardDetailsLine1>
+                <UtilizationCardDetailsLine2 id={`${id}-total`}>of {total} {unit} Provisioned</UtilizationCardDetailsLine2>
+              </UtilizationCardDetailsDesc>
+            </UtilizationCardDetails>
+            { !diskDetails &&
```
sjd78: Are we sure that we want to keep the donut chart around for VMs without a guest agent?

src/components/VmDetails/cards/UtilizationCard/index.js
```diff
@@ -45,8 +45,8 @@ const UtilizationCard = ({ vm }) => {
             }
           </Col>
           <Col>
-            { vm.has('disks')
-              ? <DiskCharts vm={vm} isRunning={isRunning} id={`${idPrefix}-disk`} />
+            { vm.has('disks') && stats.disks
```
sjd78: If `stats.disks` is there, you won't be able to show the allocated/provisioned details for VMs without a guest agent.

Plus the checks in this component are really to guard against having the base VM data only and waiting for the VM's deep fetch to complete.

bond95:
@lwrigh @sjd78 @gregsheremeta 

Running without Guest Agent
image

Running with Guest Agent
image

Not running
image

Without disks
image

lwrigh:
Looks good to me!

gregsheremeta:
@bond95 how easy is it to uppercase Windows drive letters?

`C:\`
`D:\`

vs.

`c:\`
`d:\`

bond95:
ci build please

bond95:
ci build please

leistnerova:
ci build please

gregsheremeta:
Honestly I think it looks pretty good and it's difficult to get perfect.

@lwrigh any thoughts about Lucie's comments above?

lwrigh:
Ah I see, it's always a challenge to try and get text to behave nicely when the screen size gets compressed. I don't think scaling it down any more would be a good solution so I think for the time being wrapping is a better solution in this case.

lwrigh:
@gregsheremeta @bond95 @sjd78 @leistnerova What do you guys think of these options for formatting? 
group 17
group 27

bond95:
I like it.

gregsheremeta:
Yep let's go with that. If we see issues in production we can try to enhance further.

leistnerova:
I like it too, it makes higher only the row with long name and doesn't affect other mount points.

lwrigh:
Great sounds good to me.

bond95:
@lwrigh @leistnerova @gregsheremeta 

image

bond95:
(rebased)

lwrigh:
It looks good to me.

leistnerova:
ci build please

PR #876, opened 2018-11-21
Add support for ISOs in any storage domain

### LINKED ISSUE #704: ISO images from Data domain do not show up in VM Portal
ISO Domain images do show up, data domain isos do not show up in the VM Portal.  Data domain Isos do show up in the Admin Portal.

sjd78:
Recent changes to ovirt-engine allow keeping ISO images for CDs in non-ISO type storage domains. This change accommodates ISOs in any storage domain.

Fixes #704

What changes?
  - ISOs are loaded from all storage domains types at user login.

  - A single fetch covers loading the ISOs from non-ISO domains.

  - Each ISO domain still requires its own fetch, but fetching storage domains first isn't necessary and this is only done automatically at login time.

 - When a user in on a detail or dialog/edit page, the ISO list may be refreshed by manually clicking the refresh button in the App masthead.  

Summary of code changes:

 - Login sagas load ALL storage domain types, and perform an initial load of CD ISOs

 - Sagas for storage domains and ISOs moved from `sagas.js` to `saga/storageDomains.js`

 - The `GET_ISO_STORAGE_DOMAINS` actions/sagas renamed to `GET_ISO_FILES` reflecting that it is now primarily concerned with fetching the actual images instead of the ISO storage domains

 - storageDomain reducer function `ADD_STORAGE_DOMAINS` was removed since it is no longer needed

 - ovirtapi: added function `getIsoImages` to directly fetch ISO disk images without having to go through the storage domains

 - Legacy VM edit page updated to display ISOs held in any storage domain, not just 'iso' types

 - Data refresh handling updated to support differentiating between a page change, a scheduled task and a manual refresh. This allows for the ISO files to be loaded when entering a page (`onNavigation=true, onSchedule=false`), or when the user click the manual refresh buttons (`onNavigation=false, onSchedule=false`), but not when the scheduled refresh runs (`onNavigation=false, onSchedule=true`).


michalskrivanek:
that's......creative:) I really wonder about the speed of scraping all SDs in larger environment. Is it done once on startup or periodically? I would very much prefer a dedicated API for that but if that's not available and it doesn't take ages to load it this way then it's ok. but we need to test on slower system, with 100SDs

sjd78:
On the Legacy VM edit:
screenshot_2018-11-21 vm portal 1

On Details card (for a running VM):
screenshot_2018-11-21 vm portal 2


The ISOs on webadmin in a data domain:
screenshot_2018-11-21 ovirt open virtualization manager

The ISOs on webadmin in the iso domain:
screenshot_2018-11-21 ovirt open virtualization manager 1

sjd78:
> that's......creative:) I really wonder about the speed of scraping all SDs in larger environment. Is it done once on startup or periodically?

Storage domains are loaded once at startup.  That part hasn't changed.

The ISO images are loaded at startup and uses 1 less API call (since the storage domains should already be known) then before for iso domains, and uses 1 call for all of the other domains.  

> I would very much prefer a dedicated API for that but if that's not available and it doesn't take ages to load it this way then it's ok. but we need to test on slower system, with 100SDs

Loading the ISO from the data domains is a single API call with search terms (really I was playing around on webadmin and it worked so it was an easy change).  It would be quite interesting to see how it performs on a large environment.  On my local system the `/files` call seems to consistently take longer then the `disks?search=...` call.

vm portal - mozilla firefox_001

src/sagas.js
```diff
@@ -250,16 +245,25 @@ export function* refreshMainPage ({ shallowFetch, page }) {
...
   yield getConsoleOptions(actionGetConsoleOptions({ vmId: id }))
+
+  // Load ISO images on manual refresh click
+  if (!onNavigation && !onSchedule) {
+    yield fetchIsoFiles(getIsoFiles())
+  }
 }
 
-function* refreshDialogPage ({ id }) {
+function* refreshDialogPage ({ id, onNavigation, onSchedule }) {
```
bond95: Is `onNavigation` and `onSchedule` really needed here? Because dialog page must up-to-date all the time even with ISO files.

src/ovirtapi/index.js
```diff
@@ -347,6 +349,13 @@ const OvirtApi = {
     assertLogin({ methodName: 'getStorageFiles' })
     return httpGet({ url: `${AppConfiguration.applicationContext}/api/storagedomains/${storageId}/files` })
   },
+  getIsoImages (): Promise<Object> {
```
bond95: As I see, you added new line before functions, so, please, add it here too.

src/sagas.js
sjd78: ISO file loading is potentially a very heavy operation.  Doing it every minute doesn't seem like a good idea.  Doing it once at login and then allowing it to be refreshed manually (without a page reload) feels like a good trade-off.

sjd78: Plus I'd assume the set of ISO are going to be relatively static and not in need of frequent refreshes.

src/ovirtapi/index.js
sjd78: I kinda used newlines to group a few related calls together visually.  `getIsoImages` groups with `getStorageFiles` so that's why there isn't a newline there.

Really I'd love to refactor this file a bunch.  Having `${AppConfiguration.applicationContext}` everywhere really kinda hurts my eyes.

src/sagas.js
gregsheremeta: Indeed, ISOs are very rarely changed. If it came down to it, I'd be ok with telling someone to logout and back in to get a new one that the admin had just loaded :) We won't do that because we do have a refresh button in admin portal. But, tl;dr, caching++ on this one

sjd78: **Note**: I updated the ISO caching/refresh scheme for this PR such that ISOs are loaded at login time, and then only refreshed when a user manually click refresh on a detail (view or dashboard) or dialog (edit) page.

sjd78:
**Note**: I updated the ISO caching/refresh scheme for this PR such that ISOs are loaded at login time, and then only refreshed when a user manually click refresh on a detail (view or dashboard) or dialog (edit) page.

src/ovirtapi/transform.js
```diff
@@ -468,6 +468,9 @@ const StorageDomain = {
       name: storageDomain.name,
       type: storageDomain.type,
 
+      availableSpace: convertInt(storageDomain.available),
```
gregsheremeta: this appears unrelated

sjd78: So...I added those while I was reading the API thinking that those sizes would be a nice thing to show to users about a storage domain when they're creating disks.

src/components/VmDialog/index.js
```diff
@@ -509,12 +509,10 @@ class VmDialog extends React.Component {
     const { icons, templates, clusters, storages, previousPath } = this.props
     const { bootDevices } = this.state
     const vm = this.props.vm
-    const isoStorages = storages.filter(storageDomain => storageDomain.get('type') === 'iso')
     const idPrefix = `vmdialog-${vm ? vm.get('name') : '_new'}`
 
-    let files = [{ id: '', value: '[Eject]' }]
-
-    isoStorages.toList().forEach(storageDomain => {
+    const files = [{ id: '', value: '[Eject]' }]
```
bond95: Translate, please.

leistnerova:
ci build please

leistnerova:
I uploaded iso (CentOs-7...) to data/nfs and imported iso domain, in admin portal I see all images from iso domain and the one iso uploaded in Run once mode.

screenshot from 2018-12-13 09-12-16

But in VM portal when I log with admin user - I saw no images in edit details at first but after some reload selectbox has correct values. 

As non-admin user I see all iso disks from iso and data/nfs domain (including snapshot disks) and not the one iso uploaded.

screenshot from 2018-12-13 09-07-27

Tested with ovirt-engine-4.3.0-0.4.master.20181210172657.git88f34ae.el7.noarch

leistnerova:
[logs.zip](https://github.com/oVirt/ovirt-web-ui/files/2675116/logs.zip)

sjd78:
Opened [BZ1659121](https://bugzilla.redhat.com/show_bug.cgi?id=1659121) to deal with @leistnerova 's findings.

sjd78:
[BZ1659121](https://bugzilla.redhat.com/show_bug.cgi?id=1659121) is fixed and merged in ovirt-engine master.  @leistnerova retest against a current 4.3 master engine and it should all work.

leistnerova:
LGTM
Non admin user has to have UserRole/UserVmManager on the storage domain or DC. I'm OK with that. If the permissions are not correct, please open bug for the engine.

PR #881, opened 2018-11-26
update yarn.lock

gregsheremeta:
update yarn.lock


gregsheremeta:
@bond95 @sjd78 package.json was update in the ie polyfill patch, but yarn.lock was out of date. Thus builds were failing. You made need to rebase on top of this patch.

gregsheremeta:
ci build please

Issue #875, opened 2018-11-20
Edit Disks > visual artifact when editing many disks / small window size


gregsheremeta:
We noticed that
1. the Ok and Cancel buttons appear in their own mini-box below the main box
2. with a really long disk name, the icons didn't stay aligned in a column

![selection_206](https://user-images.githubusercontent.com/2248293/48789012-0498e000-ecba-11e8-92d7-feb97dcb8216.png)

I'm not sure if this happens with many disks (3 or 4+) or a certain window size.

bond95:
Similar bug
![image](https://user-images.githubusercontent.com/3332176/49799753-ff253780-fd45-11e8-95eb-84ad4d72f3ea.png)

bond95:
![image](https://user-images.githubusercontent.com/3332176/49801244-f8002880-fd49-11e8-886c-871af5d1858c.png)

But

![image](https://user-images.githubusercontent.com/3332176/49801262-01899080-fd4a-11e8-9f2b-2243c2743ef4.png)

:laughing:

PR #890, opened 2018-12-11
Fix visual artifacts in NICs and Disks cards

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/875

image

Also adding stretching for Utilization cards, because when height of bottom card grow, it looks not very good
image

image

Without stretching
image

### DIFF
--- a/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
@@ -21,7 +21,7 @@ import style from './style.css'
 import NoLiveData from './NoLiveData'
 
 const EmptyBlock = () => (
-  <div className={style['empty-block']} />
+  <div className={style['no-history-chart']} />
 )
 
 /*
@@ -89,7 +89,7 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
             }}
           />
 
-          {/* Disks don't have historic data */}
+          {/* Disks don't have historic data but stub the space so the card stretches like the others */}
           <EmptyBlock />
         </React.Fragment>
         }
--- a/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
@@ -20,6 +20,10 @@ import style from './style.css'
 
 import NoLiveData from './NoLiveData'
 
+const EmptyBlock = () => (
+  <div className={style['empty-block']} />
+)
+
 /*
  * Disks, but intended to be in terms of guest agent reported data (file system viewpoint),
  * not in terms of storage allocation (infrastructure viewpoint - like dashboard/webadmin).
@@ -86,6 +90,7 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
           />
 
           {/* Disks don't have historic data */}
+          <EmptyBlock />
         </React.Fragment>
         }
       </CardBody>
--- a/src/components/VmDetails/cards/UtilizationCard/style.css
+++ b/src/components/VmDetails/cards/UtilizationCard/style.css
@@ -16,6 +16,8 @@
 
   display: flex;
   flex-direction: column;
+
+  justify-content: space-around;
 }
 
 .no-data-card-body {
@@ -66,4 +68,12 @@
 
 .utilization-card :global(.c3-tooltip-container) {
   white-space: nowrap;
+}
+
+.full-height {
+  height: 100%;
+}
+
+.empty-block {
+  height: 60px;
 }
--- a/src/components/VmDetails/cards/UtilizationCard/style.css
+++ b/src/components/VmDetails/cards/UtilizationCard/style.css
@@ -73,7 +73,3 @@
 .full-height {
   height: 100%;
 }
-
-.empty-block {
-  height: 60px;
-}
--- a/src/components/VmDetails/cards/UtilizationCard/style.css
+++ b/src/components/VmDetails/cards/UtilizationCard/style.css
@@ -16,8 +16,6 @@
 
   display: flex;
   flex-direction: column;
-
-  justify-content: space-around;
 }
 
 .no-data-card-body {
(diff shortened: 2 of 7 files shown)


lwrigh:
Looks good to me! +1

gregsheremeta:
UI looks good, and code looks good too.
@bond95   `EmptyBlock` -- is that the "stretching" you're referring to?

bond95:
@gregsheremeta `EmptyBlock` is a component that helping to do same align for every utilization cards (only disk card don't have history data block, thus need to add empty one, for right alignment when cards are stretching).

src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
```diff
@@ -86,6 +90,7 @@ const DiskCharts = ({ vm, isRunning, id, ...props }) => {
           />
 
           {/* Disks don't have historic data */}
+          <EmptyBlock />
```
sjd78: Just use the style from `NoHistoricData` to avoid the extra component and css class and update the comment so it says why:

````jsx
  {/* Disks don't have historic data but stub the space so the card stretches like the others */}
  <div className={style['no-history-chart']} />
````

File: src/components/VmDetails/cards/UtilizationCard/style.css [hunk 2]
```diff
@@ -16,6 +16,8 @@
 
   display: flex;
   flex-direction: column;
+
+  justify-content: space-around;
```
sjd78: I like `justify-content: center;` a bit better then `space-around` so that the card body content maintain their spacing when the cards get taller as the NICs or Disks get taller.

vm portal - google chrome_010

vm portal - google chrome_009

sjd78:
@gregsheremeta , @lwrigh , @bond95 --
One observation -- as the PR sits now, as the NIC or Disk card gets bigger, it first takes space away from the other then grows the entire row (once the other hits its min-height).  So looks like this with lots of disks and no Nics:
ezgif com-gif-maker

And like this when Disks just takes space away from Nics (initially 50% height of the row):
ezgif com-gif-maker 1

I guess the question is, in this example, should the Nics card stay taller or leave it as is (really it is a bit of an edge case anyway and trying to sort it initially drove me a bit crazy and I'd rather not drive Bohdan crazy right before Christmas :wink:)?

lwrigh:
@gregsheremeta @sjd78 @bond95 I think in this case it would be more preferable if the card headers and the content of the cards stayed static but the bottom of the card grew as more content was added to the card. The card jumping or growing into another card isn't really that ideal. I can mock up this behavior for clarity.

src/components/VmDetails/cards/UtilizationCard/style.css
bond95: With `justify-content: center` there is a lot of free space around chart, for me it looks to very good.

bond95:
@lwrigh mock up would be great.

lwrigh:
Here is the behavior of the NICs card expanding and how the Disks and Utilization card react I was thinking of: 

ezgif-1-8b10fc56f9d0

Here is the behavior of the Disks card expanding and how the Utilization card react: 

expanding disk

@bond95 @sjd78 @gregsheremeta

src/components/VmDetails/cards/UtilizationCard/style.css
sjd78: @lwrigh , @gregsheremeta - The charts on the utilization and how they change position, or not, when the Nics/Disks need to grow is an open question.  3 simple ways:
  1. Leave the charts at the top of the card and let the white space at the bottom take up the extra space (like in Laura's mockup in comment https://github.com/oVirt/ovirt-web-ui/pull/890#issuecomment-447017958)
  2. Have the card contents stretch and realign the elements to space out (as in justify-content: space-around)
  3. Have the card contents stretch but keep the elements relative position.  White space at top between the card title and content, and the white space below the contents grows equally.  (as in justify-content: center)

And last thought, the "No Data Available" already grows like #3 so it would need to be tweaked to accomplish #1.

bond95:
@lwrigh thank you for mock up, this was default behavior. I changed it, because I think it looks not very well. I don't like a lot of empty space at the bottom.

gregsheremeta:
> @lwrigh thank you for mock up, this was default behavior. I changed it, because I think it looks not very well. I don't like a lot of empty space at the bottom.

I see. I actually like the whitespace version. I like that the labels stay real close to the graph.

At any rate, since @lwrigh is our visual expert let's go with her decision. @lwrigh do you want to make the call now or check in with Liz / PF real quick?

lwrigh:
@bond95 @gregsheremeta After talking to one of the visual designers on the UXD team, he thought that the option of expanding the length of the card and having the titles and the bottoms of the cards aligned was a good option for this situation like I had suggested. PatternFly's guidelines do not specify any suggestions for the situation of cards scaling so this is something that will be considered more in PatternFly 4.

bond95:
@lwrigh @gregsheremeta @sjd78  removed stretching:
ezgif com-video-to-gif 1

lwrigh:
@bond95 LGTM!

src/components/VmDetails/cards/UtilizationCard/style.css
gregsheremeta: @sjd78 so we went with option 1 right?

gregsheremeta:
@sjd78 looks ok?

src/components/VmDetails/cards/UtilizationCard/style.css
sjd78: > @sjd78 so we went with option 1 right?

Correct.

sjd78:
@gregsheremeta, @lwrigh, @bond95 
The Utilization card looks ok now.  The Nics and Disks cards still have the edge case where they shrink to minimum content height before the utilization card starts to grow.  However, the main problem of having the list of disks render below the card's white box is fixed.  I'm good with how it is working now.

leistnerova:
ci build please

leistnerova:
LGTM
tested in FF 63.0.3 and Google Chrome 70.0

PR #897, opened 2019-01-04
Remove status from Details card, add uptime to Overview

### LINKED ISSUE #893: VM Details - remove Status field from Details Card and move uptime to Overview Card
screenshot-localhost-3000-2018 12 13-12-31-36

~~On the details card, the **Status** field shows the VM's status and the VM uptime.  Currently the formatted uptime follows the output of the unix `uptime` command.  A more user friendly formatting is preferred.~~

The Details Card duplicates the **Status** field from the Overview Card.  The field should be removed and the uptime should be added to Overview's **Status** field.  A more user friendly formatting is preferred (as discussed in the comments below).

For reference, the formatting is done by the [`moment-duration-format` package](https://github.com/jsmreese/moment-duration-format).  Currently the format is (`src/utils/format.js`):

````js
  const formatted = moment.duration(durationInMS).format('d [days], h:mm')
````

### PR DESCRIPTION
**Pull request by sjd78**
- removed the status and uptime fields from the details card

 - added uptime to the overview card

 - updated the uptime format as per the issue

Resolves #893

### DIFF
--- a/src/utils/format.js
+++ b/src/utils/format.js
@@ -26,6 +26,9 @@ export function formatUptimeDuration (
     return defaultValue
   }
 
-  const formatted: string = moment.duration(durationInMS).format('d [days], h:mm')
+  // (up for 0 days, 0 hours, 1 minute, 37 seconds)
+  const formatted: string = moment.duration(durationInMS).format(
+    'd __, h __, m __, s __'
+  )
   return formatted
 }
--- a/src/utils/format.js
+++ b/src/utils/format.js
@@ -2,6 +2,8 @@
 import moment from 'moment'
 import 'moment-duration-format'
 
+const TWO_MINUTES = 2 * 60 * 1000
+
 export function formatUptimeDuration (
   {
     interval,
@@ -26,9 +28,10 @@ export function formatUptimeDuration (
     return defaultValue
   }
 
-  // (up for 0 days, 0 hours, 1 minute, 37 seconds)
   const formatted: string = moment.duration(durationInMS).format(
-    'd __, h __, m __, s __'
+    durationInMS < TWO_MINUTES
+      ? 'd __, h __, m __, s __'
+      : 'd __, h __, m __'
   )
   return formatted
 }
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/utils/format.js [hunk 1]
```diff
@@ -26,6 +26,9 @@ export function formatUptimeDuration (
     return defaultValue
   }
 
-  const formatted: string = moment.duration(durationInMS).format('d [days], h:mm')
+  // (up for 0 days, 0 hours, 1 minute, 37 seconds)
+  const formatted: string = moment.duration(durationInMS).format(
```
bond95: I think it is better to shut off seconds after 2 minutes as @gregsheremeta mentioned in issue https://github.com/oVirt/ovirt-web-ui/issues/893#issuecomment-447343729

sjd78: Fixed

leistnerova:
ci build please

leistnerova:
Status only once in overview and correctly formatted. But it shows wrong uptime when VM was suspended, it doesn't count the pause and starts from 0 again (admin portal shows in VM grid correct uptime).

sjd78:
@leistnerova - So looking at the webadmin code for the VM table, uptime column, and it uses "elapsed time" from the VM statistics object.  That field is not currently available for VM statistics via the REST API.  Easy enough to add there, so I'm doing that.

I'll tweak the REST API to pass back the extra statistic and then use that.

While that patch is being reviewed etc, I'll update this PR to look for it and fall back to the VM start time if the statistic isn't there.  The difference being that **elapsed time** doesn't include VM suspend time while **startTime** does (I assume a suspended VM is still considered "started").

sjd78:
(rebased)

sjd78:
The uptime string will now use the **elapsedUptime** field from VmStatistics, if it is available, and **startTime** from the VM if it is not.  Checks on the Details card that require the VM status field have also been fixed (by adding back the status const).

Note: **elapsed.time** was added to the REST API for VM statistics via https://gerrit.ovirt.org/#/c/96683/ and is transformed to **statistics.elapsedUptime**.

leistnerova:
ci build please

leistnerova:
LGTM

Issue #898, opened 2019-01-05
adjust padding in overview card


gregsheremeta:
In the overview card, we need some minor padding adjustments to the uptime label. Don't have it padded left 15px, and add 5px padding bottom.

Before:

![selection_288](https://user-images.githubusercontent.com/2248293/50725382-d84eec80-10ca-11e9-92a2-049751c10a2c.png)
![selection_289](https://user-images.githubusercontent.com/2248293/50725387-e00e9100-10ca-11e9-8cea-65d39cac146d.png)

-------

After:
![selection_290](https://user-images.githubusercontent.com/2248293/50725391-ee5cad00-10ca-11e9-970f-80488d3209fd.png)

![selection_291](https://user-images.githubusercontent.com/2248293/50725394-f1f03400-10ca-11e9-9aa4-00dcea37de5b.png)

PR #906, opened 2019-01-08
Overview card - move uptime in the block with status

sjd78:
The uptime was rendered below the spacing between the status
and description fields.  This caused the uptime to look odd
being closer to the VM description rather then the status.  Moving
the uptime into the same render div as the status looks better.

Fixes: #898

Before:
screenshot-localhost-3000-2019 01 08-13-15-25

After:
screenshot-localhost-3000-2019 01 08-13-14-39


gregsheremeta:
@lwrigh @sjd78 what do you think about indent on uptime? I have no strong preference, so mostly up to @lwrigh 

Scott's coded version:
screenshot-localhost-3000-2019 01 08-13-14-39

My mockups:

selection_290

selection_291

PR #907, opened 2019-01-08
Disable legacy view support for VMs, simplify view for Pool details

### LINKED ISSUE #912: remove legacy edit page for non-pool VMs
remove legacy edit page for non-pool VMs since they use the dashboard now.

[Pool VMs will eventually also use a modified dashboard, but that is some months away.]

sjd78:
- Removed the routes for `/vm-legacy` and `/vm/{id}/edit` since both functions are now covered by `/vm`

 - Removed the pages that supported the Legacy VM details and edit

 - Adjusted the VM card list to open the `/vm` VM dashboard details page directly

 - Adjusted the actions to only support the current dashboard details page and to consistently render spacing between buttons

 - Since VM dashboard doesn't have a list navigation, the list navigation on the legacy VM details page still used for Pool details has been removed.

 - Deleted the `VmsListNavigation` component since it is no longer used.

Fixes #912


leistnerova:
ci build please

leistnerova:
LGTM

Issue #889, opened 2018-12-10
Wrong units in disk utilization tab


leistnerova:
1. 1024M -> 1G
When free space is almost 1G it shows 1024M instead
![screenshot from 2018-12-06 18-24-19](https://user-images.githubusercontent.com/29575480/49639528-894e6280-fa0b-11e8-994b-3a5722dc7629.png)

2. null
When the allocated space is so small, there is 'null' instead of 'TiB' in donut
![screenshot from 2018-12-06 18-30-07](https://user-images.githubusercontent.com/29575480/49639552-98351500-fa0b-11e8-9ebc-68b09c50e142.png)

_Originally posted by @leistnerova in https://github.com/oVirt/ovirt-web-ui/pull/855#issuecomment-445181417_

PR #914, opened 2019-01-09
Fix disk and memory chart units

bond95:
- Added precision for formating bytes, now if value of Disk or Memory will be closer to next unit more then 10 %, it will calculate next unit.
- And if change round precision according to value, if value less then 10, it give sense to show two digit after point, if less then 100 give sense to show only one digit if greater, then without floating part.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/889

image
image

### DIFF
--- a/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/DiskCharts.js
@@ -13,7 +13,7 @@ import {
   DonutChart,
 } from 'patternfly-react'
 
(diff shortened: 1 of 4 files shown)


lwrigh:
Shouldn't the 'Disk' card include KiB as the units in the text above the donut?

bond95:
@lwrigh no, because this is unused memory, which is approximately 0.99 GiB.

leistnerova:
ci build please

leistnerova:
LGTM

Issue #803, opened 2018-10-04
Overview Card > change VM name validation and set cloud-init hostname change


sjd78:
In the legacy VM edit page, changing a VM's name also includes a valid name check and updates cloud-init hostname depending on how it is set in the VM's template:

https://github.com/oVirt/ovirt-web-ui/blob/aff035eff23496834d34a7402497212459f86458/src/components/VmDialog/index.js#L241-L258

The current VM Details Overview card will change the VM name, but does not include the name check and does not update the cloud-init hostname.  I assume both of these behaviors should be added to the Overview card.

gregsheremeta:
@sjd78 is this still relevant?

PR #917, opened 2019-01-10
Added name validation and cloud-init update to overview card

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/803

When name is invalid, save button is disabled:
image

If name is valid, appear checkbox and user can choose to update cloud-init hostname or not:
image

### DIFF
--- a/src/components/VmDetails/cards/OverviewCard/index.js
+++ b/src/components/VmDetails/cards/OverviewCard/index.js
@@ -10,7 +10,7 @@ import { formatUptimeDuration } from '_/utils'
 import { editVm } from '_/actions'
 
 import { Media } from 'react-bootstrap'
-import { FormControl, FormGroup, HelpBlock, Alert, Switch } from 'patternfly-react'
+import { FormControl, FormGroup, HelpBlock, Alert, Checkbox } from 'patternfly-react'
 
 import BaseCard from '../../BaseCard'
 import VmIcon from '../../../VmIcon'
@@ -245,17 +245,17 @@ class OverviewCard extends React.Component {
                     {
                       showCloudInitCheckbox &&
                       <div className={style['vm-checkbox']}>
-                        <Switch
-                          bsSize='mini'
+                        <Checkbox
                           id={`${idPrefix}-cloud-init`}
-                          value={updateCloudInit}
-                          onChange={(e, state) => this.setState({ updateCloudInit: state })}
+                          checked={updateCloudInit}
+                          onChange={(e) => this.setState({ updateCloudInit: e.target.checked })}
                           disabled={disableHostnameToggle}
-                        />
-                        { !disableHostnameToggle
-                          ? msg.updateCloudInit()
-                          : msg.cannotUpdateCloudInitHostname()
-                        }
+                        >
+                          { !disableHostnameToggle
+                            ? msg.updateCloudInit()
+                            : msg.cannotUpdateCloudInitHostname()
+                          }
+                        </Checkbox>
                       </div>
                     }
                   </div>
--- a/src/components/VmDetails/cards/OverviewCard/index.js
+++ b/src/components/VmDetails/cards/OverviewCard/index.js
@@ -10,7 +10,7 @@ import { formatUptimeDuration } from '_/utils'
 import { editVm } from '_/actions'
 
 import { Media } from 'react-bootstrap'
-import { FormControl, FormGroup, HelpBlock, Alert, Checkbox } from 'patternfly-react'
+import { FormControl, FormGroup, HelpBlock, Alert, Switch } from 'patternfly-react'
 
 import BaseCard from '../../BaseCard'
 import VmIcon from '../../../VmIcon'
@@ -239,9 +239,15 @@ class OverviewCard extends React.Component {
                     }
                     {
                       showCloudInitCheckbox &&
-                      <Checkbox className={style['vm-checkbox']} checked={updateCloudInit} onChange={e => this.setState({ updateCloudInit: e.target.checked })}>
-                        { msg.updateCloudInit() }
-                      </Checkbox>
+                      <div className={style['vm-checkbox']}>
+                        <Switch
+                          bsSize='mini'
+                          id={`${idPrefix}-cloud-init`}
+                          value={updateCloudInit}
+                          onChange={(e, state) => this.setState({ updateCloudInit: state })}
+                        />
+                        {msg.updateCloudInit()}
+                      </div>
                     }
                   </div>
 
--- a/src/components/VmDetails/cards/OverviewCard/index.js
+++ b/src/components/VmDetails/cards/OverviewCard/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 
 import sharedStyle from '../../../sharedStyle.css'
-import { getOsHumanName, getVmIcon, isVmNameValid } from '_/components/utils'
+import { getOsHumanName, getVmIcon, isVmNameValid, isHostNameValid } from '_/components/utils'
 import { enumMsg, msg } from '_/intl'
 import { generateUnique } from '_/helpers'
 import { formatUptimeDuration } from '_/utils'
@@ -43,6 +43,7 @@ class OverviewCard extends React.Component {
 
       nameError: false,
       updateCloudInit: true,
+      disableHostnameToggle: false,
     }
     this.trackUpdates = {}
 
@@ -122,6 +123,10 @@ class OverviewCard extends React.Component {
         updates = updates.set('name', value)
         fieldUpdated = 'name'
         newState.nameError = !isVmNameValid(value)
+        newState.disableHostnameToggle = !isHostNameValid(value)
+        if (newState.disableHostnameToggle) {
+          newState.updateCloudInit = false
+        }
         break
 
       case 'description':
@@ -186,7 +191,7 @@ class OverviewCard extends React.Component {
 
   render () {
     const { vm, icons, operatingSystems, isEditable } = this.props
-    const { isEditing, correlatedMessages, nameError, updateCloudInit } = this.state
+    const { isEditing, correlatedMessages, nameError, updateCloudInit, disableHostnameToggle } = this.state
 
     const elapsedUptime = vm.getIn(['statistics', 'elapsedUptime', 'datum'], 0)
     const uptime = elapsedUptime <= 0
@@ -245,8 +250,12 @@ class OverviewCard extends React.Component {
                           id={`${idPrefix}-cloud-init`}
                           value={updateCloudInit}
                           onChange={(e, state) => this.setState({ updateCloudInit: state })}
+                          disabled={disableHostnameToggle}
                         />
-                        {msg.updateCloudInit()}
+                        { !disableHostnameToggle
+                          ? msg.updateCloudInit()
+                          : msg.cannotUpdateCloudInitHostname()
+                        }
                       </div>
                     }
                   </div>
--- a/src/components/VmDetails/cards/OverviewCard/index.js
+++ b/src/components/VmDetails/cards/OverviewCard/index.js
@@ -3,14 +3,14 @@ import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 
 import sharedStyle from '../../../sharedStyle.css'
-import { getOsHumanName, getVmIcon } from '_/components/utils'
-import { enumMsg } from '_/intl'
+import { getOsHumanName, getVmIcon, isVmNameValid } from '_/components/utils'
+import { enumMsg, msg } from '_/intl'
 import { generateUnique } from '_/helpers'
 import { formatUptimeDuration } from '_/utils'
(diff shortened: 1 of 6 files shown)


sjd78:
> One comment for when the name is invalid. I would change the text to, "Please enter a valid virtual machine name. Only lower-case and upper-case letters, numbers, and '_',"-','.' are allowed."

+1

> 
> For the cloudinit checkbox, I think this would be a better option to have when the user is first creating a VM. It feels a little awkward in the overview card. @sjd78 What do you think?

It is there because the app is coded to push the VM Name to the cloud-init hostname field (on the Details > Advanced section).  The practical effect of the chained edit is to have the OS's hostname changed by the cloud-init function the next time the VM is restarted.  In other words, the intent is to keep the VM's Name in sync with the VM's OS hostname.

So the message should convey the change and the checkbox lets the user skip it if they want.

VM Create is an entirely different beast...

src/components/VmDetails/cards/OverviewCard/index.js
```diff
@@ -88,16 +111,17 @@ class OverviewCard extends React.Component {
       this.props.onEditChange(true, true)
     }
 
+    const newState = { isDirty: true }
```
sjd78: I don't see `newState` used elsewhere.  Is it necessary or can the `isDirty: true` just be put back in the `setState` call below?

src/components/VmDetails/cards/OverviewCard/index.js
```diff
@@ -138,6 +169,11 @@ class OverviewCard extends React.Component {
       vmUpdates['description'] = stateVm.get('description')
     }
 
+    if (this.trackUpdates['name'] && updateCloudInit && this.isCloudInitHostnameUpdate()) {
+      vmUpdates['cloudInit'] = stateVm.get('cloudInit').toJS()
```
sjd78: Is the full structure necessary to partially update the cloudInit hostname?  Would it work with something like:
````js
vmUpdates['cloudInit'] = { hostName: stateVm.get('name') }
````

File: src/components/VmDetails/cards/OverviewCard/index.js [hunk 3]
```diff
@@ -185,12 +224,24 @@ class OverviewCard extends React.Component {
...
+                        { nameError &&
+                          <HelpBlock>
+                            {msg.pleaseEnterValidVmName()}
+                          </HelpBlock>
+                        }
+                      </FormGroup>
+                    }
+                    {
+                      showCloudInitCheckbox &&
+                      <Checkbox className={style['vm-checkbox']} checked={updateCloudInit} onChange={e => this.setState({ updateCloudInit: e.target.checked })}>
```
sjd78: Should the checkbox be a slider button instead like in the Details card?

sjd78:
For the checkbox ... @lwrigh , @bond95 

 - Should it be a slider button instead of the checkbox?

 - Should it be left-padded so it moves over under the VM name field a bit to help indicate it is related directly to the name field?

lwrigh:
> > One comment for when the name is invalid. I would change the text to, "Please enter a valid virtual machine name. Only lower-case and upper-case letters, numbers, and '_',"-','.' are allowed."
> 
> +1
> 
> > For the cloudinit checkbox, I think this would be a better option to have when the user is first creating a VM. It feels a little awkward in the overview card. @sjd78 What do you think?
> 
> It is there because the app is coded to push the VM Name to the cloud-init hostname field (on the Details > Advanced section). The practical effect of the chained edit is to have the OS's hostname changed by the cloud-init function the next time the VM is restarted. In other words, the intent is to keep the VM's Name in sync with the VM's OS hostname.
> 
> So the message should convey the change and the checkbox lets the user skip it if they want.
> 
> VM Create is an entirely different beast...

Ah okay that makes sense to me. In that case I guess the placement of the checkbox will do. 



> For the checkbox ... @lwrigh , @bond95
> 
> * Should it be a slider button instead of the checkbox?
> * Should it be left-padded so it moves over under the VM name field a bit to help indicate it is related directly to the name field?

+1 I think a toggle button would be more straight forward. We could have 'Yes' or 'No for the options. 
screen shot 2019-01-16 at 10 17 50 am

Also +1 for left padding it with the VM name field to make it feel more related.

src/components/VmDetails/cards/OverviewCard/index.js
bond95: Actually `isDirty: true` really can be putted in `setState`, but `newState` is used for adding name error.

src/components/VmDetails/cards/OverviewCard/index.js
bond95: It must have full structure, because it will update `cloudInit` only if cloud init is `enabled`, thus it better to leave it as it is.

bond95:
@sjd78 @lwrigh 
image

image

lwrigh:
> @sjd78 @lwrigh
> image
> 
> image

Would it possible to make the toggle smaller? It seems kind of large for the text right now.

michalskrivanek:
@bond95 
where did you come up with the valid characters?
also, this specifies guest's hostname, right? Then why do you say "valid virtual machine name"  - that's something completely different

bond95:
Made toggle smaller and added more strict hostname validation.

If hostname is not valid.
image

@michalskrivanek @lwrigh @sjd78  what do you think?

lwrigh:
> Made toggle smaller and added more strict hostname validation.
> 
> If hostname is not valid.
> image
> 
> @michalskrivanek @lwrigh @sjd78 what do you think?

If we could have the text all within one line that would be good. Maybe decrease the size of the type a little bit?

bond95:
@lwrigh font size 10px
image

lwrigh:
> @lwrigh font size 10px
> image

Can the button text size match the question text size? Thank you for making the adjustments!

gregsheremeta:
1) i way prefer the checkboxes. the toggle looks out of place, like "old" vm portal, no?
2) the language can be improved. It's a boolean, so it should not say "Do you want" ... it should lead with an imperative verb, so something like "synchronize X with Y"
3) feels like this whole thing belongs in a user settings dialog, but that's for another day

bond95:
> i way prefer the checkboxes. the toggle looks out of place, like "old" vm portal, no?

Agree, I prefer checkbox too.

bond95:
@lwrigh this is how it looks like with text same size:
image

lwrigh:
> 1. i way prefer the checkboxes. the toggle looks out of place, like "old" vm portal, no?
> 2. the language can be improved. It's a boolean, so it should not say "Do you want" ... it should lead with an imperative verb, so something like "synchronize X with Y"
> 3. feels like this whole thing belongs in a user settings dialog, but that's for another day

If we had the checkbox and the dialog read 'Sync Cloud-init hostname with VM name?' would that be a better option? 

+1 I agree, this feels more like something to include in a user settings location.

bond95:
> > feels like this whole thing belongs in a user settings dialog, but that's for another day
>
> +1 I agree, this feels more like something to include in a user settings location.

Actually, I think Overview card is the best place for this checkbox, because it doing two things:

1. Notify user that cloud init hostname will update
2. And add possibility to fast and easy disable hostname update, without settings dialog.

sjd78:
### Checkbox
Webadmin just does it without any notification.  Having the checkbox is a good step forward.  Defaulting it on a user settings page would be nice, but there isn't a good way (that I know of) to store the settings to be shared between browser instances.

For the checkbox (or slider), the text shouldn't wrap below at all.  Making the font size smaller just isn't a good solution since the German translation may still be long enough to wrap the text.  Better to make it work now rather then getting a lot of issues generated during I18N QE.

### Name Check
For the name check - the code is here and should match what webadmin does:
https://github.com/oVirt/ovirt-web-ui/blob/73916b004eb990a015727914ad1fb10ace1f8a78/src/components/utils.js#L50-L118

@michalskrivanek - I don't know off the top of my head if hostnames can be non-ascii characters (I know DNS can do it), specifically for cloud-init.

@bond95 - So maybe a secondary validation function for a valid cloud-init hostname and if the VM name is valid, but it isn't a valid hostname then show the checkbox disabled with text like "The new VM name is not a valid cloud-init based hostname."

src/components/VmDetails/cards/OverviewCard/index.js
sjd78: Ah ok, I didn't notice lines 125-129.  All good.

src/components/VmDetails/cards/OverviewCard/index.js
sjd78: It is ok to copy it forward the entire structure and just update the bit that needs it.  A bit of a deviation from the "update exactly only what is necessary to update", but that's ok.  No need to change.

src/components/VmDetails/cards/OverviewCard/index.js
```diff
@@ -88,16 +112,21 @@ class OverviewCard extends React.Component {
...
     let fieldUpdated
     switch (fieldName) {
       case 'name':
-        // TODO: add name validation?
         updates = updates.set('name', value)
         fieldUpdated = 'name'
+        newState.nameError = !isVmNameValid(value)
+        newState.disableHostnameToggle = !isHostNameValid(value)
+        if (newState.disableHostnameToggle) {
+          newState.updateCloudInit = false
```
sjd78: Cool already added the host name validation function (with its different requirements from the VM name).  Now I suggest still showing the checkbox (but disabled) and a different error message describing that the VM name cannot be used as the OS hostname.

@gregsheremeta , @lwrigh - WDYT?

gregsheremeta: makes sense

bond95:
image

WDYT? @gregsheremeta @sjd78 @lwrigh

gregsheremeta:
> image
> 
> WDYT? @gregsheremeta @sjd78 @lwrigh

1. I thought we agreed on a checkbox. If we didn't, I definitely want a checkbox :)

2. I'd change the message to: "host name cannot be synchronized with VM name because the VM name is not a valid host name"
(it always reads more clearly when we put the status / state before the reason. [also the word 'thus' is very uncommon in English -- we wouldn't ever really put that in a string :) ] )

bond95:
@gregsheremeta 
image

lwrigh:
> > image
> > WDYT? @gregsheremeta @sjd78 @lwrigh
> 
> @bond95
> 
> 1. I thought we agreed on a checkbox. If we didn't, I definitely want a checkbox :)
> 2. I'd change the message to: "host name cannot be synchronized with VM name because the VM name is not a valid host name"
>    (it always reads more clearly when we put the status / state before the reason. [also the word 'thus' is very uncommon in English -- we wouldn't ever really put that in a string :) ] )

+1 to both 1 & 2

gregsheremeta:
lgtm. @sjd78 ?

src/components/VmDetails/cards/OverviewCard/style.css
```diff
@@ -36,3 +40,22 @@
...
   white-space: pre-wrap;
 }
+
+.vm-checkbox {
+  font-size: 10px;
+  line-height: 20px;
+  padding-left: 10px;
+}
+
+.vm-checkbox :global(.bootstrap-switch) {
```
sjd78: Can remove the switch styling since it got swapped back to a checkbox.

src/components/utils.js
```diff
@@ -117,6 +117,15 @@ export function isVmNameValid (nameCandidate: string): boolean {
   return regexp.test(nameCandidate)
 }
 
+export function isHostNameValid (nameCandidate: string): boolean {
+  const asciiLetters = 'a-zA-Z'
+  const numbers = '0-9'
+  const specialCharacters = '-'
+  const regexpString = '^([' + asciiLetters + numbers + specialCharacters + '])+$'
+  const regexp = new RegExp(regexpString)
+  return regexp.test(nameCandidate)
```
sjd78: For completeness, should include a few extra rules (https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_hostnames):
  - 64 character limit
  - cannot start or end with a "-"

sjd78: I think you need: `^[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?$`

Eg. the following should match:
  - `a`
  - `a1`
  - `a-1`

gregsheremeta:
> add a few unit tests and then we won't have to guess

+1

sjd78:
@bond95 - I updated the CSS to remove the unused switch css classes.  Add a unit test for the hostname helper function and make sure it works/fails for a reasonable set of names.

src/components/utils.js
gregsheremeta: +1

leistnerova:
ci build please

samuelmacko:
Everything works fine and PR can be merged.
I have a few small nitpicks:
1. if a new VM name is not a valid host name, we could maybe tell the user why it is not valid, or what criteria there are for a valid host name.
2. criteria label for a valid VM name doesn't tell us how long can it be.

sjd78:
@bond95 - Needs a rebase and then looks like it's all set.

PR #925, opened 2019-01-17
adjust handling of src/version.js.in so src/version.js can be dropped

sjd78:
- Update TravisCI config to `autogen.sh` files prior to CI

  - Drop and ignore `src/version.js` from git since it is a generate file

  - Make 'src/version.js' a required file for the build scripts to ensure
    it gets generated before `start` and `build` scripts will run

  - Added `pretest` script to check for required files, using the same
    function used by `start` and `build`, before testing

  - Update project scripts to run `autogen.sh` as necessary


sjd78:
@bond95 , @gregsheremeta .. FYI

gregsheremeta:
failing test? weird

+1 to git ignoring that version file

bond95:
@sjd78 is `version.js` really generating? Because as see Travis failed because didn't find this file.

sjd78:
> @sjd78 is `version.js` really generating? Because as see Travis failed because didn't find this file.

It does, but I had to update the travis config to call `autogen.sh` prior to calling `yarn run build` so everything is there all nice and neat.

sjd78:
CI works again!  Yay.

sjd78:
So I requested the review just so you guys see `src/version.js` need to be generated for tests etc to work, which means running `autogen.sh` before running the yarn build.  It is a minor change but saves having to keep `version.js` current even though it gets auto generated...

gregsheremeta:
seems fine to me. Any build issues make me think about docker -- can you check that locally? `sudo docker build .`  ...  and then you can run it via the hash name

bond95:
@gregsheremeta @sjd78 maybe better to add `autogen.sh` directly to `yarn run build` command?
Something like:
`"build": "./autogen.sh --prefix=/usr --datarootdir=/share && node scripts/build.js",`

sjd78:
> @gregsheremeta @sjd78 maybe better to add `autogen.sh` directly to `yarn run build` command?
> Something like:
> `"build": "./autogen.sh --prefix=/usr --datarootdir=/share && node scripts/build.js",`

Humm. Maybe, but it doesn't need to be run every time.  I'll push another PR that is just the version update and will update this PR to just be tweaking to handle `src/version.js.in` in a better way so `src/version.js` can be dropped.

gregsheremeta:
what's the status of this one?

sjd78:
> what's the status of this one?

Got it working well now.

sjd78:
ci build please

DEVELOPERS.md
```diff
@@ -2,73 +2,118 @@
...
-    ENGINE_URL=https://my.ovirt.instance/ovirt-engine/ yarn start
+    source /usr/share/ovirt-engine-nodejs-modules/setup-env.sh
+    ./autogen.sh --prefix=/usr --datarootdir=/share
+    make rpm
 
-When asked, provide a valid username (in the form of `user@domain`) and password so
-the application can start in the context of a logged in user.
+  2. Use [mock_runner](https://ovirt-infra-docs.readthedocs.io/en/latest/CI/Using_mock_runner/index.html)
+     to run CI build artifacts locally (this method is cleanest since it runs in a chroot)
+  3. Post the comment "`ci please build`" to the GitHub Pull Request for an on-demand
```
gregsheremeta: worth noting the rpms are auto-generated on push after #950 is merged ?

sjd78: >   4. With [PR#950](https://github.com/oVirt/ovirt-web-ui/pull/950) merged, each pull request push will automatically have RPMs built (see `automation/build.sh`)

@gregsheremeta  - Sound ok?

gregsheremeta: I would leave out "With PR#950 merged" since it already is

sjd78: Reworded just to reference oVirt infra STDCI and swapped order of 3 & 4

sjd78:
rebased

Issue #900, opened 2019-01-05
VM Console single sign on modal enhancements


gregsheremeta:
RFE: VM Console SSO modal enhancements

(1) When demoing the new console single sign on modal, we found it too intrusive. Every time a user accesses a console, the popup shows (if the guest agent isn't installed, for example [and not every vm will get a guest agent]). There is nothing a user can really do except install the guest agent, or if it's already installed, contact an admin. So it should not show every time the user clicks the `Console` button.

(2) The language is currently:
`SSO logon to the VM's console failed. Do you want to continue without SSO?`
(2a) `SSO` isn't user-friendly language.
(2b) The only user impact is that user now has to log in to vm with username and password. The primary action is Connect, and not getting single sign on is probably minor, so the modal shouldn't ask -- it should simply inform with a secondary to action to back out, although user probably never would.

So:

1. Show the modal only once per VM (maybe per VM per session) -- so use local storage or session-scoped variable or something. (@michalskrivanek preference on once per VM ever or per VM per login session?)

2. Tweak the language a bit. `Single sign on failed. Please make sure the guest agent is installed on your virtual machine. Contact your administrator if the problem persists.`  and buttons would be `Connect` (primary) and `Cancel` (secondary).

gregsheremeta:
@lwrigh what do you think?

lwrigh:
![sso error](https://user-images.githubusercontent.com/20227374/50786295-23126500-1281-11e9-9d24-99899413b1e4.png)
Is this in alignment with what you were thinking of? @gregsheremeta

gregsheremeta:
@lwrigh oops I was unclear -- this language replaces the language in the dialog from #844 (but that screenshot is old -- the dialog currently says  
```
cantLogonToConsole: 'SSO logon to the VM\'s console failed. Do you want to continue without SSO?',
```
)

We noticed it during the demo that it was a bit off, experience wise.

So, this dialog appears when user clicks console. It's a different single sign on than you're thinking. Homonyms are terrible lol :) This one is "single sign me on into my vm console when i click the console button" (vs. sso me into the instance of ovirt)

gregsheremeta:
(in other words, I wouldn't have to give a windows vm my password on the windows login screen in the vm console)

lwrigh:
Ah okay that makes sense to me. I think the text for that modal and button names work for me then.

PR #929, opened 2019-01-22
Fix everytime showed console SSO modal and update modal message

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/900

image


lwrigh:
I would edit the text to say, "Single sign on failed. Please check to see if the guest agent is installed on your virtual machine. Contact your administrator if the problem continues.".

gregsheremeta:
I like Laura's text update.

Also, that button should say "Connect" instead of "Yes"  ('yes' doesn't make sense in context, since there is no question being asked)

gregsheremeta:
+1 on the language and button, thanks!

michalskrivanek:
"running" instead of "installed"?

lwrigh:
> "running" instead of "installed"?

+1 Running makes sense to me.

gregsheremeta:
> > "running" instead of "installed"?
> 
> +1 Running makes sense to me.

indeed, "running" is more correct

bond95:
@lwrigh @gregsheremeta @michalskrivanek Done.

src/saga/console/index.js
```diff
@@ -103,12 +104,18 @@ export function* getConsoleInUse (action) {
...
   if (consoleUsers.length > 0 && consoleUsers.find(user => user.id === userId) === undefined) {
     yield put(setConsoleInUse({ vmId, consoleInUse: true }))
   } else {
     yield put(setConsoleInUse({ vmId, consoleInUse: false }))
-    yield put(downloadConsole({ vmId, usbFilter, hasGuestAgent }))
+    yield put(downloadConsole({
+      vmId,
+      usbFilter,
+      hasGuestAgent,
+      force: sessionsInternal.find(s => s.user.id === userId) !== undefined,
```
sjd78: Since SPICE and VNC are always gonna have `consoleUser === true`, why not search `consoleUsers`?

And again, `force` is needed here because you want to skip SSO if the current user is already connected to a session right?

Maybe renaming `force` to `skipSSO` would make it a little easier to read.  WDYT?

bond95: > Since SPICE and VNC are always gonna have consoleUser === true, why not search consoleUsers?

I don't quite understand what do you mean?

> Maybe renaming force to skipSSO would make it a little easier to read. WDYT?

Yes, it will be more clear

gregsheremeta: @sjd78 @bond95 re `consoleUsers` all set or still need to do something?

src/components/VmActions/ConsoleConfirmationModal.js
```diff
@@ -92,6 +92,14 @@ export default connect(
...
       dispatch(setConsoleLogon({ vmId: vm.get('id'), isLogon: null }))
     },
-    onDownloadConsole: ({ usbFilter, force }) => dispatch(downloadConsole({ vmId: vm.get('id'), usbFilter, consoleId, hasGuestAgent: vm.get('ssoGuestAgent'), force })),
+    onDownloadConsole: ({ usbFilter, skipSSO, userId }) => dispatch(
+      downloadConsole({
+        vmId: vm.get('id'),
+        usbFilter,
+        consoleId,
+        hasGuestAgent: vm.get('ssoGuestAgent'),
+        skipSSO: skipSSO || vm.get('sessions').find(s => s.getIn(['user', 'id']) === userId) !== undefined, // Parameter frr skiping SSO authorization
```
sjd78: Can you extract the `vm.get('sessions').find(s => s.getIn(['user', 'id']) === userId) !== undefined` check to a utility function and then reuse it in `consoleInUse()` in `console/index.js`?

I recommend a name like `doesVmSessionExistForUserId` or similar.

gregsheremeta: +1 for readable names on utilities :)

gregsheremeta:
look good @sjd78 ?

samuelmacko:
ci build please

leistnerova:
ci build please

bond95:
ci build please

samuelmacko:
lgtm

Issue #902, opened 2019-01-05
host and FQDN fields wrap on dash, can overlap next column over


gregsheremeta:
host and FQDN fields wrap on dash, can overlap next column over

![selection_295](https://user-images.githubusercontent.com/2248293/50726238-4948d100-10d8-11e9-932a-ef460f221d54.png)

They probably shouldn't wrap on `-`
and perhaps they should ellipsis `...` with tooltip?

PR #940, opened 2019-01-29
Remove wrap for hostname and FQDN, add ellipsis with tooltip

bond95:
Fix host and FQDN fields overlapping other things by truncating and using ellipsis + tooltip

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/902

image

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -214,6 +214,10 @@ class DetailsCard extends React.Component {
     this.hotPlugNow = true
     this.nextRunUpdates = {}
     this.restartAfterSave = false
+    this.hostOverflow = false
+    this.fqdnOverflow = false
+    this.hostRef = React.createRef()
+    this.fqdnRef = React.createRef()
 
     this.handleCardOnStartEdit = this.handleCardOnStartEdit.bind(this)
     this.handleChange = this.handleChange.bind(this)
@@ -228,6 +232,8 @@ class DetailsCard extends React.Component {
     this.handleHotPlugOnApplyNow = this.handleHotPlugOnApplyNow.bind(this)
 
     this.updateOs = this.updateOs.bind(this)
+    this.updateDimensions = this.updateDimensions.bind(this)
+    this.updateOverflow = this.updateOverflow.bind(this)
   }
 
   static getDerivedStateFromProps (props, state) {
@@ -252,6 +258,36 @@ class DetailsCard extends React.Component {
     }
 
     return null
+  }
+
+  getSnapshotBeforeUpdate () {
+    this.updateOverflow()
+
+    return null
+  }
+
+  updateOverflow () {
+    this.hostOverflow = false
+    this.fqdnOverflow = false
+    if (this.hostRef.current.offsetWidth < this.hostRef.current.scrollWidth) {
+      this.hostOverflow = true
+    }
+    if (this.fqdnRef.current.offsetWidth < this.fqdnRef.current.scrollWidth) {
+      this.fqdnOverflow = true
+    }
+  }
+
+  updateDimensions () {
+    this.updateOverflow()
+    this.forceUpdate()
+  }
+
+  componentDidMount () {
+    window.addEventListener('resize', this.updateDimensions)
+  }
+
+  componentWillUnmount () {
+    window.removeEventListener('resize', this.updateDimensions)
   }
 
   componentDidUpdate (prevProps, prevState) {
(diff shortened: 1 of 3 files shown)


bond95:
@lwrigh as I know for stuff like this we usually use that type of tooltip, like example tooltip in select 
image

gregsheremeta:
> @lwrigh as I know for stuff like this we usually use that type of tooltip, like example tooltip in select

To Laura's point, we do have a mix:

selection_326

The patternfly tooltips are just bootstrap tooltips:
https://getbootstrap.com/docs/3.4/javascript/#tooltips

The problem with them is that if you use too many of them, they eat browser memory (a problem we have in the admin portal, actually). I do think they look better, though.

So maybe we need a compromise or a set of rules on when to use them. Let's take that part of the discussion to another issue and that way we can proceed on this one :)

lwrigh:
> > @lwrigh as I know for stuff like this we usually use that type of tooltip, like example tooltip in select
> 
> To Laura's point, we do have a mix:
> 
> selection_326
> 
> The patternfly tooltips are just bootstrap tooltips:
> https://getbootstrap.com/docs/3.4/javascript/#tooltips
> 
> The problem with them is that if you use too many of them, they eat browser memory (a problem we have in the admin portal, actually). I do think they look better, though.
> 
> So maybe we need a compromise or a set of rules on when to use them. Let's take that part of the discussion to another issue and that way we can proceed on this one :)

That works for me. I think it might help to do an audit of where they are currently being used and figure out some standards for them from there. I think we should definitely just stick to one style type for consistency's sake.

gregsheremeta:
@leistnerova it looks good, but can you quickly test? I'm thinking about hostnames with a single dash and lots of non-breakable characters after that -- worried that it might break at the dash. (Or another character.)

bond95:
ci build please

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -694,7 +694,7 @@ class DetailsCard extends React.Component {
                 <Col className={style['fields-column']}>
                   <Grid>
                     <FieldRow label={msg.host()} id={`${idPrefix}-host`}>
-                      { hostName || <NotAvailable tooltip={msg.notAvailableUntilRunning()} id={`${idPrefix}-host-not-available`} /> }
+                      { <span className={style['hostname']} title={hostName}>{hostName}</span> || <NotAvailable tooltip={msg.notAvailableUntilRunning()} id={`${idPrefix}-host-not-available`} /> }
```
sjd78: I know the discussion about tooltip style (or bootstrap-react tooltip vs title) got pushed, but I'd rather see the patternfly-react.  Otherwise the hidden overflow with ellipsis looks good.

@gregsheremeta - `patternfly-react` technically uses `react-bootstrap` which re-implements the bootstrap components.  I'm not sure the comment about memory leaks like in webadmin is still relevant.  For example: [Tooltip.js](https://github.com/react-bootstrap/react-bootstrap/blob/v0.32.4/src/Tooltip.js) is exported as the patternfly-react Tooltip.

sjd78:
> @leistnerova it looks good, but can you quickly test? I'm thinking about hostnames with a single dash and lots of non-breakable characters after that -- worried that it might break at the dash. (Or another character.)

The style won't have the text break at all.  It'll hit the box width and get the overflow is hidden with the ellipse shown.  The kids these days have it easy I tell ya!

sjd78:
> The patternfly tooltips are just bootstrap tooltips:
> https://getbootstrap.com/docs/3.4/javascript/#tooltips
> 
> The problem with them is that if you use too many of them, they eat browser memory (a problem we have in the admin portal, actually). I do think they look better, though.
> 
> So maybe we need a compromise or a set of rules on when to use them. Let's take that part of the discussion to another issue and that way we can proceed on this one :)

@gregsheremeta - Just to reiterate from the code review comment, the `patternfly-react` Tooltips are the `react-bootstrap` Tooltips.  Those are native react implemented components reusing the bootstrap css.

leistnerova:
ci build please

leistnerova:
Truncating looks good. But tooltip is visible also when the host/fqdn is displayed entirely. I would remove it.

screenshot from 2019-02-07 10-49-13

bond95:
@leistnerova removed tooltip in case if it displayed entirely. Please, test.

leistnerova:
ci build please

leistnerova:
The title is not responsive. When I resize window or increase font, the text get truncated, but the title appears after refresh.

bond95:
> The title is not responsive. When I resize window or increase font, the text get truncated, but the title appears after refresh.

@leistnerova fixed

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -694,7 +730,7 @@ class DetailsCard extends React.Component {
                 <Col className={style['fields-column']}>
                   <Grid>
                     <FieldRow label={msg.host()} id={`${idPrefix}-host`}>
-                      { hostName || <NotAvailable tooltip={msg.notAvailableUntilRunning()} id={`${idPrefix}-host-not-available`} /> }
+                      { <span className={style['hostname']} title={this.hostOverflow ? hostName : ''} ref={this.hostRef}>{hostName}</span> || <NotAvailable tooltip={msg.notAvailableUntilRunning()} id={`${idPrefix}-host-not-available`} /> }
```
sjd78: If you extracted a component, this could look like:

````jsx
  <TooltipSpan className={style['hostname']} tooltip={hostName}>
    {hostname}
  </TooltipSpan>
````

And put all the ref, and overflow checking in `TooltipSpan` (or something similarly named)

gregsheremeta:
@leistnerova please recheck

leistnerova:
LGTM

Issue #882, opened 2018-11-28
VMs disk is created as Thin instead of Clone when template is optimized for server


leistnerova:
Steps to reproduce:
1. in admin portal create template from VM with disk
2. set Optimized for: Server on the template
3. log into VM portal, create new VM from that template

VM is created with <format>cow</format> disk but it should be raw.

By creating VM it should sent &lt;clone&gt;true&lt;/clone&gt; when template is optimized for server (according to admin portal functionality when it automatically changes Storage Allocation depending on 'Optimized for' value).

SIN3R6Y:
I would suspect the issue is here, https://github.com/oVirt/ovirt-web-ui/blob/5a694b8422cfbc3441a16e2dc9189aa8ab2acd6b/src/components/VmDialog/index.js#L204

IIRC if you don't specify whether to make a dependent or independent disk, it defaults to dependent when this should not be the case for "server" templates.

michalskrivanek:
yes, exactly
https://github.com/oVirt/ovirt-engine-api-model/blob/12ec843f5c4486cb878d38ecf47edcccf095ba61/src/main/java/services/VmsService.java#L228

PR #941, opened 2019-01-29
Fix 'cow' and 'raw' disk bug

bond95:
Add check if template is `server` type, in this case set `clone` parameter to `true` in create VM request.

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/882


leistnerova:
ci build please

leistnerova:
LGTM

PR #952, opened 2019-02-07
Allow users without edit VM permissions to change a running VM's CD

sjd78:
- Add 'canUserChangeCd' permit check to VM transform
    - To support a normal user's ability to change a VM's CD, a field has been added to a VM's transform: `canUserChangeCd`.  This is the basis for DetailsCard permission check to be able to edit
the VM.

  - Update DetailsCard to support 'Edit CD' and 'Full Edit' modes
    - If a user only has enough permissions to be able to change the CD, the card will be editable but, apart from the CD, the individual fields will not be.
    - Only when a user has full VM edit permissions will all of the fields be editable.

  - `editVm` saga to skip VM update if only CD change is needed
    - On the `DetailsCard`, it is quite possible the only change that needs to be committed is to change the CD.  In this case, the VM edit call can be skipped.  Additionally, when the user does not have access to edit the VM, but can change the CD, it is essential to skip the VM update call.
    - The `editVm` saga now looks at the given `vm` object and if only fields required to change the CD are present, the VM update is skipped.  The ability to specify if the change CD should be only temporary on the running VM or an update to the default CD on the VM itself was added.

----
Demo for a "UserRole" user with permissions assigned directly to the VM (shows the edit on an off VM then on an up VM):
ezgif com-video-to-gif

Demo for an Admin user:
ezgif com-video-to-gif 1

Fixes: #954

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -600,11 +600,9 @@ class DetailsCard extends React.Component {
     const { vm, isEditing, correlatedMessages, clusterList, isoList, templateList } = this.state
 
     const idPrefix = 'vmdetail-details'
+
     const isPoolVm = vm.getIn(['pool', 'id']) !== undefined
-
-    const canEditDetails = (vm.get('canUserChangeCd') || vm.get('canUserEditVm')) && !isPoolVm
-    const canOnlyChangeCD = (vm.get('canUserChangeCd') && !vm.get('canUserEditVm')) && !isPoolVm
-
+    const canEditDetails = vm.get('canUserEditVm') && !isPoolVm
     const status = vm.get('status')
 
     // Host Name
@@ -682,7 +680,7 @@ class DetailsCard extends React.Component {
       />
       <BaseCard
         title={msg.cardTitleDetails()}
-        editable={canEditDetails}
+        editable={canEditDetails || canChangeCd}
         editMode={isEditing}
         idPrefix={idPrefix}
         editTooltip={msg.cardTooltipEditDetails({ vmName: vm.get('name') })}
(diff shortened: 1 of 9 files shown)


sjd78:
Details card in Edit mode for an Admin user:
screenshot-localhost-3000-2019 02 07-16-06-26

Details card in Edit mode for a UserRole user:
screenshot-localhost-3000-2019 02 07-16-07-48

gregsheremeta:
looks good, works good. UI +1
@bond95  ?

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -550,7 +550,12 @@ class DetailsCard extends React.Component {
 
     // ---- dispatch the save
     this.setState({ correlationId })
-    this.props.saveChanges(vmUpdates, this.restartAfterSave, !this.hotPlugNow, correlationId)
+    this.props.saveChanges(
+      vmUpdates,
+      this.restartAfterSave,
+      !this.hotPlugNow,
+      isVmRunning,
```
bond95: Is this parameter really needed, because you can change CD only if VM is running, so for CD this parameter always will be `true`?

sjd78: While strictly true, change CD is only currently enabled for a running VM, the code made more sense to me with the param specified in there.  Putting the param made sense at the time.

sjd78: I'll take it out.  It's less confusing that way after I had some more coffee.

sjd78:
@bond95 - You made a good argument, I took out the extra param from `DetailsCard/index.js`.  Also rebased.  Please have another glace.  Thanks!

lwrigh:
+1 Everything looks good to me!

sjd78:
(rebased)

leistnerova:
User still can't change CD for VM from pool. Shouldn't that be available too?

gregsheremeta:
> User still can't change CD for VM from pool. Shouldn't that be available too?

Yep. Good catch!

sjd78:
@leistnerova - Users can now change CD on a Pool VM.

sjd78:
@leistnerova - The latest change I pushed should take care of the various Details/Change CD edit scenarios for admin and normal users.  Change CD is limited to if the VM is "Running" / Up.  Any other state and the VM will not be allowed to change the CD.

leistnerova:
LGTM

Issue #879, opened 2018-11-22
Localized Utilization cards and Snapshots


bond95:
Utilization cards are not localized.

gregsheremeta:
"Snapshot" appears to be missing too

![2019-01-29](https://user-images.githubusercontent.com/2248293/51946471-9fffad00-23ef-11e9-8a0d-8f8fe4bb3d1a.png)

PR #953, opened 2019-02-08
Externalize strings in Utilization and Snapshots Cards

sjd78:
Some string were still hard-coded in both the `SnapshotsCard` and the `UtilizationCard`.  They've been externalized.

Fixes: #879 

By generating a dummy locale based on default English, we can see the extent of what can be localized on the details card (or anywhere in the app really):
screenshot-localhost-3000-2019 02 08-02-26-51


sjd78:
@bond95 - rebased

PR #961, opened 2019-02-20
Added to dockerfile autogen.sh

bond95:
Quick fix


gregsheremeta:
Why? What was wrong? (Please use better descriptions in the future ...)

Issue #955, opened 2019-02-13
UserSessionTimeOutInterval key doesn't take effect on VM Portal


gregsheremeta:
Originally reported at https://bugzilla.redhat.com/show_bug.cgi?id=1667489

UserSessionTimeOutInterval for engine-config is used to set timeout interval in minutes, after which inactive user sessions expire. It doesn't take effect for VM Portal, but works perfectly for Admin Portal.

See also #598

sjd78:
> UserSessionTimeOutInterval for engine-config is used to set timeout interval in minutes, after which inactive user sessions expire. It doesn't take effect for VM Portal, but works perfectly for Admin Portal.

We use the same SSO token (with the same grant type and scope) for all of the REST calls.  So the background data update task cannot be differentiated from real user generated traffic.  The token will never expire like that.  The REST calls would need to have some kind of HTTP request header available to instruct SSO token validation not to update the last access date.  I have no idea if that is currently an option.

gregsheremeta:
> I have no idea if that is currently an option.

@noriravi what do you think?

noriravi:
Currently all restapi queries will update the last access date, we would need to modify the restapi code and backend code to achieve what is requested here. I would suggest opening a BZ.

gregsheremeta:
> Currently all restapi queries will update the last access date, we would need to modify the restapi code and backend code to achieve what is requested here. I would suggest opening a BZ.

Thanks! @sjd78 can you open the BZ and describe what we need? team = infra  Thanks!

gregsheremeta:
[ blocked by REST API enhancements needed ]

michalskrivanek:
@sjd78 are you sure about the same scope for all queries. Some time ago when iportal was timing out for everyone, iirc we made a differentiation for the refresh andnother calls. Or maybe we just wanted to and didn’t do it, i don’t quite recall

sjd78:
@michalskrivanek - As I understand how the SSO works, the scope and grant type are correlated to the SSO token at the time the token is created (so login box time).  The REST code all uses the same SSO token either from the dev engine pulling it or passed from the wrapping WAR.  I also can't find any instances where API calls are made with custom headers other then content types and "Filter".

So as long as the scope for a single SSO token is constant, then yes, all the requests use the same scope.

michalskrivanek:
[...]

PR #967, opened 2019-02-22
Upgrade patternfly-react to v2.29.15

bond95:
In this version of patternfly-react appear SessionTimeout component (http://patternfly-react.surge.sh/patternfly-3/index.html?knob-displayBefore=30&knob-timeLeft=30&knob-primary%20content=Your%20session%20is%20about%20to%20expire&knob-secondary%20content=to%20continue%20click%20on%20%22Continue%20Session%22.&selectedKind=patternfly-react%2FCommunication%2FSession%20Timeout&selectedStory=Session%20Timeout&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs), which I need for fixing this issue https://github.com/oVirt/ovirt-web-ui/issues/955

PR #968, opened 2019-02-22
Added check for inactivity during session and logout after expiration

bond95:
Fixes: #955

Required: https://github.com/oVirt/ovirt-web-ui/pull/967

image


src/components/TokenExpired.js
```diff
@@ -1,25 +1,91 @@
...
+    super(props)
+    this.state = {
+      expired: false,
+      counter: MAX_INACTIVE_TIME,
+    }
+
+    this.dropCounter = this.dropCounter.bind(this)
+    this.decrementCounter = this.decrementCounter.bind(this)
+
+    document.body.addEventListener('mousemove', this.dropCounter)
```
sjd78: Is moving the mouse over the document considered interacting with the App?  That is inconsistent with session timeout counters on banking apps etc.  I would prefer requiring actual interaction with the app, for example click buttons/links/navigation or keyboard navigation or keyboard input.

@gregsheremeta, @lwrigh Do you have/know of any guide lines from UXD group around this?

src/components/TokenExpired.js
```diff
@@ -1,25 +1,91 @@
...
+    this.timer = setInterval(this.decrementCounter, 1000)
```
sjd78: Shouldn't the component really mount before the timeout starts?

src/components/TokenExpired.js
```diff
@@ -1,25 +1,91 @@
...
+    this.setState(state)
+  }
+
+  componentWillUnmount () {
+    clearInterval(this.timer)
+  }
+
+  render () {
+    const { config, onLogout } = this.props
+    if (this.state.expired) {
```
sjd78: Nit-pick: So technically the inactivity timer has expired and now the "you have 30 seconds before auto-logout" timeout modal is displayed.  They have `MAX_INACTIVE_TIME` plus `TIME_TO_DISPLAY_MODAL`. :wink:

src/components/TokenExpired.js
```diff
@@ -1,25 +1,91 @@
...
 import { connect } from 'react-redux'
+import { SessionTimeout } from 'patternfly-react'
 
 import ErrorAlert from './ErrorAlert'
 import { msg } from '_/intl'
 
+import { logout } from '_/actions'
+
+const MAX_INACTIVE_TIME = 60 * 20 // 20 minutes
+const TIME_TO_DISPLAY_MODAL = 30 // 30 seconds
```
sjd78: These settings should be at least in `ovirt-web-ui.config` and `src/config.js`  so they can be set on the server independently of a code change.  Ideally, a rest call can pull the ovirt-engine configured user session timeout, maybe even the one from webadmin.

Nit-pick: Use the expected time unit in the constant's name.  E.g. `MAX_INACTIVE_TIME_SEC`

src/components/TokenExpired.js
bond95: The problem with clicking is that some event handlers can `stopPropagation`, so document handler never get event. Mouse move means that portal tab is now active (not somewhere in background), so I think it is better solution for now.

src/components/TokenExpired.js
bond95: Actually no, because mount cycle in React has `constructor` and `componentDidMount`, timer is not depend on inner content of component, so it doesn't really matter where to have it.

src/components/TokenExpired.js
bond95: No, modal is displayed, when only `TIME_TO_DISPLAY_MODAL` seconds left. So `MAX_INACTIVE_TIME` is time to user logout. `TIME_TO_DISPLAY_MODAL` is time when modal must be shown.

bond95:
@sjd78 I think all token/session expiration stuff must be in one component, so to do it more clear, I renamed component from `TokenExpired` to `TokenExpiredTracker`.

sjd78:
> @sjd78 I think all token/session expiration stuff must be in one component, so to do it more clear, I renamed component from `TokenExpired` to `TokenExpiredTracker`.

A token expiring on the backend is a different thing and has a different use case from a user activity tracker.  How they're handled in the UI should not be the same.  I would much rather have 2 simple components than 1 complex component that needs to handle both scenarios.

src/components/TokenExpiredTracker.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  componentWillUnmount () {
+    clearInterval(this.timer)
+  }
+
+  render () {
+    const { config, onLogout } = this.props
+    if (this.state.expired) {
+      return <SessionTimeout
```
sjd78: Did you consider using `CountDownSessionTimeout`?

src/intl/messages.js
```diff
@@ -274,6 +282,7 @@ export const messages: { [messageId: string]: MessageType } = {
   },
   locked: 'Locked',
   logOut: 'Log out',
+  logOutIn30SecondsSecondary: 'You will be logged out in a 30 seconds.',
```
sjd78: "You will be logged out in 30 seconds."

But it would be nice if the number of seconds actually counted down

src/intl/messages.js
```diff
@@ -391,6 +400,10 @@ export const messages: { [messageId: string]: MessageType } = {
   secondDevice: 'Second Device',
   secondDeviceTooltip: 'Second device in order.',
   secondsShort: 's',
+  sessionExpired: {
+    message: 'Your session is about to expire',
```
sjd78: "Your session is about to timeout due to inactivity."

src/components/TokenExpiredTracker.js
```diff
@@ -0,0 +1,91 @@
...
+        timeLeft={this.state.counter}
+        displayBefore={TIME_TO_DISPLAY_MODAL}
+        continueFnc={() => this.setState({ expired: false, counter: AppConfiguration.maxUserInactiveTime })}
+        logoutFnc={onLogout}
+        primaryContent={<p className='lead'>{ msg.sessionExpired() }</p>}
+        secondaryContent={
+          <React.Fragment>
+            <p>{ msg.logOutIn30SecondsSecondary() }</p>
```
sjd78: Thoughts on ticking down the seconds on the modal instead of a static "you have 30 seconds (or less)" message?

src/components/TokenExpired.js
sjd78: Ok, I'd still like to hear from @lwrigh and @gregsheremeta about which approach makes sense from the UX side before we decide for certain.

gregsheremeta: yeah, I think clicks make more sense.

So for the things that stopPropagation, there's gotta be a hook somewhere?

package.json
```diff
@@ -77,11 +78,12 @@
...
     "intl-messageformat": "2.2.0",
     "jquery": "=3.2.1",
+    "lodash": "4.17.11",
     "moment": "2.22.2",
     "moment-duration-format": "2.2.2",
-    "patternfly": "3.53.1",
+    "patternfly": "3.58.0",
     "patternfly-bootstrap-combobox": "1.1.7",
-    "patternfly-react": "2.11.1",
+    "patternfly-react": "2.29.15",
```
sjd78: You can remove that first commit (https://github.com/oVirt/ovirt-web-ui/pull/968/commits/9d25eaa8618170c21933e3957baa855fd029447c)

src/components/TokenExpired.js
sjd78: As long as the component won't be constructed and sit around for awhile before getting mounted (for example while fetching the initial login data).

I mean even then it shouldn't matter much.  Even if the initial load takes 5 minutes and the inactivity timer is down from 30 minutes to 25, the user is probably going to start clicking once it opens.  Well that assumes a user will wait around for 5 minutes...

src/components/TokenExpired.js
sjd78: @bond95 , @gregsheremeta - I did a bit more poking around in non-react world (https://github.com/shawnmclean/Idle.js, https://github.com/kidh0/jquery.idle) to see what other people have done.  It's all similar for key events and mouse moves etc.  I'm still not fond of pure mouse motion tracking.

Other solutions attached listeners to `window` instead of `document.body`.  I'm not sure if they are subject to `e.preventDefault()` event propagation cancellation, but it is worth checking out.

src/ovirt-web-ui.config
```diff
@@ -4,5 +4,7 @@
   "applicationURL": "/ovirt-engine/web-ui",
 
   "consoleClientResourcesURL": "https://www.ovirt.org/documentation/admin-guide/virt/console-client-resources/",
-  "cockpitPort": "9090"
+  "cockpitPort": "9090",
+
+  "maxUserInactiveTimeInSeconds": 60 * 20
```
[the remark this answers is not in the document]
gregsheremeta: actually, we shouldn't hardcode 20 minutes. It should be read from the database, UserSessionTimeOutInterval

bond95:
> - Install your update on a regular engine install
> - Open webadmin
> - Open web-ui in another tab on the same instance
> - Wait for webadmin to log you out and bump you to the login screen
> - Switch back to the web-ui tab

I did as you suggested and got:
image

So the question is, does we really still need this component? Because as I see, in case of token expiration browser return auth form.

bond95:
Only after second click on `Cancel` I got this error:
image

sjd78:
> So the question is, does we really still need this component? Because as I see, in case of token expiration browser return auth form.

The auth form is unusable.  A user will need to go back to the SSO login (or in dev mode restart the server) to get a new valid token.

The app needs much better robust handling for when the token expires on the backend to detect that, stop future refresh updates (to prevent additional errors warning), show the notification (or maybe not) and then push the user to the SSO login page.  In dev mode, there is no SSO login page, so an error component or page would be nice (like `TokenExpired` :wink:).

src/components/TokenExpired.js
lwrigh: I think other products require the user to move their move across or scroll up and down the page to keep it active. From a UX perspective I would consider that to be a pretty standard practice for keeping a page active.

lwrigh:
> Fixes: #955
> 
> Required: #967
> 
> image

I would recommend changing the copy to read as: 

Title of modal: Session Expiration 

Subtitle: Your session is about to expire. 

Text: You will be logged out in 30 seconds if there is no further activity in your session. 

         To continue with your session, click on the 'Continue' button. 

Button options: Log out, Continue 

A note on the behavior, as a user I would expect the 30 second number to have a countdown behavior. So the number would countdown from 30 to 0 and then once it reaches 0 the session would be automatically logged out.

bond95:
@lwrigh updated text, except `Title of modal`, because there is no title.

 > The app needs much better robust handling for when the token expires on the backend to detect that, stop future refresh updates (to prevent additional errors warning), show the notification (or maybe not) and then push the user to the SSO login page. In dev mode, there is no SSO login page, so an error component or page would be nice (like TokenExpired wink).

@sjd78 I guess this is separate issue.

lwrigh:
> @lwrigh updated text, except `Title of modal`, because there is no title.
> 
> > The app needs much better robust handling for when the token expires on the backend to detect that, stop future refresh updates (to prevent additional errors warning), show the notification (or maybe not) and then push the user to the SSO login page. In dev mode, there is no SSO login page, so an error component or page would be nice (like TokenExpired wink).
> 
> @sjd78 I guess this is separate issue.

I would still add a title to the modal. It looks a little awkward without one.

bond95:
> I would still add a title to the modal. It looks a little awkward without one.

@lwrigh unfortunately PF-React `SessionTimeout` component has no prop for `title`, so question is if want copy `SessionTimeout` component to VM Portal, or not. Also as I see according to PF session timeout has no title https://www.patternfly.org/pattern-library/communication/session-timeout/ 

@gregsheremeta @sjd78 WDYT?

lwrigh:
> > I would still add a title to the modal. It looks a little awkward without one.
> 
> @lwrigh unfortunately PF-React `SessionTimeout` component has no prop for `title`, so question is if want copy `SessionTimeout` component to VM Portal, or not. Also as I see according to PF session timeout has no title https://www.patternfly.org/pattern-library/communication/session-timeout/
> 
> @gregsheremeta @sjd78 WDYT?

Okay in that case I would remove the gray bar that's on top of the modal since the session timeout PF component doesn't seem to have one.

bond95:
image

@lwrigh

lwrigh:
@bond95 Looks good to me!

sjd78:
> > The app needs much better robust handling for when the token expires on the backend to detect that, stop future refresh updates (to prevent additional errors warning), show the notification (or maybe not) and then push the user to the SSO login page. In dev mode, there is no SSO login page, so an error component or page would be nice (like `TokenExpired` :wink:).
> 
> @sjd78 I guess this is separate issue.

Keeping the SSO token expired handling separate is reasonable at this point.  In that case, how about put the `TokenExpired` component back to what it was and rename `TokenExpiredTracker` to `UserActivityTracker` or `SessionTimeoutTracker` or similar?

bond95:
> Keeping the SSO token expired handling separate is reasonable at this point. In that case, how about put the TokenExpired component back to what it was and rename TokenExpiredTracker to UserActivityTracker or SessionTimeoutTracker or similar?

I think make more sense to remove token expiration handling at all as it is useless and back when we do better handling https://github.com/oVirt/ovirt-web-ui/issues/991

samuelmacko:
ci build please

samuelmacko:
lgtm

Issue #965, opened 2019-02-20
add docker build and run to CI


gregsheremeta:
Feature enhancement request:

add docker build and run to CI. Several times we have changed build files / make / version scripts / etc. and the docker build has either failed or succeeded but then would fail to start.

CI should
1. build the docker container
2. make sure it starts and verify that up starting :3000 is open

PR #969, opened 2019-02-25
Added docker build to CI

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/965

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,30 @@
 language: node_js
+
+services:
+  - docker
+
 script:
   - ./autogen.sh --prefix=/usr --datarootdir=/share
   - yarn run build
+  - docker build --tag web-ui .
+  - docker run -d --rm -it -e ENGINE_URL=https://localhost/ovirt-engine -p 3000:3000 --name web-ui web-ui
+  - >
+    for i in {1..7}; do
+      docker logs web-ui
+      docker logs web-ui 2>&1  | grep -Pzl 'error'
+      if [ $? -eq 0 ]; then
+        echo 'oVirt VM Portal did not start up successfully'
+        exit 1
+      fi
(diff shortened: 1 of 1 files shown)


gregsheremeta:
Seems good. Did you test the failure path (test that it actually doesn't pass CI if something is wrong with the container)?

Also, what do you think about https://docs.docker.com/engine/reference/builder/#healthcheck ?

.travis.yml
```diff
@@ -1,6 +1,13 @@
 language: node_js
+
+services:
+  - docker
+
 script:
+  - docker build --tag web-ui .
```
sjd78: does it matter that the docker stuff is around the `yarn run build`?  seems like they're a bit independent, assuming the `docker build` includes running the `yarn build` within the docker container.

gregsheremeta: I think the container takes a bit to start, so the build is acting as a hacky delay. I didn't love that either, so that's why I asked about healthcheck

bond95:
> Also, what do you think about https://docs.docker.com/engine/reference/builder/#healthcheck ?

In this case is better to use `for` cycle which check logs for `errors` and if there is any it will end tests with fail.

gregsheremeta:
Why is a for loop better?

bond95:
> Why is a for loop better?

Because: 
 1. If there will be some `autogen` failure container will continue work
 2. VM Portal container will fully running only in case if there will be some engine, which we don't have in test environment, so it will stuck on authentication, so it is better to check logs outside docker system then in it.

gregsheremeta:
> > Why is a for loop better?
> 
> Because:
> 
> 1. If there will be some `autogen` failure container will continue work
> 2. VM Portal container will fully running only in case if there will be some engine, which we don't have in test environment, so it will stuck on authentication, so it is better to check logs outside docker system then in it.

+1

lgtm

Issue #973, opened 2019-02-27
Disable editing VM template


leistnerova:
Description of problem [include screenshots and browser console logs]:
Template can't be changed for created VM. But in VM detail it could be and then is error returned.

VM Portal version number: master

oVirt version number:
ovirt-engine-4.3.1.2-0.0.master.20190222171227.gita00fd1d.el7.noarch

Browser and version: FF 65.0.1

![screenshot from 2019-02-27 09-05-43](https://user-images.githubusercontent.com/29575480/53474951-eb4cc000-3a6e-11e9-92f7-1f0c70ce1bac.png)

sjd78:
From my trail-and-error testing:
  - Cannot change a VM's template
  - Cannot change a Pool VM's template
  - CAN change a Pool VM definition's template

It isn't allowed in webadmin and throws a rest error in web-ui.

To fix right now, I'll disable the UI from showing an editor on VMs.  I may leave some of the change code in the file (maybe a comment?) in case that Pool definition editing is implemented.  The code can be reused then.

PR #984, opened 2019-03-06
Remove editability of a VM's template

sjd78:
- Template cannot be changed by the user

  - Template change processing has not been removed for 2 reasons:

    1. Changing a VM's cluster may require changing the template if
       the template is not in the new cluster

    2. In future, we may want to allow editing the template of a Pool's
       VM definition and having the change handler code there will make
       that task easier

Fixes: #973 

Details card in edit mode for a stopped template created VM:
screenshot-localhost-3000-2019 03 06-15-23-43

Details card in edit mode for a running blank template created VM:
screenshot-localhost-3000-2019 03 06-15-25-20

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -90,26 +90,9 @@ function createClusterList (clusters) {
 }
 
 /*
- * Return a normalized and sorted list of templates ready for use in a __SelectBox__ from
- * the Map of provided templates.
- */
-function createTemplateList (templates) {
-  const templateList =
-    templates
-      .toList()
-      .map(template => ({
-        id: template.get('id'),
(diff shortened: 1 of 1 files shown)


sjd78:
> Maybe it is better to remove template from `handleChange`, and return it when we need this, because for now it has no purpose and we still have it in git.

The `handleChange` code does still use the template change in a change cluster scenario:
https://github.com/oVirt/ovirt-web-ui/blob/6e63204f01933bd7ab26560d20bc3a705b0247aa/src/components/VmDetails/cards/DetailsCard/index.js#L315-L326

bond95:
> The handleChange code does still use the template change in a change cluster scenario:

Oh, ok, I see.

michalskrivanek:
"edibility"?:)

leistnerova:
LGTM

Issue #981, opened 2019-02-27
Pressing Enter in editing vNIC name, reloads the page


leistnerova:
Description of problem [include screenshots and browser console logs]:
When you press Enter key while editing nic and you are in the name field with your cursor, the website refreshes.

VM Portal version number:
master
oVirt version number:
ovirt-engine-4.3.1.2-0.0.master.20190222171227.gita00fd1d.el7.noarch
Browser and version: FF 65.0.1

Steps to Reproduce:
1. try to create new vNIC
2. focus in name input
3. press Enter

Actual results: page is reloaded

Expected results: dialog stays

Additional info: Editing disk alias or snapshot description doesn't do that.

PR #993, opened 2019-03-28
Add form onSubmit handler to NicEditor, DiskImageEditor

sjd78:
There was behavior on the `NicEditor` form would submit when 'Enter'
was presses when the name field had focus.  This caused the page to
completely reload.  The form's new `onSubmit` handler cancels the
default behavior and prevents the page reload.

Fixes: #981

### DIFF
--- a/src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
+++ b/src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
@@ -236,7 +236,11 @@ class DiskImageEditor extends Component {
         </Modal.Header>
         <Modal.Body>
 
-          <Form horizontal>
+          <Form
+            horizontal
+            onSubmit={e => { e.preventDefault() }}
+            id={`${idPrefix}-modal-form`}
+          >
             {/* Alias */}
             <FormGroup controlId={`${idPrefix}-alias`} className={createMode && 'required'}>
               <LabelCol sm={3}>
--- a/src/components/VmDetails/cards/NicsCard/NicEditor.js
+++ b/src/components/VmDetails/cards/NicsCard/NicEditor.js
@@ -127,8 +127,8 @@ class NicEditor extends Component {
     return nic
   }
 
-  changeName (event) {
-    this.setState({ name: event.target.value })
+  changeName ({ target: { value } }) {
+    this.setState({ name: value })
   }
 
   changeVnicProfile (value) {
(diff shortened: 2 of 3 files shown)


src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -236,7 +236,11 @@ class DiskImageEditor extends Component {
         </Modal.Header>
         <Modal.Body>
 
-          <Form horizontal>
+          <Form
+            horizontal
+            onSubmit={e => { e.preventDefault() }}
```
bond95: I think this `{}` brackets are unnecessary in this case and can be removed.

src/components/VmDetails/cards/NicsCard/NicEditor.js
```diff
@@ -190,7 +190,11 @@ class NicEditor extends Component {
         </Modal.Header>
         <Modal.Body>
 
-          <Form horizontal>
+          <Form
+            horizontal
+            onSubmit={e => { e.preventDefault() }}
```
bond95: Same

src/components/VmDetails/cards/NicsCard/NicEditor.js
```diff
@@ -190,7 +190,11 @@ class NicEditor extends Component {
...
+            id={`${modalId}-form`}
```
bond95: For keeping consistency here must be `id={${modalId}-modal-form}` too

sjd78: https://github.com/oVirt/ovirt-web-ui/blob/b28c641e7bb6c5baa98935baa97f0f6e8109a65b/src/components/VmDetails/cards/NicsCard/NicEditor.js#L152-L154

It already is `props.id + '-modal' + '-form'`

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
sjd78: Technically correct, but I prefer the esthetic of including the block and that it implies that the return value is inconsequential.

src/components/VmDetails/cards/NicsCard/NicEditor.js
bond95: Oh, ok, I missed it.

samuelmacko:
ci build please

leistnerova:
ci build please

samuelmacko:
lgtm

Issue #18, opened 2016-10-04
VM quick search / selector / sort


mareklibra:
designs: https://docs.google.com/document/d/1knF5-UnYtM6a8pcPUmnzva2Q5JsBwyOxbgbIXWJ2NAo/edit

![selection_263](https://user-images.githubusercontent.com/2248293/50360507-3b1e8080-052e-11e9-8f00-46bb69860796.png)

jelkosz:
I have tried to use ovirt-web-ui on a larger setup with hundreds of VMs and getting to the VM I wanted to use was a pain (and after about 5 minutes of scroll and load next 8 vms when I was still far from mine I gave up and decided I don't really need to use that VM :) ). 
It would be hugely helpful to have a way how to make a server side filter to return only VMs where the name matches some pattern (as it can be done in webadmin).

Unfortunately, this is supported only for admin users (for a little context check this: https://bugzilla.redhat.com/show_bug.cgi?id=1534607), but it is still useful since mostly the admin users will have hundreds of VMs available so they will appreciate the most this feature.

What I propose is to have a searchbar on top of the screen for admin users. If I write something like tj-* into it, all the VM / pool related queries will only load VMs which match this pattern.

This searchbar will not be shown for regular users.

@mareklibra @jniederm what you guys think?

gregsheremeta:
designs: https://docs.google.com/document/d/1knF5-UnYtM6a8pcPUmnzva2Q5JsBwyOxbgbIXWJ2NAo/edit

ready for implementation

PR #997, opened 2019-04-04
Added Filters and Sort fields

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/18

- Added Filters and sort to toolbar 
- Rewrote Toolbar to patternfly-react component

image

image

image

@gregsheremeta @sjd78 @lwrigh @michalskrivanek


lwrigh:
Nice!! This is looking great!

bond95:
@sjd78 @lwrigh added saving state of filters and sort to local storage, please, review

lwrigh:
@bond95 Can you add an updated screenshot or a gif of it?

bond95:
@lwrigh actually nothing was changed in design, but I would like to know your opinion about sort reset, because now it saving sort state to local storage and there is no way to reset it, only change. So is there needed some button for reset or not?

lwrigh:
@bond95 For sort since there are only two options currently (alphabetical or reverse alphabetical order) and the VMs are automatically sorted in alphabetical order I do not see a need for a reset button since the user can just toggle the sort button between alphabetical and reverse alphabetical order. The user could always refresh the page if they would like to reset their sort options.

bond95:
> The user could always refresh the page if they would like to reset their sort options.

@lwrigh  No, they can't because sort is saving to local storage too.

lwrigh:
@bond95 Would there be any way to change that? It seems like a strange behavior if the user can't refresh the page to reset the sorting.

bond95:
@lwrigh yes, of course, I can remove it.

lwrigh:
@bond95 Great that sounds good to me. I think that would help to improve the behavior of it.

bond95:
> @bond95 Would there be any way to change that? It seems like a strange behavior if the user can't refresh the page to reset the sorting.

@lwrigh Done.
@sjd78  please, review

src/components/VmFilters/index.js
```diff
@@ -0,0 +1,207 @@
...
+        value={currentValue}
+        placeholder={currentFilterType.placeholder}
+        onChange={e => this.updateCurrentValue(e)}
+        onKeyPress={e => this.onValueKeyPress(e)}
+      />
+    )
+  }
+
+  componentDidUpdate () {
+    saveToLocalStorage('vmFilters', JSON.stringify(this.props.vms.get('filters').toJS()))
```
sjd78: I think this will have the problem of the filters / sorts being stored for the app and not a particular user of the app.  Login with "userA", set a filter, logout, login with "userB" and the original filter would still be applied.

sjd78:
@bond95 - the PR does need a small rebase

src/utils/filters.js
```diff
@@ -0,0 +1,23 @@
...
+  name: (value) => value,
+  os: (value) => value,
+  status: (value) => enumMsg('VmStatus', value),
+}
+
+export function filterVms (item, filters) {
+  let res = true
+  for (let name in filters) {
+    if (compareMap[name]) {
+      res &= compareMap[name](item, filters[name])
```
sjd78: Did you intend to use the bitwise AND here or just letting the value be coerced between boolean and numerics?

bond95: I'm not sure what you mean, but values returned by `compareMap` functions are boolean, so there is no coercing.

src/reducers/vms.js
```diff
@@ -26,22 +30,27 @@ import {
...
+import { loadFromLocalStorage } from '_/storage'
+
+const EMPTY_MAP = Immutable.fromJS({})
+const EMPTY_ARRAY = Immutable.fromJS([])
 
 const initialState = Immutable.fromJS({
   vms: {},
   pools: {},
+  filters: Immutable.fromJS(JSON.parse(loadFromLocalStorage('vmFilters')) || {}),
+  sort: Immutable.fromJS({ ...sortFields[0], isAsc: true }),
```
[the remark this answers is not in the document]
bond95: > Moving the loads from here to there would be consistent with the rest of the reducers.

I'm not sure that it will be consistent, because we didn't have similar case with using local storage. Also, without moving it to fetched data with no pre-reqs, we can avoid unnecessary dispatching action.

src/sagas/login.js
```diff
@@ -175,11 +179,15 @@ function* checkUserFilterPermissions () {
...
 function* initialLoad () {
+  const userId = yield select(state => state.config.getIn(['user', 'id']))
+  const filters = JSON.parse(loadFromLocalStorage(`vmFilters-${userId}`)) || {}
+
   // no data prerequisites
   yield all([
     call(fetchUserGroups, getUserGroups()),
     call(fetchAllOS, getAllOperatingSystems()),
     call(fetchAllHosts, getAllHosts()),
+    call(saveFilters, saveVmsFilters({ filters })),
```
sjd78: Thanks for moving it over here.  Would be nice to have a `loadFiltersFromLocalStorage` or similar saga function to just to keep it all in one place.

src/utils/filters.js
sjd78: `&=` is a bitwise operation not a boolean operation.  

The values of `res` on the way out will be true if no filters are applied, or numeric 0 or 1 if filters are applied.  It'll still be truthy correct, but it won't be a boolean value.

isaranova:
ci build please

isaranova:
ci build please

isaranova:
ci please test

isaranova:
ci build please

isaranova:
Sorting alphabetically by OS is not sorting correctly according to the OS labels on the VMs, probably connected to the second issue with the list of OS. 
pic1


List of OS is different from the one you choose in VM detail edit or VM create and is not sorted.
pic3
Screenshot_2019-07-24 VM Portal


List of status is not sorted.
pic4

isaranova:
The filter list with status contains one item called 'not_responding' which is not in the same format as other states.
Screenshot_2019-07-24 VM Portal(1)

The filter list with OS types containing three Linux and three Other OS, using each of them results into a different filter tag.
Screenshot_2019-07-24 VM Portal(2)
Screenshot_2019-07-24 VM Portal(4)
Screenshot_2019-07-24 VM Portal(5)

Also, you cannot have a filter with more OS types and states at once, but you can have more names at once. Not sure if this is by design or not.

Sorting VMs alphabetically by the OS type is still not correct.
Screenshot_2019-07-24 VM Portal(3)

isaranova:
LGTM

michalskrivanek:
@bond95 please change "Asleep" as well, that doesn't correspond to anything...should be called "Suspended"

TimmiORG:
Hi guys,
are you able to tell me in which version the enhancement will be in?
The current search in 1.6.0 is not really helpful as it is searching only from the beginning of the VM name. From the screen shots it looks like that this fix is supporting better search capabilities.
Best regards
Timmi

sjd78:
> Hi guys,
> are you able to tell me in which version the enhancement will be in?
> The current search in 1.6.0 is not really helpful as it is searching only from the beginning of the VM name. From the screen shots it looks like that this fix is supporting better search capabilities.
> Best regards
> Timmi

@TimmiORG - This change was merged and released in version 1.6.0.

TimmiORG:
> > Hi guys,
> > are you able to tell me in which version the enhancement will be in?
> > The current search in 1.6.0 is not really helpful as it is searching only from the beginning of the VM name. From the screen shots it looks like that this fix is supporting better search capabilities.
> > Best regards
> > Timmi
> 
> @TimmiORG - This change was merged and released in version 1.6.0.

Hi @sjd78,
then I guess this is not working as it should. 
From the design document I understood that the search string will checked at any position of the VM name. Currently it is looking only on the beginning of the VM name.
Example search for "int":
This will find the following:
intel
But not:
a0-int
Let me know if I should create a bug ticket for this one.
Best regards
Timmi

PR #990, opened 2019-03-21
Pool redesign

bond95:
Fixes: #195
Resolves: #989

image

image

image

image

PR #1001, opened 2019-04-08
Refactored VM and Pool cards

bond95:
Rewrote cards to `patternfly-react` components and added `BaseCard`.

@sjd78 WDYT?

### DIFF
--- a/src/components/VmsList/BaseCard.js
+++ b/src/components/VmsList/BaseCard.js
@@ -12,6 +12,8 @@ const ICON_NAME = 'Icon'
 const TITLE_NAME = 'Title'
 const STATUS_NAME = 'Status'
 
+const IdPrefixContext = React.createContext('')
+
 const BaseCardHeader = ({ children }) => (
   <div>
     {children}
@@ -20,6 +22,7 @@ const BaseCardHeader = ({ children }) => (
 
 BaseCardHeader.displayName = HEADER_NAME
 
+BaseCardHeader.contextType = IdPrefixContext
 BaseCardHeader.propTypes = {
   children: PropTypes.oneOfType([
     PropTypes.arrayOf(PropTypes.node),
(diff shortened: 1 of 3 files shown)


src/components/VmsList/BaseCard.js
```diff
@@ -0,0 +1,127 @@
...
+  const childs = { others: [] }
+  React.Children.forEach(children, child => {
+    if (names.includes(child.type.displayName)) {
+      childs[child.type.displayName] = child
+    } else {
+      childs['others'].push(child)
+    }
+  })
+  return (
+    <Col xs={12} sm={6} md={4} lg={3} id={`${idPrefix}-box`}>
```
sjd78: Does it still make sense to keep `Col` here or move it up to `Vms` so the card's layout in the broader grid is controlled by the grid itself?

src/components/VmsList/BaseCard.js
```diff
@@ -0,0 +1,127 @@
...
+}
+
+const names = [ HEADER_NAME, ICON_NAME, TITLE_NAME, STATUS_NAME ]
+
+/**
+ * Single icon-card in the list for a VM
+ */
+const BaseCard = ({ children, idPrefix }) => {
+  const childs = { others: [] }
+  React.Children.forEach(children, child => {
```
sjd78: This is pretty slick to collect the children so they can be rendered in the correct order regardless as to how they're specified by the containing component.  Nice!

src/components/VmsList/Pool.js
```diff
@@ -28,34 +27,17 @@ const Pool = ({ pool, icons, onStart }) => {
...
-        </div>
-      </div>
-    </div>
+    <BaseCard idPrefix={idPrefix}>
+      <BaseCard.Header>
+        <span className={sharedStyle['operating-system-label']}>{ osName }</span>
+      </BaseCard.Header>
+      <BaseCard.Icon url={`/pool/${pool.get('id')}`} icon={icon} />
+      <BaseCard.Title idPrefix={idPrefix} url={`/pool/${pool.get('id')}`} name={pool.get('name')} />
+      <BaseCard.Status idPrefix={idPrefix}>
```
sjd78: I've seen patternfly-react and react-bootstrap code use a context conatiner to pass things down like id prefixes.  It may be worth looking into it and having nested contexts automatically add id postfixes depending on how the components are nested.  Just a thought.

bond95: Nice idea, I'll look at it.

src/components/VmsList/BaseCard.js
bond95: Yes, it would be better to move it to `Vms`

src/components/VmsList/BaseCard.js
```diff
@@ -12,6 +12,8 @@ const ICON_NAME = 'Icon'
 const TITLE_NAME = 'Title'
 const STATUS_NAME = 'Status'
 
+const IdPrefixContext = React.createContext('')
```
sjd78: :+1:

src/components/VmsList/BaseCard.js
sjd78: Do you mean that since `Col` has the `` id={`${idPrefix}-box`} `` parameter that it's better to leave the `Col` here?

bond95: > Do you mean that since `Col` has the `` id={`${idPrefix}-box`} `` parameter that it's better to leave the `Col` here?

I mean, that `BaseCard` is component that responsible for creating card. Card this is not only structure (arrangement of single items), but also shared style. `Col` is container that responsible for width of card, which is style, so it must be in `BaseCard` component.

sjd78:
@bond95 - Should we merge this PR and then you can rebase the card changes from #990 on top of the refactoring?

bond95:
> @bond95 - Should we merge this PR and then you can rebase the card changes from #990 on top of the refactoring?

True, I'll merger this.

PR #1007, opened 2019-04-11
Fixed broken infinite scroll

bond95:
Infinite scroll was broken again because of desynchronization initial VM load with infinite scroll VM load. Initial load fetch VMs after infinite scroll and this broke whole system. So initial VM load was removed because it is useless.

### DIFF
--- a/src/sagas/login.js
+++ b/src/sagas/login.js
@@ -31,7 +31,6 @@ import {
   getAllStorageDomains,
   getAllTemplates,
   getAllVnicProfiles,
-  getByPage,
   getIsoFiles,
   getUserGroups,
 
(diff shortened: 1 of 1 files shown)


src/sagas/login.js
```diff
@@ -229,9 +228,6 @@ function* initialLoad () {
 
   // requires storage domains to be in redux store
   yield put(getIsoFiles())
-
-  // --- store is populated with user & system info, now pull VMs and Pools
-  yield put(getByPage({ page: 1 }))
```
sjd78: Maybe just leave a comment here...
````js
  // The `Vms` card view component will take care of loading pages of VMs and Pools as needed.
  // Loading VMs and Pools here is not necessary and will cause issues with `Vms`'s loading.
````

Issue #1009, opened 2019-04-15
Update login Api version check to limit to ovirt 4.2


sjd78:
A REST API version check is performed at login.  Currently, as long as the `ovirt-engine` is version 4.0 or higher, the check passes.  With `ovirt-web-ui` 1.5, a few matching changes were also made on the REST API.

Instead of attempting to support a 4.2 or lower API, I propose the version check is updated to require at least ovirt 4.3.  This will also allow us to refactor some sagas that attempt to maintain compatibility with older API versions (pre 4.2 follow requests as an example).

michalskrivanek:
@sjd78 I wouldn't kill 4.2, it's still widely used. We have all the API changes in latest 4.2.z...so why dropping 4.2? Any other reasons?

sjd78:
@michalskrivanek  - I'm ok with putting the lower limit at ovirt-engine 4.2 for master web-ui (v1.5).  However, I don't think master has been tested against 4.2.  The web-ui 1.4 branch is more specifically for 4.2, even if it is ugly.

Do you know if anyone has tried installing ovirt-web-ui 1.5, or using the docker container, on a 4.2.Z engine?  I haven't tried that yet.

PR #1010, opened 2019-04-15
Set minimum ovirt-engine REST API to version 4.2

sjd78:
Resolves: #1009 (initial check)
Additional work to refactor sagas can be done in future.


sjd78:
Changed the limit to 4.2 as per discussion with @michalskrivanek on #1009

sjd78:
@bond95 , @sgratch - Any issues with limiting master to >= 4.2 ovirt?

Issue #975, opened 2019-02-27
Checkbox is not compatible with check in removing VM


leistnerova:
Description of problem [include screenshots and browser console logs]:
Checkbox by Preserve disks is smaller than check that appears after clicking on it.

VM Portal version number:
master
oVirt version number:
ovirt-engine-4.3.1.2-0.0.master.20190222171227.gita00fd1d.el7.noarch
Browser and version: FF 65.0.1

![screenshot from 2019-02-27 09-14-20](https://user-images.githubusercontent.com/29575480/53475389-1a176600-3a70-11e9-8bfa-d72b847fb2ba.png)

PR #1011, opened 2019-04-18
Fixed checkbox size in Remove VM confirmation dialog

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/975

image


lwrigh:
Would we be able to change the 'Yes' button to a red destructive button that says 'Remove' since it's a destructive action?

bond95:
> Would we be able to change the 'Yes' button to a red destructive button that says 'Remove' since it's a destructive action?

Yes, but I think this is for another issue, because color of button is not related to this issue.

lwrigh:
> > Would we be able to change the 'Yes' button to a red destructive button that says 'Remove' since it's a destructive action?
> 
> Yes, but I think this is for another issue, because color of button is not related to this issue.

That works for me. I can open up an issue for the remove button.

bond95:
> > > Would we be able to change the 'Yes' button to a red destructive button that says 'Remove' since it's a destructive action?
> > 
> > 
> > Yes, but I think this is for another issue, because color of button is not related to this issue.
> 
> That works for me. I can open up an issue for the remove button.

It would be nice, thank you. =)

bond95:
> Thoughts on just killing this `Checkbox` and use only the one from patternfly-react in the remove VM confirm dialog? As far as I can tell, the remove VM confirm is the only place where this component is use web-ui.

Yes, remove confirmation is the only place, because we don't use anywhere else checkbox at all. But this is discussion for separate issue. This issue little bit not about removing checkbox =)

sjd78:
> > Thoughts on just killing this `Checkbox` and use only the one from patternfly-react in the remove VM confirm dialog? As far as I can tell, the remove VM confirm is the only place where this component is use web-ui.
> 
> Yes, remove confirmation is the only place, because we don't use anywhere else checkbox at all. But this is discussion for separate issue. This issue little bit not about removing checkbox =)

Why would it need a separate issue?  The issue is to fix the render of the checkbox right?  Remove the crap check box, use the patternfly-react one, kill the CSS tweak you had to make and done.  That is cleaner than tweaking a component that doesn't need to even exist anymore.

sjd78:
@bond95 - Any followup?

bond95:
Replaced checkbox @lwrigh @sjd78  
image

isaranova:
ci build please

leistnerova:
ci build please

isaranova:
LGTM

PR #1014, opened 2019-04-18
Refactor logout handling

sjd78:
This change will allow logout and session timeout handling to work
gracefully on both dev server and normal production deployment.  Summary
of changes:

  - The `NoLogin` component is styled based on the `GlobalErrorBoundary`

  - Added a key "applicationLogoutURL" to the config to make the logout
    URL an explicitly configured entity

  - The `logout` saga won't navigate away from the page when the logout
    URL isn't defined, leaving a dev server user on the "Please login"
    screen instead of a broken URL/page

  - The `logout` saga will cancel the background refresh tasks since
    they aren't needed when the user is logged out

  - `SessionActivityTracker` was updated to behave when it gets unloaded
    and loaded based on the App login status


sjd78:
This is what the dev mode logged out page looks like currently:
screenshot-localhost-3000-2019 04 18-19-23-32

lwrigh:
> This is what the dev mode logged out page looks like with the change:
> screenshot-localhost-3000-2019 04 18-19-23-32

I think it would be helpful to provide the user with a message explaining why they got logged out. It could be something along the lines of, 'Your session has timed out. Please log back in if you would like to continue.' + 'Log in' link.

sjd78:
@lwrigh - I should have mentioned it before ... this page is only displayed for any length of time to developers when they either manually "logout" or if the session activity tracker logs them off.  In a normal install situation, the page quite literally blinks by very quickly.  Right now it is a gray box.

Give it a shot on a normal install (brq maybe).  Login, see the VM cards, click logout from the user icon in the masthead and see the gray box blink by on the way to the sso screen.  Let me see if I can cap a video of what it currently looks like for a normal user on my install...
ezgif com-video-to-gif

I could differentiate between a manual and auto logout and display a different message accordingly.  I could also render a different message for a dev server vs a prod install.

lwrigh:
@sjd78 I gave it a shot on brq and I see what you mean by it going really quickly from the grey log in screen with the log in message back to the initial log in screen. It hung a little longer for me but that might just be my internet connection. In that case I would recommend giving the user a brief message saying that they will be redirected to the log in page momentarily.

src/App.js
```diff
@@ -14,25 +13,36 @@ import ToastNotifications from './components/ToastNotifications'
...
-        <span className='pficon pficon pficon-user' />
+    <div>
+      <nav className='navbar obrand_mastheadBackground obrand_topBorder navbar-pf-vertical'>
+        <div className='navbar-header'>
+          <a href='/' className='navbar-brand obrand_headerLogoLink' id='pageheader-logo'>
+            <img className='obrand_mastheadLogo' src={resourcesUrls.clearGif} />
+          </a>
+        </div>
+      </nav>
+      <div className={`container text-center`}>
```
bond95: I think this braces can be removed `<div className='container text-center'>`

src/App.js
```diff
@@ -14,25 +13,36 @@ import ToastNotifications from './components/ToastNotifications'
...
+        <h1 className='bolder'>{msg.pleaseLogInTripleDot()}</h1>
+        <div>
+          <a href={AppConfiguration.applicationURL}>Log in</a>
```
bond95: Please, translate

src/components/SessionActivityTracker.js
```diff
@@ -41,23 +36,32 @@ class SessionActivityTracker extends React.Component {
...
+    document.body.addEventListener('mousemove', this.dropCounter)
+    this.timer = setInterval(this.decrementCounter, 1000)
+  }
+
   componentWillUnmount () {
-    clearInterval(this.timer)
+    document.body.removeEventListener('mousemove', this.dropCounter)
+    if (this.timer) {
+      clearInterval(this.timer)
+      this.timer = null
```
bond95: `componentWillUnmount` is last stage of component's life cycle, so after this call component will be destroyed, so this line is unnecessary.

src/components/SessionActivityTracker.js
```diff
@@ -41,23 +36,32 @@ class SessionActivityTracker extends React.Component {
...
-      const state = {
-        counter: this.state.counter - 1,
-      }
-      if (this.state.counter <= TIME_TO_DISPLAY_MODAL && !this.state.showTimeoutModal) {
-        state.showTimeoutModal = true
-      }
-      if (this.state.counter <= 0) {
+    const { counter } = this.state
+    if (counter !== null) {
+      this.setState({
```
bond95: I think in this case it better to rewrite it like this 
```JavaScript
decrementCounter () {
  this.setState(state => {
    const { counter } = state
    if (counter !== null) {
      return {
        counter: counter - 1,
        showTimeoutModal: counter <= TIME_TO_DISPLAY_MODAL,
      }
    }
    return null
  }, () => {
    if (this.state.counter <= 0) {
      this.props.onLogout()
    }
  })
}
```

sjd78: I like that style but had to adjust the `appReady` condition in the `App` component to flip only after the config values have been loaded and `counter` won't default to `undefined`.

sjd78:
When logged out by the session activity tracker (on development):
screenshot-localhost-3000-2019 04 24-18-48-36

When logged out manually (on development):
screenshot-localhost-3000-2019 04 24-18-49-07

On prod (i.e. not on the dev server), the secondary message will be, "You will be redirected to the log in page momentarily."

(@lwrigh, @bond95)

lwrigh:
@sjd78 The addition of the messages is much more informative and look good to me!

Issue #980, opened 2019-02-27
vNIC profile is mandatory, but can stay <Empty>


leistnerova:
Description of problem [include screenshots and browser console logs]:
VNIC Profile is mandatory, but can stay <Empty>. Check by saving should be added that profile is other than that.

VM Portal version number:
master
oVirt version number:
ovirt-engine-4.3.1.2-0.0.master.20190222171227.gita00fd1d.el7.noarch
Browser and version: FF 65.0.1

Steps to Reproduce:
1. add new vNIC for VM
2. let profile stay <Empty>
3. save

Actual results: dialog is closed, error appears in notification

Expected results: can't save dialog with &lt;Empty&gt; profile

sjd78:
@leistnerova , @bond95 - In webadmin it is perfectly valid to create a nic without a VNIC profile (it shows as "<Empty>".

sjd78:
Here is the actual error from the REST api:

![screenshot-chopper local dickerson-web com-2019 05 16-19-40-14](https://user-images.githubusercontent.com/3985964/57893858-7b9ad580-7812-11e9-9000-11307d3ddf49.png)

So when `<Empty>` is selected, somehow something is getting the the API where it thinks a `vnicProfile` should be valid but isn't.  From my comment https://github.com/oVirt/ovirt-web-ui/pull/1017#issuecomment-493266050, web-ui is sending an empty string for the vnicProfile id, and that doesn't exist.  The empty string shouldn't be sent at all in this case.

PR #1017, opened 2019-04-29
Fixed saving NIC with empty VNIC Profile

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/980

image

After change value, error disappear:
image

### DIFF
--- a/src/components/VmDetails/cards/NicsCard/NicEditor.js
+++ b/src/components/VmDetails/cards/NicsCard/NicEditor.js
@@ -11,7 +11,6 @@ import {
   FormGroup,
   Modal,
   Radio,
-  HelpBlock,
 } from 'patternfly-react'
 import SelectBox from '../../../SelectBox'
 import ExpandCollapseSection from '../../../ExpandCollapseSection'
@@ -67,11 +66,12 @@ class NicEditor extends Component {
       showModal: false,
 
       id: undefined,
-      name: props.nextNicName,
-      vnicProfileId: EMPTY_ID,
-      interface: NIC_INTERFACE_DEFAULT,
-      linked: true,
-      errors: {},
+      values: {
+        name: props.nextNicName,
+        vnicProfileId: EMPTY_ID,
+        interface: NIC_INTERFACE_DEFAULT,
+        linked: true,
+      },
     }
 
     this.open = this.open.bind(this)
@@ -93,22 +93,24 @@ class NicEditor extends Component {
         showModal: true,
 
         id: nic.id,
-        name: nic.name,
-        vnicProfileId: nic.vnicProfile.id,
-        interface: nic.interface,
-        linked: nic.linked,
-        errors: {},
+        values: {
+          name: nic.name,
+          vnicProfileId: nic.vnicProfile.id,
+          interface: nic.interface,
+          linked: nic.linked,
+        },
       })
     } else {
       this.setState({
         showModal: true,
 
         id: undefined,
-        name: this.props.nextNicName,
-        vnicProfileId: EMPTY_ID,
-        interface: NIC_INTERFACE_DEFAULT,
-        linked: true,
-        errors: {},
+        values: {
+          name: this.props.nextNicName,
+          vnicProfileId: EMPTY_ID,
+          interface: NIC_INTERFACE_DEFAULT,
+          linked: true,
+        },
       })
     }
   }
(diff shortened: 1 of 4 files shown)


lwrigh:
I would suggest changing the error text to read, 'VNIC profile is required, please select a VNIC profile."

bond95:
@lwrigh 
image

lwrigh:
LGTM!

sjd78:
@lwrigh , @bond95 - Why the even both showing the message?  The field is already marked as required.  Why not just disable the "Ok" button until the fields (name and VNIC profile) are valid?  That would be more consistent with the rest of the UI I think.

lwrigh:
@sjd78 @bond95 Scott brings up a good point since create new VM and similar actions follow the pattern that a user must fill out the necessary information before they can complete the task and the primary action button remains disabled until they complete the necessary fields.

bond95:
> @sjd78 @bond95 Scott brings up a good point since create new VM and similar actions follow the pattern that a user must fill out the necessary information before they can complete the task and the primary action button remains disabled until they complete the necessary fields.

I think for user it will be not so clear, why button is disabled. So I think it is better not to disable buttons, but tell user explicitly why NIC cannot be created.

bond95:
@lwrigh WDYT?

image
image

lwrigh:
@bond95 I think in this case it's a little confusing with the <Empty> since it looks like a selection has been made. I think it would be more helpful to the user if the text in the drop down said something along the lines of, "Select a VNIC profile" so it's a little bit more clear that an option needs to be selected.

bond95:
@lwrigh ?
image

lwrigh:
@bond95 I would recommend getting rid of the '< >' and have the text be a regular weight, not bold if they are.

sjd78:
@bond95 - Selecting "<Empty>" should be valid.  It can be done in webadmin, but it is an explicit selection as a vnic profile is auto-selected initially.  The selection is probably based on logic like "use the same vnic profile as other nics" or "if only one vnic profile is available, select that" or "select the first available vnic profile".

When I tried to create a nic in web-ui, the request payload come out like:
````json
{
   "name":"_demo_",
   "linked":true,
   "interface":"virtio",
   "vnic_profile":{
      "id":""
   }
}
````
The problem is in the `vnic_profile` field with the blank `id`!

If I run `curl` like this against the same rest endpoint, it works creating the nic:
````
curl -k -u"admin@internal:*****" \
    --request POST \
    -H"Accept: application/json" \
    -H"Content-Type: application/json" \
    --data '{"name":"_demo_","linked":true,"interface":"virtio"}' \
    "https://chopper.local.dickerson-web.com/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics"
````

In web-ui before:
screenshot-chopper local dickerson-web com-2019 05 16-19-20-39

Output of the curl run:
````json
{
  "interface" : "virtio",
  "linked" : "true",
  "mac" : {
    "address" : "56:6f:d5:d8:00:0d"
  },
  "plugged" : "true",
  "actions" : {
    "link" : [ {
      "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics/7934aa80-e576-4f90-877d-8f3782166794/activate",
      "rel" : "activate"
    }, {
      "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics/7934aa80-e576-4f90-877d-8f3782166794/deactivate",
      "rel" : "deactivate"
    } ]
  },
  "name" : "_demo_",
  "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics/7934aa80-e576-4f90-877d-8f3782166794",
  "id" : "7934aa80-e576-4f90-877d-8f3782166794",
  "vm" : {
    "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a",
    "id" : "10ebb6ac-645d-42ec-8031-c844aa14666a"
  },
  "link" : [ {
    "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics/7934aa80-e576-4f90-877d-8f3782166794/networkfilterparameters",
    "rel" : "networkfilterparameters"
  }, {
    "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics/7934aa80-e576-4f90-877d-8f3782166794/reporteddevices",
    "rel" : "reporteddevices"
  }, {
    "href" : "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/nics/7934aa80-e576-4f90-877d-8f3782166794/statistics",
    "rel" : "statistics"
  } ]
}
````

In web-ui after:
screenshot-chopper local dickerson-web com-2019 05 16-19-21-47

So really, the error is somewhere in the nic object creating or the saga/api call creating the json object to be posted.  Looks like the problem is between object creation by the `NicEditor` popup:
https://github.com/oVirt/ovirt-web-ui/blob/7e3b8b5b39fae13bd9bdb057f27b4d25b2de7c3a/src/components/VmDetails/cards/NicsCard/NicEditor.js#L116-L128

and transforming for POST:
https://github.com/oVirt/ovirt-web-ui/blob/7e3b8b5b39fae13bd9bdb057f27b4d25b2de7c3a/src/ovirtapi/transform.js#L626-L641

The `vnicProfileId` is coming in as an empty string and is incorrectly being added to the transformed json.  Adjust either the `NicEditor` to exclude the `vnicProfile` when it is selected as "<Empty>" or adjust the transform to consider `undefined` and empty string, or do both to be safe.

bond95:
> @bond95 - Selecting `<Empty>` should be valid. It can be done in webadmin, but it is an explicit selection as a vnic profile is auto-selected initially. The selection is probably based on logic like "use the same vnic profile as other nics" or "if only one vnic profile is available, select that" or "select the first available vnic profile".

Is there any purposes to let regular user create NICs with empty VNIC Profile?

sjd78:
> > @bond95 - Selecting `<Empty>` should be valid. It can be done in webadmin, but it is an explicit selection as a vnic profile is auto-selected initially. The selection is probably based on logic like "use the same vnic profile as other nics" or "if only one vnic profile is available, select that" or "select the first available vnic profile".
> 
> Is there any purposes to let regular user create NICs with empty VNIC Profile?

I don't know the use cases for an empty vnic profile, and I think it is the wrong question.  The issue says the field is required, but the issue is wrong.  The error message is that the `vnicProfileId` cannot be blank, and that happens because of a bug in generating the JSON object for the REST call.

IMO, the fix should be 2 parts.  First, fix the REST call so selecting `<Empty>` is a valid choice.  Second, update the dialog to auto-select the vnic (If only 1 vnic is available, that one. Otherwise default to empty or default to the most commonly used vnic across any other vnics like disks do for storage domains).

sjd78:
@bond95 - Any followup?

bond95:
@sjd78 please, review

isaranova:
ci build please

isaranova:
LGTM

Issue #976, opened 2019-02-27
Validation of disk name after closing the dialog


leistnerova:
Description of problem [include screenshots and browser console logs]:
Validation of disk name - after closing dialog. When adding a disk with space means that the dialog is closed, notification is received and then user have to invoke the dialog and fill/set all its form fields again.

VM Portal version number:
master
oVirt version number:
ovirt-engine-4.3.1.2-0.0.master.20190222171227.gita00fd1d.el7.noarch
Browser and version: FF 65.0.1

Steps to Reproduce:
1. have VM with edit permissions
2. try to create new disk
3. in alias enter 'disk_name1 '
4. save

Actual results: dialog closes, no disk is created, error appears in notification

Expected results: validation stays in dialog as for e.g. VM name

PR #1018, opened 2019-04-30
Added validation for Disk name

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/976

It working similar to https://github.com/oVirt/ovirt-web-ui/pull/1017

If name invalid:
image

If name empty:
image


lwrigh:
If the name is invalid, I would recommend changing the message to be more straight forward. I think we have similar messages in other areas of the UI so I would suggest reusing the message if we already have one. If not I would suggest, 'Disk name is required and can include letters, numbers, or "-, _ , .". 

If the field is empty I would recommend, "This field is required".

bond95:
@lwrigh 
image

I got message from VM name validation message
image

lwrigh:
Both of these are looking good! One nit pick, I would add a period to the end of the first 'This field is required.' sentence.

sjd78:
> Both of these are looking good! One nit pick, I would add a period to the end of the first 'This field is required.' sentence.

I'd leave the field black when empty or valid.  Error border and help text when the text isn't valid.  I don't think the "This field is required." text is necessary.

lwrigh:
Okay so after taking a closer look at PatternFly recommendations for both client and server side validation errors it sounds like PF recommends disabling the action button until the necessary fields have correct inputs. I would follow that recommendation for disabling and enabling the OK button. Adding syntax hints for more specific input types looks like a good option for us as well. Especially for the inputs that have special criteria like the 'Name' field. 

So looking at the use of asterisks next to the fields if they are required, it looks like they will be unnecessary if all of the fields are required. Disabling the action button until all of the fields are filled out correctly would be enough of an indicator based off of PF recommendations. 

I can mock up these suggestions for the 'Create New Disk' dialog. @sjd78 also mentioned adding a guide for validation standards to the web ui repo which I thought was a good idea that we can base off of the validation decisions we make.

bond95:
> I can mock up these suggestions for the 'Create New Disk' dialog.

@lwrigh It will be nice, thank you.

@sjd78 what do you think about using https://jaredpalmer.com/formik/ for validation, it has nice and clear mechanism for working and validating forms.

lwrigh:
Normal input fields before content is added:
Screen Shot 2019-05-13 at 3 39 17 PM

Input fielded filled out with no errors and action button becomes activated: 
Screen Shot 2019-05-13 at 3 39 26 PM

Input field error and action button is disabled: 
Screen Shot 2019-05-13 at 3 39 35 PM

Input is corrected and action button becomes activated: 
Screen Shot 2019-05-13 at 3 39 42 PM

@bond95 @sjd78 Here are the updated validation wireframes for the new disk dialog, what do you think?

bond95:
@lwrigh looks good to me, thank you.

bond95:
@lwrigh WDYT?

image
image
image

lwrigh:
@bond95 If an error appears in one of the input fields the 'OK' button should remain disabled until the error has been fixed. Besides that it's looking good.

bond95:
@lwrigh done

image

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -396,6 +411,10 @@ class DiskImageEditor extends Component {
             id={`${idPrefix}-modal-ok`}
             bsStyle='primary'
             onClick={this.handleSave}
+            disabled={
```
sjd78: Create a new function for this check `isFormValid` or similarly named so it is easier to document and maintain.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -191,24 +205,24 @@ class DiskImageEditor extends Component {
   }
 
   changeAlias ({ target: { value } }) {
-    this.setState({ alias: value })
+    this.setState((state) => ({ values: { ...state.values, alias: value }, errors: { ...state.errors, 'alias': '' } }))
```
sjd78: This would allow for immediate feedback as the user types instead of waiting for them to click the ok button:
```suggestion
    this.setState((state) => ({ 
      values: { ...state.values, alias: value }, 
      errors: { ...state.errors, 'alias': isDiskNameValid(value) ? '' : msg.pleaseEnterValidDiskName() },
    }))
```

leistnerova:
ci build please

isaranova:
LGTM

Issue #258, opened 2017-06-26
templates should be filtered by user permission on New and Edit VM pages


jelkosz:
There is a chance that the user is allowed to see a template from a cluster but is not allowed to create a VM based on it. This templates should be marked somehow or not shown in the new VM dialog at all.

[GS: edit] Also, on the Edit dashboard, if there are templates a user doesn't have permissions to apply, they should be hidden.

michalskrivanek:
@mareklibra still targeted to 1.3? doesnt' seem realistic

mareklibra:
No, I see "Future release". Might be moved to 1.4 within next planing

doron-fediuck:
@sgratch still relevant? If so it should be prioritized (and we need a bz).

michalskrivanek:
The only tracker for ovirt-web-ui is the github issue tracker

doron-fediuck:
> The only tracker for ovirt-web-ui is the github issue tracker

Which makes it hard to track as a part of the overall picture.

isaranova:
ci build please

PR #1019, opened 2019-04-30
Added template filtering for VM create dialog

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/258

### DIFF
--- a/src/sagas/index.js
+++ b/src/sagas/index.js
@@ -150,6 +150,7 @@ import {
   canUserUseCluster,
   canUserUseVnicProfile,
   getUserPermits,
+  canUserUseTemplate,
 } from '_/utils'
 
 const vmFetchAdditionalList =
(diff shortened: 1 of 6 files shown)


src/sagas/index.js
```diff
@@ -835,7 +836,15 @@ function* fetchAllTemplates (action) {
   const templates = yield callExternalAction('getAllTemplates', Api.getAllTemplates, action)
 
   if (templates && templates['template']) {
-    const templatesInternal = templates.template.map(template => Api.templateToInternal({ template }))
+    const templatesInternal = (yield all(
+      templates.template
+        .map(function* (template) {
+          const templateInternal = Api.templateToInternal({ template })
+          templateInternal.permits = yield fetchPermits({ entityType: PermissionsType.TEMPLATE_TYPE, id: template.id })
```
sjd78: I hate that we have to do this.  Maybe a better RFE to writeup would be to have the API be able to directly give us a user's permits on an entity instead of having to do through the permissions and roles?

@sgratch , @michalskrivanek - Do you think an approach like that wold be more likely to be implemented over a three level follow as the [current RFE BZ1639784](https://bugzilla.redhat.com/show_bug.cgi?id=1639784) asks for?

isaranova:
ci build please

isaranova:
When user has a pool VM created from template in cluster that user doesn't have any permissions for, the template name for the VM is N/A instead of the name of the template.

bond95:
> When user has a pool VM created from template in cluster that user doesn't have any permissions for, the template name for the VM is N/A instead of the name of the template.

This PR fixing filtering for select in `Create VM Dialog`, detail page has no such select, so I think it will be nice to create separate issue for it.

isaranova:
LGTM

PR #1020, opened 2019-05-18
Fetch roles and permit at login and do entity permission to permit mapping on frontend

sjd78:
Resolves: #1066 

### Part 1
At login, fetch all roles and permits available to the user.  By fetching all of the roles and permits available to a user at login time, the set of permits based on an entity's permission set can be calculated on the front end without having to fetch the permits directly.

We can't fetch the set of permits directly when requesting an entity since the permits are on the 3rd tier down from the entity and the REST 'follow' parameter doesn't traverse that far down. See [BZ1639784]( https://bugzilla.redhat.com/show_bug.cgi?id=1639784)

Now, we can fetch an entities permissions and cross reference each permission's role with the one in the redux store and build a permit list from there without having to go back to the API.

### Part 2
Refactor user permission / permit / 'canUser*' calculations.  Each entity's permits can be checked without an additional REST call since they're already loaded in redux state.

Note: This is currently only applied to redux data fetched during
      login.  VM and Disk fetches will need to be updated in a similar
      fashion in future.

The authorization (permit/canUser*) refactoring has been applied to:
  - Clusters
  - Templates
  - Vnic Profiles
  - Data Centers
  - Storage Domains

Summary of changes:
  - Refactor the redux read-only fetches to `base-data.js`

  - User permits are calculated from entity permissions + user groups
    (fetched at login) + roles (fetched at login).  Entity permissions
    map through the user and groups to roles and permits.   A user's
    authorization to access or perform a function is based on their
    net permits.


sjd78:
@bond95, @michalskrivanek, @sgratch - I think this technique may be a viable way to reduce the number of REST calls we need to make.  Currently, for every entity we want to check permits, the permissions (with roles and permits followed) are fetched directly.  With this PR, I'm prefetch all the roles and each entity can just add a follows=permissions, cross reference in the fetch saga and be done.

Any thoughts on why this wouldn't be a valid way to go?

michalskrivanek:
it would certainly help a lot!

sjd78:
@bond95 - What do you think about this approach instead of the extra calls that are currently being doing (like in #1019)?

src/sagas/storageDomains.js
```diff
@@ -20,111 +20,119 @@ import {
...
-        Object.assign(merged.statusPerDataCenter, storageDomain.statusPerDataCenter)
-      } else {
-        accum[storageDomain.id] = storageDomain
-      }
-      return accum
-    },
-    {}
-  )
+function* fetchDataAndIsoStorageDomains () {
+  const storageDomainsApi = yield callExternalAction('getStorages', Api.getStorages, { payload: {} })
```
bond95: `payload` is not necessary too

src/sagas/storageDomains.js
```diff
@@ -20,111 +20,119 @@ import {
...
-          files.push(Api.fileToInternal({ file: disk }))
-        })
-        return acc
-      },
-      {}
-    )
+  yield all([
+    call(function* () {
+      // fetch ISO disk images and distribute them to their storage domains as files
+      const images = yield callExternalAction('getIsoImages', Api.getIsoImages, { payload: {} })
```
bond95: Here too.

src/sagas/storageDomains.js
```diff
@@ -20,111 +20,119 @@ import {
...
-    .keySeq()
-    .toArray()
+      const isoStorageDomains = storageDomains
+        .filter(storageDomain => storageDomain.get('type') === 'iso')
+        .keySeq()
+        .toArray()
 
-  const isoFilesFetches = isoStorageDomains.map(isoStorageDomain => fetchAllFilesForISO(isoStorageDomain))
-  yield all(isoFilesFetches)
+      const isoFilesFetches = isoStorageDomains.map(isoStorageDomain => fetchAllFilesForISO(isoStorageDomain))
```
bond95: I think here enough will be `isoStorageDomains.map(fetchAllFilesForISO)`

src/sagas/utils.js
```diff
@@ -165,9 +167,38 @@ export function* fetchPermits ({ entityType, id }) {
...
+ * explicitly checked.
+ */
+export function* permissionsToUserPermits (permissions) {
+  const userFilter = yield select(state => state.config.get('filter'))
+  const userGroups = yield select(state => state.config.get('userGroups'))
+  const userId = yield select(state => state.config.getIn(['user', 'id']))
+  const roles = yield select(state => state.roles)
+
+  const permitNames = []
+  for (const permission of (Array.isArray(permissions) ? permissions : [permissions])) {
```
bond95: I think it would be nice to move `(Array.isArray(permissions) ? permissions : [permissions])` to separate constant

src/actions/roles.js
```diff
@@ -0,0 +1,20 @@
...
+    payload: {
+      roles,
+    },
+  }
+}
+
+export function getRoles () {
+  return {
+    type: GET_ROLES,
+    payload: {},
```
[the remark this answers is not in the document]
sjd78: I still think that should be done under its own RFE issue.  There are too many moving part where action objects get passed around for me to feel comfortable making that changes along side any other change.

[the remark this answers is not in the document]
sjd78: I opened #1059 to cover any of these kinds of changes.

src/sagas/storageDomains.js
sjd78: I opened #1059 to cover any of these kinds of changes.

src/sagas/storageDomains.js
sjd78: I opened #1059 to cover any of these kinds of changes.

sjd78:
What kind of difference does this change make?  Hitting the brq-dev engine, I have the following numbers from initial app request to the APP_CONFIGURED action:

Code Base | XHR Req Count | XHR Size | Load time
----- | ----- | ----- | -----
master | 46 | 99.7 KB | 16.4s
this pr | 24 | 44.1 KB | 11.3s

The request count and size different is due to not requesting each entities's permits explicitly.  The number will vary depending on how many things each user has access to in their environment.  Given that caveat, for admin @ brq-dev, it is a 52% request count reduction, 44% request size reduction and 68% of the original time taken.  Decent changes!

src/sagas/storageDomains.js
```diff
@@ -20,111 +20,119 @@ import {
...
-        disk.storage_domains.storage_domain.forEach(({ id }) => {
-          const files = acc[id] = acc[id] || []
-          files.push(Api.fileToInternal({ file: disk }))
-        })
-        return acc
-      },
-      {}
-    )
+  yield all([
+    call(function* () {
```
bond95: I think it would be nice to move this function outside, because it is not very clear purpose of this function.

src/sagas/storageDomains.js
```diff
@@ -20,111 +20,119 @@ import {
...
+        yield all(updates)
       }
-    )
-    yield all(updates)
-  }
+    }),
 
-  // fetch 'files' from ISO storage domains
-  const storageDomains = yield select((state) => state.storageDomains)
+    call(function* () {
```
bond95: Here too.

sjd78:
ci test please

sjd78:
(rebased)

sjd78:
(rebased)

sjd78:
(rebased)

sjd78:
(rebased)

sjd78:
(rebased)

pnovotny:
ci build please

sjd78:
Rebased and is working again on top of current master.
(@sgratch)

sjd78:
@sgratch, @michalskrivanek - CI is failing here because nodejs-modules isn't build correctly for el7 and fc30 on a my pre-seed patch https://gerrit.ovirt.org/#/c/107309/.  Once nodejs-modules actually can run a clean re-merge, CI will work again.

sjd78:
ci test please

pnovotny:
ci build please

pnovotny:
LGTM

Issue #1024, opened 2019-05-30
VM Portal cannot sign in with "yarn start" 


thaorell:
VM portal does not automatically sign in in development mode. Even when clicking "Sign In", the user is directed back to the same error page. 

ovirt-version: 4.4.0-0.0.master.20190522130703.git401db7f.el7

Browser and version: Chrome Version 74.0.3729.169 (

Steps to Reproduce:
```
ENGINE_URL=https://[your engine]/ovirt-engine yarn start
```

![Screenshot from 2019-05-29 15-50-11](https://user-images.githubusercontent.com/25762021/58640728-b3e8ec00-82c7-11e9-95a7-a2886f6ef459.png)

Temporary fix: 
Change the 'userSessionTimeoutInterval' in src/components/SessionActivityTracker.js

gregsheremeta:
> Change the 'userSessionTimeoutInterval' in src/components/SessionActivityTracker.js
@thaorell  change it to what?

thaorell:
@gregsheremeta change this.props.config.get('userSessionTimeoutInterval') to a high number, like 1000, should work

sjd78:
When the ovirt-engine rest endpoint is close / fast, this issue doesn't tend to occur.  If there is any network delay, it always happens.

PR #1025, opened 2019-05-30
Resolve issues with immediate logout when app starts

sjd78:
When the app is starting up, the `SessionActivityTracker` would immediately load. If the tracker started before the timeout value is loaded, as part of the login initial data fetch, it was considering a timeout value of `null` as 0 and immediately logging out the user.

This has been fixed in two ways:
  1. The `SessionActivityTracker` isn't loaded on the `App` until the app is ready (i.e. once enough initial data has been fetched)

  2. `SessionActivityTracker` will not tick down its counter when the counter is set to `null` (i.e. the `userSessionTimeoutInterval` hasn't been fetched yet)

Fixes: #1024

### DIFF
--- a/src/components/SessionActivityTracker.js
+++ b/src/components/SessionActivityTracker.js
@@ -18,7 +18,7 @@ class SessionActivityTracker extends React.Component {
       counter: props.config.get('userSessionTimeoutInterval'),
     }
 
-    this.dropCounter = this.dropCounter.bind(this)
+    this.resetTimeoutCounter = this.resetTimeoutCounter.bind(this)
     this.decrementCounter = this.decrementCounter.bind(this)
   }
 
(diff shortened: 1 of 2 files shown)


src/components/SessionActivityTracker.js
```diff
@@ -31,36 +31,37 @@ class SessionActivityTracker extends React.Component {
...
   }
 
   decrementCounter () {
-    this.setState(state => {
-      const { counter } = state
-      if (counter !== null) {
-        return {
-          counter: counter - 1,
-          showTimeoutModal: counter <= TIME_TO_DISPLAY_MODAL,
+    if (this.state.counter === null) { // counter is null if timeout value hasn't been fetched yet
```
thaorell: Nice catch!

PR #1030, opened 2019-06-04
toolbar action buttons alignments, snapshots dates, and chart alignments

thaorell:
Fixes 3 issues: 
* toolbar action buttons jumping from left to right when screen hits 768px
* snapshot created time segmented
* chart alignment disorganized in narrow screens


src/components/Toolbar/style.css
```diff
@@ -5,3 +5,9 @@
 .toolbar {
   width: 100%;
 }
+
+@media only screen and (max-width : 768px){
+  .toolbar :global(.toolbar-pf-action-right){
+    float: right
```
bond95: Did you try to put this class outside `@media` ?

thaorell: You're right. I don't think it would make a difference. Nice catch!

src/components/VmDetails/cards/SnapshotsCard/style.css
```diff
@@ -76,7 +76,8 @@ div.popover {
 
 .snapshot-item-time {
   color: #808080;
-  padding-left: 5px;
+  display: block;
```
sjd78: This will make ALL snapshot times appear below the description, even when there is plenty of space for the time.
screenshot-chopper local dickerson-web com-2019 06 13-17-07-15

Perhaps you should use `inline-block`?

And the 5px padding between the name and the time should still be there...maybe a `padding-right: 5px` on the name, or just leave it in there and let the time push over to the right if the block has to flow to the next line...

isaranova:
ci build please

isaranova:
LGTM

PR #987, opened 2019-03-12
Create VM Wizard

sjd78:
## Summary
Remove create VM dialog page, and replace it with a new Create VM Wizard.  `CreateVmWizard` has the following steps:
  - Basic Settings
  - Networking
  - Storage
  - Summary Review

After the final review step, the VM will be created.  Success or failure of the creation will be detailed on the review step.

A number of refactors have been done to support this PR:
  - Move most list creation functions to `components/utils` to allow sharing between create and edit components
  - Updated API calls, API transforms and helper functions around:
    - OS
    - CloudInit
    - Templates (specifically to support fetching information about template defined disks and template defined NICs)

#### Pre-work PRs:
  - ~#1036 (import patternfly-react CSS)~
  - ~#1037 (refactor GridComponents)~
  - ~#1038 (refactor components/utils.js)~
  - ~#1039 (refactor DetailsCard)~

Resolves: #923
Resolves: #1015
Resolved: #922 
[Design Document](https://docs.google.com/document/d/1UtFhNa_1byMV5mLEJ1MbdvZj-uMpmsD-_bJ37JDcyjU)

## Basic details about the change:
  - Remove create VM dialog page, replace with create VM wizard
  - Since the Create VM Wizard is a modal dialog, a route is no longer needed
  - Create VM is accessed from the VM list page
  - Change page type to 'DIALOG_PAGE_TYPE' when CreateVmWizard is opened
  - Establish a base default settings for a new VM managed by
    the wizard component `CreateVmWizard`:
    - Auto-select a cluster if the user only has access to 1
    - If the auto-selected cluster has only 1 template available, the Blank template, change the default provision source to 'iso'
    - If the auto-selected cluster has >1 template available, do not auto-select a template
  - The wizard and the components for each step share the new VM data PropType shapes

## Create VM saga
  - VM compose is a saga now (so it is easy to query any other info needed from redux state)
  - Refactored sagas related to CRUD operations on VM to `vmChanges.js`
  - Adjusted reducers and sagas to allow correlation results from vm creation when a vmId is not available before the task (for use in SummaryReview)

## Wizard Step: Basic Settings
  - This is a controlled component, all of the field state is held in the containing Wizard itself
  - Every change notifies the wizard of 2 things.  First, the validity status of the BasicSettings (all required fields filled out and all entered data is valid).  Second, the partial VM data changes just made by the user interaction.
  - Changing provision source clears the currently selected ISO, clears the selected Template and resets the VM's nic and disk lists.

[...]

Issue #923, opened 2019-01-16
Redesign Create VM


michalskrivanek:
redesign Create VM flow

design document:
https://docs.google.com/document/d/1UtFhNa_1byMV5mLEJ1MbdvZj-uMpmsD-_bJ37JDcyjU/edit

![screenshot from 2019-02-13 11-43-34](https://user-images.githubusercontent.com/2248293/52728137-bcb9ea00-2f84-11e9-9535-6d3302c22b49.png)

![screenshot from 2019-02-13 11-43-40](https://user-images.githubusercontent.com/2248293/52728146-c04d7100-2f84-11e9-9ad9-4fd19a33aa24.png)

See also: #695

lwrigh:
@sjd78 Let me know if you would rather do an in person or recorded design review of the wizard. I'd be happy to do either.

PR #1037, opened 2019-06-06
Refactor `GridComponents` to enable use outside `VmDetails`

sjd78:
Changes:
 - Move `GridComponents.js` from **src/components/VmDetails** to **src/components/Grid**
 - Added Grid/index.js in the same style as **patternfly-react**
 - Adjust existing `import` statements to match the new location

Pre-work for PR #987, Issue #923


src/components/Grid/GridComponents.js
```diff
@@ -47,7 +47,7 @@ const Col = ({ className = '', cols = -1, style, children, content = 'expand', .
   )
 }
 Col.propTypes = {
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
```
bond95: Is there any reason to removing `isRequired`? Because as I understand this component, this is container for other components, so it should get some children.

src/components/Grid/index.js
```diff
@@ -0,0 +1,11 @@
+import { Grid, Row, Col } from './GridComponents'
+
+Grid.Row = Row
+Grid.Col = Col
+
+export default Grid
```
bond95: I understand that this is similar way to export as in bootstrap, but is this really necessary to export `Grid` as default? Because I don't any use of this, so maybe better will be to do like this:

```JavaScript
export {
  Grid,
  Row,
  Col,
} from './GridComponents'
```
?

sjd78: Makes sense.

src/components/Grid/GridComponents.js
sjd78: The reason is if you want a `Col` to take up space but don't want to put an empty tag inside...
````js
  <Grid>
    <Row>
      <Col cols={4} />
      <Col cols={8}>
        ...content that will render on the right most 8 columns....
      </Col>
    </Row>
  <Grid>
````

Comment (bond95) [reply src/components/Grid/GridComponents.js]: Hmmm... true, but maybe in this case better use similar props like in patternfly-react: http://patternfly-react.surge.sh/patternfly-3/index.html?knob-fluid=true&selectedKind=patternfly-react%2FLayouts%2FGrid&selectedStory=Grid&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

In this case your example will look like this:

```jsx
<Grid>
    <Row>
      <Col xsOffset={4} cols={8}>
        ...content that will render on the right most 8 columns....
      </Col>
    </Row>
  <Grid>
```

sjd78: Ok I can see that.  We don't use that approach anywhere currently, but I've added it regardless.

sjd78:
@bond95 - I added an `offset` to the `Col` along the same lines as what you pointed to.  I even added some fancy validation functions for the PropTypes that check ranges on `cols` and `offset` so they make sense, at least on a single `Col`.

Issue #1023, opened 2019-05-30
VM's CDROM list is empty 


michalskrivanek:
Run a VM and try Edit and change CD. The list is always empty unless you manually hit the reload button.

michalskrivanek:
originally submitted as https://bugzilla.redhat.com/show_bug.cgi?id=1714942

michalskrivanek:
@sjd78 addressed by #789? Even if so, this is a better issue description:)

sjd78:
> @sjd78 addressed by #789? Even if so, this is a better issue description:)

Yes, this issue will be fixed when #789 is fixed.

This has also been logged in [BZ1687521](https://bugzilla.redhat.com/show_bug.cgi?id=1687521)

gematik-fue:
Will it also fix the issue, the list contains images from different clusters/datacenters too?

sjd78:
> Will it also fix the issue, the list contains images from different clusters/datacenters too?

That is a good point.

With #987, I'll make sure the ISO list on the cd-rom select box on the VM details page, and on create new VM are filtered to show only the ISOs in selected cluster's storage domains that are also permissioned to the user.

sjd78:
> > Will it also fix the issue, the list contains images from different clusters/datacenters too?
> 
> That is a good point.
> 
> With #987, I'll make sure the ISO list on the cd-rom select box on the VM details page, and on create new VM are filtered to show only the ISOs in selected cluster's storage domains that are also permissioned to the user.

See #1040

PR #1039, opened 2019-06-06
Refactor Details Card, limit CD/ISO list by data center, template name render with version

sjd78:
Includes #1038.  Pre-work for PR #987, Issue #923.

Refactor the VM dashboard view Details Card:
  - Moved `SelectBox` creating list functions out of DetailsCard.  As part of `components/utils` they can now be reused by `CreateVmWizard`.

  - Filter cd-rom ISO list based on VM's cluster/data center (see https://github.com/oVirt/ovirt-web-ui/issues/1023#issuecomment-499098974)

  - Use `templateNameRenderer` helper to render the template name with the version.  This will make the template names consistent across the app and allow differentiation between multiple versions of the same template.


src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -253,14 +194,16 @@ class DetailsCard extends React.Component {
     }
 
     if (prevProps.storageDomains !== this.props.storageDomains) {
-      this.setState({ isoList: createIsoList(this.props.storageDomains) }) // eslint-disable-line react/no-did-update-set-state
+      const vmClusterId = this.props.vm.getIn(['cluster', 'id'])
```
bond95: I think it would better to move it outside condition, because cluster id is used a few times in this function.

src/components/utils/build-select-box-lists.js
```diff
@@ -0,0 +1,109 @@
...
+      .toJS()
+
+  return dataCenterList
+}
+
+/**
+ * Return a normalized and sorted list of clusters ready for use in a __SelectBox__ from
+ * the Map of provided clusters.
+ */
+function createClusterList (clusters, dataCenterId = false) {
```
bond95: To be more consistent in types of argument, I think here would be better to use `dataCenterId = ''` or `dataCenterId = null`

sjd78: It's really 6 of one, half dozen of another.  The param will be evaluated as 'truthy' is one way or another.  Explicitly defaulting to null at least set a consistent interpretation.

src/components/VmDetails/cards/DetailsCard/index.js
sjd78: Done.

src/components/utils/build-select-box-lists.js
```diff
@@ -0,0 +1,109 @@
...
+ */
+function createIsoList (storageDomains, dataCenterId = null) {
+  const list = []
+
+  storageDomains
+    .toList()
+    .filter(storageDomain =>
+      storageDomain.has('files') &&
+      storageDomain.get('canUserUseDomain') &&
+      (dataCenterId ? storageDomain.getIn([ 'statusPerDataCenter', dataCenterId ]) === 'active' : true)
```
thaorell: ternary operator could get a little confusing, consider using multiline ternary operators?
```
(dataCenterId ?
    (storageDomain.getIn([ 'statusPerDataCenter', dataCenterId ]) === 'active') : 
    true)
```

sjd78: I think it is the same amount of confusing on a single line v multiple lines in this case.  But if I change the condition a touch to make the "value of null means to skip that part of the check", a multiline looks ok:
````js
    (dataCenterId === null
      ? true
      : storageDomain.getIn([ 'statusPerDataCenter', dataCenterId ]) === 'active')
````

isaranova:
ci build please

isaranova:
When in VM detail and editing Change CD, the list has only Empty. After force refresh the list contains other CDs.

sjd78:
> When in VM detail and editing Change CD, the list has only Empty. After force refresh the list contains other CDs.

That is a known issue (#1023) and is addressed in PR #1049

isaranova:
ci build please

isaranova:
LGTM

PR #1044, opened 2019-06-14
Default memory unit to MiB while editing VM memory on details card

### LINKED ISSUE #999: When editing the amount of memory for a VM, memory units are automatically changed from MiB to GiB
When editing the amount of memory for a VM, memory units are automatically changed from MiB to GiB, which may lead to a VM which is not able to be scheduled

Version-Release number of selected component (if applicable):
4.3

How reproducible:
100%

Steps to Reproduce:
1. Provision a VM and edit the memory amount
2. Change 384 MiB (for example) to 512 MiB

Actual results:
MiB is changed to GiB

Expected results:
Units stay the same unless it's >1000 and makes sense to scale
——
Found here:

https://github.com/oVirt/ovirt-web-ui/commit/7f39063f#diff-1ed030b4b0f12b219c73b9e938e03286R363

Did we document this anywhere/ Assuming it's GB for templates is probably ok, but I'm not seeing a tooltip or anything. It's probably better to just assume it's MiB for 'normal' VMs, or we could put a little more effort into it and change:

if (isNumber(value) && ...) {
...
}
// Addition
else if (isParsableByte(value)) {
...
}

function isParsableByte(input) {
    types = {
        "MiB":
          ["MiB", "MB"],
        "GiB":
          ["GiB", "GB"]
    }
    matched = false

    Object.keys(types).forEach(function (size) {
        types[size].some(function(x) {
          if (input.match(x)) {
              matched = size
              return true
          }   
        }) 
    }) 
    console.log(matched)
    return matched 
}

Or, y'know, just default to MiB ;) Or document it

thaorell:
Fixes #999

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -5,13 +5,11 @@ import { List } from 'immutable'
 
 import * as Actions from '_/actions'
 import { MAX_VM_MEMORY_FACTOR } from '_/constants'
-import { generateUnique, localeCompare, filterOsByArchitecture, isWindows } from '_/helpers'
+import { generateUnique, localeCompare, filterOsByArchitecture, isWindows, userFormatOfBytes } from '_/helpers'
 import { msg, enumMsg } from '_/intl'
 
 import {
-  convertValue,
   isNumber,
-  round,
 } from '_/utils'
 import {
   canChangeCluster as vmCanChangeCluster,
@@ -324,7 +322,7 @@ class DetailsCard extends React.Component {
 
               // fields that are editable on the card
               changeQueue.push(
-                { fieldName: 'memory', value: template.get('memory') / (1024 ** 3) }, // input assumed to be GiB
+                { fieldName: 'memory', value: template.get('memory') / (1024 ** 2) }, // input assumed to be MiB
                 { fieldName: 'cpu', value: template.getIn(['cpu', 'vCPUs']) },
                 { fieldName: 'bootMenuEnabled', value: template.get('bootMenuEnabled') },
                 { fieldName: 'os', value: templateOs && templateOs.get('id') },
@@ -455,7 +453,7 @@ class DetailsCard extends React.Component {
 
         case 'memory':
           if (isNumber(value) && value > 0) {
-            const asBytes = value * (1024 ** 3) // input assumed to be GiB
+            const asBytes = value * (1024 ** 2) // input assumed to be MiB
             updates = updates.setIn(['memory', 'total'], asBytes)
             fieldUpdated = 'memory'
             this.hotPlugUpdates['memory'] = true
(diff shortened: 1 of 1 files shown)


ovirt-infra:

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -691,7 +690,8 @@ class DetailsCard extends React.Component {
     const osType = vm.getIn(['os', 'type'])
     const osId = operatingSystems && operatingSystems.find(os => os.get('name') === osType).get('id')
 
-    const { unit: memoryUnit, value: memorySize } = convertValue('B', vm.getIn(['memory', 'total']))
```
bond95: I think it will be better leave this line, and change units to `MiB` only in edit mode.

thaorell: this line causes all memory to be rounded to GB during edit. for example: if only edit mode is set to MiB,  1023, 1024 and 1050 MB are all rounded to 1GB, which is inaccurate

bond95: Yeah, but if we want to be 100% accurate we should use bytes =) Actually we already using formatting with precision in utilization cards, so you'll get not `1 GiB` for `1023 MiB`, but `0.99 GiB`.
https://github.com/oVirt/ovirt-web-ui/blob/e83d2f9357104f988a42168d7cd6ed6e0115cdea/src/components/VmDetails/cards/UtilizationCard/MemoryCharts.js#L39-L41

thaorell: Noted

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -840,17 +840,17 @@ class DetailsCard extends React.Component {
                       }
                     </FieldRow>
                     <FieldRow label={msg.memory()} id={`${idPrefix}-memory`}>
-                      { !isFullEdit && `${round(memorySize)} ${memoryUnit}` }
+                      { !isFullEdit && `${round(memorySize / (1024 ** 3), 2)} ${'GiB'}` }
```
bond95: I meant not to hard-code `GiB`, but use function `userFormatOfBytes` to calculate most preferable unit with precision.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -840,17 +840,17 @@ class DetailsCard extends React.Component {
...
                           <FormControl
                             id={`${idPrefix}-memory-edit`}
                             className={style['memory-input']}
                             type='number'
-                            value={round(memorySize)}
+                            value={(memorySize / (1024 ** 2))}
                             onChange={e => this.handleChange('memory', e.target.value)}
                           />
-                          {memoryUnit}
+                          {'MiB'}
```
bond95: It not needed using brackets, you can just write:
```JavaScript
/>
MiB
```

src/components/VmDetails/cards/DetailsCard/index.js
thaorell: didn't know this existed!

isaranova:
LGTM

PR #1046, opened 2019-06-18
Refactor `FieldValue` to be more generic as `EllipsisValue` with better type handling

sjd78:
When rendering a VM that is not running, the VM's __host__ is undefined.  This causes the `FieldValue` proptype checker to throw an error on the browser console.  The child value is in the JSX code, but the value is `undefined`.

Updated the component's prop type such that `children` and `tooltips` must both be defined or must both be undefined.  If they are not defined, don't bother rendering anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,12 +68,13 @@
     "whatwg-fetch": "1.0.0"
   },
   "dependencies": {
-    "babel-polyfill": "6.26.0",
     "@patternfly/react-console": "1.10.22",
+    "babel-polyfill": "6.26.0",
     "blob-util": "1.2.1",
     "bootstrap": "3.4.1",
     "bootstrap-select": "1.13.1",
     "bootstrap-switch": "3.3.4",
+    "classnames": "2.2.6",
     "connected-react-router": "4.3.0",
     "history": "4.7.2",
     "immutable": "3.8.2",
--- a/src/components/EllipsisValue/index.js
+++ b/src/components/EllipsisValue/index.js
@@ -1,9 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { xor } from '_/propTypeExtras'
 import style from './style.css'
 import classnames from 'classnames'
 
-class FieldValue extends React.Component {
+class EllipsisValue extends React.Component {
   constructor (props) {
     super(props)
     this.ref = React.createRef()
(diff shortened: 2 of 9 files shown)


src/components/EllipsisValue/index.js
```diff
@@ -0,0 +1,79 @@
...
+      return null
+    }
+  }
+}
+
+FieldValue.propTypes = {
+  className: PropTypes.string,
+  id: PropTypes.string,
+  children: function (props, propName, componentName, ...rest) {
+    if ((props.children && !props.tooltip) || (!props.children && props.tooltip)) {
```
bond95: I think it will be nice to extract this validation to extra validator, something like:

```JavaScript
const xor = (validator, otherPropName) => (
  (props, propName, componentName, ...rest) => {
    if ((props[propName] && !props[otherPropName]) || (!props[propName] && props[otherPropName])) {
      return new Error(`Props ${propName} and ${otherPropName} are both required for component ${componentName}`)
    }
    return validator(props, propName, componentName, ...rest)
  }
)
```

Then propTypes will look like that:
```JavaScript
FieldValue.propTypes = {
  className: PropTypes.string,
  id: PropTypes.string,
  children: xor(PropTypes.oneOfType([ PropTypes.string, PropTypes.node ]), 'tooltip'),
  tooltip: xor(PropTypes.string, 'children'),
}
```

src/components/EllipsisValue/index.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import style from './style.css'
+import classnames from 'classnames'
+
+class FieldValue extends React.Component {
```
bond95: I think component should be renamed to `EllipsisValue`

sjd78: Done.

src/components/EllipsisValue/index.js
sjd78: Good point.  I externalized it and added some unit tests.

package.json
```diff
@@ -68,12 +68,13 @@
     "whatwg-fetch": "1.0.0"
   },
   "dependencies": {
-    "babel-polyfill": "6.26.0",
     "@patternfly/react-console": "1.10.22",
+    "babel-polyfill": "6.26.0",
```
thaorell: Does it matter which package goes first?

sjd78: Technically no.  But `yarn add` will sort the entire dependency list alphabetically.

Issue #789, opened 2018-09-25
Address login saga initial data load concurrency issues


sjd78:
When loading and populating the initial redux store at login time, everything is currently done in parallel with async dispatches.

At the very least, the app's initial data load needs to be gated such that ALL of the user's data is loaded in the redux store BEFORE permissions are calculated on clusters and VMs (and other entities in the future).

Beyond the base gating that needs to be done, the flag used to load the App's routes and pages should be re-evaluated.  Often time a page is rendered basic and then re-rendered when the base data finally arrives and is loaded in the store.  Gating the app to start page render after a minimum set of data is loaded to minimize this behavior would be ideal.

Proposed flow (left most bullets happen sequentially):
  - Login and User data (as a set of actions)
    - verify SSO token
    - verify API version
    - fetch and store user data (group membership, admin/default filtering, etc?)
  - Base non-VM redux store loading (fetch, transform and store each concurrently since they don't depend on each other)
    - Clusters
    - Hosts
    - Operating Systems (and icons)
    - Templates
    - Storage Domains, ISO Storage Domains and ISO files
    - VNIC Profiles
  - Load the first page of VMs and Pools
  - Start the background data refresh scheduler

Note: May need a quick UX design for an "Initial App State is loading" design to display to users after login and user data is verified and fetched but before the redux store is ready for page render.

gregsheremeta:
:+1: Nice write up.

Roughly, do you have an idea for the gating mechanism at the code level? I suppose my redux-saga knowledge is lacking here. I'm guessing it won't be as simple as chaining some .then()'s together :smile:

sjd78:
This article [Common patterns with Redux-Saga](https://engineering.universe.com/common-patterns-with-redux-saga-ed68f89dfecf) has concepts that are very close to what I'd adopt.  Perhaps mix in a bit of [concurrent task queue](https://medium.freecodecamp.org/how-to-use-a-concurrent-task-queue-in-your-redux-sagas-39e598c4fcae) for good measure.

In summary, need to do something similar to a Barrier in java (which is inherently multi-threaded) but with the event/message driven concurrency in redux/saga on js (which is single threaded with promises).  Plus the whole process is a bit more complicated when the sagas have chained dispatches (a saga function that does a `put()` -- a non-blocking async operation).

Issue #991, opened 2019-03-21
Better handling for token expiration


bond95:
The app needs much better robust handling for when the token expires on the backend to detect that, stop future refresh updates (to prevent additional errors warning), show the notification (or maybe not) and then push the user to the SSO login page.  In dev mode, there is no SSO login page, so an error component or page would be nice (like `TokenExpired` :wink:).

_Originally posted by @sjd78 in https://github.com/oVirt/ovirt-web-ui/pull/968#issuecomment-469367125_

sgratch:
@bond95 @sjd78 do we still have work to do here?

bond95:
@sgratch yes, we still don't have robust token expiration handling.

lwrigh:
![Logged out empty state](https://user-images.githubusercontent.com/20227374/64281045-4ed51000-cf20-11e9-8f7c-ee152ceecee3.png)

Here's a PatternFly 4 pass at an improved logged out empty state.

PR #1049, opened 2019-06-24
Resolve login data load concurrency issues (allows ISOs to load at login)

sjd78:
Fixes: #1023, Fixes: #789 

Refactored the login sagas such that they load in an order that allows all required data to be present in the redux store prior to it being needed but other operations.  For example, a user's groups need to be loaded before permissions can be appropriately calculated.  Also, specifically for #1023, the set of storage domains for a user needs to be loaded before the ISOs can be properly fetched.

Details:
  - Login/logout sagas better encapsulated in sagas/login.js

  - Moved server config/options to sagas/server-configs.js

  - Data load concurrency issues have been resolved.  Fetches happen in the order needed to ensure all required data is loaded prior to it being needed.

  - Each stage is dependent on (gated by) the previous stage's values:
    - Login information
    - User information
    - Server stored configurations
    - Data with no other pre-reqs
    - Data requiring user groups (and permissions/permits)
    - Data requiring storage domains

  - When possible at each stage, fetches are run as parallel tasks.

  - Added "appConfigured" flag to config reducer.  The login process will set the `config` reducer's `appConfigured` flag once enough configuration data has been loaded and the App can properly render.

  - The user domain is now part of the login action and doesn't require a separate action.

Updated configuration of redux dev tools:
  - Configure REDUX_DEVTOOLS to keep 100 actions in history.  The default is 50 but that doesn't let us see the first few actions at App start-up time.  100 works better for now.

Updated trace logging to make it easier to trace multiple requests happening in parallel:
  - Trace logging on the transport has been enhanced to make it easier to match up requests and responses looking at console output
  - `callExternalAction` debug logging formatting updated to match that of the transport logging

Additional refactoring work:
  - Reorganize sagas/utils imports
  - Move `doCheckTokenExpired` from sagas/login to sagas/utils
  - Sagas that need to query multiple times now use `delayInMsSteps()`
    to generate query intervals
  - Remove unused 'shallowFetch' from getAllTemplates() action


isaranova:
LGTM

src/ovirtapi/transport.js
```diff
@@ -57,27 +57,32 @@ type GetRequestType = { url: string, custHeaders?: Object}
 type InputRequestType = { url: string, input: string, contentType?: string }
 type DeleteRequestType = { url: string, custHeaders?: Object }
 
+let getCounter = 0
+const logHeaders = (headers) => JSON.stringify({ ...headers, 'Authorization': '*****' })
+
 function httpGet ({ url, custHeaders = {} }: GetRequestType): Promise<Object> {
-  logger.log(`_httpGet start: url="${url}"`)
+  const myCounter = getCounter++
```
bond95: I don't see any usage of this, so think it can be removed.

src/sagas/login.js
```diff
@@ -37,26 +29,37 @@ import {
...
+import { fetchDataCentersAndStorageDomains, fetchIsoFiles } from './storageDomains'
 
 /**
  * Perform login checks, and if they pass, perform initial data loading
  */
-export function* login (action) {
-  const { payload: { token, userId, credentials: { username } } } = action
+function* login (action) {
+  const { payload: { token, userId, username, domain } } = action
+  console.group('Login Verification')
```
bond95: Nice, I like using of groups.

src/sagas/login.js
```diff
@@ -173,106 +138,83 @@ function* checkOvirtApiVersion (oVirtMeta) {
   return passed
 }
 
-export function* logout () {
-  if (AppConfiguration.applicationLogoutURL && AppConfiguration.applicationLogoutURL.length > 0) {
-    window.location.href = AppConfiguration.applicationLogoutURL
+function composeIncompatibleOVirtApiVersionMessage (oVirtMeta) {
```
bond95: I think this function must be in `utils` too

src/ovirtapi/transport.js
sjd78: It's being used by the logger function lower down.  I added it so the request and responses can be matched in the log when fetches are being run in parallel.

src/sagas/login.js
sjd78: Perhaps, but it is only ever used by the `login()`, and only if the version check fails.  I don't see the function being used anywhere else, so I went with keeping it here.

src/ovirtapi/transport.js
bond95: Oh, yes, I missed it, thank you.

michalskrivanek:
seems to me there are no fundamental comment. Anything to wait for, @bond95 ?

michalskrivanek:
@sjd78 ready to be merged?

src/actions/index.js
```diff
@@ -39,6 +40,13 @@ export * from './console'
 export * from './userMessages'
 export * from './disks'
 
+export function appConfigured () {
+  return {
+    type: APP_CONFIGURED,
+    payload: {},
```
bond95: From time to time if payload is empty we just don't write it (like in https://github.com/oVirt/ovirt-web-ui/blob/master/src/actions/storageDomains.js). I like that style more than writing unnecessary payload.

src/actions/templates.js
```diff
@@ -13,11 +13,9 @@ export function setTemplates (templates) {
...
 }
 
-export function getAllTemplates ({ shallowFetch = false }) {
+export function getAllTemplates () {
   return {
     type: GET_ALL_TEMPLATES,
-    payload: {
-      shallowFetch,
-    },
+    payload: {},
```
bond95: Same here.

src/ovirtapi/transport.js
```diff
@@ -57,27 +57,32 @@ type GetRequestType = { url: string, custHeaders?: Object}
...
     'Authorization': `Bearer ${_getLoginToken()}`,
     'Accept-Language': AppConfiguration.queryParams.locale, // can be: undefined, empty or string
     'Filter': Selectors.getFilter(),
-    'Accept': 'application/json',
-  }, custHeaders)
-  logger.log(`_httpGet: url="${url}", headers="${JSON.stringify(headers)}"`)
+    ...custHeaders,
+  }
 
+  logger.log(`http GET[${myCounter}] -> url: "${url}", headers: ${logHeaders(headers)}`)
```
bond95: As I see in other files `logger` was rewritten to `console` is there any reason to leave it here?

src/ovirtapi/transport.js
```diff
@@ -57,27 +57,32 @@ type GetRequestType = { url: string, custHeaders?: Object}
...
   return $.ajax(url, {
     type: 'GET',
     headers,
   })
     .then((data: Object): Object => {
       notifyStop(requestId)
+      logger.log(`http GET[${myCounter}] <- data:`, data)
```
bond95: Same

src/sagas/login.js
bond95: As well as `doCheckTokenExpired`, that used only in `index.js`. So I still think that better move it to `utils`.

src/actions/index.js
sjd78: `storageDomains.js` contains the only 2 actions in the app that don't have a payload.  And for the sagas attached to those actions, they push a payload into the action, or define their own, before calling the external action.  I'll leave the `payload: {}` here.

bond95: Yes, but still this unnecessary string, so why not to start refactor code from here, and remove it completely in the future? Even in sagas examples they don't use `payload: {}` where it not needed. https://github.com/redux-saga/redux-saga/blob/master/examples/shopping-cart/src/actions/index.js

sjd78: Ok, then file a separate issue.  I'm not going to make any further changes to the actions in this PR.

src/sagas/login.js
sjd78: Any `doCheckTokenExpired` and `callExternalAction` work should probably be done under the context of #991.  I'll leave tweaking these functions for that PR.

PR #1050, opened 2019-06-28
Remove logger.js in favor of straight console logging

sjd78:
`logger.js` is structured in such a way that standard console logging
is enhanced and is identical in functionality to using the exported
logger.  To maintain consistency across the code base for logging,
the exported version of logger has been removed and its previous usage
has been replaced by traditional console logging.


sjd78:
> Please, remove export for function `enhanceOutput` in `logger.js`, as it is not used.

It is used by `config/jest/chalk.logger.js` to enhance the logging for jest tests so they look like the console output.

Issue #1055, opened 2019-07-03
VMs with qemu guest agent don't render utilization disk chart


sjd78:
For a VM with qemu guest agent, disk file system data may not be reported and the utilization graph is empty.

VM Portal version number: 1.5.2-1 (and 1.5.3-0 / master)
oVirt version number: 4.3 and 4.4
Browser and version: Firefox and Chrome, any version

Steps to Reproduce:
1. Have a VM with the qemu guest agent installed
2. Open VM details
3. Witness no Disk utilization chart

Screen shot: 
![screenshot-chopper local dickerson-web com-2019 07 03-14-25-17 (1)](https://user-images.githubusercontent.com/3985964/60616215-f908ae00-9d9e-11e9-9fc9-ae19779f1333.png)

VM's `disks.usage` statistics:
````json
{
  "kind": "gauge",
  "type": "string",
  "unit": "none",
  "values": {
    "value": [
      {
        "detail": "[]"
      }
    ]
  },
  "vm": {
    "href": "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a",
    "id": "10ebb6ac-645d-42ec-8031-c844aa14666a"
  },
  "name": "disks.usage",
  "description": "Disk usage, in bytes, per filesystem as JSON (agent)",
  "href": "/ovirt-engine/api/vms/10ebb6ac-645d-42ec-8031-c844aa14666a/statistics/c5dd0086-d5f0-3abf-9628-4674abb3f5ac",
  "id": "c5dd0086-d5f0-3abf-9628-4674abb3f5ac"
}

````

Guest VM info:
````
[sdickers@centos7_1 ~]$ cat /etc/system-release
CentOS Linux release 7.6.1810 (Core) 
[sdickers@centos7_1 ~]$ yum info qemu-guest-agent
Loaded plugins: fastestmirror
Repodata is over 2 weeks old. Install yum-cron? Or run: yum makecache fast
Loading mirror speeds from cached hostfile
 * base: ewr.edge.kernel.org
 * extras: mirror.clarkson.edu
 * updates: mirror.clarkson.edu
Installed Packages
Name        : qemu-guest-agent
Arch        : x86_64
Epoch       : 10
Version     : 2.12.0
Release     : 2.el7
Size        : 293 k
Repo        : installed
From repo   : base
Summary     : QEMU guest agent
URL         : http://www.qemu.org/
License     : GPLv2
[...]

PR #1056, opened 2019-07-03
Only show file system chart when there is >0 file systems

sjd78:
Fixes: #1055

There is a permutation where the standard QEMU guest agent
will not report actual per file system data.  In this case, the
VM's `disks.usage` statistic is an empty array.  This case is
now tested, and if the usage array is empty, the unallocated vs
provisioned donut chart is displayed (as if the guest agent is
not installed).

Showing the same VM used to report #1055:
screenshot-localhost-3000-2019 07 03-15-16-49


isaranova:
ci build please

bond95:
ci build please

isaranova:
LGTM

lwrigh:
LGTM

PR #1057, opened 2019-07-08
Refactored Selectors

bond95:
- Replaced Selectors in saga workers by `select`
 - Replaced some Selectors in components
 - Removed unusable Selectors


src/sagas/index.js
```diff
@@ -1021,12 +1038,13 @@ function* schedulerWithFixedDelay (delayInSeconds = AppConfiguration.schedulerFi
     } else {
       logger.log(`⏰ schedulerWithFixedDelay[${myId}] running after delay of: ${delayInSeconds}`)
 
-      const oVirtVersion = Selectors.getOvirtVersion()
+      const oVirtVersion = yield select(state => state.config.get('oVirtApiVersion'))
       if (oVirtVersion.get('passed')) {
+        const page = yield select(state => state.vms.get('page'))
```
sjd78: Nit-pick: No real need for a constant here - could be inline

PR #1061, opened 2019-07-16
Allow starting dev server without opening browser

sjd78:
Simple change to the `start.js` such that if the BROWSER env var is set to "none", the script will not attempt to open a browser window/tab to the newly started server.  This is useful when running the server on a remote box that has a GUI so random tabs aren't open every time the server is restarted (and freaking out whoever is sitting at the remote box).

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -58,6 +58,15 @@ to accomplish this. To start the server:
 When asked, provide a valid user name (in the form of `user@domain`) and password so
 the application can start in the context of a logged in user.
 
+When the dev server is started, it will attempt to open a new browser window/tab on
+your system default browser to the app's running URL.  This behavior can be modified
+by specifying the `BROWSER` environment variable.  Possible values are:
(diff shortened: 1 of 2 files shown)


DEVELOPERS.md
```diff
@@ -58,6 +58,14 @@ to accomplish this. To start the server:
 When asked, provide a valid user name (in the form of `user@domain`) and password so
 the application can start in the context of a logged in user.
 
+When the dev server is started, it will attempt to open a new browser window/tab on
+your system default browser to the app's running URL.  This behavior can be modified
+by specifying the `BROWSER` environment variable.  Possible values are:
+    BROWSER=none               # disable the feature
+    BROWSER=google-chrome      # open a new tab in chrome on Linux
+    BROWSER='google chrome'    # open a new tab in chrome on MaxOS
```
bond95: Nit-pick: I think it would be nice also add version for windows:

`BROWSER=chrome`

PR #1063, opened 2019-07-19
Refactor App init code (redux store create and config) and OS icon caching

sjd78:
Redux store refactoring:
 - All of the redux store creation and configuration, including saga middleware and ovirt API activity listening, is now done only in **store.js**.

Made the OS icon handling more consistent with the other entities:
 - Moved the code that loads cached OS icons (from browser local storage) to the login saga.
 - Refactored the OS icon handling to **src/saga/osIcon.js**.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,12 +22,9 @@ import { getSelectedMessages, locale } from '_/intl'
 import configureStore from '_/store'
 import Selectors from '_/selectors'
 import AppConfiguration, { readConfiguration } from '_/config'
-import { loadStateFromLocalStorage } from '_/storage'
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -111,20 +91,9 @@ function addLinkElement (rel: string, href: string) {
...
-      return
-    }
-    if (eventType === 'STOP') {
-      store.dispatch(delayedRemoveActiveRequest(requestId))
-    }
-  })
-}
-
-function SagaErrorBridge () {
+function SagaErrorBridge (storeRootTask: any) {
```
bond95: Please, use more specific type. Otherwise flowtype is just useless.

Issue #946, opened 2019-01-31
Make dashboard more responsive


bond95:
It looks like the VM name and status get lost once the screen reaches a certain size however and that seems like important information to still have available. My one question is would it make sense to put the action buttons into a kebab menu if the screen size reaches a certain, narrow size?
https://github.com/oVirt/ovirt-web-ui/pull/934#pullrequestreview-198140711

@gregsheremeta @michalskrivanek @sjd78 @lwrigh

gregsheremeta:
The action button row gets chopped pretty good, too.

sjd78:
We never did a design of dashboard on a narrow screen.  A stopgap would be to set a minimum width on the screen and do horizontal scrolling (yuck) if the browser window gets too narrow.  

Although, a lot of the cards could have min-widths set to make a decent accommodation for narrow screens.  They're based purely on percentages right now.

lwrigh:
@bond95 @gregsheremeta @michalskrivanek @sjd78 I can take a pass at a narrower dashboard view if you like. I can also see what a mobile dashboard might look like.

lwrigh:
Here's the link design doc that I put together with some design suggestions for making the dashboard more responsive: https://docs.google.com/document/d/1p_MKG55VBWHPzxYl383nf5mU9Sj5Af11Naj6UXtRJ-4/edit?usp=sharing

PR #1067, opened 2019-07-23
Add kebob actions menu and changed Shutdown button type to default in Dashboard

thaorell:
Fixes: suggestions in #946 and #1069 
Prerequisite: #1030 

Enhancements: For narrow mobile screens (<= 600px), automatically retract the Action menu to a kebob menu
Before: 
screenshot-localhost_3000-2019 07 23-15_32_43

After: 
screenshot-localhost_3000-2019 07 25-12_54_33
screenshot-localhost_3000-2019 07 25-12_54_46

### DIFF
--- a/src/components/VmActions/Action.js
+++ b/src/components/VmActions/Action.js
@@ -88,16 +88,16 @@ Button.propTypes = {
   id: PropTypes.string.isRequired,
 }
 
-const MenuItemAction = ({ id, confirmation, shortTitle, icon, onClick }) => {
+const MenuItemAction = ({ confirmation, shortTitle, icon, onClick, ...rest }) => {
   return <Action confirmation={confirmation}>
     <MenuItem
-      id={id}
       onClick={() => {
         onClick && onClick()
         document.dispatchEvent(new MouseEvent('click'))
       }}
+      {...rest}
     >
-      <span>{shortTitle}</span> { icon }
+      <span>{shortTitle}</span> {icon}
     </MenuItem>
   </Action>
 }
@@ -108,6 +108,8 @@ MenuItemAction.propTypes = {
   shortTitle: PropTypes.string.isRequired,
   icon: PropTypes.node,
   onClick: PropTypes.func,
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
 }
 
 const ActionButtonWraper = (props) => {
@@ -133,4 +135,46 @@ ActionButtonWraper.propTypes = {
   ...Button.propTypes,
 }
 
-export { ActionButtonWraper, MenuItemAction }
+const ActionMenuItemWrapper = (props) => {
+  const { items, actionDisabled, confirmation, shortTitle } = props
+
+  const dangerText = (!actionDisabled && props.className === 'btn btn-danger') ? style['menu-item-danger'] : ''
+
+  // For console button
+  if (items && items.filter(i => i !== null).length > 0) {
+    if (actionDisabled) {
+      return <MenuItemAction
+        shortTitle={shortTitle}
+        id='console-selector'
+        disabled={actionDisabled}
+      />
+    } else {
+      return (<React.Fragment>
+        <MenuItem divider />
+        {items.filter(i => i !== null && !i.actionDisabled).map(item => {
+          return <MenuItemAction key={item.id} {...item} />
+        })
+        }
+        <MenuItem divider />
+      </React.Fragment>)
+    }
+  }
+
+  return (
+    <MenuItemAction
+      id={shortTitle}
+      key={shortTitle}
+      confirmation={!actionDisabled && confirmation}
+      disabled={actionDisabled}
+      className={dangerText}
+      shortTitle={shortTitle}
+    />)
+}
+
+ActionMenuItemWrapper.propTypes = {
+  confirmation: PropTypes.node,
+  items: PropTypes.array,
+  ...Button.propTypes,
+}
+
+export { ActionButtonWraper, MenuItemAction, ActionMenuItemWrapper }
--- a/src/components/VmActions/Action.js
+++ b/src/components/VmActions/Action.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { excludeKeys, DropdownButton, MenuItem } from 'patternfly-react'
+import { excludeKeys, DropdownButton, MenuItem, noop } from 'patternfly-react'
 
 import { hrefWithoutHistory } from '_/helpers'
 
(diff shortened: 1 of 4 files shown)


src/components/VmActions/index.js
```diff
@@ -94,9 +95,34 @@ class VmActions extends React.Component {
...
+    this.updateScreenType()
+  }
+
+  componentWillUnmount () {
+    window.removeEventListener('resize', this.updateScreenType)
+  }
+
+  updateScreenType () {
+    const state = { isMobile: false }
+    if (window.innerWidth <= 600) {
```
bond95: I'm not sure that using this check all the time is a good idea. I think better will be using Bootstrap CSS helpers for hiding some parts for certain resolution: https://getbootstrap.com/docs/3.4/css/#responsive-utilities In that case you don't need to adding a lot of handlers and checks, you'll just hide needless component.

src/components/VmActions/style.css
```diff
@@ -61,3 +61,10 @@
     height: 27px;
 }
 
+.menu-item-danger a {
+    color: #cc0000 !important;
```
bond95: I prefer to avoid `!important`, here it is harmless, but it could soon become a habit. Also `!important` can be replaced with adding additional classes to declaration.

https://uxengineer.com/css-specificity-avoid-important-css/

thaorell: Thanks! Good read!

src/components/VmActions/index.js
```diff
@@ -279,8 +280,21 @@ class VmActions extends React.Component {
         confirm={{ title: msg.yes(), onClick: () => onRemove({ preserveDisks: this.state.removePreserveDisks }) }}
       />)
 
-    return (
-      <div className={`actions-line ${style['left-padding']}`} id={idPrefix}>
+    return (<React.Fragment>
+      <DropdownKebab id='kebab' pullRight className={'visible-xs'}>
```
bond95: It is unnecessary use brackets here, you can write just: `className='visible-xs'`

src/components/VmActions/Action.js
```diff
@@ -133,4 +138,42 @@ ActionButtonWraper.propTypes = {
   ...Button.propTypes,
 }
 
-export { ActionButtonWraper, MenuItemAction }
+const ActionTextWrapper = (props) => {
```
bond95: Nit-pick: maybe better name will be `ActionMenuItemWrapper`?

src/components/VmActions/style.css
bond95: I still think it will be better to remove `!important`

src/components/VmActions/Action.js
thaorell: Thanks! I wasn't sure what a good name for it

lwrigh:
In this case "Shutdown" doesn't need to be red since it's not a destructive action. Only "Remove" should be red. The structure and placement looks good though.

thaorell:
@lwrigh but wouldn't that make the design inconsistent since Shutdown has always been a Danger button??

lwrigh:
> @lwrigh but wouldn't that make the design inconsistent since Shutdown has always been a Danger button??

A danger button shouldn't be there so that's another issue we should address. We can open another issue for it.

thaorell:
@lwrigh I've made the changes. take a look

lwrigh:
Cool that looks good to me. It feels less out of place now to me.

src/components/VmActions/index.js
```diff
@@ -279,8 +280,21 @@ class VmActions extends React.Component {
         confirm={{ title: msg.yes(), onClick: () => onRemove({ preserveDisks: this.state.removePreserveDisks }) }}
       />)
 
-    return (
-      <div className={`actions-line ${style['left-padding']}`} id={idPrefix}>
+    return (<React.Fragment>
+      <DropdownKebab id='kebab' pullRight className='visible-xs'>
+        {actions.map((action) => {
```
sjd78: Suggest:
```jsx
{ actions.map(action => <ActionMenuItemWrapper key={action.id} {...action} />) }
```

src/components/VmActions/style.css
```diff
@@ -61,3 +61,10 @@
     height: 27px;
 }
 
+.menu-item-danger a {
```
sjd78: This works:
```css
:global(.dropdown-menu) > .menu-item-danger > a {
    color: #cc0000;
}
```

sjd78:
As per @thaorell, my #1 issue with the kebab on the left is already handled by #1030.

thaorell:
screenshot-localhost_3000-2019 08 05-10_44_35

lwrigh:
@thaorell Can you include a screenshot that includes the kebab icon in it? It's had to tell if it's aligned with the icon without it in the screenshot. Thank you!

src/components/VmActions/Action.js
```diff
@@ -88,14 +88,16 @@ Button.propTypes = {
   id: PropTypes.string.isRequired,
 }
 
-const MenuItemAction = ({ id, confirmation, shortTitle, icon, onClick }) => {
+const MenuItemAction = ({ id, confirmation, shortTitle, icon, onClick, disabled, className }) => {
```
bond95: Maybe will be better to replace some props with rest props, like:  
```JavaScript
const MenuItemAction = ({ confirmation, shortTitle, icon, onClick, ...rest }) => {
  return <Action confirmation={confirmation}>
    <MenuItem
      onClick={() => {
        onClick && onClick()
        document.dispatchEvent(new MouseEvent('click'))
      }}
     {...rest}
    >
      <span>{shortTitle}</span> { icon }
    </MenuItem>
  </Action>
}
```

sjd78:
@thaorell - also needs a rebase

lwrigh:
> ezgif-5-fd694631d689
> 
> @sjd78 does this look better?
> @lwrigh it seems like the kebab icon is toggled off when clicked

I wouldn't do a tooltip on top of a tool tip, it's kind of overwhelming. I would just list out all of the console options instead or we could have a console button with a dropdown that lists out all of the console options and then have all of the other actions in the kebab menu.

thaorell:
Latest update: 
ezgif-5-418059c75579

isaranova:
The kebob dots show on the left instead on the right where the menu appears.
Screenshot_2019-08-27 VM Portal(3)
firefox-68.0-4.fc30.x86_64

sjd78:
> The kebob dots show on the left instead on the right where the menu appears.
> Screenshot_2019-08-27 VM Portal(3)
> firefox-68.0-4.fc30.x86_64

@isaranova - PR #1030 (which I merged earlier) should have taken care of the kebob icon being on the left. I rebased this PR so it should work now.  Please re-test.

isaranova:
ci build please

isaranova:
@sjd78 No, unfortunately, it's still on the left even though the package has been updated.
ovirt-web-ui-1.6.1-0.20190909git747b7d7.el7.noarch
ovirt-engine-4.4.0-0.0.master.20190907174703.gitd1c8e76.el7.noarch

Screenshot_2019-09-09 VM Portal

isaranova:
ci build please

isaranova:
Kebob is still on the left.
Screenshot_2019-09-20 VM Portal(1)

sjd78:
@thaorell , @isaranova - Fixed the layout issues and updated the PR. Good to test again.

screenshot-localhost_3000-2019 09 25-17_23_13

isaranova:
So, I've found another issue - when shutdowning the VM through the kebob, the VM is down as expected, however the kebob contents change into the old version and the only available option 'run' does not work. VM cannot be started unless I resize the window to the non-kebob size and click on the menu run button.

Screenshot_2019-10-07 VM Portal

firefox-69.0.1-3.fc30.x86_64

sjd78:
@isaranova , @thaorell - Fixed the no-confirm-dialog actions on the kebab modal (i.e. Run works now).  Please test one more time.  Thanks!

ovirt-infra:
All tests passed

isaranova:
LGTM

Issue #977, opened 2019-02-27
Disabled delete button on disks is not gray enough


leistnerova:
Description of problem [include screenshots and browser console logs]:
Disabled delete button on disks is not gray as in snapshots (on running VM). So the difference is bad visible.

VM Portal version number:
master
oVirt version number:
ovirt-engine-4.3.1.2-0.0.master.20190222171227.gita00fd1d.el7.noarch
Browser and version: FF 65.0.1

![screenshot from 2019-02-27 09-23-17](https://user-images.githubusercontent.com/29575480/53475884-63b48080-3a71-11e9-8eae-57888601c395.png)
![screenshot from 2019-02-27 09-23-28](https://user-images.githubusercontent.com/29575480/53475888-67480780-3a71-11e9-88c7-afcc144e36a3.png)

yodem:
@lwrigh Would you like the trash icon in the disks section to be as gray as the snapshots' color?

lwrigh:
> @lwrigh Would you like the trash icon in the disks section to be as gray as the snapshots' color?

I would recommend using the grey featured in snapshots card action buttons since that's the correct grey to use if an action is disabled.

PR #1079, opened 2019-08-05
Disabled delete button on disks and NICs is gray enough

yodem:
Fixes: #977 

Disabled delete button on disks and NICs is as gray as the play button in snapshots.

Snapshots

disks


yodem:
@lwrigh Can you please check this commit? I cant add you as a reviewer so im tagging you here :)

sgratch:
@lwrigh @yodem 
The issue here is: 
1. Does changing the delete option to a lighter grey in case of a running VM (with a tooltip explaining why removing is disabled) is applied also with PF4 conventions? Currently it's colored in a darker grey
2. If 1 is approved by Laura then I guess we'll need to change it for NICs as well

lwrigh:
> @lwrigh @yodem
> The issue here is:
> 
> 1. Does changing the delete option to a lighter grey in case of a running VM (with a tooltip explaining why removing is disabled) is applied also with PF4 conventions? Currently it's colored in a darker grey
> 2. If 1 is approved by Laura then I guess we'll need to change it for NICs as well

We definitely need some sort of tooltip with an explanation explaining why the button is disabled if the button persists. What exact use case is this for? Does this happen when the VM comes from a pool or is it because of permissions?  

I would avoid applying PF4 to just one button here. If we do one button in PF4 we should do all of them. 

@sgratch @yodem

sgratch:
@lwrigh The scenario is just trying to delete a disk or a NIC for a running VM. 
The current status is that there is a tool-tip and the delete option is disabled as required, but the delete option color is a more darker grey than in other places

sgratch:
Current state:

Screenshot from 2019-08-06 14-38-09

sgratch:
@yodem  please change title to "Disabled delete button on disks **and nics**...."

lwrigh:
> @lwrigh The scenario is just trying to delete a disk or a NIC for a running VM.
> The current status is that there is a tool-tip and the delete option is disabled as required, but the delete option color is a more darker grey than in other places

Okay that makes sense to me.

michalskrivanek:
"can't delete at this time" is not very helpful though....not much different from nothing. I mean...you try and you see that you can't right now do that:) would be better to actually describe why

lwrigh:
> "can't delete at this time" is not very helpful though....not much different from nothing. I mean...you try and you see that you can't right now do that:) would be better to actually describe why

What about "NICs cannot be deleted when the VM is running."?

michalskrivanek:
@lwrigh much better! (if it's true, of course:). Now it explains why and gives a hint to the user what needs to be done

sjd78:
@yodem - This PR needs a rebase (on upstream/master so the commit for "Asleep Changed to Suspended" can be dropped since it already got merged) and commits to be squashed.

lwrigh:
@yodem Could you post current screenshots of it? Thank you!

yodem:
Nothing has seriously changed, besides the tooltip text. 
GrayText

lwrigh:
> Nothing has seriously changed, besides the tooltip text.
> GrayText

Awesome it looks good to me!

isaranova:
LGTM

sgratch:
Verified by QE and therefore can be merged

PR #1084, opened 2019-08-08
About Dialog - Report issues URL added to branding

yodem:
So far, both in RHV and oVirt, the -"About" window dialog included a reference to github to report issues. However In RHV the customers supposed to report to access.redhat.com in case of a problem.
Now the report reference depends on what brand do you use.

Reports1

Fixed  https://bugzilla.redhat.com/show_bug.cgi?id=1724959

### DIFF
--- a/branding/fixed-strings.json
+++ b/branding/fixed-strings.json
@@ -3,5 +3,7 @@
   "LEGAL_INFO": "Released under",
   "LEGAL_INFO_LINK_TEXT": "Apache License 2.0",
   "LEGAL_INFO_LINK_URL": "https://github.com/oVirt/ovirt-web-ui/blob/master/LICENSE",
-  "DOCUMENTATION_LINK": "https://ovirt.org/documentation"
+  "DOCUMENTATION_LINK": "https://ovirt.org/documentation",
+  "ISSUES_TRACKER_URL": "https://github.com/oVirt/ovirt-web-ui/issues",
+  "ISSUES_TRACKER_TEXT": "Github Issue Tracker"
 }
--- a/src/GlobalErrorBoundary.js
+++ b/src/GlobalErrorBoundary.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { msg } from '_/intl'
 import AppConfiguration from './config'
 import ErrorContent from '_/components/ErrorContent'
-import { resourcesUrls } from '_/branding'
+import * as branding from '_/branding'
 
 class GlobalErrorBoundary extends React.Component {
   constructor (props) {
@@ -24,28 +24,36 @@ class GlobalErrorBoundary extends React.Component {
   }
 
   render () {
+    const trackText = branding.fixedStrings.ISSUES_TRACKER_TEXT || 'Github Issue Tracker'
+    const trackUrl = branding.fixedStrings.ISSUES_TRACKER_URL || 'https://github.com/oVirt/ovirt-web-ui/issues'
+    // const title = msg.globalErrorBoundaryTitle()
+    const descr = msg.globalErrorBoundaryDescription({
+      bugUrl: `<a href='${trackUrl}'>${trackText}</a>`,
+    })
+    const logOut = msg.logOut()
+    const refresh = msg.refresh()
+    const refreshUrl = AppConfiguration.applicationURL
+    const logoutUrl = AppConfiguration.applicationURL + '/sso/logout'
     if (this.state.hasError) {
       return (
         <div>
           <nav className='navbar obrand_mastheadBackground obrand_topBorder navbar-pf-vertical'>
             <div className='navbar-header'>
               <a href='/' className='navbar-brand obrand_headerLogoLink' id='pageheader-logo'>
-                <img className='obrand_mastheadLogo' src={resourcesUrls.clearGif} />
+                <img className='obrand_mastheadLogo' src={branding.resourcesUrls.clearGif} />
               </a>
             </div>
           </nav>
           <ErrorContent
             title={msg.globalErrorBoundaryTitle()}
-            description={msg.globalErrorBoundaryDescription({
-              bugUrl: `<a href='https://github.com/oVirt/ovirt-web-ui/issues'>${msg.gitHub()}</a>`,
-            })}
+            description={msg.globalErrorBoundaryDescription(descr)}
             leftButton={{
-              href: AppConfiguration.applicationURL,
-              title: msg.refresh(),
+              href: logoutUrl,
+              title: logOut,
             }}
             rightButton={{
-              href: AppConfiguration.applicationURL + '/sso/logout',
-              title: msg.logOut(),
+              href: refreshUrl + '/sso/logout',
+              title: refresh,
             }}
           />
         </div>
--- a/src/components/About.js
+++ b/src/components/About.js
@@ -37,6 +37,11 @@ class AboutDialog extends React.Component {
 
     const { oVirtApiVersion } = this.props
     const idPrefix = `about`
+    const trackText = fixedStrings.ISSUES_TRACKER_TEXT || 'Github Issue Tracker'
+    const trackUrl = fixedStrings.ISSUES_TRACKER_URL || 'https://github.com/oVirt/ovirt-web-ui/issues'
+    const reportLink = msg.aboutDialogReportIssuesLink({
+      link: `<strong><a href='${trackUrl}' target='_blank' id='${idPrefix}-issues-link'>${trackText}</a></strong>`,
+    })
 
     let apiVersion = 'unknown'
     if (oVirtApiVersion && oVirtApiVersion.get('major')) {
@@ -67,6 +72,9 @@ class AboutDialog extends React.Component {
                       <span dangerouslySetInnerHTML={{ __html: docLink }} />
                     </li>
                   }
+                  <li id={`${idPrefix}-issues`}>
+                    <div dangerouslySetInnerHTML={{ __html: reportLink }} />
+                  </li>
                 </ul>
               </div>
 
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -43,6 +43,7 @@ export const messages: { [messageId: string]: MessageType } = {
   authorizationExpired: 'Authorization expired. The page is going to be reloaded to re-login.',
   automaticPoolsNotEditable: 'The pool type of {poolName} is automatic so the details of this virtual machine are not editable.',
   availableVmsFromPool: 'Available VMs from this Pool',
+  aboutDialogReportIssuesLink: 'Please report issues on {link}',
   bootMenu: 'Boot Menu',
   bootMenuTooltip: 'Boot menu allows to select bootable device. It is accessible from a console.',
   bootMenuWarning: 'All changes will take effect after reboot only.',


src/intl/messages.js
```diff
@@ -221,7 +221,7 @@ export const messages: { [messageId: string]: MessageType } = {
     description: 'Being hibernated. One of states of a virtual machine. Other are e.g. Up, Down, Powering-Up',
   },
   enum_VmStatus_suspended: {
-    message: 'Asleep',
+    message: 'Suspended',
```
bond95: I guess this commit is from other PR

src/components/VmDetails/CardEditButton.js
```diff
@@ -44,9 +46,11 @@ class CardEditButton extends React.Component {
     const myClick = editEnabled ? noop : this.enableEditHandler
 
     return (
-      <a title={tooltip} onClick={(e) => { e.preventDefault(); myClick() }} className={classes} id={id}>
-        <Icon type='pf' name='edit' />
-      </a>
+      <OverlayTrigger overlay={<Tooltip id={`${id}-edit-tooltip`}>{tooltip}</Tooltip>} placement='top' id={id} onClick={(e) => { e.preventDefault(); myClick() }} >
```
bond95: This one too.

yodem:
I working on in, i figured out what is my mistake and i'll fix it :)

branding/fixed-strings.json
```diff
@@ -2,5 +2,8 @@
   "BRAND_NAME": "oVirt",
   "LEGAL_INFO": "Released under",
   "LEGAL_INFO_LINK_TEXT": "Apache License 2.0",
-  "LEGAL_INFO_LINK_URL": "https://github.com/oVirt/ovirt-web-ui/blob/master/LICENSE"
+  "LEGAL_INFO_LINK_URL": "https://github.com/oVirt/ovirt-web-ui/blob/master/LICENSE",
+  "GITHUB_ISSUES_TRACKER_URL": "https://github.com/oVirt/ovirt-web-ui/issues",
```
bond95: I think will be enough leave attributes with name `ISSUES_TRACKER_URL` and `ISSUES_TRACKER_NAME`. And then use it in `About` component.

src/components/About.js
```diff
@@ -57,7 +57,7 @@ class AboutDialog extends React.Component {
                 <ul className='list-unstyled'>
                   <li id={`${idPrefix}-version`}>Version <strong id={`${idPrefix}-version-value`}>{Product.version}-{Product.release}</strong></li>
                   <li id={`${idPrefix}-apiversion`}>{fixedStrings.BRAND_NAME} API Version <strong id={`${idPrefix}-apiversion-value`}>{apiVersion}</strong></li>
-                  <li id={`${idPrefix}-issues`}>Please report issues on <strong><a href='https://github.com/oVirt/ovirt-web-ui/issues' target='_blank' id={`${idPrefix}-issues-link`}>GitHub Issue Tracker</a></strong></li>
+                  <li id={`${idPrefix}-issues`}>Please report issues on <strong><a href={fixedStrings.BRAND_NAME === 'oVirt' ? fixedStrings.GITHUB_ISSUES_TRACKER_URL : fixedStrings.REDHAT_CUSTOMER_PORTAL_URL} target='_blank' id={`${idPrefix}-issues-link`}>{fixedStrings.BRAND_NAME === 'oVirt' ? 'GitHub Issue Tracker' : 'Redhat Customer Portal'}</a></strong></li>
```
bond95: Like this, for example:
```JavaScript
<li id={`${idPrefix}-issues`}>Please report issues on <strong><a href={fixedStrings.ISSUES_TRACKER_URL} target='_blank' id={`${idPrefix}-issues-link`}>{fixedStrings.ISSUES_TRACKER_NAME}</a></strong></li>
```

yodem: ohhh and then it will automatically change because RHV fixedStrings will override oVirt fixedStrings?

branding/fixed-strings.json
bond95: Please, update `fixed-strings.json`, I mean, remove `REDHAT_CUSTOMER_PORTAL_URL` and `GITHUB_ISSUES_TRACKER_URL` and add `ISSUES_TRACKER_URL`, `ISSUES_TRACKER_TEXT`

src/components/About.js
bond95: RHV fixedStrings replace oVirt

yodem: oh I forgot to add fixedstrings to the commit...

sgratch:
Can you please add a screenshot of how it looks like in RHV now?

michalskrivanek:
then there should be a condition too in case they are not defined. we can't easily add a package dependency

michalskrivanek:
...though, if it doesn't explode the whole page it's probably fine. it's not too important

yodem:
Is there anything else to do about this issue?

michalskrivanek:
well, to check if it does or does not explode when not defined:)

yodem:
checks added :)

src/components/About.js
```diff
@@ -57,7 +57,9 @@ class AboutDialog extends React.Component {
                 <ul className='list-unstyled'>
                   <li id={`${idPrefix}-version`}>Version <strong id={`${idPrefix}-version-value`}>{Product.version}-{Product.release}</strong></li>
                   <li id={`${idPrefix}-apiversion`}>{fixedStrings.BRAND_NAME} API Version <strong id={`${idPrefix}-apiversion-value`}>{apiVersion}</strong></li>
-                  <li id={`${idPrefix}-issues`}>Please report issues on <strong><a href='https://github.com/oVirt/ovirt-web-ui/issues' target='_blank' id={`${idPrefix}-issues-link`}>GitHub Issue Tracker</a></strong></li>
+                  {fixedStrings.ISSUES_TRACKER_URL && fixedStrings.ISSUES_TRACKER_TEXT &&
+                    <li id={`${idPrefix}-issues`}>Please report issues on <strong><a href={fixedStrings.ISSUES_TRACKER_URL} target='_blank' id={`${idPrefix}-issues-link`}>{fixedStrings.ISSUES_TRACKER_TEXT}</a></strong></li>
```
sjd78: The "Please report issues on {link}" text should probably be externalized to `src/intl/messages.js`

src/GlobalErrorBoundary.js
```diff
@@ -25,9 +25,9 @@ class GlobalErrorBoundary extends React.Component {
 
   render () {
     const title = msg.globalErrorBoundaryTitle()
-    const descr = msg.globalErrorBoundaryDescription({
-      bugUrl: `<a href='https://github.com/oVirt/ovirt-web-ui/issues'>${msg.gitHub()}</a>`,
-    })
+    const descr = branding.fixedStrings.ISSUES_TRACKER_URL ? msg.globalErrorBoundaryDescription({
```
sjd78: This is a bit awkward.
Like in the About box, if the tracker isn't in branding, the "report issues" bit should be hidden instead of an empty link.

Maybe have a `const showDescription = branding.fixedString.ISSUES_TRACKER URL && .....ISSUES_TRACKER_TEXT` and then the `descr` and the html would be wrapped by that constant.

src/components/About.js
yodem: i agree on the  "Please report issues on " but i dont think that you need to add the link. The link has <a>,<strong> tags and id which the "Please report issues on " dont have. and to insert it all as a json object and to pass it through a function make the code more complex instead of leave it as is, which is readable and clear

src/GlobalErrorBoundary.js
yodem: the descr need to be wrapped as well? i cant wrap him inside the return section.
I can check if showDescription exists before creating the descr, but it needs to be before the return statement.
right?

bond95:
I think, it will be nice to add default values in case if `ISSUES_TRACKER_URL` and `ISSUES_TRACKER_TEXT` are not defined.
@yodem @sjd78 @michalskrivanek WDYT?

yodem:
> I think, it will be nice to add default values in case if `ISSUES_TRACKER_URL` and `ISSUES_TRACKER_TEXT` are not defined.
> @yodem @sjd78 @michalskrivanek WDYT?

I thought about it, but what would be the default value?

bond95:
> I thought about it, but what would be the default value?

As it was before, ovirt-web-ui github.

yodem:
> As it was before, ovirt-web-ui github.

ok

src/GlobalErrorBoundary.js
```diff
@@ -24,9 +24,11 @@ class GlobalErrorBoundary extends React.Component {
   }
 
   render () {
+    const trackText = branding.fixedStrings.ISSUES_TRACKER_TEXT ? branding.fixedStrings.ISSUES_TRACKER_TEXT : 'https://github.com/oVirt/ovirt-web-ui/issues'
```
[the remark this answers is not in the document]
sjd78: You have the default value as the URL instead of the text.

src/GlobalErrorBoundary.js
```diff
@@ -24,9 +24,11 @@ class GlobalErrorBoundary extends React.Component {
...
+    const trackUrl = branding.fixedStrings.ISSUES_TRACKER_URL ? branding.fixedStrings.ISSUES_TRACKER_URL : 'Github Issue Tracker'
```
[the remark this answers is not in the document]
sjd78: You have the default value as the text instead of the URL.

src/components/About.js
```diff
@@ -37,6 +37,8 @@ class AboutDialog extends React.Component {
 
     const { oVirtApiVersion } = this.props
     const idPrefix = `about`
+    const trackText = fixedStrings.ISSUES_TRACKER_TEXT ? fixedStrings.ISSUES_TRACKER_TEXT : 'https://github.com/oVirt/ovirt-web-ui/issues'
```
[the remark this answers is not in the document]
sjd78: You have the default value as the URL instead of the text.

src/components/About.js
```diff
@@ -37,6 +37,8 @@ class AboutDialog extends React.Component {
...
+    const trackUrl = fixedStrings.ISSUES_TRACKER_URL ? fixedStrings.ISSUES_TRACKER_URL : 'Github Issue Tracker'
```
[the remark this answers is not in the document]
sjd78: You have the default value as the text instead of the URL.

yodem:
> Thanks for making the changes to the translation key and text.
> 
> @yodem - Any particular reason to avoid the style change away from the ternary to just the `||` for the default `trackText` and `trackUrl` constants? The form `const a = data.value || 'constant'` is quite common for default values when a variable is not set.

Good to know, I'll change it.

sgratch:
ci build please

sjd78:
@yodem - There are lint errors causing the tests (and build) to fail: https://travis-ci.org/oVirt/ovirt-web-ui/builds/607522440#L275

sgratch:
@yodem moving the ON_QA label since QE can't test till build is passed

sgratch:
@leistnerova @isaranova Do you want to verify this fix before merging or do you prefer to test it only after a d/s build is ready?

src/components/NoLogin/index.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-
-import { Button } from '@patternfly/react-core'
+import { Button } from 'patternfly-react'
```
sjd78: This is an unrelated change that should be removed.

src/index.js
```diff
@@ -13,7 +13,6 @@ import { type Task } from 'redux-saga'
 
 import '_/logger' // initialize our console logging overlay
 
-import '@patternfly/react-core/dist/styles/base.css'
```
sjd78: This is an unrelated change that should be removed.

leistnerova:
> @leistnerova @isaranova Do you want to verify this fix before merging or do you prefer to test it only after a d/s build is ready?
We will test it on d/s so it can be merged now.

Issue #949, opened 2019-02-05
Change hover texts in masthead


leistnerova:
Hover text over bell icon is 'Messages' but drawer head contains 'Notifications'.
Hover text over user icon doesn't contain profile like in admin portal (e.g. admin X admin@internal-authz)
And refresh icon doesn't have any hover text.

![screenshot from 2019-02-05 09-48-33](https://user-images.githubusercontent.com/29575480/52262126-d6778380-292b-11e9-89b0-89ffe2e60614.png)

yodem:
@bond95 @sjd78 What do you think we should do? I mean, the problem is clear, but what is supposed to be the solution? To add hover text? And what text?

sgratch:
This issue is depended on issue #1095

yodem:
> This issue is depended on issue #1095

I solved it using patternfly 4 conventions as laura said

yodem:
> This issue is depended on issue #1095

So i'll wait for the merge of this issue and then i'll change the commit

PR #1086, opened 2019-08-12
Hover texts in masthead fixed and changed

yodem:
Fixes: #949

The texts on the masthead changed and designed like patternfly 4 tooltips.
admin_Hovering
Notifications_Hover
Refresh_Hovering

### DIFF
--- a/src/components/VmsPageHeader/UserMenu.js
+++ b/src/components/VmsPageHeader/UserMenu.js
@@ -1,6 +1,5 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-
 import { connect } from 'react-redux'
 
 import { logout } from '_/actions'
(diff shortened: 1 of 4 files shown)


sjd78:
> I think the `OverlayTrigger` should directly wrap the `a` tag in each case to be consistent and make sure it doesn't stay there with About/Options dropdown menu.

Location of the Tooltips look good now.

> Generally looks good, but now we have a lot of cases where we using `Tooltip` with `OverlayTrigger`, what means a lot of similar code. So I think it will be better to create component with that code and use it instead.

@yodem - You can create a component like this, drop it in `src/components`, and use it as Bohdan suggests...
```jsx
const OverlayTooltip = ({ id, tooltip, placement = 'bottom', children, ...rest }) =>
  <OverlayTrigger
    overlay={
      <Tooltip id={id}>{tooltip}</Tooltip>
    }
    placement={placement}
    {...rest}
  >
    { children }
  </OverlayTrigger>
```

sgratch:
@yodem @lwrigh Regarding hover text over user icon: do we want it to include only "admin" or also the provider, as appears on webAdmin:
Screenshot from 2019-08-25 15-04-23

yodem:
> @yodem @lwrigh Regarding hover text over user icon: do we want it to include only "admin" or also the provider, as appears on webAdmin:
> Screenshot from 2019-08-25 15-04-23

I think that admin is enough, adding the provider make it more complex and I think its unnecessary.

yodem:
> > I think the `OverlayTrigger` should directly wrap the `a` tag in each case to be consistent and make sure it doesn't stay there with About/Options dropdown menu.
> 
> Location of the Tooltips look good now.
> 
> > Generally looks good, but now we have a lot of cases where we using `Tooltip` with `OverlayTrigger`, what means a lot of similar code. So I think it will be better to create component with that code and use it instead.
> 
> @yodem - You can create a component like this, drop it in `src/components`, and use it as Bohdan suggests...
> 
> ```js-jsx
> const OverlayTooltip = ({ id, tooltip, placement = 'bottom', children, ...rest }) =>
>   <OverlayTrigger
>     overlay={
>       <Tooltip id={id}>{tooltip}</Tooltip>
>     }
>     placement={placement}
>     {...rest}
>   >
>     { children }
>   </OverlayTrigger>
> ```

see #1098

lwrigh:
I would tend to agree with @yodem with just having "Admin" as the hover text. Having the provider also seems like overkill.

src/components/VmsPageHeader/UserMenu.js
```diff
@@ -13,9 +14,11 @@ const UserMenu = ({ config, onLogout }) => {
   const idPrefix = 'usermenu'
   return (
     <li className='dropdown'>
-      <a className='dropdown-toggle nav-item-iconic' href='#' data-toggle='dropdown' title={config.getIn(['user', 'name'])} id={`${idPrefix}-user`}>
-        <i className='pficon pficon-user' /><span className='caret' />
-      </a>
+      <OverlayTrigger overlay={<Tooltip>{config.getIn(['user', 'name'])}</Tooltip>} placement='bottom'>
```
sjd78: This should be updated to use `OverlayTooltip` like the others.

isaranova:
LGTM

PR #1087, opened 2019-08-13
UserSessionTimeOut automatic logout fixed

### LINKED ISSUE #1085: Handle engine option "UserSessionTimeOutInterval" value of <=0 appropriately
VM Portal version number: 1.5.3
oVirt version number: 4.3.5
Browser and version: Any

Steps to Reproduce:
1. Set the "UserSessionTimeOutInterval" to -1 on the engine
2. Login to VM Portal
3. User is auto-logged out immediately

Actual results: If the "UserSessionTimeOutInterval" is set to -1, VM Portal will log out the user immediately.

Expected results: With a value of -1, it is expect the user is never auto logged out.

Additional info:
As per the config value definition in ovirt-engine, "A negative value indicates that sessions never expire.": https://github.com/oVirt/ovirt-engine/blob/e0940bd9b768cb52c78f9e0d0c97afd6de7ac8a5/packaging/etc/engine-config/engine-config.properties#L218-L220

`SessionActivityTracker` needs to be modified to handle the -1 value correctly.

yodem:
Fixes: #1085 

If the user decides that his UserSessionTimeOut = -1, he wouldn't logout.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -82,7 +82,7 @@ function isBrowserUnsupported () {
  * Main App component. Wrap the main react-router components together with
  * the various dialogs and error messages that may be needed.
  */
-const App = ({ history, config, appReady }) => {
+const App = ({ history, config, appReady, activateSessionTracker }) => {
   if (isBrowserUnsupported()) {
     return <UnsupportedBrowser />
   }
@@ -98,7 +98,7 @@ const App = ({ history, config, appReady }) => {
         <OvirtApiCheckFailed />
         <LoadingData />
         <ToastNotifications />
-        { appReady && <SessionActivityTracker /> }
+        { appReady && activateSessionTracker && <SessionActivityTracker /> }
         { appReady && renderRoutes(getRoutes()) }
       </div>
     </ConnectedRouter>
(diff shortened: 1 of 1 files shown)


src/components/SessionActivityTracker.js
```diff
@@ -41,18 +41,19 @@ class SessionActivityTracker extends React.Component {
...
-
-    this.setState(
-      state => ({
-        counter: state.counter - 1,
-        showTimeoutModal: state.counter <= TIME_TO_DISPLAY_MODAL,
-      }),
-      () => {
-        if (this.state.counter <= 0) {
-          this.props.onLogout()
+    if (!(this.props.config.get('userSessionTimeoutInterval') <= 0)) {
```
sjd78: I recommend you don't wrap the `this.setState` call.  Just have another:
```js
if (configInterval <= 0) {
  return
}
```
above the `this.setState` block.

bond95:
> I also recommend updating `src/App.js` so `{ appReady && <SessionActivityTracker /> }` looks more like `{ appReady && enableActivityTracker && <SessionActivityTracker /> }`. And setup `enableActivityTracker` as a prop on `App` with a test in the `connect` call similar to how it does `appReady` but check the appropriate config entry...

Agree, I think it will be better to move check of `userSessionTimeoutInterval` to App.js, because if `userSessionTimeoutInterval <= 0` then `SessionActivityTracker` is completely useless.

sgratch:
@bond95 @sjd78 please review this

pnovotny:
ci build por favor

pnovotny:
ci build please

PR #1092, opened 2019-08-15
Full-Screen console

### LINKED ISSUE #1032: [RFE] webvnc console should have a "full screen" mode
When I tested webvnc console from my smaller screened Chrome book, the page looks a bit dense:
Screenshot 2019-05-22 at 6 20 29 PM

Screenshot 2019-05-22 at 6 31 14 PM

_Originally posted by @sjd78 in https://github.com/oVirt/ovirt-web-ui/issues/490#issuecomment-494998708_

Adding functionality to be able to have the console display take up the full browser window would be a good start.  It does not need to jump from console with the app masthead etc around it to true full screen.  Just taking out the masthead etc would be good.  The user can put their browser in full screen mode themselves to get "true" full screen.

yodem:
Fixes #1032 
'Full Screen console' feature is now available - 

Before:
FullScreen1

After:
FullScreen2


sgratch:
Please review this

lwrigh:
Once the user enters full screen do they have the option to go back to the original screen size?

yodem:
Yes

lwrigh:
Desktop HD
I would suggest adding an alert that lingers for like 5 seconds or so at the top of the screen when the user enters full screen just so they know that they can press the "Esc" button to exit full screen mode.

yodem:
I think that the browser does it by himself

lwrigh:
Can you test it out on all of the browsers just to be sure? There might be some differences depending on the browser.

yodem:
Ok, not all of the browsers pop a notification, but the exit way is to click F11.
So can we perhaps add a tooltip instead of a popup?

yodem:
Ok, and yes F11 is the alternative.

yodem:
Console

this is how it looks now, the notification pop up is toasted only at the first time

bond95:
> Console
> 
> this is how it looks now, the notification pop up is toasted only at the first time

I think it will be better if you set for this alert `position: absolute` and with `top:` property move it slightly lower.

lwrigh:
I would change the message to the one that's pictured in the wireframe and make the toast notification less wide.

sjd78:
ci test please

bond95:
image
@lwrigh

lwrigh:
> image
> @lwrigh

Can the text be left aligned instead of center aligned?

bond95:
@lwrigh 
image

sjd78:
Ok, so the functionality that is implemented is to go from VM Portal with the full mast head + console actions + console to full screen like hitting F11.  Here is my question.  

What about the console taking up the full browser page / tab?  That is what I actually intended in #1032.

@lwrigh , @yodem , @bond95 , @sgratch

lwrigh:
> Ok, so the functionality that is implemented is to go from VM Portal with the full mast head + console actions + console to full screen like hitting F11. Here is my question.
> 
> What about the console taking up the full browser page / tab? That is what I actually intended in #1032.
> 
> @lwrigh , @yodem , @bond95 , @sgratch

Full screen is my mind means taking up the full width of the screen so I would agree with Scott's comment.

bond95:
@lwrigh @sjd78 I don't think that to make console a little bit bigger in tab is so very needed. Because for me full-screen mode means that user will use one console for a long time, that's why I think that full-screen is more useful than full-tab mode.

sjd78:
> @lwrigh @sjd78 I don't think that to make console a little bit bigger in tab is so very needed. Because for me full-screen mode means that user will use one console for a long time, that's why I think that full-screen is more useful than full-tab mode.

@bond95 - For a large screen workstation, full screen may not even be that useful.  Launching a SPICE client is far more performant in that case.  But consider this case (screen shot from a small screen chromebook):
Screenshot 2019-05-22 at 6 31 14 PM

As a user, I would like to see as much of the console as possible but still be able to retain use of the browser tabs etc.  A user can already "F11" the browser to full screen mode, if they like once the console is full page (or fuller paged).

@lwrigh - Can you mockup a "full page" mode based on the above screenshot?  Also, can you ask around the UXD folks to see what they think?

lwrigh:
Desktop HD Copy 11

Here's what I'm thinking the full screen would look like.

sjd78:
> 
> Here's what I'm thinking the full screen would look like.

@lwrigh .. Ok I finally understand the confusion @yodem was having with making the console "full screen".  Your mockup implies changing the zoom and/or aspect ratio in the existing content pane space such that there would be no letter boxing (black bars on top/bottom) or pillar boxing (black bars on left/right).

The current solution that pushes the console to be full screen like hitting F11.  I'm ok with this as it does the job.  (Try using the VNC browser console in cockpit with it can be painful when the resolution on the VM is greater than the available space on the page - it's painful).

src/components/CounterAlert.js
```diff
@@ -0,0 +1,59 @@
...
+    const state = {
+      counter: this.state.counter - 1,
+    }
+    if (this.state.counter <= 0) {
+      clearInterval(this.timer)
+      this.timer = null
+      this.handleDismiss()
+    }
+    this.setState(state)
+  }
```
sjd78: Should probably use the same `decrementCounter` style as in `SessionActivityTracker`, where the checking code is done in the `setState` callback...
https://github.com/oVirt/ovirt-web-ui/blob/840e846f852971d411b6687f7f1f07bac14a5e79/src/components/SessionActivityTracker.js#L40-L56

```jsx
```

isaranova:
LGTM

PR #1098, opened 2019-08-21
Tooltip and OverlayTrigger component

### LINKED ISSUE #1095: Tooltip Code Refactor
As @sjd78 and @bond95 said -
Every time we use tooltip we use OverlayTrigger tag which includes the tooltip tag.
it makes the code more complicated and if we have a case where we use specific tags together couple of time we can create a component of the both of them together.
See https://github.com/oVirt/ovirt-web-ui/pull/1086#pullrequestreview-275397387

yodem:
Fixed #1095 

Tooltip tag and OverlayTrigger tag combined to a new component.

### DIFF
--- a/src/components/OverlayTooltip/index.js
+++ b/src/components/OverlayTooltip/index.js
@@ -1,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { OverlayTrigger, Tooltip } from 'patternfly-react'
+
+const OverlayTooltip = ({ id, tooltip, placement = 'left', children }) => {
+  return (
+    <OverlayTrigger
+      overlay={
+        <Tooltip id={id}>{tooltip}</Tooltip>
+      }
+      placement={placement}
+    >
+      { children }
+    </OverlayTrigger>
+  )
+}
+OverlayTooltip.propTypes = {
+  id: PropTypes.string.isRequired,
+  tooltip: PropTypes.string.isRequired,
+  placement: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+}
+
+export default OverlayTooltip
--- a/src/components/VmDetails/cards/DisksCard/DiskListItem.js
+++ b/src/components/VmDetails/cards/DisksCard/DiskListItem.js
@@ -10,10 +10,11 @@ import { escapeHtml } from '../../../utils'
 import itemStyle from '../../itemListStyle.css'
 import style from './style.css'
 
-import { Icon, OverlayTrigger, Tooltip, Label } from 'patternfly-react'
+import { Icon, Label } from 'patternfly-react'
 import DeleteConfirmationModal from '../../../VmModals/DeleteConfirmationModal'
 import DiskStateIcon from './DiskStateIcon'
 import DiskImageEditor from './DiskImageEditor'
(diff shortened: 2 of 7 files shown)


sgratch:
@bond95 @sjd78  Please review

src/components/OverlayTooltip/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { OverlayTrigger, Tooltip } from 'patternfly-react'
+
+class OverlayTooltip extends React.Component {
+  // constructor (props) {
```
bond95: Remove, please, comments

src/components/OverlayTooltip/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { OverlayTrigger, Tooltip } from 'patternfly-react'
+
+class OverlayTooltip extends React.Component {
```
bond95: I think this component can be a function

src/components/VmDetails/cards/DisksCard/DiskListItem.js
```diff
@@ -95,29 +93,23 @@ const DiskListItem = ({
           storageDomainList={storageDomainList}
           onSave={onEdit}
           trigger={
-            <OverlayTrigger
-              overlay={<Tooltip id={`${idPrefix}-action-edit-tooltip`}>{msg.diskEditTooltip()}</Tooltip>}
-              placement='left'
-            >
+            <OverlayTooltip id={`${idPrefix}-action-edit-tooltip`} tooltip={msg.diskEditTooltip()} placement='left'>
```
bond95: As I see `placement='left'` is the most popular, so I think it will be better to make this as default value, for `placement`

src/components/OverlayTooltip/index.js
```diff
@@ -0,0 +1,33 @@
...
+  //   var id = props.id
+  //   var tooltip = props.tooltip
+  //   var placement = props.placement
+  //   var children = props.children
+  // }
+  render () {
+    return (
+      <OverlayTrigger
+        overlay={
+          <Tooltip id={this.props.id}>{this.props.tooltip}</Tooltip>
```
bond95: nit-pick: I think it will be more readable if you use `const { id, tooltip, placement, children } = this.prop`, instead of writing `this.props` every time.

src/components/OverlayTooltip/index.js
```diff
@@ -0,0 +1,33 @@
...
+        { this.props.children }
+      </OverlayTrigger>
+    )
+  }
+}
+OverlayTooltip.propTypes = {
+  id: PropTypes.string.isRequired,
+  tooltip: PropTypes.string.isRequired,
+  placement: PropTypes.string.isRequired,
+  children: PropTypes.node,
```
bond95: Why children is not required? Because I don't see any case of empty `children` in the code.

src/components/VmDetails/cards/DisksCard/DiskListItem.js
bond95: `placement` can be removed as well as down below, because `left` is default value.

sgratch:
@sjd78 @bond95 does this PR require QE testing?

Issue #1006, opened 2019-04-11
Extend By Size tooltip


michalskrivanek:
when editing disk you can extend the size by xyz GB. That's all good, but people may not realize it doesn't resize the filesystem. It would be great to have a tooltip explaining that after the resize you are responsible to grow the filesystem inside the guest.

yodem:
@lwrigh @sjd78 @bond95 What do you think the text is supposed to be?

yodem:
@lwrigh  Where do you think the tooltip should be?(ignore the text content, its for demonstration)
![ExtendTooltip](https://user-images.githubusercontent.com/13103729/63222707-ccc0b980-c1b3-11e9-8af5-b32610c28527.png)

And is it right for it to be also in the creation window? I think that letting the user know he supposed to do it later on, its important as well.
![ExtendsDisk2](https://user-images.githubusercontent.com/13103729/63222732-1ad5bd00-c1b4-11e9-8806-599d3cd1d65e.png)

yodem:
@michalskrivanek ?

lwrigh:
For the "Edit Disk" modal, we can add an info tip next to the "Extend Size By" label and the info tip text could say, "You can extend the size of the disk but it does not resize the filesystem. After resizing the disk you must expand the file system inside the guest separately." @michalskrivanek Does that sound accurate?

sgratch:
@michalskrivanek @lwrigh @yodem 
(1) Why we need this tool-tip added in vm portal and not in webadmin as well? 
(2) Will the VM portal's user know how to expand the file system? I would add to the tool-tip something like: "...you will probably need to use the fdisk utility for resizing the partitions and file systems as required. Please refer your administrator for more information on how to do that."

sgratch:
> @lwrigh @sjd78 @bond95 
> #846 Can I combine these two issues in the same PR?
Please see https://github.com/oVirt/ovirt-web-ui/issues/846#issuecomment-522634332

michalskrivanek:
Yotam, your text lgtm

sjd78:
@lwrigh - To be consistent with the rest of the dialog, shouldn't the new text be a Field Level help (the circle-i icon) next to the "Extend Size by" field?  That way it'll match the storage domain and disk type fields.

lwrigh:
@sjd78 Yup that sounds good to me so it matches storage domain and disk type.

PR #1104, opened 2019-08-25
Disk Edit Dialog - Add Tooltip

yodem:
Fixes: #1006

Added tooltip to Edit and Create disk dialog - 

Create - 

ResizeTooltipCreate

Edit - 

ResizeTooltipEdit


lwrigh:
@yodem Can you add a screenshot of when the tooltip is exposed?

yodem:
> @yodem Can you add a screenshot of when the tooltip is exposed?

@lwrigh this is it, when you hover this field

yodem:
Current state -
ExtendSizeTooltip

yodem:
oh i uploaded the wrong image
This is after the changes -
ExtendSizeTooltip2

lwrigh:
> oh i uploaded the wrong image
> This is after the changes -
> ExtendSizeTooltip2

No problem! It looks good to me!

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -332,13 +333,15 @@ class DiskImageEditor extends Component {
...
                 { !createMode && msg.diskEditorResizeLabel() }
               </LabelCol>
-              <Col sm={9}>
-                <FormControl
-                  type='number'
-                  value={this.state.values.size}
-                  onChange={this.changeSize}
-                />
-              </Col>
+              <OverlayTooltip id={`${idPrefix}-resize-tooltip`} tooltip={msg.diskEditorSizeTooltip()} placement='right'>
```
bond95: Maybe better to wrap in `OverlayTooltip` only `FormControl`?

yodem: why?

sjd78: > why?

Because the tooltip is for the FormControl itself and not the entire column.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -332,13 +333,15 @@ class DiskImageEditor extends Component {
                 }
                 { !createMode && msg.diskEditorResizeLabel() }
```
sjd78: I'm thinking that instead of an onHover Tooltip on the field, displaying the message via a FieldLevelHelp popover would be more consistent with the dialog.  So really it would be just copying the style above and using the new tooltip message as the help's content.

src/intl/messages.js
```diff
@@ -134,10 +134,11 @@ export const messages: { [messageId: string]: MessageType } = {
     message: 'Thin Provision',
     description: 'Select list option for qcow / thin provisioned format disk images',
   },
-  diskEditorResizeLabel: 'Extend Size By (GiB)',
+  diskEditorResizeLabel: 'Increase Size By (GiB)',
```
sjd78: Why the wording change?

src/intl/messages.js
```diff
@@ -134,10 +134,11 @@ export const messages: { [messageId: string]: MessageType } = {
...
   diskEditorSizeCantChangeHelp: 'Disk size cannot be extended for this type of disk.',
   diskEditorSizeCreateHelp: 'After you create a disk, you may only extend its size if you make edits.',
-  diskEditorSizeLabel: 'Size (GiB)',
+  diskEditorSizeLabel: 'Current Size (GiB)',
```
sjd78: Changing this key will also change the label on the field for the Create Disk dialog.  To have different text, you'll need another key and adjust the use in `DiskImageEditor`.

For **Create New Disk**, the label should be "Size (GiB)" as it is now.

For **Edit Disk**, the label can say "Current Size (GiB)" or stick with "Size (GiB)".  I could go either way with the one.

@lwrigh , @sgratch - WDYT about the new size / current size field labels?

src/intl/messages.js
yodem: https://github.com/oVirt/ovirt-web-ui/pull/1104#pullrequestreview-283618258

sjd78:
@lwrigh, @bond95, @yodem, @sgratch - I'm thinking that instead of an onHover `OverlayTooltip` on the field, displaying the message via a `FieldLevelHelp` popover would be more consistent with the dialog.

bond95:
> @lwrigh, @bond95, @yodem, @sgratch - I'm thinking that instead of an onHover `OverlayTooltip` on the field, displaying the message via a `FieldLevelHelp` popover would be more consistent with the dialog.

Hmm... Not sure, because mostly `FieldLevelHelp` describe function of field or field state (for example, if field is disabled). But this is not really that case, so I think it is better to leave as it is.

yodem:
@sjd78 @bond95 @lwrigh 
So lets decide -

With the tooltip - 
ResizeTooltipEdit

with the field level help -
ExtendSizeHelper
ExtendSizeHelper2

I think the the field help is more consistent with the dialog, however the user must know the content of the field help so a tooltip would be a better solution. What do you think?

michalskrivanek:
+1 on tooltip

bond95:
+1 on tooltip too

sjd78:
Make the size input field narrower so the tooltip shows on the dialog itself, and that will be good.

lwrigh:
> Make the size input field narrower so the tooltip shows on the dialog itself, and that will be good.

I agree with Scott for decreasing the size of the field so the tool tip fits onto the dialog box. I would also change the wording to say, "After resizing the disk, you must also increase the size of the guest's filesystem."

yodem:
Current state - 

Enough2

isaranova:
LGTM

Issue #713, opened 2018-08-11
notifications unread and show flags cleanup


gregsheremeta:
https://github.com/oVirt/ovirt-web-ui/pull/660#issuecomment-412133377

> The `unread` and `show` flags aren't used anywhere and aren't updated in every case. Either they should be removed, or they should be updated on every operation.

Issue #712, opened 2018-08-11
use uuid instead of Date.now() as a unique id for the messages


gregsheremeta:
https://github.com/oVirt/ovirt-web-ui/pull/660#issuecomment-412133377

> We could do better than Date.now() for a unique id for the messages. Maybe at the very least, wrap it in a uuid() function so we can adjust it to something else in the future. Date.now()'s precision may lead to records that have the same timestamp: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now#Reduced_time_precision

PR #1111, opened 2019-09-02
User messages refactor

bond95:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/712
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/713

 - Removed `unread` and `show` from `userMessages` reducer
 - Added `eventId` to user messages

### DIFF
--- a/src/reducers/userMessages.js
+++ b/src/reducers/userMessages.js
@@ -28,8 +28,6 @@ function addLogEntry ({ state, message, type = 'ERROR', failedAction }) {
 
 const initialState = Immutable.fromJS({
   records: [],
-  unread: false,
-  show: false,
 })
 
 const userMessages = actionReducer(initialState, {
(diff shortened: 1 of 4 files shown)


src/reducers/userMessages.js
```diff
@@ -8,6 +8,7 @@ import {
   SET_USER_MESSAGES,
 } from '_/constants'
 import { actionReducer } from './utils'
+import uniqueId from 'lodash/uniqueId'
```
sjd78: Why not just use:
```js
import { generateUnique } from '_/helpers'
```

bond95: Because `Math.random` could return same numbers, but `uniqueId` use `idCounter` variable so, id will be unique within one session.

sjd78: 🤷‍♂ 

`generateUnique` actually follows the UUID v4 algo.  Shouldn't be any problem there.

bond95: Yes, chance that 2 UUID will be same is very small, but I prefer to leave `uniqeId`, because in this case chance is 0, and we already using lodash.

sjd78: Makes sense.

michalskrivanek:
no QE feedback, but we don't want this to rot

PR #1114, opened 2019-09-05
Allow use of alternate branding by the DEV server

sjd78:
Added instructions for how to run the dev server with alternate branding.


sjd78:
For example, here is my dev server with RHV branding:
screenshot-localhost_3000-2019 09 05-17_24_29

PR #1116, opened 2019-09-07
Update autogen/automake configurations and yarn integration

sjd78:
- Update automake to clean up mock_runner generated files

  - Update yarn scripts to run `./autogen.sh` before `yarn install`

  - Added script `yarn clean` to run automake's `distclean`

  - No need for travis CI to invoke autogen.sh since `yarn install`
    is doing that now

Now when you install `node_modules/`, automake will also update. This
will help cover changes in automake configs and help reduce any related
errors.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,13 +41,21 @@ DISTCLEANFILES = $(PACKAGE)-$(VERSION).tar.gz \
 	configure \
 	install-sh \
 	missing \
-	Makefile.in
+	Makefile.in \
+	.bash_history \
+	mocker-*.cfg
 
 DISTCLEANDIRS = autom4te.cache \
 	engine-plugin \
 	tmp.repos \
 	build \
-	dockerbuild
+	extra \
+	.zanata-cache \
+	dockerbuild \
+	.node-gyp \
+	exported-artifacts \
+	rpmbuild \
+	node_modules
 
 TMPREPOS = tmp.repos
 RPMBUILD_ARGS :=
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ services:
   - docker
 
 script:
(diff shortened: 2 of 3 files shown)


Makefile.am
```diff
@@ -41,13 +41,20 @@ DISTCLEANFILES = $(PACKAGE)-$(VERSION).tar.gz \
 	configure \
 	install-sh \
 	missing \
-	Makefile.in
+	Makefile.in \
+	.bash_history \
```
michalskrivanek: should probably skip this

sjd78: All it will do is let `make distclean` or `yarn clean` be able to clean up extra `mock_runner.sh` generated files.  It is an entirely manual step.  I really only added them here so I could clean up my own project folder after running the `mock_runner.sh`.

michalskrivanek: I've seen complaints on vdsm project about the same...it's annoying shortcoming of stdci.
it's fine with me...just make sure it dosn't wipe out your own user's bash history:)

Makefile.am
```diff
@@ -41,13 +41,20 @@ DISTCLEANFILES = $(PACKAGE)-$(VERSION).tar.gz \
...
 	engine-plugin \
 	tmp.repos \
 	build \
-	dockerbuild
+	extra \
+	.zanata-cache \
+	dockerbuild \
+	.node-gyp \
+	exported-artifacts \
+	rpmbuild
```
bond95: I think it will be nice to add here `node_modules` with `yarn.lock` as well. WDYT?

sjd78: Not `yarn.lock` since it is part of normal source.

`node_modules` does make some sense as you could do...
```
yarn clean
yarn install
yarn build
```
...for a nice really clean build.

sjd78: (done in the current commit)

PR #1118, opened 2019-09-16
Fixed Windows VM creation bug

bond95:
Now if template time zone format or VM time zone format is not compatible with VM OS it will change time zone to default. For Windows default time zone if: `GMT Standard Time`, for other OSs: `Etc/GMT`


src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -327,6 +329,17 @@ class DetailsCard extends React.Component {
...
+
+          const timeZoneName = updates.getIn(['timeZone', 'name'])
+          const isWindowsTimeZone = timezones.find(timezone => timezone.id === timeZoneName)
+          const isWindowsVm = isWindows(os.get('name'))
+
+          if (isWindowsVm && !isWindowsTimeZone) {
+            changeQueue.push({ fieldName: 'timeZone', value: 'GMT Standard Time' })
+          }
+          if (!isWindowsVm && isWindowsTimeZone) {
+            changeQueue.push({ fieldName: 'timeZone', value: 'Etc/GMT' })
```
sgratch: We should not hard-coded the timezone defaults since they are declared as config values "vdcDefaultGeneralTimeZone", "DefaultWindowsTimeZone" on vdc_options table.
I suggest that in case we want to set the default then set the timeZone to null and let the engine backend choose the value for you as set in https://github.com/oVirt/ovirt-engine/blob/2cc0ab0ff6b00b5b51975497eaf1747b29d2a69c/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java#L800.

src/components/VmDialog/index.js
```diff
@@ -316,6 +324,34 @@ class VmDialog extends React.Component {
...
+  checkTimeZone ({ osType }) {
+    let timeZone = null
+    const template = this.getTemplate()
+    if (template) {
+      if (template.getIn(['timeZone', 'name'])) {
+        timeZone = timeZone || template.get('timeZone').toJS()
+        const isWindowsTimeZone = timezones.find(timezone => timezone.id === timeZone.name)
+        const isWindowsVm = isWindows(osType)
+
+        if (isWindowsVm && !isWindowsTimeZone) {
```
sgratch: What if OS is non-windows type (isWindowsVm=false) and timezone is set to windows type? Don't you need to check both like in .

src/components/VmDialog/index.js
```diff
@@ -316,6 +324,34 @@ class VmDialog extends React.Component {
...
+          timeZone = {
+            name: 'GMT Standard Time',
```
sgratch: We should not hard-coded the timezone defaults since they are declared as config values  "vdcDefaultGeneralTimeZone", "DefaultWindowsTimeZone" on vdc_options table.
I suggest that in case we want to set the default then set the timeZone to null and let the engine backend choose the value for you as set in https://github.com/oVirt/ovirt-engine/blob/2cc0ab0ff6b00b5b51975497eaf1747b29d2a69c/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/VmHandler.java#L800.

src/components/VmDialog/index.js
```diff
@@ -316,6 +324,34 @@ class VmDialog extends React.Component {
...
+          }
+        }
+        if (!isWindowsVm && isWindowsTimeZone) {
+          timeZone = {
+            name: 'Etc/GMT',
```
sgratch: Same comment here regarding the hard-coded constants

src/components/VmDetails/cards/DetailsCard/index.js
michalskrivanek: absolutely. 
And only send the new default one(from vdc_options indeed) when you switch OS Type to a different family than what's in the template and the template has non-null TZ.

src/components/VmDialog/index.js
michalskrivanek: indeed, need to work both ways

bond95: It is already works in both ways
https://github.com/oVirt/ovirt-web-ui/blob/71335fa5c1c0e457aae11e6b2235dd8e828e34b9/src/components/VmDialog/index.js#L336-L346

src/sagas/server-configs.js
```diff
@@ -63,6 +65,15 @@ function* fetchUserSessionTimeoutInterval () {
   yield put(setUserSessionTimeoutInternal(parseInt(userSessionTimeout, 10)))
 }
 
+function* fetchDefaultTimeZoneConfig () {
+  const defaultGeneralTimezone = yield callGetOption('DefaultGeneralTimeZone', 'general', 'Etc/GMT')
+  const defaultWindowsTimezone = yield callGetOption('DefaultWindowsTimeZone', 'general', 'GMT Standard Time')
+  yield put(setDefaultTimezone({
```
sjd78: Alternatively, as in `fetchCpuTopologyOptions`:
```jsx
  const [ defaultGeneralTimezone, defaultWindowsTimezone ] =
    yield all([
      yield callGetOption('DefaultGeneralTimeZone', 'general', 'Etc/GMT'),
      yield callGetOption('DefaultWindowsTimeZone', 'general', 'GMT Standard Time'),
    ])
```

isaranova:
ci build please

isaranova:
If you choose blank template, the timezone object in the VM legacy creator is set to null, won't change according to the selected OS type and only changes to an object with name (and offset) if you pick another template, where the timezone object default is correctly changing according to the OS type. However, if you move back to blank template, the timezone object remains set to the Linux default timezone and won't change with selecting Windows OS type. Also going back from blank template to Windows template does not change the timezone, which remains as Linux default timezone instead of being Windows default timezone, even though it changes when going from Linux template to Windows template.

isaranova:
When you get to the Create VM legacy dialog, the default template (Blank in my case) does not have a default timezone assigned, instead I see that the timezone is null. Otherwise, it seems to be working correctly.

bond95:
> When you get to the Create VM legacy dialog, the default template (Blank in my case) does not have a default timezone assigned, instead I see that the timezone is null. Otherwise, it seems to be working correctly.

`Blank` template has `Other OS`, and if user want to change it will update timezone to correct one. So this is not a bug.

isaranova:
LGTM

Issue #1123, opened 2019-09-24
Disk create, Disk edit, Nic create, Nic edit modals don't open


sjd78:
Due to refactoring tooltips, the `onClick` event handler for opening `DiskImageEditor` (used for create and edit) and `NicEditor` (also used for create and edit) do not get attached to `a` tags as appropriate.  The modals will not open.

Best solution is to convert the `trigger` prop on both of the editors from a simple element to a render function that takes the `onClick` handler as a parameter.  That way, when declaring the trigger prop, the `onClick` can be put in the correct place.

PR #1124, opened 2019-09-24
Refactor various dialog components' trigger prop

sjd78:
Converted the `trigger` prop on dialog components the use the technique to be a render function instead of an element.  The `trigger` render function takes `onClick` as an argument and needs to assign that event passed handler to the correct tag for the dialogs to open.

Applied the refactor to:
  - `DiskImageEditor`
  - `NicEditor`
  - `DeleteConfirmationModal`
  - `RestoreConfirmationModal`

Fixes: #1123

**NOTE**: The `Action` component in `src/components/VmActions/Action.js` uses a very similar pattern but with the children prop.  I'm not changing the `Action` component at this time since its use looks ok.

### DIFF
--- a/src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
+++ b/src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
@@ -249,7 +249,7 @@ class DiskImageEditor extends Component {
     const diskSize = disk && (disk.get('lunSize') ? disk.get('lunSize') : disk.get('provisionedSize'))
 
     return <React.Fragment>
-      {React.cloneElement(trigger, { onClick: this.open })}
+      { trigger({ onClick: this.open }) }
 
       <Modal
         dialogClassName={style['editor-modal']}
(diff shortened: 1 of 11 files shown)


sgratch:
What about action-delete for both NICs and Disks? It seems that the issue is reproduced for delete button as well...

sjd78:
> What about action-delete for both NICs and Disks? It seems that the issue is reproduced for delete button as well...

Good catch.  I'll get those updated as well.

sjd78:
> Does it integrate with the `...rest` prop on OverlayTooltip?

@yodem - That change on `OverlayTooltip` will not impact or be impacted by this change.  This change really just makes it very explicit where the events handlers are coming from (the `onClick` parameter give to the trigger render function) and how they're attached to DOM elements (`onClick={onClick}`)

sjd78:
(rebased)

sgratch:
@leistnerova when do you plan to test this?

leistnerova:
> @leistnerova when do you plan to test this?

I did not know, we should test it. So no plan yet.

pnovotny:
ci build please

pnovotny:
LGTM

Verified with PR build ovirt-web-ui-1.6.1-0.20200113.git1af5cce.el7
Disk create/edit, NIC create/edit dialogs open on click and work correctly.

PR #1131, opened 2019-10-17
Fix action menu render error and console error on VmsList page

### LINKED ISSUE #1130: Regression - Action menus on VM cards, console errors
In current master, there are a few active regressions:

  1. The split button action menus do not render with the correct styles.  For example:

Selection_024

  2. There are a number of error warnings on the browser console:
Selection_025
Selection_026
Selection_027

sjd78:
Fixes #1130 

Issues addressed:
  - Action menu on the VM Cards list page were showing some actions
    as normal buttons in a split button, instead of as a normal menu
    item.  Item `classNames` have been explicitly set to fix this.

  - Console error messages were being logged for two items:

    - Using `<p>` tags on the `BaseCardTitle` and `BaseCardStatus`
      components were causing errors with child content.  Converted
      the tag to `<div>` and adjusted styles as necessary.

    - Because of the use of `{...rest}` in a few `VmActions`
      components, extra properties were being passed to child
      components causing errors.  `VmDropdownActions` now explicitly
      passes down the props needed.

VM Card action menu is back to normal after this PR:
Selection_028

### DIFF
--- a/src/components/VmActions/index.js
+++ b/src/components/VmActions/index.js
@@ -48,36 +48,46 @@ EmptyAction.propTypes = {
   isOnCard: PropTypes.bool.isRequired,
 }
 
-class VmDropdownActions extends React.Component {
-  render () {
-    const { actions, id } = this.props
-    let actionsCopy = [...actions]
-    if (actionsCopy.length === 0) {
-      return null
-    }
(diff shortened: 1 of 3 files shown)


isaranova:
LGTM

PR #1132, opened 2019-10-30
Edit Disk's Bootable Option

### LINKED ISSUE #1103: Can't set disks as bootable
Disks cannot be set/unset as bootable. 
Also the first added disk should be set as bootable

yodem:
fix  #1103 

User can now edit the bootable option of a disk through disk image editor dialog(only when the vm is down):

Bootable


lwrigh:
Screen Shot 2019-10-31 at 9 14 56 AM

I would recommend moving the 'Bootable' label more directly next to the checkbox so it's not as disconnected.

bond95:
ci test please

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -234,6 +235,13 @@ class DiskImageEditor extends Component {
     }
   }
 
+  changeBootable (event) {
+    const target = event.target
+    const value = target.type === 'checkbox' ? target.checked : target.value
```
bond95: Is this condition really needed? Because `changeBootable` is used only for `Checkbox`, so value always will be in `target.checked`

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -234,6 +235,13 @@ class DiskImageEditor extends Component {
...
+    this.setState((state) => ({ values: { ...state.values, bootable: value } }))
```
bond95: Nit-pick: if rename `value` to `bootable`, then line will look like that `this.setState((state) => ({ values: { ...state.values, bootable } }))`

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -245,7 +253,7 @@ class DiskImageEditor extends Component {
   }
 
   isFormValid () {
-    const isNotAllFieldFilled = Object.values(this.state.values).reduce((acc, value) => !value ? true : acc, false)
+    const isNotAllFieldFilled = Object.values(this.state.values).reduce((acc, value) => !(value !== undefined) ? true : acc, false)
```
bond95: I'm not sure about this condition `!(value !== undefined)`, maybe will be better to rewrite it like that `value === undefined`?

bond95:
Also I can't change bootable disk if VM already has bootable disk. So to make another disk bootable I need to unbootable previous one and only after that make disk bootable. Maybe will be better to do it in one step? I mean if user want to make another disk bootable it will automatically unbootable previous one.

yodem:
I get what you are saying, I'll try to.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -90,7 +91,7 @@ class DiskImageEditor extends Component {
     this.state = {
       showModal: false,
       storageDomainSelectList: storageDomainsToSelectList(props.storageDomainList),
-
+      bootable: DISK_DEFAULTS.bootable,
```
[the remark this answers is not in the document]
sjd78: @yodem - This line needs to be moved down a few lines into the `values` block.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -423,8 +431,24 @@ class DiskImageEditor extends Component {
...
               </Col>
             </FormGroup>
+            {/* Disk Bootable */}
+            { !createMode &&
+              <FormGroup controlId={`${idPrefix}-bootable`}>
+                <LabelCol sm={3}>
+                  { msg.diskEditorBootableLabel() }
+                </LabelCol>
+                <Col sm={9}>
+                  {
```
sjd78: The extra block is not necessary, please remove it and have `Checkbox` be the direct child of `Col`.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
sjd78: Agree with @bond95, this is awkward.  Probably doing explicit checks per field would be easier even if more verbose.

For example:
```jsx
isFormValid () {
  const isErrors = Object.values(this.state.errors).reduce((acc, value) => value ? true : acc, false)
  if (isErrors) {
    return false
  }

  // make sure that required fields have data
  const { values } = this.state
  return this.props.disk
    // edit mode needs: a change, an alias and a disk size
    ? this.changesMade && !!values.alias && values.size > 0 
    // create mode needs: name, size, storage domain, disk type
    : values.alias && values.size > 0 && values.storageDomain && values.format
}
```

yodem:
I have some issues with adding tooltip to disabled element, and then I saw this - https://github.com/patternfly/patternfly-react/issues/1894 
It basically says thats there is a problem with adding tooltip to disabled elements. Do you know something about it?

lwrigh:
> I had some issues with adding tooltip to disabled element, and then I saw this - [patternfly/patternfly-react#1894](https://github.com/patternfly/patternfly-react/issues/1894)
> It basically says thats there is a problem with adding tooltip to disabled elements. Do you know something about it?

From an accessibility point of view, adding a tooltip to a disabled element can be misleading so it might be best to just have the disabled element with no explanation.

sgratch:
> From an accessibility point of view, adding a tooltip to a disabled element can be misleading so it might be best to just have the disabled element with no explanation.

@lwrigh @yodem we do have that for icons (e.g. delete a disk/nic for a running VM) - the icon is disabled but there is a tooltip attached for explaining why (e.g. "Disks cannot be deleted when the VM is running"). 
My concern is that the user won't understand why it is disabled.. so maybe it is preferable to leave it as it is now (receiving an informative error from server in case updating it) then disabling without an explanation?

yodem:
I added a popover tooltip next to the bootable label when the vm is running + disabled the checkbox

lwrigh:
I think in this case the action should definitely be disabled. I'm fine with having the text explanation for why it's disabled for the time being. If PatternFly comes out with official standards for it we can always revisit it.

src/intl/messages.js
```diff
@@ -43,6 +43,7 @@ export const messages: { [messageId: string]: MessageType } = {
   authorizationExpired: 'Authorization expired. The page is going to be reloaded to re-login.',
   automaticPoolsNotEditable: 'The pool type of {poolName} is automatic so the details of this virtual machine are not editable.',
   availableVmsFromPool: 'Available VMs from this Pool',
+  bootableEditTooltip: 'Can not change bootable option while vm is running',
```
sgratch: "Can not change bootable option when the VM is running"

src/sagas/disks.js
```diff
@@ -31,6 +31,8 @@ function* createDiskForVm (action) {
     yield put(setNewDiskDialogDone())
   }
   yield put(setNewDiskDialogProgressIndicator(false))
+
+  yield clearBootableFlagOnVm(vmId, result)
```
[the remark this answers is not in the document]
sjd78: (I applied this change)

src/sagas/disks.js
```diff
@@ -77,6 +81,28 @@ function* editDiskOnVm (action) {
   yield fetchDisks({ vms: [ { id: vmId } ] })
 }
 
+function* clearBootableFlagOnVm (vmId, currentDisk) {
+  const vmDisks = yield select(state => state.vms.getIn([ 'vms', vmId, 'disks' ]))
+  const bootableDisk = vmDisks.find(disk => disk.get('bootable'))
+
+  if (bootableDisk && bootableDisk.get('id') !== currentDisk.id) {
```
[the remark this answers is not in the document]
sjd78: (I applied this change)

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -246,9 +254,18 @@ class DiskImageEditor extends Component {
...
+      return false
+    }
+
+    // make sure that required fields have data
+    const { values } = this.state
+    return this.props.disk
+      // edit mode needs: a change, an alias and a disk size
+      ? this.changesMade && !!values.alias // add size + bootable or leave like that
+      // create mode needs: name, size, storage domain, disk type
+      : values.alias && values.size > 0 && values.storageDomain && values.format
```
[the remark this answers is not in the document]
sjd78: (I applied this change)

sjd78:
> Few small things:
> 
> 1. Trying to create a new disk with size > 1 GiB and bootable, while there is an existing bootable disk already assigned to this VM => sometimes the disk is created as bootable and sometimes an error that a bootable disk is already existed is raised.
> 
> 2. In case changing a disk to bootable then we cancel other disk from being bootable without notifying the user.
>

Both addressed in the `sagas/disks.js`
 
>
> 3. I'm still not sure about changing the disks order as appears in vm dashboard, so that the first one will be the bootable, Thoughts on that?
> 
Disk list sorting is done in `sortDisksForDisplay` in file `src/components/VmDisks/utils.js`.  We can adjust it there if needed in future.  I vote for just adding a new issue to handle any debate and change to the sort.

sgratch:
> > 2. In case changing a disk to bootable then we cancel other disk from being bootable without notifying the user.
> 
> Both addressed in the `sagas/disks.js`

@sjd78 @lwrigh  Regarding Issue #2 above- 
I meant that there is no notification for letting the user know that there is an already bootable disk attached for this vm and by setting other disk as bootable then you cancel the previous one. Do you agree that it is required?

> 
> > 3. I'm still not sure about changing the disks order as appears in vm dashboard, so that the first one will be the bootable, Thoughts on that?
> 
> Disk list sorting is done in `sortDisksForDisplay` in file `src/components/VmDisks/utils.js`. We can adjust it there if needed in future. I vote for just adding a new issue to handle any debate and change to the sort.

+1

lwrigh:
> > > 1. In case changing a disk to bootable then we cancel other disk from being bootable without notifying the user.
> > 
> > 
> > Both addressed in the `sagas/disks.js`
> 
> @sjd78 @lwrigh Regarding Issue 2 above-
> I meant that there is no notification for letting the user know that there is an already bootable disk attached for this vm and by setting other disk as bootable then you cancel the previous one. Do you agree that it is required?
> 
> > > 1. I'm still not sure about changing the disks order as appears in vm dashboard, so that the first one will be the bootable, Thoughts on that?
> > 
> > 
> > Disk list sorting is done in `sortDisksForDisplay` in file `src/components/VmDisks/utils.js`. We can adjust it there if needed in future. I vote for just adding a new issue to handle any debate and change to the sort.
> 
> +1

We might want to include the name of the selected bootable disk in the details card and have the user have the ability to change it there. What do you guys think of that as an option? @sgratch @sjd78

sgratch:
> We might want to include the name of the selected bootable disk in the details card and have the user have the ability to change it there. What do you guys think of that as an option? @sgratch @sjd78

@lwrigh @sjd78 I think it's preferable to display/set all disks info within the disks card only, to avoid complexity...If the user need to check disks info then he will need to go to disks card. So to me it looks ok as it is now. 
So to summarize my two concerns:
1. Displayed disks order. Should we always jump bootable disks to first in line and avoid disks naming order? I think it's confusing (so we can open a separate issue on that and handle that later on).
2. Since only one disk can be bootable per a VM  and since we allow to set a disk as bootable even if there is already existed bootable one,  then we should notify the user on that. Currently the user can set another disk as bootbable and by doing that he cancels the previous bootable one without any notification/tooltip etc.

lwrigh:
> > We might want to include the name of the selected bootable disk in the details card and have the user have the ability to change it there. What do you guys think of that as an option? @sgratch @sjd78
> 
> @lwrigh @sjd78 I think it's preferable to display/set all disks info within the disks card only, to avoid complexity...If the user need to check disks info then he will need to go to disks card. So to me it looks ok as it is now.
> So to summarize my two concerns:
> 
> 1. Displayed disks order. Should we always jump bootable disks to first in line and avoid disks naming order? I think it's confusing (so we can open a separate issue on that and handle that later on).
> 2. Since only one disk can be bootable per a VM  and since we allow to set a disk as bootable even if there is already existed bootable one,  then we should notify the user on that. Currently the user can set another disk as bootbable and by doing that he cancels the previous bootable one without any notification/tooltip etc.

We could add a 'View all disks' button or toggle in the disks card so the user can expand the list of disks if they like. I don't think all of the disks should be displayed by default since that could make the card very long. 

Going with just the naming order as the primary way of ordering the disks works for me. I think the bootable label is enough to let the user know that the disk is bootable. We could also break out the list of disks into two different sections, bootable disks and other disks.  

If the user is changing the bootable disk to another disk there should probably be some sort of confirmation dialog that pops up before the new bootable disk is set.

sjd78:
> > @lwrigh @sjd78 I think it's preferable to display/set all disks info within the disks card only, to avoid complexity...If the user need to check disks info then he will need to go to disks card. So to me it looks ok as it is now.
> > So to summarize my two concerns:
> > 
> > 1. Displayed disks order. Should we always jump bootable disks to first in line and avoid disks naming order? I think it's confusing (so we can open a separate issue on that and handle that later on).

A new issue is the way to go for the sort on the disks card.

> > 2. Since only one disk can be bootable per a VM  and since we allow to set a disk as bootable even if there is already existed bootable one,  then we should notify the user on that. Currently the user can set another disk as bootable and by doing that he cancels the previous bootable one without any notification/tooltip etc.
> 
> We could add a 'View all disks' button or toggle in the disks card so the user can expand the list of disks if they like. I don't think all of the disks should be displayed by default since that could make the card very long.
>

While in theory you can attach a huge number of disks to a VM, in practice I doubt there would be more than 4.  There is more than enough space to display 4 disks.  Even displaying 10 disks wouldn't look horrible since they don't take much space.  Adding a "show more" when you get to 4 items in a list is just annoying.  I'd rather leave the card alone and address it if a specific user request comes in.

> 
> Going with just the naming order as the primary way of ordering the disks works for me. I think the bootable label is enough to let the user know that the disk is bootable. We could also break out the list of disks into two different sections, bootable disks and other disks.
> 
> If the user is changing the bootable disk to another disk there should probably be some sort of confirmation dialog that pops up before the new bootable disk is set.

A dialog with a confirmation dialog once you Ok the change?  What about how the overview card does the cloud-init host name change notification?  Basically if the bootable box is checked, and the flag will get reassigned, just throw a warning message on the dialog itself.  "The bootable flag will be removed from {diskName} and placed on this disk."  I'm not a fan of chains of modals.

lwrigh:
+1 to new issue and leaving it alone until a specific user requests it so we don't over complicate it. 

Yeah I want to avoid overlapping modals :/ We could provide the user with this information on an inline alert: 
Screen Shot 2019-12-03 at 1 25 11 PM

sjd78:
Created issue #1146 to handle the disk sorting changes.

isaranova:
LGTM

isaranova:
LGTM

sjd78:
@lwrigh , @sgratch - I added the warning alert (albeit the PF3 version):

For creating a new disk with an existing disk already marked bootable on the VM:
screenshot-localhost_3000-2019 12

For editing a disk and marking it bootable when a different disk is currently marked bootable:
screenshot-localhost_3000-2019 12 (1)

lwrigh:
> @lwrigh , @sgratch - I added the warning alert (albeit the PF3 version):
> 
> For creating a new disk with an existing disk already marked bootable on the VM:
> screenshot-localhost_3000-2019 12
> 
> For editing a disk and marking it bootable when a different disk is currently marked bootable:
> screenshot-localhost_3000-2019 12 (1)

I would edit the message to say, "Only one disk can be bootable at a time. The bootable label will be removed from  '_B_Disk1' and placed on this disk."

Would there be anyway to use a PF4 inline alert for this or would that require a lot of work? @sjd78

sjd78:
> 
> I would edit the message to say, "Only one disk can be bootable at a time. The bootable label will be removed from '_B_Disk1' and placed on this disk."
> 
> Would there be anyway to use a PF4 inline alert for this or would that require a lot of work? @sjd78

Not difficult.  Looks a bit off:
screenshot-localhost_3000-2019 12 (4)

sgratch:
> Not difficult. Looks a bit off. Here is an example (I haven't push the update to the PR yet):
> screenshot-localhost_3000-2019 12 (4)

@sjd78 @lwrigh LGTM

sgratch:
@sjd78 there is an exception when trying to create/edit a disk to be bootable (while other non bootable disks already existed for that VM).

lwrigh:
I'm cool with the PF4 inline alert for notifying the user since this is a checkbox and helper text isn't the most appropriate for checkboxes.

rhv-qe-github:
ci build please

rhv-qe-github:
ci build please

sjd78:
@lwrigh Here is the update (finally):

screenshot-localhost_3000-2020 04

sjd78:
> @sjd78 there is an exception when trying to create/edit a disk to be bootable (while other non bootable disks already existed for that VM).

All fixed!  This PR is finally good to go.

lwrigh:
> @lwrigh Here is the update (finally):
> 
> screenshot-localhost_3000-2020 04

Awesome this looks good to me!

rhv-qe-github:
ci build please

isaranova:
Unfortunately, it seems to be a bit broken for me. When I have a VM with bootable disk and create a new disk and check the bootable option, the expected warning message underneath the checkbox appears - I proceed with creating the disk; However, an error message appears in the VM portal itself preventing the new disk from being created, though the first disk is stripped of the bootable label. The same exact thing happens when you have two disks, one bootable and try to edit the not bootable one to bootable. Otherwise, if the disk is edited on its own without any of the passing the bootability to another disk, it seems to work okay.
Screenshot_2020-05-04 VM Portal

sgratch:
> Unfortunately, it seems to be a bit broken for me. 

@isaranova I verified and It works fine for me. I tried to reproduce mentioned use cases but it works as expected. Did you test it with relevant build ovirt-web-ui-1.6.2-0.20200504.gitda55ecc.{platform}.noarch.rpm?

isaranova:
@sgratch Yes, I have the correct package ovirt-web-ui-1.6.2-0.20200504.gitda55ecc.el8.noarch. I reproduced the issue on ovirt-engine-4.4.0-0.8.master.20200430104406.gitd08c1c40c72.el8.noarch, so I tried to see if the problem persists on newer version ovirt-engine-4.4.0.1-0.0.master.20200507171742.gitbf0c2d50fbf.el8.noarch and I can still reproduce it there as well. I did engine-setup, updated the whole engine, rebooted, removed cache from browser and checked console for any error messages, but there is nothing except for the same error message in the messages. I added all disk permissions to the user on system level and I can still reproduce the issue. I recorded it and have the video ready, so I can send it to you via email or chat.

sgratch:
> @sgratch . I recorded it and have the video ready, so I can send it to you via email or chat.

@isaranova please send if you can. I

sjd78:
@isaranova , @sgratch - I made a few tweaks to the order of operations to make sure the bootable flag moves prior to adding it to an existing disk, or prior to creating a new disk.  All of the my tests (moving the flag, creating a new disk and flagging it bootable) are working on my dev instance.

sgratch:
> @isaranova , @sgratch - I made a few tweaks to the order of operations to make sure the bootable flag moves prior to adding it to an existing disk, or prior to creating a new disk. All of the my tests (moving the flag, creating a new disk and flagging it bootable) are working on my dev instance.T

@sjd78 @isaranova I think that the problem with that solution is that since it's not done within one transaction block then no rollbacks are supported - i.e. in case of a failure in creating a new bootable disk/adding bootable to a disk, the previous bootable flag was already removed.

We can solve that by either adding an explaining of that behavior of no rollbacks as part of the msg "Only one disk can be bootable at a time..." or (preferable) by removing that option of changing a bootable disk by removing it from other disks. You can only set a disk as bootable in case there is no other bootable disk already existed for that VM. If there is then we can display a message of "Only one disk can be bootable at a time" while bootable checkbox is disabled.

sjd78:
> > @isaranova , @sgratch - I made a few tweaks to the order of operations to make sure the bootable flag moves prior to adding it to an existing disk, or prior to creating a new disk. All of the my tests (moving the flag, creating a new disk and flagging it bootable) are working on my dev instance.T
> 
> @sjd78 @isaranova I think that the problem with that solution is that since it's not done within one transaction block then no rollbacks are supported - i.e. in case of a failure in creating a new bootable disk/adding bootable to a disk, the previous bootable flag was already removed.
> 
> We can solve that by either adding an explaining of that behavior of no rollbacks as part of the msg "Only one disk can be bootable at a time..." or (preferable) by removing that option of changing a bootable disk by removing it from other disks. You can only set a disk as bootable in case there is no other bootable disk already existed for that VM. If there is then we can display a message of "Only one disk can be bootable at a time" while bootable checkbox is disabled.

Or detect the failed disk create and set the bootable flag back to where it was within the enclosing  saga.  That should be reasonably easy to do.

sjd78:
@sgratch, @isaranova - commit 470c5f2 will reapply the bootable flag to the original disk if the disk creation fails.  It is easy to test by attempting to create a preallocated disk with a size larger than the remaining size of the storage domain.

isaranova:
LGTM

PR #1137, opened 2019-11-18
Card list, running VM actions default to open VM's console

yodem:
On VM card list page, if a VM is running, change default action item to open the VM console.

Options

Ref: [BZ1724959](https://bugzilla.redhat.com/show_bug.cgi?id=1724959)

### DIFF
--- a/src/components/VmActions/index.js
+++ b/src/components/VmActions/index.js
@@ -56,16 +56,15 @@ const VmDropdownActions = ({ actions, id }) => {
   if (actions.length === 1) {
     return <ActionButtonWraper {...actions[0]} className='btn btn-default' />
   }
-
-  const primaryAction = actions[0]
   const secondaryActions = []
-  for (const action of actions.slice(1)) {
+  for (const action of actions) {
     if (action.items) {
       secondaryActions.push(...action.items.filter(item => item && !item.actionDisabled))
     } else {
       secondaryActions.push(action)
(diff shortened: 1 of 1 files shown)


sjd78:
> What is the default in case of a headless VM (i.e. no graphic console is enabled)?

The console actions for a VM are built from the VM's definition so a headless VM w/o consoles shouldn't be an issue.

src/components/VmActions/index.js
```diff
@@ -56,10 +56,17 @@ const VmDropdownActions = ({ actions, id }) => {
...
-  const primaryAction = actions[0]
+  var tempPrimaryAction
+  if (actions[0].items) {
+    tempPrimaryAction = actions[0].items[0]
+    actions[0].items.shift()
+  } else {
+    tempPrimaryAction = actions[0]
+    actions.splice(1)
+  }
+  const primaryAction = tempPrimaryAction
```
sjd78: That is a bit over complicated and doesn't work for headless VMs.  Try this:

```jsx
  const secondaryActions = []
  for (const action of actions) {
    if (action.items) {
      secondaryActions.push(...action.items.filter(item => item && !item.actionDisabled))
    } else {
      secondaryActions.push(action)
    }
  }

  const primaryAction = secondaryActions.shift()
```

isaranova:
LGTM

PR #1141, opened 2019-11-27
Tooltips refactoring

### LINKED ISSUE #1094: Different Tooltip Types
In the VM Details page, they are different types of tooltips, and i think we should decide what tooltip do we want.
TooltipPush
Tooltip type3
Tooltip type 2
Tooltips types4

@lwrigh what do you think we need to do?

yodem:
fix #1094

According to [Tooltips Document](https://docs.google.com/document/d/1ouQ7b4tKH1ksmB54hnPz69rkHcynv7qhG3hDXf342XA/edit?usp=sharing).


sjd78:
This work is now being handled in #1289

PR #1148, opened 2019-12-17
Fix DiskImageEditor storage domain list

sjd78:
A regression was introduced in the create VM code such that
the storage domain list on the disk editor/creator modal showed
as blank.  This has been fixed by adding the proper data center
filtering to the list generation.

Now, the storage domain list will only show storage domains available
for use by the logged in user in the VM's cluster's data center.


pnovotny:
ci build please

pnovotny:
LGTM

Verified with ovirt-web-ui-1.6.1-0.20200113.git1af5cce.el7.noarch
All storage domains are now listed in the select box. Adding a new disk works also properly.

PR #1158, opened 2020-02-02
CreateVmWizard - Network and Storage table edit row action buttons updates

sjd78:
Custom table edit row component so the location of the the confirm and cancel buttons during a Nic or Disk edit can be controlled.


sjd78:
Before:
VM Portal - Google Chrome_031

After:
VM Portal - Google Chrome_030

PR #1165, opened 2020-03-01
Minor cleanup to web.xml and branding.jsp

sjd78:
- Remove JSTL reference from branding.jsp since it isn't used.

  - Remove unused servlet mappings and added some comments about
    index.jsp and how those mappings need to allow all of the
    routes used by the app or it'll break in a RPM deploy to
    ovirt-engine.


sjd78:
ci test

PR #1167, opened 2020-03-09
Changes the disks' display order 

### LINKED ISSUE #1146: Change VM details Disks card disk sort order
> > > 1. Displayed disks order. Should we always jump bootable disks to first in line and avoid disks naming order? I think it's confusing (so we can open a separate issue on that and handle that later on).
>
> A new issue is the way to go for the sort on the disks card.

_Originally posted by @sjd78 in https://github.com/oVirt/ovirt-web-ui/pull/1132#issuecomment-561291986_

On the VM Details, Disks card, the bootable disk is sorted before any other disk.  If the bootable flag is changed by the user (once PR #1132 is merged), the disks will be resorted.  To avoid confusion with the sorting change, simply sort the disks by disk name.  The "bootable" label should be enough to easily identify which disk is the VMs bootable disk.
This is the default behaviour on webadmin as well.

For example, the disks list are curenttly displayed on web_ui like this:
VM Portal-disks-ordering

While on webadmin, exactly the same disks will be sorted by name as a default:
webadmin-disks-ordering

benams1:
Hi,
I changed the function that sorts the disks' display order in the disks card component.
the change solving issue #1146 
now the disks sorted only by their name.

Fixes #1146

### DIFF
--- a/src/components/VmDisks/utils.js
+++ b/src/components/VmDisks/utils.js
@@ -4,7 +4,7 @@ import { localeCompare } from '_/helpers'
 
 /*
  * Sort an Immutable List of Maps (set of disks) for display on the VmDisks list.
- * Bootable drives sort first, then sorted number aware alphabetically.
+ * Drives Sorted by disk name.
  */
 export function sortDisksForDisplay (disks, locale = appLocale) {
   return disks.sort((a, b) => { return localeCompare(a.get('name'), b.get('name'), locale) })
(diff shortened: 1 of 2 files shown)


ovirt-infra:

sjd78:
ci test please

sjd78:
@benams1 - Please squash your commits.  Change is small enough to only need a single commit.  

`git rebase -i` and then set "s" for all but the first commit.

src/components/VmDisks/utils.js
```diff
@@ -7,12 +7,5 @@ import { localeCompare } from '_/helpers'
  * Bootable drives sort first, then sorted number aware alphabetically.
```
sgratch: Please change the comment accordingly.

benams1: @sgratch - I changed the comment according the function behavior.

sjd78:
> Please consider displaying the same disks sorting order for create VM based on a template as well.

I'll take care of the disk sorting on the Create VM Wizard disk step.  This change looks good to me.

PR #1172, opened 2020-03-18
Tweak the navbar icons

sjd78:
- Main icons layout correctly

  - Mouse hover background highlighting takes the same spaces for all icons (not a few pixels too short)

  - Drop down caret icons are spaced the same distance from their nav icons as in ovirt webadmin


sjd78:
Before:
VM Portal - Google Chrome_002

After:
VM Portal - Google Chrome_003

Issue #1128, opened 2019-10-14
Main page crash after taking pool VM


bond95:
I found one more issue, but it occurs only with SuperUser role (admin@internal).

Version:
ovirt-web-ui-1.6.1-0.20190923.git5bd255e.el7.noarch
ovirt-engine-4.3.6.7-0.1.el7.noarch

Browsers: Firefox 69.0.2, Chromium 77.0.3865.90

If you take or run a pool VM from the VM grid and refresh the page in browser (F5),
the web-ui crashes during loading the VM cards. 
Running a regular VM doesn't have this effect.

Browser console log:
Firefox:
```
[debug]  Reducing action: {"type":"UPDATE_VMPOOLS_COUNT"} utils.js:47:14
[error]  TypeError: "n is undefined"
    default index.js:324
    Redux 8
    qt react-dom.production.min.js:132
    Dn react-dom.production.min.js:167
    Rn react-dom.production.min.js:180
    xr react-dom.production.min.js:232
    wr react-dom.production.min.js:233
    Hr react-dom.production.min.js:249
    Vr react-dom.production.min.js:248
    Ir react-dom.production.min.js:245
    Er react-dom.production.min.js:243
    enqueueSetState react-dom.production.min.js:130
    setState React
    Redux 12
    _ middleware.js:25
    Redux 14
    jQuery 2
react-dom.production.min.js:198:186
```
Chromium:
```
[debug]  Reducing action: {"type":"UPDATE_VMPOOLS_COUNT"} utils.js:47
[error]  TypeError: Cannot read property 'get' of undefined
    at Function.mapToProps (index.js:324)
    at r (wrapMapToProps.js:54)
    at Function.r.mapToProps (wrapMapToProps.js:63)
    at r (wrapMapToProps.js:54)
    at o (selectorFactory.js:37)
    at selectorFactory.js:85
    at Object.run (connectAdvanced.js:43)
    at c.initSelector (connectAdvanced.js:195)
    at new Connect(t) (connectAdvanced.js:136)
    at qt (react-dom.production.min.js:132)
```

_Originally posted by @pnovotny in https://github.com/oVirt/ovirt-web-ui/pull/990#issuecomment-541106637_

sjd78:
Looks to be the same issue as [BZ1771793](https://bugzilla.redhat.com/show_bug.cgi?id=1771793).

PR #1174, opened 2020-03-23
Fix Pool handling and refactor VM loading

sjd78:
Fixes: #1128 
Fixes: [BZ1771793](https://bugzilla.redhat.com/show_bug.cgi?id=1771793)

A few related changes:

  - UPDATE_VMPOOLS_COUNT reducer fixed.  If a pool VM is fetched and processed before the Pool, the reducer would break and cause UI errors trying to load a Pool that does not yet exist.  Now a missing pool is skipped.  The reducer will process properly once the Pool is loaded.

  - VM saga updates:
    - Removed code for ovirt-engine versions < 4.2.  The version is already blocked at app startup time so the code is dead.

    - VM permits and canUse* mappings updated following the frontend method used by Clusters etc (see #1020)

  - There is no visual change to the display of a Pool card.  However, this change does leave a scenario where a VM could momentarily appear as a normal VM but the get tagged as a Pool VM once the Pool is actually fetched.  To mitigate this confusion, all Pool VMs whose Pool is unknown (typically it hasn't been fetched yet) will not be displayed.  The Pool and all of its VMs will be rendered together once they are all available.


sjd78:
Example of the Pool loading progression.  For this video, a 2 second delay was added in the code to force VMs to load before Pools.

ezgif com-video-to-gif

sjd78:
Now that this is merged, testing can be done from the master branch build artifacts CI: https://jenkins.ovirt.org/job/oVirt_ovirt-web-ui_standard-on-ghpush/lastSuccessfulBuild/artifact

PR #1179, opened 2020-04-06
ovirt-web-ui 1.6.1-1

sgratch:


michalskrivanek:
dropping compatibility with 4.3, adding brand new New VM dialog, moving to PF4...doesn't it deserve a 1.7 at least?

sgratch:
> dropping compatibility with 4.3, adding brand new New VM dialog, moving to PF4...doesn't it deserve a 1.7 at least?

@michalskrivanek Based on Zanata project version #1162, it should stay as 1.6. We can change that on next build

sgratch:
ci build please

Issue #1186, opened 2020-04-14
VM Portal fails to initialize if the environment contains an Export storage domain


sjd78:
Given an engine storage domain setup like this:
![Red Hat Virtualization Manager Web Administration - Google Chrome_020](https://user-images.githubusercontent.com/3985964/79263475-e6102780-7e60-11ea-99c3-b2a7560c5056.png)

When VM Portal is loading, there is a reported saga error and the error page is displayed.
![VM Portal - Google Chrome_021](https://user-images.githubusercontent.com/3985964/79263629-312a3a80-7e61-11ea-903b-e61d0290dd09.png)

This error only happens when an Export type storage domain is available to the user.

Text copy of the browser console error message:
```
Uncaught saga error (store.js):  TypeError: Cannot read property 'statusPerDataCenter' of undefined
    at fetchDataCentersAndStorageDomains$ (storageDomains.js:34)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:296)
    at Generator.prototype.<computed> [as next] (runtime.js:114)
    at next (proc.js:322)
    at currCb (proc.js:399)
    at checkEffectEnd (proc.js:623)
    at chCbAtKey (proc.js:638)
    at Object.currCb [as cont] (proc.js:399)
    at end (proc.js:367)
```

sjd78:
Another example from a prod build:
![Screenshot from 2020-04-14 21-23-35](https://user-images.githubusercontent.com/3985964/79266207-325d6680-7e65-11ea-935d-cd0c663606a4.png)

PR #1187, opened 2020-04-14
Fix `statusPerDataCenter` calculation in storageDomains sagas

sjd78:
Fixes: #1186

Since only 'data' and 'iso' storage domains are fetched in the
`fetchDataAndIsoStorageDomains` saga, if a data center contained
another type of storage domain (i.e. 'export'), the code that
calculates a storage domain's status per data center would fail.

Now, the code in `fetchDataCentersAndStorageDomains` will only
cross reference statuses for 'data' and 'iso' type storage domains.


michalskrivanek:
well, export domains are on their way out so probably not really important. moreover they're not usab;e/visible in VM Portal so it's fine...

PR #1157, opened 2020-01-31
Update branding and navbar masthead

sjd78:
Update the `brand.css` and the PF3 navbar masthead with recent
changes to ovirt-engine 4.4.  This will allow alternative branding
for webadmin and web-ui to continue to use the same CSS class names.

Related ovirt-engine patches:
  - https://gerrit.ovirt.org/106579
  - https://gerrit.ovirt.org/106694

PR #1188, opened 2020-04-14
Apply ovirt 4.4 branding changes to GlobalErrorBoundary

sjd78:
Follows up changes to `Header.js` in PR #1157 and applies the same masthead changes to the error page.


sjd78:
Before:
VM Portal - Google Chrome_023


After:
VM Portal - Google Chrome_022

PR #1200, opened 2020-05-05
Increase parallel calls during login data fetch

### LINKED ISSUE #1210: Enhancement: Opportunity to do additional parallel data fetches at login
The login code needs to fetch a lot of information to populate the redux store before the app can be allowed to run normally.  A lot of the process is already done with parallel fetches.  There is a small gain to be made by:

  1. Fetch **user permissions** and **server config values** in parallel
  2. Fetch all **server config values** in parallel

This should max out the number of fetches that can be done at the same time (a browser limit).

sjd78:
Resolves #1210

Increase parallel calls during login data fetch:
  - Fetch user permissions and server config values in parallel
  - Server config values are all fetched in parallel

This should max out the number of fetches that can be done at the same time (a browser limit).

### DIFF
--- a/src/sagas/login.js
+++ b/src/sagas/login.js
@@ -93,9 +93,12 @@ function* login (action) {
 
   // API checks passed.  Load user data and the initial app data
   console.group('Login Data Fetch')
+
   console.group('user checks and server config')
-  yield checkUserFilterPermissions()
-  yield fetchServerConfiguredValues()
+  yield all([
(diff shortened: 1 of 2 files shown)


rhv-qe-github:
ci build please

michalskrivanek:
LGTM too, though at least a vague note about impact on load speed in large scale env would be good

sgratch:

Issue #1214, opened 2020-05-19
Create VM Wizard - the Storage Domain menu field on a new disk row is not aligned


sgratch:
On the "Create Virtual Machine" wizard -> "Storage" step, when adding/editing a Disk by clicking the checkmark box, the "Storage Domain" field looks like overriding a value beneath:

![image](https://user-images.githubusercontent.com/18169498/82301957-a25d9000-99c1-11ea-9e26-33607469d591.png)

PR #1217, opened 2020-05-19
added explicitly width value to disk size FormControl field

bamsalem:
"On the "Create Virtual Machine" wizard -> "Storage" step, when adding/editing a Disk by clicking the checkmark box, the "Storage Domain" field looks like overriding a value beneath:"

the problem happens only in FF browser, in chrome it looks good.

the problem was the <FormControl>  component's width, the component took 100% of the table's cell and "pushed" the <span> tag with the units to the left and the tag covered by the "Storage Domain" field.

I added explicitly width value to the component.
now it's not looks like the "Storage Domain" field overriding a value beneath.

Chrome:
image

FF:
image

Fixes: #1214

### DIFF
--- a/src/components/CreateVmWizard/steps/style.css
+++ b/src/components/CreateVmWizard/steps/style.css
@@ -111,6 +111,10 @@
   display: flex;
   align-items: center;
 }
+.disk-size-form-control-edit {
+  width: 0;
+  flex-grow: 1;
+}
 
 .disk-size-edit .disk-size-edit-label {
   margin: 0 3px;
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -201,6 +201,7 @@ class Storage extends React.Component {
               id={`${idPrefix}-${value}-size-edit`}
               type='number'
               defaultValue={sizeGiB}
+              className={style['disk-size-form-control-edit']}
               onBlur={e => this.handleCellChange(rowData, 'size', e.target.value)}
             />
             <span className={style['disk-size-edit-label']}>GiB</span>


ovirt-infra:

rhv-qe-github:
ci build please

bamsalem:
Hi @sjd78  thanks for the flex-box cheatsheet it was really helpful :)
I checked what is the cause for this issue and I found that the `FormControl` component inherit `width: 100%` attribute.
I tried using flex-grow and flex-shrink as you mentioned and it still spread over the whole cell, the solution I found is to overwrite the width attribute in the 'disk size' field to 0 and then using flex-growth attribute affects on field's width and it looks good.

what do you think about the solution?

sgratch:
ci add to whitelist

sjd78:
> Hi @sjd78 thanks for the flex-box cheatsheet it was really helpful :)
> I checked what is the cause for this issue and I found that the `FormControl` component inherit `width: 100%` attribute.
> I tried using flex-grow and flex-shrink as you mentioned and it still spread over the whole cell, the solution I found is to overwrite the width attribute in the 'disk size' field to 0 and then using flex-growth attribute affects on field's width and it looks good.
> 
> what do you think about the solution?

That looks much better!  Did you try `width: auto;` in the style?

src/components/CreateVmWizard/steps/style.css
```diff
@@ -111,6 +111,10 @@
   display: flex;
   align-items: center;
 }
+.disk-size-form-control-edit {
+  width: 0;
```
sjd78, line 115: Last question.

Does the layout work with `width: auto` instead of `width: 0`?  The `auto` value is a bit more typical to see when you want to reset a style that gets inherited.

bamsalem: it works but using `width:auto` stretches the whole column and I thought it will be better if the table structure doesn't change between modes. 
That how it looks on edit mode:
image
That how it looks on view mode:
image

pnovotny:
ci build please

pnovotny:
LGTM

sgratch:
Approved by QE so merging and moving TM to 4.4.1

Issue #1151, opened 2020-01-05
When creating a new disk,  the pre-allocated allocation policy is ignored and treated as if it is Thin Provision.


sgratch:
The "preallocated" disk type value is treated the same as the "Thin Provision" value.

**Current behaviour**:
- When creating a new disk for an existing VM via the **web-ui** and setting allocation policy/disk type field to "preallocated" value (default), the disk is actually created with "Thin Provision" type. 

- When creating a new disk for an existing VM via the **webAdmin** with policy/disk type field set to "Thin Provision" value, the disk is displayed on web-ui with the "preallocated" type. 

- In addition, the default disk type value on web-ui for a new disk is "preallocated".

**Expected behaviour**:
-  The disk's type should be displayed on web-ui with the actual required value.

- The default disk type value for a new disk should be "Thin Provision" (the same as on webAdmin).

sgratch:
It seems that this issue is solved and not reproduced anymore so closing

PR #1218, opened 2020-05-24
Fix DiskImageEditor default format and size tooltip

sjd78:
Fixes: #1151

### Disk size tooltip refactoring
Don't show the "resize client filesystem" tooltip on the size field when creating a disk.

### Disk type refactoring
Disks in web-ui use the term 'Disk Type' as either 'Thin Provision' or 'Preallocated'.  This gets translated to 'sparse' or 'not sparse'.  The refactoring changes all the use of 'format' and 'sparse' to 'diskType' except where the data needs to be converted for the REST API.

Note: The disk's format attribute can be either 'raw' or 'cow'. Webadmin seems to always create disks with a format of `raw`.  The way to tell a preallocated disk from a thin provisioned disk is to look at the disk's `sparse` field.  This will work for disks created in webadmin and web-ui.

Update disk creation and editing to use 'diskType' in component state:
  - Initialize the value from an existing disk's `sparse` value

  - Use the 'diskType' value to derive both a disk's `sparse` attribute for disk creation

    - diskType of 'Thin Provision' maps to sparse 'true'

    - diskType of 'Preallocated' maps to sparse 'false'

  - Always use 'raw' for the disk's `format` attribute for disk creation to follow web-ui's behavior

  - Edit only needs to read the value, the `sparse` and `format` attributes can't be updated once a disk is created

### Rename diskType constants to 'thin' and 'pre'

To maintain consistency with the refactoring to __diskType__ to store "Thin Provision" vs "Preallocate", change the constants:
  - 'cow' to 'thin'
  - 'raw' to 'pre'

There are a few places where 'cow' and 'raw' are still used, but not in the context of __diskType__.  They're used in the context
of the REST API Disk's `format` attribute.

### Rename i18n keys for disks "Format" to "DiskType"
  - diskEditorFormatOptionRaw to diskEditorDiskTypeOptionPre
  - diskEditorFormatOptionCow to diskEditorDiskTypeOptionThin
  - diskEditorFormat(.*) to diskEditorDiskType\1


sjd78:
> it looks OK now except one misbehavior (mentioned on issue):
> When creating a new disk for an existing VM via the webAdmin with disk type field set to "Thin Provision" value, the disk is displayed on web-ui with the "preallocated" type.

Interesting.  When I create disks on webadmin, no matter if I pick "Thin Provisioned" or "Preallocated", the disks come back from the REST api as format == "raw".  Works the same way on 4.3 and 4.4 engines:

Disks:
oVirt Open Virtualization Manager - Mozilla Firefox (Private Browsing)_031
oVirt Open Virtualization Manager - Mozilla Firefox (Private Browsing)_032

REST response (parse to easily see the relevant data):
```
> curl -k -u"admin@internal:XXXX" -H"Accept: application/json" "https://ghost.local.dickerson-web.com:8443/ovirt-engine/api/vms/879ab4ca-fe2d-4d3d-97ab-5efee82d9cee/diskattachments?follow=disk" \
    | jq '.disk_attachment[].disk | { alias, sparse, format, actual_size, provisioned_size }' 

{
  "alias": "DiskTestVM_Disk1_PRE",
  "sparse": "false",
  "format": "raw",
  "actual_size": "1073741824",
  "provisioned_size": "1073741824"
}
{
  "alias": "DiskTestVM_Disk1_THIN",
  "sparse": "true",
  "format": "raw",
  "actual_size": "4096",
  "provisioned_size": "2147483648"
}
```

I'll adjust the dialog code to look at the `sparse` field instead of the `format` field to make sure its displayed correctly.

sjd78:
I created 2 new disk on my test VM in web-ui.  One preallocated and one thin provisioned.  Disk names are `/.*Disk2_(PRE|THIN)/`.  Here is how the VM's disks look like via REST api:
```
> curl -k -u"admin@internal:XXXX" -H"Accept: application/json" "https://ghost.local.dickerson-web.com:8443/ovirt-engine/api/vms/879ab4ca-fe2d-4d3d-97ab-5efee82d9cee/diskattachments?follow=disk" \
    | jq '.disk_attachment[].disk | { alias, sparse, format, actual_size, provisioned_size }'
{
  "alias": "DiskTestVM_Disk2_THIN",
  "sparse": "true",
  "format": "cow",
  "actual_size": "200704",
  "provisioned_size": "1073741824"
}
{
  "alias": "DiskTestVM_Disk2_PRE",
  "sparse": "false",
  "format": "raw",
  "actual_size": "1073741824",
  "provisioned_size": "1073741824"
}
{
  "alias": "DiskTestVM_Disk1_THIN",
  "sparse": "true",
  "format": "raw",
  "actual_size": "4096",
  "provisioned_size": "2147483648"
}
{
  "alias": "DiskTestVM_Disk1_PRE",
  "sparse": "false",
  "format": "raw",
  "actual_size": "1073741824",
  "provisioned_size": "1073741824"
}
```

and how the disks look in the disk image edit dialog on web-ui:
VM Portal - Mozilla Firefox (Private Browsing)_036
VM Portal - Mozilla Firefox (Private Browsing)_035
VM Portal - Mozilla Firefox (Private Browsing)_034
VM Portal - Mozilla Firefox (Private Browsing)_033

sgratch:
> I'll adjust the dialog code to look at the `sparse` field instead of the `format` field to make sure its displayed correctly.

This adjustment is correct because 'format' field is the volumeFormat on backend and it's based on the storage type of the disk (IMAGE, LUN, CINDER, MANAGED_BLOCK_STORAGE), while the 'sparse' field is the VolumeType on backend based on the user input for deciding if the physical storage for the disk should be pre-allocated or not.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -120,7 +120,10 @@ class DiskImageEditor extends Component {
           bootable: disk.get('bootable'),
           size: 0,
           storageDomain: disk.get('storageDomainId'),
-          format: disk.get('format'),
+
+          // NOTE: webadmin always uses raw when creating disks.  We need to look at
```
sgratch, line 124: This is no so accurate (check https://github.com/oVirt/ovirt-web-ui/pull/1218#issuecomment-638751656 #). It can be either raw or cow.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
```diff
@@ -120,7 +120,10 @@ class DiskImageEditor extends Component {
...
+          //       the sparse flag to determine the proper format to display.
+          format: disk.get('sparse') ? 'cow' : 'raw',
```
sgratch, line 126: I think it's better to change the field name to "type" or "sparse" instead of "format" and the values to "prealocated"/"thin provision" or boolean accordingly. 
Leaving the of current "format", "cow","raw" values is wrong because it refers to a different field.

sjd78:
> > I'll adjust the dialog code to look at the `sparse` field instead of the `format` field to make sure its displayed correctly.
> 
> This adjustment is correct because 'format' field is the volumeFormat on backend and it's based on the storage type of the disk (IMAGE, LUN, CINDER, MANAGED_BLOCK_STORAGE), while the 'sparse' field is the VolumeType on backend based on the user input for deciding if the physical storage for the disk should be pre-allocated or not.

The `format` field within web-ui correlates to a REST `Disk` (http://ovirt.github.io/ovirt-engine-api-model/master/#types/disk) format attribute.  The format attribute is a `DiskFormat` (http://ovirt.github.io/ovirt-engine-api-model/master/#types/disk_format) whose values are `cow` or `raw`.

I think you are talking about `VolumeFormat`.

Either way, I can update the name of the field in the component's state and deal with the data fields in the transform fields.

src/components/VmDetails/cards/DisksCard/DiskImageEditor.js
sjd78: I renamed the form state field to `diskFormat` to match the label on the form.

sjd78:
> The disk type default on create vm wizard should change from "preallocated" to "thin provision" as well.

On Create VM Wizard, the default "Disk Type" is driven by the optimized for field on the basic setting step.  That behavior should not change.

See:
https://github.com/oVirt/ovirt-web-ui/blob/f67e9a64f4a2cb1c706fb5fb086a97a4b8f3cae0/src/components/CreateVmWizard/steps/Storage.js#L30-L37

and:
https://github.com/oVirt/ovirt-web-ui/blob/f67e9a64f4a2cb1c706fb5fb086a97a4b8f3cae0/src/components/CreateVmWizard/steps/Storage.js#L361-L389

hstastna:
@sgratch @sjd78 But there's one thing which worries me. It's not related to changes in this PR but it's related to Disks: when I make some changes in _Disks_ card (no matter if adding or deleting Disks), the changes are made and they look already saved despite clicking on "X" to cancel the changes in a _Disks_ card. I would expect a different behavior. It's present also on master. Isn't that another bug or am I missing something? Thanks! :)
cancel_disks

sjd78:
> @sgratch @sjd78 But there's one thing which worries me. It's not related to changes in this PR but it's related to Disks: when I make some changes in _Disks_ card (no matter if adding or deleting Disks), the changes are made and they look already saved despite clicking on "X" to cancel the changes in a _Disks_ card. I would expect a different behavior. It's present also on master. Isn't that another bug or am I missing something? Thanks! :)

That was a design decision to keep all the cards consistent, and it has come up in multiple discussions previously.  The X and the Check, in this case, both server only to take the card out of edit mode.

Probably the best time to revisit this particular design decision will be when we update web-ui to PF4 and touch everything again.

sgratch:
> The `format` field within web-ui correlates to a REST `Disk` (http://ovirt.github.io/ovirt-engine-api-model/master/#types/disk) format attribute. The format attribute is a `DiskFormat` (http://ovirt.github.io/ovirt-engine-api-model/master/#types/disk_format) whose values are `cow` or `raw`.
> 
> I think you are talking about `VolumeFormat`.
> 
> Either way, I can update the name of the field in the component's state and deal with the data fields in the transform fields.

The 'format' REST attribute is mapped to the disk 'VolumeFormat' on backend (https://github.com/oVirt/ovirt-engine/blob/5a3e5ae1f8789e8b25c1213e86ceee1f8dd734a0/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/types/DiskMapper.java#L202), so we are referring to the same value.

sgratch:
> On Create VM Wizard, the default "Disk Type" is driven by the optimized for field on the basic setting step. That behavior should not change.

This seems wrong. 

You are totally right that the provisioning is based on the "optimized for" field in a way that: 
if if it's Server or High Performance  then provisioning=true (i.e. disks from the template are cloned) 
if it's Desktop then provisioning=false (i.e. disks from template are thin provisioned).
But it's only relevant for the template's disks. 
It's not relevant for new created disks. For new disks the default is always "thin provision" - so we need to fix that.

In addition,  it seems that it's not working correctly for existed template's disks as well:
if I create a VM based on a template with "optimized for" field set to Server/High Performance then the VM's disk should be created as Clone/Independent and uses the exact same Allocation Policy as the disk's template.  But it's not working as expected because the disk is always created by default as thin provision regardless.  

By looking at https://github.com/oVirt/ovirt-engine/blob/2aa3593f2b70871c7dd48408933b56d4d9f9d80a/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendVmsResource.java#L226 it seems that we need to use the CLONE parameter. Please also check:  http://ovirt.github.io/ovirt-engine-api-model/master/#services/vms/methods/add/parameters/clone

sgratch:
@sjd78 @hstastna  moving back to review state - please check my comments above.

sjd78:
@sgratch - The issue #1151 doesn't refer to Create VM behavior, so I'm not going to pull it apart at this point.  We can create a new issue and we can discuss (again) what the Create VM behavior should be for disk creation from template vs ISO and how the selected optimized for flag should drive the new disk creation of both template defined disks and new user defined disks.  We went rounds with Lucie and everyone before, I just don't remember all the discussed details.

As for the REST code, that `format` field still maps through only RAW, COW or null: https://github.com/oVirt/ovirt-engine/blob/5a3e5ae1f8789e8b25c1213e86ceee1f8dd734a0/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/types/DiskMapper.java#L252-L274

And `format` is a required field, so needs to be specified (https://github.com/oVirt/ovirt-engine/blob/5a3e5ae1f8789e8b25c1213e86ceee1f8dd734a0/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendDisksResource.java#L98).  I don't know the implications of raw vs cow and leaving `sparse` undefined or raw vs cow with `sparse` specified.

In terms of the REST api, specifying both `format` and `sparse` seems to be the correct thing to do to guarantee the desired results.  Lining it up with webadmin disks makes things annoying.  I can go ahead and rename this field and the component state fields to `diskType` and handle the `sparse` and `format` attributes in the `composeDisk()` function.  That'll probably make it a bit more obvious how the UX relates to the REST call.

sgratch:
> @sgratch - The issue #1151 doesn't refer to Create VM behavior, so I'm not going to pull it apart at this point. We can create a new issue and we can discuss (again) what the Create VM behavior should be for disk creation from template vs ISO and how the selected optimized for flag should drive the new disk creation of both template defined disks and new user defined disks. We went rounds with Lucie and everyone before, I just don't remember all the discussed details.

Sure, done. I opened a new issue for handling new disks for a new VM: #1237

> In terms of the REST api, specifying both `format` and `sparse` seems to be the correct thing to do to guarantee the desired results. Lining it up with webadmin disks makes things annoying. I can go ahead and rename this field and the component state fields to `diskType` and handle the `sparse` and `format` attributes in the `composeDisk()` function. That'll probably make it a bit more obvious how the UX relates to the REST call.

Right, sounds OK to me. So for creating new disks for an existing VM, we should specify both `format` and `type` properties.
The `type` is based on the "Disk Type" field chosen by the user in the UI. The default is always "thin provision" - so we need to fix that.
I'm not sure if setting the `format` is mandatory BTW (since based on storage domain format anyway), but we can continue using the "RAW" disk format as the default for now and fix that if issues will be raised. So just need to rename the fields as you suggested.

sjd78:
> I'm not sure if setting the `format` is mandatory BTW (since based on storage domain format anyway), but we can continue using the "RAW" disk format as the default for now and fix that if issues will be raised. So just need to rename the fields as you suggested.

Setting the `format` field on the disk is required to create a disk via REST.  It must be sent in the data.

isaranova:
LGTM

Issue #1205, opened 2020-05-10
Create VM Wizard - Validate add/edit NIC fields on row save action


sgratch:
On the "Create Virtual Machine" wizard -> "Networking" step, when adding/editing a NIC by clicking the checkmark box, need to validate all fields before saving and avoid saving in case of invalid values. Specifically need to make sure that the NIC name is unique and valid.
Consider display an error in case of a validation failure.

bamsalem:
In this dialog there is no error handling yet, so I can't use existing notification/ error message component so we need to add a new error message. I tried to add the error message but it looks bad:

![pasted image 0](https://user-images.githubusercontent.com/64131213/82205283-54b72980-98d4-11ea-9369-d33b2ab39807.png)

There is one error message for now, the error is about the NIC's name, something like: "the NIC's name should be unique with maximum X  characters and contains only A-Z...."

For now I can add "bubble message" or tooltip with the error message and color the field in a red color, but in a broad view I guess that in the future we will add more validations (disks' name field, etc...), so maybe we need to consider adding infrastructure for error handling messages which will serve us in the next validations.
there are no validation which uses a dedicated message box (like in the image above - but designed in a better way), dedicated error message box will create uniformity in handling errors in this dialog.

What would be the best solution for the error message component?

@lwrigh I would like to hear you opinion about that...

sgratch:
@lwrigh @bamsalem I guess we will use Patternfly 4 guidelines for that, right? It will be used for all steps so we need a general design for fields validation of displaying an error on current dialog step while disabling moving to next step as long as that error occurs. WDYT?

PR #1219, opened 2020-05-24
Create VM, Network - add validation to NIC name

bamsalem:
Fixes: #1205 

Hi there,
I added new validations to the "NIC's name" field, to notify the user about the validation status I made few visual changes too.
the changes are:
1. New information tooltip in the "NIC's name" cell which explains to the users the NIC's name rules (maybe we need new message formulation).

image

2. On edit/ create row state I used patternfly 4 components instead the old components (`TextInput`, `InputGroup`,  `FormSelect`, `FormSelectOption`), the `TextInput` component lets us to notify users that the name failed in the validation checks, the rest of components used for uniformity.

image

3. I added 2 tests for the propriety of the name and its uniqueness. the tests are the `isNicsNameValid` and `isNICsNameUnique` functions.

@sgratch @sjd78 @lwrigh  what do you think about that changes?

### DIFF
--- a/src/components/CreateVmWizard/steps/Networking.js
+++ b/src/components/CreateVmWizard/steps/Networking.js
@@ -10,6 +10,8 @@ import {
   createVNicProfileList,
   sortNicsDisks,
   suggestNicName,
+  isNicNameUnique,
+  isNicNameValid,
 } from '_/components/utils'
 
 import {
@@ -18,6 +20,8 @@ import {
   EmptyState,
   FieldLevelHelp,
   FormControl,
+  FormGroup,
+  HelpBlock,
   Label,
   MenuItem,
   Table,
@@ -73,8 +77,13 @@ class Networking extends React.Component {
     this.onDeleteRow = this.onDeleteRow.bind(this)
     this.onEditNic = this.onEditNic.bind(this)
     this.rowRenderProps = this.rowRenderProps.bind(this)
+    this.isVnicNameUniqueAndValid = this.isVnicNameUniqueAndValid.bind(this)
 
     this.state = {
+      editingErrors: {
+        nicInvalidName: false,
+        nicDuplicateName: false,
+      },
       editing: {},
       creating: false,
     }
@@ -136,14 +145,24 @@ class Networking extends React.Component {
         },
         editView: (value, { rowData }) => {
           const row = this.state.editing[rowData.id]
+          const { nicInvalidName, nicDuplicateName } = this.state.editingErrors
+          const validAndUniqueName = !nicInvalidName && !nicDuplicateName
 
           return (
-            <FormControl
-              id={`${idPrefix}-${value}-name-edit`}
-              type='text'
-              defaultValue={row.name}
-              onBlur={e => this.handleCellChange(rowData, 'name', e.target.value)}
-            />
+            <FormGroup
+              validationState={validAndUniqueName ? null : 'error'}
+              className={style['form-group-edit']}
+            >
+              <FormControl
+                id={`${idPrefix}-${value}-name-edit`}
+                type='text'
+                defaultValue={row.name}
+                onChange={e => this.handleCellChange(rowData, 'name', e.target.value)}
+              />
+              {!validAndUniqueName &&
+              <HelpBlock>{msg.createVmWizardNetVNICNameRules()}</HelpBlock>
+              }
+            </FormGroup>
           )
         },
       },
@@ -278,6 +297,11 @@ class Networking extends React.Component {
     ]
   }
 
+  isVnicNameUniqueAndValid (editingRow) {
+    const { nics } = this.props
+    return { unique: isNicNameUnique(nics, editingRow), valid: isNicNameValid(editingRow.name) }
+  }
+
   onCreateNic () {
     const newId = generateUnique('NEW_')
     const nextNicName = suggestNicName(this.props.nics)
@@ -305,6 +329,7 @@ class Networking extends React.Component {
         },
       },
     }))
+    this.props.onUpdate({ valid: false })
   }
 
   onEditNic (rowData) {
@@ -314,6 +339,7 @@ class Networking extends React.Component {
         [rowData.id]: rowData,
       },
     }))
+    this.props.onUpdate({ valid: false })
   }
 
   onDeleteRow (rowData) {
@@ -322,9 +348,20 @@ class Networking extends React.Component {
 
   handleCellChange (rowData, field, value) {
     const editingRow = this.state.editing[rowData.id]
+
     if (editingRow) {
       editingRow[field] = value
+      const editingErrors = {}
+      if (field === 'name') {
+        const { unique, valid } = this.isVnicNameUniqueAndValid(editingRow)
+        editingErrors.nicInvalidName = !valid
+        editingErrors.nicDuplicateName = !unique
+      }
       this.setState(state => ({
+        editingErrors: {
+          ...state.editingErrors,
+          ...editingErrors,
+        },
         editing: {
           ...state.editing,
           [rowData.id]: editingRow,
@@ -338,10 +375,15 @@ class Networking extends React.Component {
     const actionCreate = !!this.state.creating && this.state.creating === rowData.id
     const editedRow = this.state.editing[rowData.id]
 
-    // TODO: Add field level validation for the edit or create fields
+    let editingErrors = false
+    for (const errorKey of this.state.editingErrors) {
+      editingErrors = editingErrors || this.state.editingErrors[errorKey]
+    }
 
-    this.props.onUpdate({ [actionCreate ? 'create' : 'update']: editedRow })
-    this.handleRowCancelChange(rowData)
+    if (!editingErrors) {
+      this.props.onUpdate({ [actionCreate ? 'create' : 'update']: editedRow })
+      this.handleRowCancelChange(rowData)
+    }
   }
 
   // Cancel the creation or editing of a row by throwing out edit state
@@ -351,10 +393,15 @@ class Networking extends React.Component {
       const editing = state.editing
       delete editing[rowData.id]
       return {
+        editingErrors: {
+          nicInvalidName: false,
+          nicDuplicateName: false,
+        },
         creating: false,
         editing,
       }
     })
+    this.props.onUpdate({ valid: true })
   }
 
   // Create props for each row that will be passed to the row component (TableInlineEditRow)
--- a/src/components/utils/validation.test.js
+++ b/src/components/utils/validation.test.js
@@ -1,5 +1,5 @@
 /* eslint-env jest */
-import { isHostNameValid, isVmNameValid } from './validation'
+import { isHostNameValid, isVmNameValid, isNicNameValid, isNicNameUnique } from './validation'
 
 describe('check host names', function () {
   it('valid host names', function () {
@@ -34,5 +34,45 @@ describe('check VM names', function () {
     expect(isVmNameValid('AC/DC')).toEqual(false)
     expect(isVmNameValid('  abc  ')).toEqual(false)
     expect(isVmNameValid('veryveryveryveryveryveryveryveryveryveryveryveryveryveryverylongname')).toEqual(false)
+  })
+})
+
+describe('check NIC names', function () {
+  it('valid NIC names', function () {
+    expect(isNicNameValid('SomeName')).toEqual(true)
+    expect(isNicNameValid('1Ab')).toEqual(true)
+    expect(isNicNameValid('A_c')).toEqual(true)
+    expect(isNicNameValid('c1-b')).toEqual(true)
+    expect(isNicNameValid('A10-_.b')).toEqual(true)
+  })
+  it('invalid NIC names ', function () {
+    expect(isNicNameValid('SomeBadName?')).toEqual(false)
+    expect(isNicNameValid('The Worst name Ever!')).toEqual(false)
+    expect(isNicNameValid('')).toEqual(false)
+    expect(isNicNameValid('N@me$)')).toEqual(false)
+    expect(isNicNameValid('A1 b1')).toEqual(false)
+  })
+})
+describe('check NIC name uniqueness', function () {
+  const nicList = [
+    { id: '1', name: 'nic1' },
+    { id: '2', name: 'nic2' },
+    { id: '3', name: 'nic3' },
+    { id: '4', name: 'nic4' },
+  ]
+  it('duplicate name on create new NIC', function () {
+    expect(isNicNameUnique(nicList, { id: 'a', name: 'nic4' })).toEqual(false)
+  })
+  it('duplicate name on update NIC', function () {
+    expect(isNicNameUnique(nicList, { id: '4', name: 'nic3' })).toEqual(false)
+  })
+  it('unique name on update NIC same name', function () {
+    expect(isNicNameUnique(nicList, nicList[3])).toEqual(true)
+  })
+  it('unique name on update NIC new name', function () {
+    expect(isNicNameUnique(nicList, { id: '4', name: 'nic5' })).toEqual(true)
+  })
+  it('unique name on create new NIC', function () {
+    expect(isNicNameUnique(nicList, { id: '5', name: 'nic5' })).toEqual(true)
   })
 })
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -165,6 +165,7 @@ export const messages: { [messageId: string]: MessageType } = {
   createVmWizardStepTitleNetwork: 'Networking',
   createVmWizardStepTitleReview: 'Review',
   createVmWizardStepTitleStorage: 'Storage',
+  createVmWizardNetVNICNameRules: 'NIC name must be unique, 50 or less alphanumeric characters or "-_."',
   currentlyInsertedIsoInCdDrive: 'Currently inserted ISO in CD drive',
   customIcon: 'Custom icon of the virtual machine.',
   customScript: 'Custom script',
--- a/src/components/CreateVmWizard/steps/style.css
+++ b/src/components/CreateVmWizard/steps/style.css
@@ -147,6 +147,10 @@ table tbody tr td.kebab-menu-cell {
   padding: 2px 10px 1px;
 }
 
+.form-group-edit{
+  margin-bottom: 0;
+}
+
 /*
  * Summary Review
  */
--- a/src/components/utils/validation.js
+++ b/src/components/utils/validation.js
@@ -90,3 +90,23 @@ export function isVmNameValid (nameCandidate: string): boolean {
 export function isDiskNameValid (nameCandidate: string): boolean {
   return isVmNameValid(nameCandidate)
 }
+
+/**
+ * check if the name's length between 1 to 50 characters
+ * and contains only uppercase, lowercase, hyphens and underscores
+ */
+export function isNicNameValid (name: string): boolean {
+  return /^[a-zA-Z0-9_\-\\.]{1,50}$/.test(name)
+}
+
+/**
+ * check if the nicEditing name is unique to the given nicList
+ */
+export function isNicNameUnique (nicList: Object, nicEditing: Object): boolean {
+  const nicsMatchEditingName = nicList.filter((nic) => nic.name === nicEditing.name)
+
+  const nameUnique = nicsMatchEditingName.length === 0 ||
+    (nicsMatchEditingName.length === 1 && nicsMatchEditingName[0].id === nicEditing.id)
+
+  return nameUnique
+}


sgratch:
@hstastna  
>     1. How should it work regarding the buttons, after the name turned to be invalid?
>        I still can click on the _Next_ button and to ignore the unfinished creating/editing of a NIC, and this is little bit confusing. Do we want it like that?

I agree, we should disable the button.

> 
>     2. We are adding PF 4 components in this PR and making the Create VM Wizard inconsistent. The drop downs in other steps are different etc.. It does not feel good. Yes, I understand that using PF 4 components here has some advantages but also some disadvantages.. Are we ok with that?
> 
>     3. We are adding `QuestionCircleIcon` here to inform the user about how to choose a valid NIC's name. But for example in the Storage step, we use a different icon information_source (in a circle, I think it's the `FieldLevelHelp` component). And I think the purpose is the same: to inform the user what they can/cannot do, regarding the specific field(s). Is there any special reason to use a different icon here than in the Storage step? It's just a detail but it feels little bit inconsistent with what we can experience for example in the Storage step. But maybe it's because of PF 4 ...

If we will use the PF4 tooltip only in case of a validation error on create VM dialog (check my review comment), then I think it'll look much better since currently there is no error validation at all for that dialog. So adding something new can use PF4. 
Regardless, we have a task for refactoring all tooltips/icons to be more consistent: https://github.com/oVirt/ovirt-web-ui/issues/1094

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -132,14 +142,26 @@ class Networking extends React.Component {
         },
         editView: (value, { rowData }) => {
           const row = this.state.editing[rowData.id]
+          const nicNameErrFlag = this.state.editedRowNameErrorsFlag
```
sjd78, line 145: This is inconsistent with how the edit data is being stored.  The state data is setup to potentially allow simultaneous editing of rows.  This probably won't ever be the case, so the error flags can be done without regards to supporting multiple edit rows being active at the same time.

Something like `this.state.editingErrors.nicInvalidName` and/or `this.state.editingErros.nicDuplicateName` would be easier to understand at first reading.

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -132,14 +142,26 @@ class Networking extends React.Component {
...
+                  isValid={!nicNameErrFlag}
+                  id={`${idPrefix}-${value}-name-edit`}
+                  type='text'
+                  className={style['nic-edit-row']}
+                  defaultValue={row.name}
+                  onBlur={e => this.handleCellChange(rowData, 'name', e.target.value)}
+                />
+                <OverlayTooltip id={`${idPrefix}-nics-name-rules`} placement={'bottom'} tooltip={msg.createVmNetVnicNameRules()}>
+                  <Button variant={'control'} aria-label={'popover for input'}>
+                    <QuestionCircleIcon />
```
sjd78, line 160: Could you use `FieldLevelHelp` here instead?

src/components/utils/utils.js
```diff
@@ -30,3 +31,23 @@ export function escapeHtml (s: string): string {
...
+ * and contains only uppercase, lowercase, hyphens and underscores
+ */
+export function isNicsNameValid (name: string): boolean {
+  return /^([a-z]|[A-Z]|[0-9]|_|-){1,21}$/.test(name)
+}
+
+/**
+ * check if the name of a given nic's row is unique to the given array of rows
+ */
+export function isNICsNameUnique (nicsToCompareWith: Object, MaxOccurrence: number, actionCreate: boolean, comparedRow: Object): boolean {
```
sjd78, line 46: vars and parameters should start with a lowercase letter

bamsalem:
@sgratch @sjd78 @hstastna 
Regarding the commit message I will edit that when I finish to fix all the rejects :)
Regarding to VNIC validations I took that from the link you guys sent me on slack: https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/virtual_machine_management_guide/Virtual_Machine_Network_Interface_dialogue_entries
I got confused between network interface and Network interface card.
now I think that I found the correct one: https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/administration_guide/sect-virtual_network_interface_cards 
image
regarding all the rest... back to work :)
thanks guys

bamsalem:
Hi friends I want to consult you,
I have few options/ideas where to place the error message and I want to hear your opinion, or maybe you have another ideas,if you have so I would like to hear...
so there are 3 options:
Alert at the bottom of the table:
alert
Alert at the bottom of the table with description:
alertWithDesc
The disadvantage of these is in case of long list the user might miss the error message.
popover for the name field:
image
All the options are not fully implemented yet, I want to hear your opinion before I keep further if one of them...
WDYT?
@sgratch @sjd78 @hstastna

src/intl/messages.js
```diff
@@ -118,6 +118,7 @@ export const messages: { [messageId: string]: MessageType } = {
   createVmNetEmptyInfo: 'No NICs have been defined for this VM.',
   createVmNetEmptyTitle: 'No NICs Defined',
   createVmNetNoEditHelpMessage: 'This NIC is defined by your selected template and cannot be edited or deleted at creation.',
+  createVmNetVnicNameRules: "Nic's name must be unique and up to 21 characters consisting only lowercase(a-z), uppercase(A-Z), numeric(0-9) , '_', '-'",
```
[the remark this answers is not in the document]
bamsalem: Thanks, I didn't found anything about '.' in the documentation:
https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/administration_guide/sect-virtual_network_interface_cards
are you sure '.' is allowed?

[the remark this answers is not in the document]
bamsalem: Thanks

src/components/utils/utils.js
```diff
@@ -30,3 +31,23 @@ export function escapeHtml (s: string): string {
 export function maskForElementId (name: string): string {
   return name && name.replace(/[\W\s]+/g, '_')
 }
+
+/**
+ * check if the name's length between 1 to 21 characters
+ * and contains only uppercase, lowercase, hyphens and underscores
+ */
+export function isNicsNameValid (name: string): boolean {
```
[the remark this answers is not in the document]
bamsalem: Thanks, moved to src/components/utils/validation.js

src/components/utils/utils.js
```diff
@@ -30,3 +31,23 @@ export function escapeHtml (s: string): string {
...
+  return /^([a-z]|[A-Z]|[0-9]|_|-){1,21}$/.test(name)
+}
+
+/**
+ * check if the name of a given nic's row is unique to the given array of rows
+ */
+export function isNICsNameUnique (nicsToCompareWith: Object, MaxOccurrence: number, actionCreate: boolean, comparedRow: Object): boolean {
+  let occurrences = nicsToCompareWith.filter((value) => localeCompare(value.name, comparedRow.name) === 0)
```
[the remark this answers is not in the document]
bamsalem: done.

hstastna:
In my opinion, I would try to inspire myself by this:
warn
But as you've mentioned, there's the disadvantage in case of long list of NICs. Yes, I think we should care about that, too. But anyway, I'm not sure how many NICs can be created there.

Your first two suggestions look similar to the screenshot I've posted above but I can see another issue with such an approach:
err
The buttons displayed above the error message look little bit weird. I think nothing should be displayed above the error message. If we were able to place the error message between the NICs in the list - we would solve the issues but I still wouldn't be sure if that's ok. I haven't seen anything like that.

So maybe those suggestions are not the best. The popover seems to be better but I haven't seen anything like that in Create VM Wizard yet so for me it would feel little bit inconsistent with other stuff. But maybe is similar to our info icons and tooltips in the Wizard.

To sum up, I am not sure right now what's the best. I wish we had some help of @lwrigh

bamsalem:
I have one more idea:

valid & unique name:
image

invalid or duplicated name (error):
image

WDYT?

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -9,20 +9,30 @@ import {
...
+  isNICsNameUnique,
+  isNicsNameValid,
 } from '_/components/utils'
 
+import {
+  TextInput,
+  InputGroup,
+  FormSelect,
+  FormSelectOption,
+} from '@patternfly/react-core'
```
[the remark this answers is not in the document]
bamsalem: like I said in of our standups the React-PF3 documentation is missing, so it's reasonable I missed PF3 error state capabilities...
thanks for the link, it was really helpful :)

sgratch:
> image

@bamsalem  I would prefer the shorter version: "The NIC name must be unique, not more than 50 chars and formed of alphanumeric characters, numbers or "-_."

sgratch:
@bamsalem  
I would prefer the popover option and also consider adding a red asterisk next to the field name as @lwrigh suggested on a separate mail for the storage domain field:
image

Please note that this is already implemented for the VM name field  (without a popover tooltip solution-need to fix that as well):
image

hstastna:
@sgratch @bamsalem @lwrigh  
I would go for [this](https://github.com/oVirt/ovirt-web-ui/pull/1219#issuecomment-648680541) as user would still have access to the information they need and they would be less distracted without the popover when clicking in the field. But that's just my opinion.

Also I am not sure how exactly you would like to use the red asterisk to communicate that the NIC's name is required. I think it would be misleading to put the asterisk next to the whole row. Or did you mean to put the asterisk next to the "Name" in the header of the table? :)

sjd78:
@sgratch , @hstastna , @bamsalem - I would not bother with a '*' to denote a required field.  There isn't really a good place to put it, and the field is already prefilled.  I think it is enough to have 3 error conditions (name is required, name is valid, name is unique on the VM).

In all 3 cases, highlight the field in red (the error state).  Chatting with @lwrigh yesterday, and she recommended making the row taller and displaying the error message below the field in an error condition.  That makes sense to me.  Something like this example:

patternfly - react - Forms and Controls - Forms - Forms ( FormGroup , FormControl , ControlLabel ) ⋅ Storybook - Mozilla Firefox_021

Mocked up looks like:
Selection_022

sgratch:
> Mocked up looks like:
> Selection_022

This looks ok to me. The only disadvantage of displaying the error message below the field instead of using a tooltip is that in case of a long message (more than 3-5 words) it will become problematic to display all. Since we will use that design decision for displaying all error messages within the wizard, and sometimes errors are longer than 3 words then it might become too dense. And I don't think it worthy splitting each error to specific sub errors just to handle that (name is required, name is valid, name is unique on the VM). I prefer one error message per specific field validation (so that in case validation logic is changed on backend, we will need to fix only one message).

lwrigh:
I would not go for the popover option for displaying the error message since it feels quite jarring. I would suggest adding error text below the field like the PF3 form with an error suggests: https://www.patternfly.org/v3/pattern-library/widgets/#forms

hstastna:

bamsalem:
Hi guys I made the changes like you asked:
image
In my opinion I think we need to find another solution or shorten the message because it spread over too many lines...
What do you think?
@sgratch @hstastna @lwrigh @sjd78 @rszwajko

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -318,14 +344,27 @@ class Networking extends React.Component {
...
-          ...state.editing,
-          [rowData.id]: editingRow,
-        },
-      }))
+      if (field === 'name') {
+        const { unique, valid } = this.isVnicNameUniqueAndValid(editingRow)
+        const validAndUniqueName = unique && valid
+        editingErrors.nicInvalidName = !valid
+        editingErrors.nicDuplicateName = !unique
+        this.props.onUpdate({ valid: validAndUniqueName }, false)
```
sjd78, line 355: Shouldn't need the `false` here if you update the `onUpdate` related functions as suggested in `CreateVmWizard`.

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -346,10 +386,16 @@ class Networking extends React.Component {
...
+        ...state,
+        editingErrors: {
+          nicInvalidName: false,
+          nicDuplicateName: false,
+        },
         creating: false,
         editing,
       }
     })
+    this.props.onUpdate({ valid: true }, false)
```
sjd78, line 398: Same here, probably don't need the `false`

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -149,9 +149,13 @@ class CreateVmWizard extends React.Component {
...
             nics={this.state.steps.network.nics}
 
-            onUpdate={({ valid = true, ...updatePayload }) => {
-              this.handleListOnUpdate('network', 'nics', updatePayload)
+            onUpdate={({ valid = true, ...updatePayload }, updateList = true) => {
               this.wizardStepsMap.network.preventExit = !valid
+              if (updateList) {
+                this.handleListOnUpdate('network', 'nics', updatePayload)
+              } else {
+                this.forceUpdate()
```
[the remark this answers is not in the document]
bamsalem: yes, we dealt with this in a few issues...
I asked about it in another issue:
https://github.com/oVirt/ovirt-web-ui/pull/1233#discussion_r451315065
maybe we should consider to change it...

michalskrivanek:
> "NIC name must be unique, less than 50 alphanumeric characters, numbers or "-_."
alphanumeric covers numbers, so it can shorter;)

Also, is "." really valid on its own? I doubt I can create a NIC named "..-..__..-..", but all right, it's not the point to describe exact regexp

bamsalem:
> Also, is "." really valid on its own? I doubt I can create a NIC named "..-..__..-..", but all right, it's not the point to describe exact regexp


@michalskrivanek 
[Documentation](https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/administration_guide/sect-virtual_network_interface_cards)
According to the documentation and web admin, Yes:
image

src/components/CreateVmWizard/CreateVmWizard.js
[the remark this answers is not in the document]
sjd78: It's not that big of an issue.  Really you could test if `preventExit` will be changed and only call `forceUpdate()` if it has been changed.

If we want to change the `preventExit` bits as an explicit part of state, we'd do it in another PR and for all wizard steps.

src/components/utils/validation.test.js
```diff
@@ -36,3 +36,69 @@ describe('check VM names', function () {
...
+      vnicProfileId: 'af4d8ef3-b379-4044-832b-cb086e1384e4',
+    },
+  ]
+  const createdDuplicateName = { deviceType: 'virtio', id: 'NEW_3849c219-bc68-4c3e-a851-a7f6e0tsv9l3', name: 'nic4', vnicProfileId: '1d32a4e5-7851-43e1-903c-25c1fd0a792d' }
+  const editedUniqueNewName = { deviceType: 'virtio', id: 'NEW_3849c219-bc68-4c3e-a851-a7f6e0bc4ae1', name: 'nic5', vnicProfileId: 'af4d8ef3-b379-4044-832b-cb086e1384e4' }
+  const createUniqueName = { deviceType: 'virtio', id: 'NEW_3849c219-bc68-4c3e-a851-a7f6e0tsv9l3', name: 'nic5', vnicProfileId: '1d32a4e5-7851-43e1-903c-25c1fd0a792d' }
+  const editedDuplicate = { deviceType: 'virtio', id: 'NEW_3849c219-bc68-4c3e-a851-a7f6e0bc4ae1', name: 'nic3', vnicProfileId: 'af4d8ef3-b379-4044-832b-cb086e1384e4' }
+  const editedUniqueSameName = nicList[3]
+  it('duplicate name on create new NIC', function () {
+    expect(isNicNameUnique(nicList, createdDuplicateName)).toEqual(false)
```
sjd78, line 90: Tests look ok, but I wonder if a different format would make them easier to read.  Something like:
```js
    expect(isNicNameUnique(nicList, { id: 'A', name: 'nic4' })).toEqual(false)
```

assuming adjustments to `nicList` more like this, since it's easier to look at test data with simple ids:
```js
const nicList = [
  { id: '1', name: 'nic1' },
  { id: '2', name: 'nic2' },
  { id: '3', name: 'nic3' },
  { id: '4', name: 'nic4' },
]
```

The other tests can be adjusted in the same way.

sjd78:
@hstastna , @bamsalem - The button jumping is a new issue that we should deal with, but on another PR.  The handling of row height with the error message will need to be updated to keep the buttons at the bottom of the row at all times.  Probably will need to look at `src/components/CreateVmWizard/steps/_TableInlineEditRow.js` to make it happen.

src/components/utils/validation.test.js
bamsalem: I agree, my first thought was to use real data, but it really makes it clearer

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -388,11 +392,15 @@ class CreateVmWizard extends React.Component {
...
+    const moveBackward = newStepIndex < activeStepIndex
     const activeStep = this.wizardSteps[activeStepIndex]
     const newStep = this.wizardSteps[newStepIndex]
 
+    // enable leaving the current step in case of moving backward
+    const preventExit = moveBackward ? false : activeStep.preventExit
+
     // make sure we can leave the current step and enter the new step
-    if (activeStep.preventExit || newStep.preventEnter) return
+    if (preventExit || newStep.preventEnter) return
```
sjd78, line 403: I'm not sure it makes sense to touch the GoToStep handling in this PR.  It goes along with the back button hanlding that @hstastna was talking about.  All that you're doing here is allowing the back button to be clicked always.

Let's just leave current functionality in place.  If the editing is invalid, force the user to fix and save or cancel the change.  We can handle more complex navigation consistently across steps in future issues (if an issue for it doesn't already exist).

hstastna: If we force the user to fix and save or cancel the change, we should implement the same behavior also in https://github.com/oVirt/ovirt-web-ui/pull/1221 to be consistent.

sjd78: The `this.props.onUpdate({ valid: validAndUniqueName })` work can easily be replicated to #1221 after this PR is merged.

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -149,9 +149,13 @@ class CreateVmWizard extends React.Component {
             clusterId={this.state.steps.basic.clusterId}
             nics={this.state.steps.network.nics}
 
-            onUpdate={({ valid = true, ...updatePayload }) => {
-              this.handleListOnUpdate('network', 'nics', updatePayload)
+            onUpdate={({ valid = true, ...updatePayload }, updateList = true) => {
```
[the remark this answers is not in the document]
sjd78: @bamsalem - I know you dropped `updateList`, but you didn't make any accommodations for it being empty to avoid rerenders with no deep state updates.

Please add the extra checks I detailed above to `handleListOnUpdate()`.

src/components/CreateVmWizard/CreateVmWizard.js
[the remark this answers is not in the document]
sjd78: See #1260. These changes can be dropped after rebasing on that commit.

src/components/CreateVmWizard/CreateVmWizard.js
sjd78: With #1260 both this PR and #1221 can use the same `onUpdate` calls.

sjd78:
> There's one more thing I would like to talk about and it's related also to some other steps and validation in the Create VM Wizard. I don't understand, why we can click on the _Next_ button if there's any unsaved NIC:
> unsaved
> I click on the button and I am in the next step and the unsaved NIC was just thrown away. Okay, less clicking for me. But when I have more NICs in the _Networking_ step and some has invalid name, I cannot click on the _Next_ button and simply to throw it away and to move forward.
> 
> In one of my my previous [comments](https://github.com/oVirt/ovirt-web-ui/pull/1219#pullrequestreview-432330541) I was not sure about what should happen in this situation. @sgratch has answered that we should disable the _Next_ button in this case. But I was thinking and now I am asking: really? Should we disable the button in this case? What's the difference between throwing away unsaved NIC and throwing away unsaved (and invalid) NIC? An invalid and unsaved NIC does not matter. We still can create a VM successfully. I would expect not to be able to move to the next step only if there's any really necessary thing I should do to be able to move to the next step and to create a VM successfully.
> 
> In short, to achieve consistency, I think that if we don't want user to move to the next step if anything unsaved is invalid, we should implement the same behavior also for any unsaved and valid NIC (and also Disk). Or to let the user go back and forth anytime if there isn't any crucial thing for creating a VM properly. Or if we had some clear UX design for (not only) this feature somewhere so everybody would be sure what's the expected behavior.

Ok, so currently you can't exit from Basic until the form is valid.

So, adding the same behavior to Networks and Storage is easier.  If editing or creating a Nic or Disk, don't allow the user to exit the step.  Ultimately, we want to the Nic/Disk create/edit save button to be blocked until the fields are all valid.  When the create/edit is done (save or cancel), then users can exit the step.

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -318,14 +342,27 @@ class Networking extends React.Component {
...
+        const { unique, valid } = this.isVnicNameUniqueAndValid(editingRow)
+        const validAndUniqueName = unique && valid
+        editingErrors.nicInvalidName = !valid
+        editingErrors.nicDuplicateName = !unique
+        this.props.onUpdate({ valid: validAndUniqueName })
+      }
+      this.setState(state => {
+        const newState = {
+          ...state,
+          editingErrors: merge({}, state.editingErrors, editingErrors),
```
sjd78, line 358:
```suggestion
          editingErrors: {
            ...state.editingErrors,
            ...editingErrors,
          },
```

bamsalem:
Please review it again after the last changes

hstastna (changes requested):
Right after clicking on _Create NIC_ in the _Networking_ step:
prop
Otherwise, validation of the NIC names works as expected.

---

There's one little thing which we could improve: the order of NICs in the **Review** step is different than in the **Networking** step:
nics
review
Probably not related to the NIC name validation but related to the NIC names. I'm not sure what's the proper order of NICs but it's inconsistent.

hstastna:
Note that sorting NICs will be fixed via https://github.com/oVirt/ovirt-web-ui/issues/1281 as there's the same issue regarding Disks created in the Storage step.

hstastna (changes requested):
Alphanumeric characters include numbers, so we can make the message even shorter. Some related comment: https://github.com/oVirt/ovirt-web-ui/pull/1219#issuecomment-661026881

Also the message says that the name has to be **less** than 50 chars long, but I was able to successfully save NIC with the 50 chars long name, it was valid. Only 51 and more chars long name was invalid, regarding the length. If we allow 50 chars long NIC names, maybe the message should be changed to "... 50 and less chars long" or "less than 51". If not, the message should be changed appropriately :)

rszwajko (changes requested):
Save/cancel buttons jump after correcting an error. See below:
nic_validation

rszwajko (approved):

hstastna (approved):
LGTM.

@sjd78 , @sgratch Just a question: We wanted to add this validation only in the CreateVmWizard. Right?
For already existing VM, when editing _nic5_, I was able to choose already existing name _nic2_ and to click on _OK_ button without any warning or errors (which will be inconsistent with the CreateVmWizard with this new validation):
nic
Do we have plans to add the validation also here? Thanks!

bamsalem:
@sgratch @sjd78 
I think it looks ready, WDYT?

sjd78:
> @sjd78 , @sgratch Just a question: We wanted to add (not only) this validation only in the CreateVmWizard. Right?
> For already existing VM, when editing _nic5_, I was able to choose already existing name _nic2_ and to click on _OK_ button without any warning or errors (which will be inconsistent with the CreateVmWizard with this new validation):
> nic
> Do we have plans to add the validation also here? Thanks!

We could do front end validation in VM Detail Nic editing as well.  Just needs to be added as a separate issue so this PR doesn't wait any longer. @hstastna, go ahead and raise an issue and reference this PR and your comment.

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -337,10 +375,11 @@ class Networking extends React.Component {
     const actionCreate = !!this.state.creating && this.state.creating === rowData.id
     const editedRow = this.state.editing[rowData.id]
 
-    // TODO: Add field level validation for the edit or create fields
-
-    this.props.onUpdate({ [actionCreate ? 'create' : 'update']: editedRow })
-    this.handleRowCancelChange(rowData)
+    const { nicInvalidName, nicDuplicateName } = this.state.editingErrors
```
sjd78, line 378: You could do this generically, and just check to make sure every key in `editingErrors` is not true...

```jsx
  const editingErrors = false
  for (const errorKey of this.state.editingErrors) {
    editingErrors = editingErrors || this.state.editingErrors[errorKey]
  }

  if (!editingErrors) {
    ...
  }
```

sjd78 (dismissed):
A few minor changes

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -350,10 +389,16 @@ class Networking extends React.Component {
       const editing = state.editing
       delete editing[rowData.id]
       return {
+        ...state,
+        editingErrors: {
```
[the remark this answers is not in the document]
sjd78: Prefer to leave it as-is?

sjd78 (approved):
LGTM

sgratch:
> > @sjd78 , @sgratch Just a question: We wanted to add (not only) this validation only in the CreateVmWizard. Right?
> > For already existing VM, when editing _nic5_, I was able to choose already existing name _nic2_ and to click on _OK_ button without any warning or errors (which will be inconsistent with the CreateVmWizard with this new validation):
> > nic
> > Do we have plans to add the validation also here? Thanks!
> 
> We could do front end validation in VM Detail Nic editing as well. Just needs to be added as a separate issue so this PR doesn't wait any longer. @hstastna, go ahead and raise an issue and reference this PR and your comment.

I don't think it's necessary. Please check comment https://github.com/oVirt/ovirt-web-ui/issues/1294#issuecomment-697294796

isaranova:
LGTM

Issue #1198, opened 2020-05-05
Create VM Wizard - VM time zone format should be compatible with VM OS and also displayed on the review report step


sgratch:
This includes two parts for handling the VM timezone:

1. Apply the fix done for previous create VM  #1118, on existing create VM flow: 
If template's time zone format or VM's time zone format is not compatible with VM OS then it should change time zone to default. For Windows default time zone if: GMT Standard Time, for other OSs: Etc/GMT

2. Consider displaying current timezone in the review report step (not just for windows but for all OS types) so that the user will know which timezone was actually set.

hstastna:
Just a note that this is because Windows and Linux use different time zone names and the (default or otherwise) template time zone could be in the wrong format.

hstastna:
@sgratch What time zone should I choose when user sets ISO as a Provision Source? I would not set any time zone and let the backend to choose. Creating VM works fine when I don't choose any time zone in this case.

sgratch:
> @sgratch What time zone should I choose when user sets ISO as a Provision Source? I would not set any time zone and let the backend to choose. Creating VM works fine when I don't choose any time zone in this case.

You should choose the default based on the OS, as done on current 1220 PR fix.

PR #1220, opened 2020-05-25
Make VM time zone format compatible with the VM OS when creating a VM, display time zone in the Review report step

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1198

In this PR I:
- make VM time zone format compatible with the selected VM OS by implementing the [fix](https://github.com/oVirt/ovirt-web-ui/pull/1118) previously done for editing an existing VM - this change is related to the "Basic Settings" (first) step of the Create VM Wizard. I've used the `checkTimeZone` function for the case when:
  - changing or selecting the VM OS
  - changing _Provision Source_ to _ISO_ (set the default time zone according to the OS, as we don't have any specific info about the selected iso image)
  - changing or selecting the Template, because we need to check template's timezone compatibility with the template's OS and then to set the timezone corresponding to the template's OS if needed
- make the current Time zone displayed in the "Review" report step of the Create VM Wizard (last step)

---

**Before:**
tz_before

**After:**
tz_after

### DIFF
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -255,7 +255,6 @@ class BasicSettings extends React.Component {
       case 'templateId':
         const template = this.props.templates.find(t => t.get('id') === value)
         changes.templateId = template.get('id')
-        changes.timeZone = template.get('timeZone').toJS()
 
         changes.memory = template.get('memory') / (1024 ** 2) // stored in bytes, input in Mib
         changes.cpus = template.getIn([ 'cpu', 'vCPUs' ])
@@ -264,6 +263,8 @@ class BasicSettings extends React.Component {
           .find(os => os.get('name') === template.getIn([ 'os', 'type' ]))
           .get('id')
 
+        // Check template's timezone compatibility with the template's OS, set the timezone corresponding to the template's OS
+        changes.timeZone = this.checkTimeZone(changes.operatingSystemId, changes.templateId)
         changes.cloudInitEnabled = template.getIn(['cloudInit', 'enabled'])
         changes.initHostname = template.getIn(['cloudInit', 'hostName'])
         changes.initSshKeys = template.getIn(['cloudInit', 'sshAuthorizedKeys'])
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import { localeCompare } from '_/helpers'
+import { localeCompare, isWindows } from '_/helpers'
 import { msg } from '_/intl'
 import { isNumberInRange } from '_/utils'
 import { BASIC_DATA_SHAPE } from '../dataPropTypes'
@@ -18,6 +18,8 @@ import {
 import { Form, FormControl, Checkbox } from 'patternfly-react'
 import { Grid, Row, Col } from '_/components/Grid'
 import SelectBox from '_/components/SelectBox'
+
+import timezones from '_/components/utils/timezones.json'
 
 import style from './style.css'
 
@@ -113,6 +115,7 @@ export const optimizedForMap = {
 class BasicSettings extends React.Component {
   constructor (props) {
     super(props)
+    this.checkTimeZone = this.checkTimeZone.bind(this)
     this.handleChange = this.handleChange.bind(this)
     this.validateForm = this.validateForm.bind(this)
     this.validateVmName = this.validateVmName.bind(this)
@@ -125,6 +128,35 @@ class BasicSettings extends React.Component {
         'startOnCreation', 'cloudInitEnabled',
       ],
     }
+  }
+
+  checkTimeZone (osId, templateId) {
+    const { defaultGeneralTimezone, defaultWindowsTimezone } = this.props
+    const template = templateId ? this.props.templates.get(templateId) : undefined
+
+    let timeZone = {
+      name: defaultGeneralTimezone,
+    }
+    const osType = this.props.operatingSystems.getIn([ osId, 'name' ])
+
+    if (template && template.getIn(['timeZone', 'name'])) {
+      timeZone = template.get('timeZone').toJS()
+    }
+
+    const isWindowsTimeZone = timeZone && timezones.find(timezone => timezone.id === timeZone.name)
+    const isWindowsVm = isWindows(osType)
+
+    if (isWindowsVm && !isWindowsTimeZone) {
+      timeZone = {
+        name: defaultWindowsTimezone,
+      }
+    }
+    if (!isWindowsVm && isWindowsTimeZone) {
+      timeZone = {
+        name: defaultGeneralTimezone,
+      }
+    }
+    return timeZone
   }
 
   validateForm (dataSet) {
@@ -213,11 +245,17 @@ class BasicSettings extends React.Component {
         changes.cpus = this.props.defaultValues.cpus
         changes.optimizedFor = this.props.defaultValues.optimizedFor
         changes.cloudInitEnabled = this.props.defaultValues.initEnabled
+
+        // when changing Provision type to ISO, set the default time zone according to the OS
+        if (changes.provisionSource === 'iso') {
+          changes.timeZone = this.checkTimeZone(changes.operatingSystemId, changes.templateId)
+        }
         break
 
       case 'templateId':
         const template = this.props.templates.find(t => t.get('id') === value)
         changes.templateId = template.get('id')
+        changes.timeZone = template.get('timeZone').toJS()
 
         changes.memory = template.get('memory') / (1024 ** 2) // stored in bytes, input in Mib
         changes.cpus = template.getIn([ 'cpu', 'vCPUs' ])
@@ -231,6 +269,11 @@ class BasicSettings extends React.Component {
         changes.initSshKeys = template.getIn(['cloudInit', 'sshAuthorizedKeys'])
         changes.initTimezone = template.getIn(['cloudInit', 'timezone'])
         changes.initCustomScript = template.getIn(['cloudInit', 'customScript'])
+        break
+
+      case 'operatingSystemId':
+        changes[field] = value
+        changes.timeZone = this.checkTimeZone(value, this.props.data.templateId)
         break
 
       case 'memory':
@@ -524,6 +567,9 @@ BasicSettings.propTypes = {
   maxMemorySizeInMiB: PropTypes.number.isRequired,
 
   onUpdate: PropTypes.func.isRequired,
+
+  defaultGeneralTimezone: PropTypes.string.isRequired,
+  defaultWindowsTimezone: PropTypes.string.isRequired,
 }
 
 export default connect(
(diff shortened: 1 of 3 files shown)


ovirt-infra:

sgratch:
ci add to whitelist

hstastna:
> Looks OK to me except one issue:
> When selecting a template as a source, the VM's OS is set based on the template's timezone while it should be the other way around. and set based on the template's OS only
> I.e. in case the chosen template's timezone is not compatible with the template's OS for some reason, we should set VM's OS based on template's OS and change the timezone value to be compatible to the OS if needed.

I am not sure if I understand this issue. When we choose between templates, the VM's OS is changed automatically - this behavior was already there, it is not my choice. So I consider ok to get the template's timezone because I expect to be in harmony with the VM's OS. And when we choose template as a provision source, nothing happens, we have to choose a specific template or the OS to make some change in the timezone to happen. And if I applied the same [code](https://github.com/oVirt/ovirt-web-ui/pull/1220/files#diff-cef40fc8d8b38f14df16a31106f8d660R243-R247) also on the situation when user switches from ISO to Template provision source, we would end up with the same result regarding the chosen time zone.

michalskrivanek:
> When selecting a template as a source, the VM's OS is set based on the template's timezone while it should be the other way around

I would say a Template selection should reset the previous values. It carries TZ as well as other things and it is to be expected that those fields are replaced

hstastna:
@sgratch When I look into the actual VM time zone format fix when editing an existing VM, the time zone is [here](https://github.com/oVirt/ovirt-web-ui/pull/1118/files#diff-9c54d51933c15d7dabcb3a1008b6b1caR332) set according to the chosen template and not VM's OS (it any template was chosen, of course).

hstastna:
> I would say a Template selection should reset the previous values. It carries TZ as well as other things and it is to be expected that those fields are replaced

So you are saying that you would check TZ no matter we choose as a Provision Source? I think it is unnecessary but I can make appropriate change.

When we choose Template as a Provision Source, we have to choose a specific Template (or ISO) to continue adding a VM. And when we choose a specific Template, the TZ is set, and the VM OS is set with respect to the template. We really need to check the TZ only when we change the VM OS.

sgratch:
> Looks OK to me except one issue:
> When selecting a template as a source, the VM's OS is set based on the template's timezone while it should be the other way around. and set based on the template's OS only
> I.e. in case the chosen template's timezone is not compatible with the template's OS for some reason, we should set VM's OS based on template's OS and change the timezone value to be compatible to the OS if needed.

@hstastna @michalskrivanek All you mentioned seems OK and working as expected. 

The specific scenario I raised here is that in case we try to create a VM based on a template with an invalid tmz (e.g. a template set with Windows OS and non-windows TMZ or vice versa) then on webAdmin it fails with the relevant error and the same goes on previous web-admin code, 
but on current code after this fix we create the VM with the template's timezone (by changing the OS to be compatible) instead of creating based on the template's OS and changing the tmz to be compatible. We can also fail the VM creation as well. As long as we don't change the template's OS because of the tmz which is wrong to me.

hstastna:
@sgratch Thank you for raising the specific scenario we should definitely take care of. But I think I still don't understand it. Are you saying that we do want to see [this](https://user-images.githubusercontent.com/13417815/82832322-051bc380-9ebb-11ea-83fe-8980d387144c.png) error?

> ... we create the VM with the template's timezone (by changing the OS to be compatible) instead of creating based on the template's OS and changing the tmz to be compatible. We can also fail the VM creation as well. As long as we don't change the template's OS because of the tmz which is wrong to me.

Do you mean that we should change the template's OS somehow in the process of creating a VM? I would expect that there's nothing wrong with any chosen template and that the template's OS is in harmony with the template's TZ. I thought that the issue I am trying to fix here is compatibiliy of template's TZ with the chosen VM OS, not compatiblity of the template's TZ with the template's OS.

sgratch:
> Do you mean that we should change the template's OS somehow in the process of creating a VM? 

No, we shouldn't change the template, only the VM based on that template. The VM's OS will be the same as the template's OS and then we should check the template's TZ. If it fits then use that. If it doesn't fit then change to the correct default TZ or leave it as is and it will fail on backend. 

> I would expect that there's nothing wrong with any chosen template and that the template's OS is in harmony with the template's TZ. I thought that the issue I am trying to fix here is compatibiliy of template's TZ with the chosen VM OS, not compatiblity of the template's TZ with the template's OS.

You are right and it's a specific case. But this fix should handle all cases. 
On webAdmin it validates the template's OS<->TZ compatibility before applying and fails with an error in case it's not ok, so that the user will be able to manually fix it.

On web-ui the user can't fix the TZ since it's auto done for him but we should still be consistent. 
Before this fix, the VM creation failed on this, the same as on webAdmin. 
Now after this fix it creates it wrongly by changing the OS taken from the template. This is wrong.

sjd78:
> > Do you mean that we should change the template's OS somehow in the process of creating a VM?
> 
> No, we shouldn't change the template, only the VM based on that template. The VM's OS will be the same as the template's OS and then we should check the template's TZ. If it fits then use that. If it doesn't fit then change to the correct default TZ or leave it as is and it will fail on backend.
> 
> > I would expect that there's nothing wrong with any chosen template and that the template's OS is in harmony with the template's TZ. I thought that the issue I am trying to fix here is compatibiliy of template's TZ with the chosen VM OS, not compatiblity of the template's TZ with the template's OS.
> 
> You are right and it's a specific case. But this fix should handle all cases.
> On webAdmin it validates the template's OS<->TZ compatibility before applying and fails with an error in case it's not ok, so that the user will be able to manually fix it.
> 
> On web-ui the user can't fix the TZ since it's auto done for him but we should still be consistent.
> Before this fix, the VM creation failed on this, the same as on webAdmin.
> Now after this fix it creates it wrongly by changing the OS taken from the template. This is wrong.

If I'm following things correctly, in summary, changing the template changes the OS and the TZ.  OS and TZ are only tied together because Linux and Window have different formats for TZ.  So yeah, when progressing off the Basic settings step, the correct TZ format needs to be applied based on the OS, even if the TZ is wrong on the Template.

Also, makes sense the only way you could have the Template be OS Windows and a TZ Linux is if the template is super old or if you tweak the DB directly.  I would expect that editing a template in webadmin will not allow you to save it if the OS and TZ are not in harmony.

hstastna:
@sgratch @sjd78 And what about the actual fix for editing a VM? The TZ is set according to the template there. Is this correct?

hstastna:
@sgratch @sdj So I've made a decision and appropriate change - the timezone will be set corresponding to the template's OS to prevent VM creation fail when choosing template with an invalid timezone. So the user will be able to create a VM with a proper TZ. But I think it could be very useful for the user to display some warning about invalid TZ of the chosen template and about the appropriate TZ change, before creating a VM. And if they agreed, then they could create a VM immediately, without any fail. If not, then they could change the VM OS or template, before creating a VM. What do you think?

sgratch:
> @sgratch @sdj So I've made a decision and appropriate change - the timezone will be set corresponding to the template's OS to prevent VM creation fail when choosing template with an invalid timezone. So the user will be able to create a VM with a proper TZ. But I think it could be very useful for the user to display some warning about invalid TZ of the chosen template and about the appropriate TZ change, before creating a VM. And if they agreed, then they could create a VM immediately, without any fail. If not, then they could change the VM OS or template, before creating a VM. What do you think?

I'm not sure it worth the change since it's an extreme case, so as long as we won't create an invalid VM as a result then it's ok to change the tz without letting the user know. Please note that the user can see the created VM timezone in the  "review" report step, so it's not hidden from him.

sdj:
@hstastna @sgratch Please correct your @-mentions from sdj to sjd78, I am not this user and am receiving your misdirected notifications. Feel free to remove this comment after correcting (or I can).

sgratch:
> @hstastna @sgratch Please correct your @-mentions from sdj to sjd78, I am not this user and am receiving your misdirected notifications. Feel free to remove this comment after correcting (or I can).

Done, sorry for the mistake.

hstastna:
@sgratch So then I think the PR is ready for another review, the specific case should be fixed by my last commit and user should see the created VM timezone in the "review" report step :)

hstastna:
@sgratch Is there anything I could improve regarding this PR? Thanks! :)

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -530,5 +583,6 @@ export default connect(
     storageDomains: state.storageDomains,
     maxNumOfVmCpus: state.config.get('maxNumOfVmCpus', 384),
     maxMemorySizeInMiB: 4194304, // TODO: 4TiB, no config option pulled as of 2019-Mar-22
+    config: state.config,
```
sjd78, line 586: Follow the same pattern as for `maxNumOfVmCpus` and pull the value out of `config` instead of pushing the entire config object down.

The more we can minimize redux/immutablejs code in components themselves, the easier it is to refactor the store in future.

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -250,6 +294,14 @@ class BasicSettings extends React.Component {
     }
   }
 
+  getTemplate () {
```
sjd78, line 297: I'm not sure this function is really needed since you're only calling it from `checkTimeZone()`.  Could just be inlined (like it is in `validateForm()` line 251) and be ok.

sjd78:
@hstastna - You need to rebase this PR after we did a few merges today.

hstastna:
I've found a bug in my changes, so I am working on fixing it. I will update the PR asap. Thanks.

hstastna:
> The default value changes look good with a few more minor tweaks.
> 
> How about also handling the TZ drop down values? You're already 75% there...
> 
> https://github.com/oVirt/ovirt-web-ui/blob/2744ed666a7e5cc3cfc8d3298ef32792668ebd39/src/components/CreateVmWizard/steps/BasicSettings.js#L477-L485

Hi @sjd78, what do you mean by handling the TZ drop down values? There isn't any drop down in the Create VM Wizard, regarding the TZ, even when I enable Cloud-init/Sysprep. Thanks! :)

sjd78:
> > The default value changes look good with a few more minor tweaks.
> > How about also handling the TZ drop down values? You're already 75% there...
> > https://github.com/oVirt/ovirt-web-ui/blob/2744ed666a7e5cc3cfc8d3298ef32792668ebd39/src/components/CreateVmWizard/steps/BasicSettings.js#L477-L485
> 
> Hi @sjd78, what do you mean by handling the TZ drop down values? There isn't any drop down in the Create VM Wizard, regarding the TZ, even when I enable Cloud-init/Sysprep. Thanks! :)

There is a sysprep field that is a text input box that should be converted to a dropdown:
screenshot-localhost_3000-2020 07

An empty field/drop down is allowed, it'll just pickup TZ from the template.

You can probably just pick up this code from VmDetails:
https://github.com/oVirt/ovirt-web-ui/blob/4ab783a8549f2e12fbf85a7a7991bb5b3c1f239d/src/components/VmDetails/cards/DetailsCard/CloudInit/SysprepForm.js#L39-L49

which looks like:
screenshot-localhost_3000-2020 07 (1)

hstastna:
@sjd78 Thank you for the perfect explanation! And I would add the TZ dropdown box in sys-prep in a new PR.

pnovotny:
LGTM

For VMs created from Blank template or ISO:
- for Windows OS types: TZ is set to GMT Standard Time
- for Linux & other OS types: TZ is set to Etc/GMT

For VM created from a Windows or Linux-based template, the template OS TZ is used. 
However when using for example a Windows-based template and changing VM OS to Linux, the VM TZ is then set to the Linux OS default TZ (Etc/GMT).

Issue #1215, opened 2020-05-19
Create VM Wizard - setting the disk size by using arrows doesn't work properly


sgratch:
On the "Create Virtual Machine" wizard -> "Storage" step, when adding/editing a disk and setting disk size by using arrows then after applying the size is set back to previous value instead of using the new value:

![image](https://user-images.githubusercontent.com/18169498/82303841-29136c80-99c4-11ea-8be6-3708746cf55c.png)

Note that in case of entering the size value then it works OK.

PR #1221, opened 2020-05-27
Create VM, Storage - add support for arrow keys in edit disk size field

bamsalem:
Hi there,
I was investigate the issue #1215  and I found why the value is not changed when using arrows, the described behavior refers only to FF (in chrome it worked before the changes and it will keep work after them).

In the disk size field we are using `onBlur` event to update the state and the arrows don't focus the field (in chrome it focus the field) so when the user change another field or confirm the changes and save the row the state is not updated because no event fired.

So why not using `onChange` event? Using `onChange` event instead `onBlur` is a good solution in case of using the arrows , but it cause another problem in case the user setting the field by typing, for any single character typed the page will re-render and the field will lose focus so the user need to refocus the field for any single character typed, and it's annoying.

So the solution I found is to update the state in `onChange` event only if the disk size field is not focused (using arrows), in case it's focused (using typing) the state will not update on `onChange` event but will update when the `onBlur` event handler will be triggered.

It a little bit complex solution but it works great and gives the requested behavior.

Fixes: #1215

### DIFF
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -18,6 +18,7 @@ import {
   DropdownKebab,
   EmptyState,
   FieldLevelHelp,
+  FormGroup,
   FormControl,
   MenuItem,
   Table,
@@ -239,15 +240,24 @@ class Storage extends React.Component {
         editView: (value, { rowData }) => {
           const row = this.state.editing[rowData.id]
           const sizeGiB = row.size / (1024 ** 3)
+          const { invalidSizeValue } = row
 
           return <div className={style['disk-size-edit']}>
-            <FormControl
-              id={`${idPrefix}-${value}-size-edit`}
-              type='number'
-              defaultValue={sizeGiB}
-              className={style['disk-size-form-control-edit']}
-              onBlur={e => this.handleCellChange(rowData, 'size', e.target.value)}
-            />
+            <FormGroup
+              validationState={invalidSizeValue && 'error'}
+              className={style['form-group-edit']}
+            >
+              <FormControl
+                id={`${idPrefix}-${value}-size-edit`}
+                type='number'
+                min={this.props.minDiskSizeInGiB}
+                max={this.props.maxDiskSizeInGiB}
+                step={1}
+                value={sizeGiB}
+                className={style['disk-size-form-control-edit']}
+                onChange={e => this.handleCellChange(rowData, 'size', e.target.value)}
+              />
+            </FormGroup>
             <span className={style['disk-size-edit-label']}>GiB</span>
             <FieldLevelHelp
               inline
@@ -437,9 +447,17 @@ class Storage extends React.Component {
 
   onCreateDisk () {
     const newId = generateUnique('NEW_')
+    const {
+      minDiskSizeInGiB: diskInitialSizeInGib,
+      storageDomains,
+      dataCenterId,
+      vmName,
+      disks,
+      optimizedFor,
+    } = this.props
 
     // If only 1 storage domain is available, select it automatically
-    const storageDomainList = createStorageDomainList(this.props.storageDomains, this.props.dataCenterId)
+    const storageDomainList = createStorageDomainList(storageDomains, dataCenterId)
     const storageDomainId = storageDomainList.length === 1 ? storageDomainList[0].id : '_'
 
     // Setup a new disk in the editing hash
@@ -449,7 +467,7 @@ class Storage extends React.Component {
         ...state.editing,
         [newId]: {
           id: newId,
-          name: suggestDiskName(this.props.vmName, this.props.disks),
+          name: suggestDiskName(vmName, disks),
 
           diskId: '_',
           storageDomainId,
@@ -457,12 +475,13 @@ class Storage extends React.Component {
           bootable: false,
           iface: 'virtio_scsi',
           type: 'image',
-          diskType: getDefaultDiskType(this.props.optimizedFor),
+          diskType: getDefaultDiskType(optimizedFor),
 
-          size: 0,
+          size: (diskInitialSizeInGib * 1024 ** 3),
         },
       },
     }))
+    this.props.onUpdate({ valid: false })
   }
 
   onEditDisk (rowData) {
@@ -472,6 +491,7 @@ class Storage extends React.Component {
         [rowData.id]: rowData,
       },
     }))
+    this.props.onUpdate({ valid: false })
   }
 
   onDeleteRow (rowData) {
@@ -479,13 +499,19 @@ class Storage extends React.Component {
   }
 
   handleCellChange (rowData, field, value) {
+    const editingRow = this.state.editing[rowData.id]
     if (field === 'size') {
-      if (isNumber(value) && value > 0 && value <= this.props.maxDiskSizeInGiB) {
-        value = +value * (1024 ** 3) // GiB to B
+      if (!isNumber(value)) return
+
+      const { minDiskSizeInGiB, maxDiskSizeInGiB } = this.props
+      if (value >= minDiskSizeInGiB && value <= maxDiskSizeInGiB) {
+        delete editingRow.invalidSizeValue
+      } else {
+        editingRow.invalidSizeValue = true
       }
+      value = +value * (1024 ** 3) // GiB to B
     }
 
-    const editingRow = this.state.editing[rowData.id]
     if (editingRow) {
       editingRow[field] = value
       this.setState(state => ({
@@ -501,6 +527,8 @@ class Storage extends React.Component {
   handleRowConfirmChange (rowData) {
     const actionCreate = !!this.state.creating && this.state.creating === rowData.id
     const editedRow = this.state.editing[rowData.id]
+
+    if (editedRow.invalidSizeValue) return
 
     // if the edited disk is set bootable, make sure to remove bootable from the other disks
     if (editedRow.bootable) {
@@ -529,6 +557,7 @@ class Storage extends React.Component {
         editing,
       }
     })
+    this.props.onUpdate({ valid: true })
   }
 
   // Create props for each row that will be passed to the row component (TableInlineEditRow)
@@ -647,7 +676,7 @@ Storage.propTypes = {
   cluster: PropTypes.object.isRequired,
   storageDomains: PropTypes.object.isRequired,
   maxDiskSizeInGiB: PropTypes.number.isRequired,
-
+  minDiskSizeInGiB: PropTypes.number.isRequired,
   onUpdate: PropTypes.func.isRequired,
 }
 
@@ -656,5 +685,6 @@ export default connect(
     cluster: state.clusters.get(clusterId),
     storageDomains: state.storageDomains,
     maxDiskSizeInGiB: 4096, // TODO: 4TiB, no config option pulled as of 2019-Mar-22
+    minDiskSizeInGiB: 1,
   })
 )(Storage)
--- a/src/components/CreateVmWizard/steps/style.css
+++ b/src/components/CreateVmWizard/steps/style.css
@@ -112,7 +112,6 @@
   align-items: center;
 }
 .disk-size-form-control-edit {
-  width: 0;
   flex-grow: 1;
 }
 
@@ -154,6 +153,10 @@ table tbody tr td.kebab-menu-cell {
 .review-content {
   margin: 0 auto;
   width: 80%;
+}
+
+.form-group-edit{
+  margin-bottom: 0;
 }
 
 .review-progress {
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -464,7 +464,11 @@ class CreateVmWizard extends React.Component {
           <Button bsStyle='default' onClick={this.showCloseWizardDialog}>
             { msg.createVmWizardButtonCancel() }
           </Button>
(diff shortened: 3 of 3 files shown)


ovirt-infra:

sgratch:
ci add to whitelist

bamsalem:
Oh oops,
my bad, it's been removed.

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -201,7 +201,17 @@ class Storage extends React.Component {
...
               defaultValue={sizeGiB}
-              onBlur={e => this.handleCellChange(rowData, 'size', e.target.value)}
+              onFocus={() => { this.isDiskSizeFocused = true }}
+              onBlur={e => {
+                this.isDiskSizeFocused = false
+                this.handleCellChange(rowData, 'size', e.target.value)
+              }}
+              onChange={e => {
+                if (!this.isDiskSizeFocused) {
+                  this.handleCellChange(rowData, 'size', e.target.value)
```
sgratch, line 211: Please note that the down arrow is not disabled when 0 value is set (and after clicking the -1 value is set).
In general, negative values are not allowed to disk size either by clicking arrows or by entering the text.

bamsalem: I wasn't focused in the field validation, I assumed the  validation is all right, I'll be more critical next time
Thanks

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -360,6 +370,7 @@ class Storage extends React.Component {
 
   onCreateDisk () {
     const newId = generateUnique('NEW_')
+    const diskInitialSize = 1
```
sjd78, line 373: This should be a prop to the component and probably driven by a constant in `CreateNewVm` where the `Storage` component is used.

sjd78:
> * Max value for disk size field is currently 4096GB (4 TiB). In case you enter a bigger value then it jumped back to previous value. Why is that @sjd78? On webadmin and on web-ui edit vm->create a new disk - there is no such limitation on disk size.

I don't remember the reason, but it would probably be either a "why does an end-user need >4TiB disk size" or similar. I'd have to look back at notes and docs to see if I can find that.

bamsalem:
@sgratch 
> * Please check to see how we handle the disk size field for vm portal -> exisiting VM -> "Create new Disk". No negative value can be entered either by setting the value or by using the arrows.


I found the difference and I fixed that, Thanks


>   On current flow, setting a negative value causes too much issues like when editing disk's size by entering a positive value, clicking outside the field and then setting a negative one then the original/last saved disk size is lost which is wrong etc. instead of adding tooltips then it worth behaving like on edit an existing vm.

I can't understand what you meant because when user type invalid (negative or higher than max value) it keeps the last edited state and the field value back to last valid value that entered in this edit... so I'm not sure what you mean here

> * In addition, when editing disk's size by entering the value (and not using the arrows) then after clicking the "V" button for saving, nothing is done. The user is required to to click outside the field focus and then saving or clicking the "V" button twice for saving.

I can't reproduce it, it works for me on FF...

> * Max value for disk size field is currently 4096GB (4 TiB). In case you enter a bigger value then it jumped back to previous value. Why is that @sjd78? On webadmin and on web-ui edit vm->create a new disk - there is no such limitation on disk size.

I used the exists value, is it need to change?

hstastna:

sjd78:
> 1. on Chrome arrows are not visible until you move pointer over them (see gif below)

That is normal behavior for the browser `input type="number"` tag being used.  Not much can be done there.

> 2. on FF arrows  have no styles for on focus/on click states

Another "feature" of the browser field behavior.  Not much can be done there.

> 3. fractional values are incorrectly rounded (0.123 GiB becomes 125.952 MiB - see below)

Storage of the value is in bytes, the display is in , the value gets converted around a bunch, and use the 1024 multiplier:
```
0.123 GiB * (1024 ** 3) = 132,070,244.352 Bytes
132,070,244.352 / (1024 ** 2) = 125.952 MiB
```
also
```
0.123 GiB * 1024 = 125.953 MiB
```
The calcs are correct, but decimal sizes shouldn't be allowed anyway.

>  4. factional values lower then min disk size are accepted (0.123 GiB)
> 
Yep!  No fractional values should be allowed to be saved!

>
> 5. you can only reach first input in the row by Tab key

Yep - that's partially due to the way the table uses `onBlur` to update values.  Also, that would be easier solved in another issue as it applies in both Storage and Network tabs.

bamsalem:
> Ideally, I think that right now, having the cell change value but then add validation to `handleRowConfirmChange` will close the major gaps in input.
> 
> Doing more probably requires reworking the way editing is handled on the table itself.

Hi @sjd78  I'm not sure what do you meaning...
if you are meaning to the if statement so it's there to avoid from the size field get unfocused on every change when the user change the field using keyboard.
I mentioned it above in my first comment :)

bamsalem:
> Also I am able to reproduce the issue Sharon has already mentioned in the comment. Try these steps:
> 
> 1. Create some Disk in the _Storage_ step, save it
> 2. Click on "..." and then on _Edit_ to be able to edit the Disk
> 3. **Write** there some value (any value or letter, it doesn't matter) for the Disk _Size_, don't use the arrow keys
> 4. Click on the blue heavy_check_mark button to save it
>    => nothing happens, the Disk is not saved, you have to click on the button once again to save the Disk. This is an extra clicking. This does not happen when just creating a new Disk, only **when editing** an existing one.

Hi @hstastna @sgratch regarding to Sharon issues:
In this dialog we are using `onBlur` event, if we will use `onChange` event the page will re-render and the field will lose focus and in that case the user should focus it for every character, and it's annoying I'm not sure why it happened but I found a lot of issues regarding that.
so the user can type negative values but when he tries to submit it, the value verified before saving or the border turned red:
image

WDYT?

hstastna:
@bamsalem I am not sure about that. For consistency reasons I would recommend to implement the same behavior as it already is for editing Disks of an existing VM: I cannot write there any negative value or a letter in the field and I don't see any issues with losing focus or submitting/saving the info:
disk

bamsalem:
> onBlur is the only reasonable way to handle the focus issue for inline editing of the table with the current library.  I spent A LOT of time on this when building create vm initially.

> PR 1221 is not going to result in "perfect" functioning of the size input field.  As long as editing it ultimately doesn't allow fractional, non-number, or negative numbers to be saved to the edit, it doesn't matter if the field is verified while typing in, or making it work like the disk input dialog.

> The sad truth is that, with the PF3 table component, input fields and the wizard component, the input field will not work as it does on the disk edit modal on VM details.  The only real way to make editing "really nice" would be to do record editing outside of the table itself (edit dialog from the wizard dialog?  == nope).  I'm not even sure if PF4 really even recommends inline table editing as a pattern.

hstastna:
> ... so the user can type negative values but when he tries to submit it, the value verified before saving or the border turned red

@bamsalem But why cannot we to verify it and save it at once? Something like "if the value is valid, then also save it". User wouldn't need to click on "Save" twice. Maybe it's a good reason for that and I'm still just missing it.

Anyway I think it should be doable. And if not, then I think that we should solve it by completely other way. Maybe to have some another button, something like "Verify" there. And after the user verified it, they could save it. Because it is confusing to click on "Save" button and to see that nothing happens. It feels buggy. And if something is confusing for us.. 

Anyway, in the some other project I was working on in the past, we solved this situation with a presence of "Verify" button and after successful verifying, the "Save" button became enabled. But this is maybe for a longer discussion also with the UX designers.. :angel:  :)

src/components/CreateVmWizard/steps/style.css
```diff
@@ -111,6 +111,11 @@
   display: flex;
   align-items: center;
 }
+
+.form-group-edit{
```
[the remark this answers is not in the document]
bamsalem: I changed it in #1219 PR

bamsalem:
Guys, @sjd78 @hstastna @rszwajko @sgratch 
I added one more commit to handle typing of non-numeric characters in size field, please let me know WDYT.
I found that if the user type any non-numeric character the VM Portal crashes, so I handled that in this [commit](https://github.com/oVirt/ovirt-web-ui/pull/1221/commits/d25cd0504c860e6117e910bc59f464d0a9b2e10c).

hstastna:
@bamsalem The solution for dealing with the letters in the _Disk Size_ field works for me but I am not sure about the expected behavior. Actually it "jumps" to the minimal disk size value when typing a letter there so the problem there is "fixed" but it feels little bit confusing for me. As a user, I have no idea what was going on, it's not clear to me. I would expect to see the red frame - the same as with a wrong invalid number typed in the field + some info that I cannot put any letters there, only the numbers. @yfrimanm @sjd78 @rszwajko @sgratch WDYT? Thanks! :)

yfrimanm:
Hi @hstastna +1. 
I would also recommend keeping the consistency of what you already have in cases like this, as with a wrong invalid number typed in the field. So if what you have is adding a red frame +  a notification below that input field telling the users they cannot put any letters there, but only numbers - then I would stick to that behavior.

cc @mcarleto@redhat.com

bamsalem:
the current PR should deal with using the arrows to set the size value.
but I add the size validation because it was a small fix, but I think dealing with non numeric characters get bigger enough to be dealt as another issue (we need to change proptypes and etc...) WDYT guys?

hstastna:
I think you can fix it within this issue. It shouldn't be that big :)

bamsalem:
@hstastna the changes requested to create the described behavior is to pass a string in a prop that should be a number so it may create a problem, so we need to deal with all the cases that the user typed non-numeric character and numeric character, what do you think about set the value to 0 (invalid value) and then all the validation will work like in any other case, because technically it wrong to do it but if the UX Requires it דם I'll do it, WDYT?

hstastna:
@bamsalem I am not sure.. What about setting `editingRow.invalidSizeValue = true` also in [this case](https://github.com/oVirt/ovirt-web-ui/pull/1221/commits/ae96175409d241e470fbe61c11305e9b5b6ea73d#diff-82b3e1a31b05cef34fe3e9cd5a4f1720R446)? We would easily achieve the same behavior as it already is there when deleting a value from the field.

IMHO, the best would be to see something like this:
aa
The user would see what they've typed in the field and the tooltip message would clearly inform the user about what to do + red frame indicating about the invalid input.

This is what I can see when I type some letters in the field but when I lose focus or click anywhere else, the value in the field changes to 0. And this leads me to ask if this is really the behavior we want. Why does the string in the field changes to 0, when I delete the number in the field? I know it has some reasons but does it really have to be like that? Isn't there any other option to achieve the expected behavior? Maybe I am missing something but I would better to see in the field exactly what I typed in there, and not changing it to 0 in the case of invalid input. @yfrimanm what do you think about it?

Also it makes more sense to me to provide [this calculation](https://github.com/oVirt/ovirt-web-ui/pull/1221/commits/ae96175409d241e470fbe61c11305e9b5b6ea73d#diff-82b3e1a31b05cef34fe3e9cd5a4f1720R448) only when the input is valid and if it is a number.

matthewcarleton:
> @bamsalem I am not sure.. What about setting `editingRow.invalidSizeValue = true` also in [this case](https://github.com/oVirt/ovirt-web-ui/pull/1221/commits/ae96175409d241e470fbe61c11305e9b5b6ea73d#diff-82b3e1a31b05cef34fe3e9cd5a4f1720R446)? We would easily achieve the same behavior as it is already there when deleting a value from the field.
> 
> IMHO, the best would be to see something like this:
> aa
> The user would see what they've typed in the field and the tooltip message would clearly inform the user about what to do + red frame indicating about the invalid input.
> 
> This is what I can see when I type some letters in the field but when I lose focus or click anywhere else, the value in the field changes to 0. And this leads me to ask if this is really the behavior we want. Why does the string in the field changes to 0, when I delete the number in the field? I know it has some reasons but does it really have to be like that? Isn't there any other option to achieve the expected behavior? Maybe I am missing something but I would better to see in the field exactly what I typed in there, and not changing it to 0 in the case of invalid input. @yfrimanm what do you think about it?
> 
> Also it makes more sense to me to provide [this calculation](https://github.com/oVirt/ovirt-web-ui/pull/1221/commits/ae96175409d241e470fbe61c11305e9b5b6ea73d#diff-82b3e1a31b05cef34fe3e9cd5a4f1720R448) only when the input is valid and if it is a number.

I don't think adding a valid value of `0` is useful. It should remain what they've typed and remain invalid and require them to correct it before they can proceed.

hstastna (changes requested):
Now it's not possible for me to use arrow keys at all, also not possible to type in some other value for the Disk size, when **editing** some existing Disk. It's like frozen. Tried in both Firefox and Chromium.

In addition, it's not possible to edit the bootable flag, also when editing an existing Disk (blank template was chosen in the 1st step) - however this works fine on master branch. Or better said, it's possible but user does not see any change when editing (except editing disk name, storage domain and disk type). The changes are obvious only after saving the Disk. Same for both bootable flag and Disk size. To add another detail about this, it reacts only to first one of my clicks. So for example, if I want to increase the Disk size from 4 to 8 GiB, I click four times on the arrow key but after saving the disk, the disk size increased from 4 to only 5 GiB.

Example: just started editing some Disk:
start1
After 7 clicks on the upper arrow key: the disk size hasn't visually changed at all (same with bootable flag)
start2

hstastna (approved):
LGTM

rszwajko (approved):
Looks OK.

michalskrivanek:
looks ready

PR #1222, opened 2020-05-28
refactor Makefile to prevent unexpected symlinks

sjd78:
Changes:
  - Wrap install path names in quotes to avoid unexpected errors

  - Add `-n` param to `ln` to avoid unexpected symlinks if install is run multiple times (e.g. a developer installing to engine multiple times)

  - Only create target directories if they don't exist

  - Move the contents of "branding/ovirt/" up one level to be installed directly in "branding/"

Now installing a prod build of web-ui to a dev engine should be clean,follow the normal make command:
```
   make all install-data DESTDIR=$HOME/ovirt-engine
```

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -100,17 +100,29 @@ ovirt-web-ui:
 	cp -rpv packaging/* build/
 	mv build/static build/index.jsp build/ovirt-web-ui.config build/ovirt-web-ui.war/
 
+#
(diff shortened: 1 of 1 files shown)


sjd78:
> Looks good to me but I didn't test the fix (I was not sure how to do this)

Install VM Portal to your dev engine multiple times and make sure there aren't any symlinks there that were not there on the first install.

Issue #1224, opened 2020-06-02
Error in UPDATE_VMS reducer in scheduled refresh after manual refresh


sjd78:
VM Portal version number: master

oVirt version number: 4.4-master

Browser and version: Any

Steps to Reproduce:
1. Open VM Portal
2. While the card list is showing, click the manual refresh button
3. Wait (up to a minute) for the scheduled refresh to happen

Actual results: Error visible only in the console showing a failure in the UPDATE_VMS reducer.

Expected results: No errors on UI or in console.

![VM Portal - Google Chrome_037](https://user-images.githubusercontent.com/3985964/83543298-37847c80-a4ca-11ea-808c-e53da7ba09e7.png)

PR #1225, opened 2020-06-02
Update UPDATE_VMS reducer to handle `copySubResources` properly

sjd78:
Fixes: #1224 

Simplify UPDATE_VMS reducer handling of 'copySubResources':
  - The incoming vms are plain JS objects.  It is easier to handle copying of existing sub-resources from the current copy of the VM in redux to the incoming VM updates when they're both JS objects.

Minor refactor of VM fetch additional/follow items:
  - This makes the list of additional items to be followed when a VMis fetched slightly more explicit.

### DIFF
--- a/src/reducers/vms.js
+++ b/src/reducers/vms.js
@@ -28,13 +28,10 @@ import {
 import { actionReducer, removeMissingItems } from './utils'
 import { sortFields } from '_/utils'
 
-const EMPTY_MAP = Immutable.fromJS({})
-const EMPTY_ARRAY = Immutable.fromJS([])
-
 const initialState = Immutable.fromJS({
   vms: {},
   pools: {},
-  filters: EMPTY_MAP,
+  filters: {},
   sort: { ...sortFields[0], isAsc: true },
 
   page: 1,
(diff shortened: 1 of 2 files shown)


sjd78:
Note: after the fix, here is the screen shot with the UPDATE_VMS reducer succeeding:

VM Portal - Google Chrome_038

sjd78:
> LGTM, the error in UPDATE_VMS reducer is not present anymore. I also like the small refactoring of VM fetch additional items. It makes sense to me. I just wonder - why it was `true` [there](https://github.com/oVirt/ovirt-web-ui/pull/1225/commits/a76c0aca7935f2c8caf64bf9871cd6fee12e81dc#diff-3a773596665964819aa579d52d9feb94L80)? Thank you :)

Well the simple answer is that it is mirroring the default values in the transform:
https://github.com/oVirt/ovirt-web-ui/blob/f67e9a64f4a2cb1c706fb5fb086a97a4b8f3cae0/src/ovirtapi/transform.js#L177-L183

But really it shouldn't matter in practice because we always fetch the permits and then translate them here:
https://github.com/oVirt/ovirt-web-ui/blob/f67e9a64f4a2cb1c706fb5fb086a97a4b8f3cae0/src/sagas/index.js#L256-L266

sgratch:

PR #1226, opened 2020-06-02
Remove Components that are no longer used

sjd78:
After all of the various redesigns in the last ~2 years, we still have the old components hanging around.

These can all be safely removed since they aren't actually used anywhere.

Changes:
  - Removed components that are not used since before VM dashboard
  - Removed components that are not used since before CreateVmWizard
  - Refactored a few utils and components to different locations
  - Minor makefile tweaks
  - Minor config file tweaks


sjd78:
**NOTE**: Anticipated QE is simply a general regression test to ensure the base app still works as intended with the components removed.

Issue #1216, opened 2020-05-19
Create VM Wizard - disk size units on a new disk row are not displayed before saving


sgratch:
On the "Create Virtual Machine" wizard -> "Storage" step, when adding/editing a disk by clicking the checkmark box, the disk size units are not displayed. How the user will know which value to set without knowing the units. 

It worth changing the column title to include units, and use GiB instead of TiB, the same as done for create new disk on VM's card. 
Please consider adding the tooltip for size extending as well.

The create new disk on VM's card:
![image](https://user-images.githubusercontent.com/18169498/82305289-1ef26d80-99c6-11ea-8827-76e4d62521a7.png)

bamsalem:
Hi @sgratch I want to draw your attention, it seems like this issue related to  issue #1214  and the units covered by the "Storage Domain" field.
do you still want the changes requested above?
if the units will be shown after the changes which fix issue #1214 will be merged we can consider just add `FieldLevelHelp`.
I think changing the column title is not necessary after fixing issue #1214.
what do you think?

sgratch:
@bamsalem @lwrigh I agree. So the question is if to add a tooltip and where to locate it (column title or field itself).

lwrigh:
I would suggest if the units are going to be changed and moved outside of the label or the dropdown field that they are consistent across the board. Both the unit of measurement and the placement of the unit of measurement in the field.

sgratch:
@lwrigh Not sure I understand, do you suggest moving both units ("GiB") and the tooltip (as mentioned on the description above) to the "Size" column title or leave it as is in the field?

Current appearance (without the tooltip):
![image](https://user-images.githubusercontent.com/18169498/83426542-fba0d700-a437-11ea-9736-3cbe2be2f7b4.png)

lwrigh:
I would keep it at the field level and not hide it in the info tip if I'm understanding the question correctly.

sgratch:
> I would keep it at the field level and not hide it in the info tip if I'm understanding the question correctly.

@lwrigh My question was where to add the tooltip for the Size field of: "After you create a disk, you may only extend his size when you make edits". I guess, based on your previous answer, that it should be kept as a tooltip at the field level as well

PR #1229, opened 2020-06-08
Create VM Wizard - added tooltip message to Disk size field on edit mode

bamsalem:
Fixes: #1216  

The fix related to issue #1214  and PR #1217 , so the changes from PR #1217 can be found in this PR in commit 6bbd3e0.

In the Create VM Wizard > storage step I added tooltip to disk size field.
I used the same message as used in "create new disk" on VM's card.
after the tooltip addition the `FormControl` was too small, so I increased the column width from 10% to 15 % and I removed the last column width value which was 40px, so now the width value is the default value (`auto`).

 before (10% with the new tooltip):
- view mode:
image
- edit mode:
image

after (15%):
- view mode:
image
- edit mode:
image


isaranova:
ci build please

isaranova:
LGTM

sjd78:
@bamsalem - This PR needs to be rebased and the merge conflict resolved.  The conflict is in the width of the action column.  You will probably want to retain the width, but only testing it will verify.

bamsalem:
@sjd78 , resolved.

Issue #1199, opened 2020-05-05
Create VM Wizard - Storage step - enable setting/unsetting disks as bootable


sgratch:
Disks should be set/unset as Bootable in Storage step of Create VM wizard.
Also, the first added disk should be set as Bootable by default.

We added this feature to "edit VM" flow #1103 and now we should support that for "create VM" as well.
Note that in case this flag is taken of a template's disk then we should updated disks as appropriate

hstastna:
@sgratch @lwrigh @michalskrivanek There's a bunch of UX things I would like to talk about regarding enabling setting/unsetting disks as bootable feature, not only in the Storage step, but also when adding/editing Disks for an existing VM:

**1.** I don't know how we'd like to display the bootable option in the Storage step when adding/editing a Disk, so a couple of possibilities came to my mind. For example:
a) displaying the checkbox together with the 'Bootable' label **under the list/table of Disks** (similarly as it is already when editing an existing VM's Disks in our web UI), 
![bootable_under2](https://user-images.githubusercontent.com/13417815/83773063-5dee1780-a684-11ea-8656-9a33f1ae5861.png)
  - Later, we could add more options there. They would disappear after user was satisfied and saved the change.

b) or **next to the table of Disks** - similarly as it is in webadmin (where Bootable option is next to the other options):
![edit_disk_webadmin](https://user-images.githubusercontent.com/13417815/83775513-4f552f80-a687-11ea-9b58-2311a17cf68c.png)
c) displaying the checkbox **in the new column** for a new/edited Disk (for Disks which are just displayed, there's already "bootable" label displayed next to the Disk Name):
![boot_column](https://user-images.githubusercontent.com/13417815/83776769-dce54f00-a688-11ea-8516-95ec69f84b78.png)

  - I would go for the last (c) option as I think it's the best one.

---

**2.** I think the last column for Disks takes too much space and in some cases, there have to be 2 lines in the row because of not enough space for example for "select a Storage Domain":
![storage_domain](https://user-images.githubusercontent.com/13417815/83778032-6f3a2280-a68a-11ea-9714-8f11a561cff2.png)
[...]

PR #1232, opened 2020-06-12
Create VM, Storage - allow (un)setting Disks as bootable

hstastna:
**Fixes**: https://github.com/oVirt/ovirt-web-ui/issues/1199

**In this PR we:**
- **display the column (including _Bootable_ header)** only when adding/editing some Disk ([option 1c](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-638970474)) 
- display the checkbox in the appropriate disk's row and the _Bootable_ column  to allow editing the bootable flag when it is allowed (absence of a template based bootable disk)
- ~~make the 'Bootable' checkbox disabled when there's another Disk already set as bootable~~
- add the icon - tooltip in the Bootable column and appropriate row to inform user about:
  - why they cannot set any disk as bootable if they cannot (presence of a bootable template disk)
  - that only one disk at a time can be set as bootable
  - ~~that they have to 'unbootable' another (and not a template based) Disk before they can set some Disk as bootable~~
  - that the bootable flag will be removed from an original bootable disk and placed on the actual disk if possible
- edit the width of the columns of the table in the Storage step little bit to prevent unnecessary displaying the text for the Storage Domain cell in the two lines, achieve a better look of the table (see the point 2 from [this comment](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-638970474))

---

_In this PR we don't:_
- deal with the point 3 and 4 from [here](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-638970474), the design mentioned in the point 3 seems to be [intentional](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-640279790) (however I don't consider it as the best), the point 4 will be solved by another PR

_TODO:_
- [x] consistency with moving the bootable flag between Disks of an existing VM (see [the comment](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-642017889))
- [x] situation with none of the bootable Disks (more [here](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-643002269)) - see also [this](https://github.com/oVirt/ovirt-web-ui/issues/1199#issuecomment-656917491)

---

**Before:**
Viewing the Disks:
before2
Editing a Disk:
before1

**After:**
Viewing the Disks (no _Bootable_ column):
after_view1
~~Creating new Disk while there's already some template Disk which is bootable:~~ see the [new screenshots](https://github.com/oVirt/ovirt-web-ui/pull/1232#issuecomment-657836558)
after_edit1
~~Inform the user that they cannot set any Disk as bootable:~~
after_edit2
~~Editing some Disk:~~
after_edit3
Enabled checkbox and inform the user (same for editing bootable Disk which is not template based):
after_edit4
~~Inform the user to remove bootable flag from other (non template based) Disk first:~~
after_edit5


src/components/CreateVmWizard/steps/style.css
```diff
@@ -120,6 +120,12 @@
   margin: 0 3px;
 }
 
+.disk-bootable-edit {
+  display: flex;
+  margin: auto;
+  width: 50%;
+}
+
```
sjd78, line 128: Add another style for the checkbox to remove the ugly top/bottom margins and keep the row on a single line...

```css
.disk-bootable-edit :global(div.checkbox) {
  margin: 0;
  display: flex;
  align-items: center;
}
```

Note: the `:global()` bit is an instruction to the webpack CSS loader module to keep the rule selector as-is.  Otherwise the CSS loader generates unique names for things

sjd78:
I do like the bootable column with the checkbox.  Good trade-off between all of the other options.

hstastna:
**Updated screenshots:**

Creating/editing a disk while there's already some template based disk which is bootable:


Inform the user that they cannot set any disk as bootable now:


Inform the user to move the bootable flag between the disks (if applicable):

hstastna:
@sjd78 @sgratch @rszwajko @bamsalem This is ready for the review. Thanks! :sparkle:

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -124,8 +127,12 @@ class Storage extends React.Component {
 
       renderValue: (value, additionalData) => {
         const { column } = additionalData
+
+        // prepare suitable className to achieve hiding/displaying the cell if we are in a cell of 'Bootable' column and in viewing/editing mode
+        const bootableClassname = column.header.label === 'Bootable' && Object.entries(this.state.editing).length === 0 ? 'hidden' : ''
```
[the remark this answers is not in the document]
hstastna: Ahh, you are right. This definitely should be changed. Thank you for pointing out to this, Scott!

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -176,6 +183,47 @@ class Storage extends React.Component {
...
+            },
+          ],
+          props: {
+            style: {
+              width: '5%',
+            },
+          },
+        },
+        cell: {
+          formatters: [inlineEditFormatter],
```
rszwajko, line 204: can we customize cell formatters here similar as it was done for header formatters? This would keep local(for this column only) logic into leaking to global scope (for all columns)

hstastna: I've tried so I know we can but I'm not sure if it's a better option. I found some disadvantages with such an approach and ... Sometimes it's good to have everything (related to one thing, for example cell formatters) in one place: you know where you find all the related stuff and you can change something easily for all of the columns at once. But yeah, this has also some disadvantages as you've mentioned above.

bamsalem:
Hi Hilda except Scott's and Radek's review I found something that might be confusing to user's,
the bootable disk sorted before other disks, when the user changes the bootable disk it will resorted and it might be confusing...
it's not a bug it's just a small improvement.
we did'n notice that until this PR because we can't set the bootable option before.
you can see the changes I did in `VmDetailsPage` component: #1167 .

image
image

I would like to see something like that:
image

everything else LGTM. great work :)

hstastna:
I think that we definitely should make the UI consistent. So if we sort the Disks in the VM's details page in some way, we should definitely implement the same also in the Create VM Wizard. But I am not sure what's the expected behavior. Is there any clear UX decision of how we should sort the Disks? Thanks.

src/components/CreateVmWizard/steps/Storage.js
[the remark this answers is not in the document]
hstastna: I think I've finally fixed that ;)

src/components/CreateVmWizard/steps/Storage.js
hstastna: Done. And it also helped me with fixing a potential issue regarding checking column's `label` :)

hstastna:
@sjd78 @sgratch @rszwajko @yfrimanm
Regarding the Ben's comment above, I think that we definitely should make the UI consistent. So if we sort the Disks in the VM's details page in some way, we should definitely implement the same also in the Create VM Wizard. But I am not sure about the expected behavior. Is there any clear UX decision of how we should sort the Disks? Because I don't understand, why it was changed only in the VM's details page or why I can see [such comments](https://github.com/oVirt/ovirt-web-ui/pull/1232/files#diff-82b3e1a31b05cef34fe3e9cd5a4f1720R554) in the code. Thanks! :sparkle:

hstastna:
@sjd78 @sgratch @rszwajko @yfrimanm
Regarding the Ben's comment above, I think that we definitely should make the UI consistent. So if we sort the Disks in the VM's details page in some way, we should definitely implement the same also in the Create VM Wizard. But I am not sure about the expected behavior. Is there any clear UX decision of how we should sort the Disks? Because I don't understand, why it was changed only in the VM's details page or why I can see [such comments](https://github.com/oVirt/ovirt-web-ui/pull/1232/files#diff-82b3e1a31b05cef34fe3e9cd5a4f1720R554) in the code. Thanks! :sparkle:

rszwajko:
> Is there any clear UX decision of how we should sort the Disks? 

IMHO automatic sorting was OK when data was non-editable. When editing is allowed the list starts to change dynamically without user actions. This makes the UI look unreliable. I would rather drop this feature (or allow user triggered sorting).
However this is not the core of this PR and not a blocker to me.

hstastna:
@yfrimanm WDYT? :) Thanks!

matthewcarleton:
from a UX perspective I think following what is done in the details view should be carried over to the wizard. The Bootable label is clear enough and doesn't need to prioritized further. It also leads to some very odd behaviour when resorting happens so that should be avoided.

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -362,6 +402,33 @@ class Storage extends React.Component {
     ]
   }
 
+  // return boolean value to answer if we are editing a Disk or not
+  isEditingMode = () => Object.keys(this.state.editing).length > 0
```
sjd78, line 406: Please convert this to the less cool function-bound-in-constructor style to maintain consistency in the component.

hstastna:
Sorting Disks updated (bootable flag does not matter):
sort1
The same order of the disks in the _Review_ step:
sort2

isaranova:
Hello, I found some issues, though they might not be relevant to this PR.

- Default option when creating disk isn't Thin provision. 
- Changing between Thin provision and Preallocated options changes the size of the table slightly. 
- Setting the bootable flag resets the size of the disk to the previous value (either 0 or the value it had when the disk had been set as bootable); however, so does the change of the disk type. 
- The final overview does list the info without any space from the divider. 
 Screenshot_2020-08-19 VM Portal
- Creating the VM is successful but adding the disk isn't and the error message points out to API documentation which is very confusing.
Screenshot_2020-08-19 VM Portal(1)

matthewcarleton:
hey @isaranova can you provide some more on this. Hard to know exactly what you are referring to without more context. I'm not sure since this PR has been merged if we would want to continue this here or in a follow up PR. I'll leave that to you to decide.

isaranova:
I'll create a new issue for this then, thanks for pointing out the merge.

sjd78:
> Hello, I found some issues, though they might not be relevant to this PR.
> 
>     * Default option when creating disk isn't Thin provision.

Preallocated is the default for **server** type/"optimized for" VMs.  Thin provisioned is default for **desktop**.  The default in code is "desktop", but when a Template is selected, it overrides the type.  My Blank template has a default type of **server**.  So any template defined VM will default type to server.  The only way to stick with the app default of **desktop** is to use an ISO provision source or to manually select **desktop** when you pick a template.

Throw a bit of auto select Template provision source if the cluster has more than the blank template and it is easy to over look.

> 
>     * Changing between Thin provision and Preallocated options changes the size of the table slightly.

Meh. Very minor issue.

> 
>     * Setting the bootable flag resets the size of the disk to the previous value (either 0 or the value it had when the disk had been set as bootable); however, so does the change of the disk type.

That should get fixed up over in #1221.

> 
>     * The final overview does list the info without any space from the divider.

Meh, also a very minor issue.

> 
>     * Creating the VM is successful but adding the disk isn't and the error message points out to API documentation which is very confusing.

Fixed already in #1272

isaranova:
@sjd78 thanks a lot for detailed answer, no problems then!

LGTM

sgratch:
> > Hello, I found some issues, though they might not be relevant to this PR.
> > 
> > * Default option when creating disk isn't Thin provision.
> 
> Preallocated is the default for **server** type/"optimized for" VMs. Thin provisioned is default for **desktop**. The default in code is "desktop", but when a Template is selected, it overrides the type. My Blank template has a default type of **server**. So any template defined VM will default type to server. The only way to stick with the app default of **desktop** is to use an ISO provision source or to manually select **desktop** when you pick a template.

@isaranova @sjd78   This is not totally correct. The assumption of the disk type value (Preallocated/Thin) derived from the "optimized for" value (server/desktop) is correct only for disks "inherited" from the template, which we currently can't edit anyway via the create VM wizard.

For all other new disks, the current behaviour and default value are not correct (it should definitely be "Thin provision") and there is an issue for handling that: #1237 

In addition, it seems that it's not working correctly for existed template's disks as well:
If I create a VM based on a template with "optimized for" field set to Server then the VM's disk should be created as Clone/Independent and uses the **exact same** disk type as the disk's template. But currently the disk is always created by default as thin provision regardless.

Anyway, all of this should be handled on #1237

Issue #1230, opened 2020-06-08
Create VM Wizard - validate the disk's storage domain for the current user


sgratch:
On the "Create Virtual Machine" wizard -> "Storage" step, for both template's and new disks we need to make sure that the disk’s storage domain is a valid domain for the current user.  If it isn’t, force selection of another storage domain.

bamsalem:
I started to work on a solution for this issue,
I want to ask you what do you think about its look:
![image](https://user-images.githubusercontent.com/64131213/84258860-a32ba280-aae5-11ea-8d9d-698be059f22a.png)
In case that the storage domain is not valid for the current user the disk row will appear with an "UNKNOWN" storage domain, in my solution that case will be handled as shown in the screenshot above, the user won't be able to proceed to the next step till he choose storage domain from the dropdown list, and then the "Next" button will be enabled.
 what do you think?

PR #1233, opened 2020-06-15
Create VM, Storage - force user to choose a storage domain for template disks if necessary

bamsalem:
Fixes: #1230 

Hi I added validation to storage domain for template defined disks.
in case the template's disk should be located in storage domain which not allowed by the current user permissions the disks table in the Create Vm Wizard will be displayed with 'UNKNOWN' storage domain:
image

The user can create this vm with no validation or Error handling.
My new feature will prevent them from create the vm and force them to choose new storage domain according to their permissions, it will disable the next button as long as the user didn't change the storage domain:
Dropdown:
image
Dropdown options:
image
After choosing:
image
What do you think?

### DIFF
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -8,7 +8,7 @@ import { List } from 'immutable'
 import * as Actions from '_/actions'
 import { generateUnique } from '_/helpers'
 import { msg } from '_/intl'
-import { createTemplateList } from '_/components/utils'
+import { createStorageDomainList, createTemplateList } from '_/components/utils'
 
 import NavigationConfirmationModal from '../NavigationConfirmationModal'
 import BasicSettings from './steps/BasicSettings'
@@ -345,6 +345,7 @@ class CreateVmWizard extends React.Component {
         }
 
         if (resetDisks) {
+          const filteredStorageDomainList = createStorageDomainList(this.props.storageDomains, this.state.steps.basic.dataCenterId)
           draft.steps.storage = {
             updated: (draft.steps.storage.updated + 1),
             disks: !template
@@ -357,6 +358,7 @@ class CreateVmWizard extends React.Component {
                   diskId: disk.get('id'),
                   storageDomainId: disk.get('storageDomainId'),
                   canUserUseStorageDomain: this.props.storageDomains.getIn([ disk.get('storageDomainId'), 'canUserUseDomain' ], false),
+                  createdFromInvalidTemplate: !filteredStorageDomainList.find(sd => sd.id === disk.get('storageDomainId')),
 
                   bootable: disk.get('bootable'),
                   iface: disk.get('iface'),
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -105,6 +105,12 @@ class Storage extends React.Component {
     this.isBootableDiskTemplate = this.isBootableDiskTemplate.bind(this)
     this.isEditingMode = this.isEditingMode.bind(this)
 
+    const invalidDisk = props.disks.find(disk => (disk.isFromTemplate && !disk.canUserUseStorageDomain))
+
+    if (invalidDisk) {
+      props.onUpdate({ valid: false })
+    }
+
     this.state = {
       editing: {},
       creating: false,
@@ -286,11 +292,26 @@ class Storage extends React.Component {
           const {
             storageDomainId: id,
             storageDomain: sd,
+            createdFromInvalidTemplate,
+            isFromTemplate,
           } = rowData
 
-          // TODO: if the disk's storage domain is not available to the user, highlight it
-          // TODO: like a validation error and force editing the SD before allowing the user to move
-          // TODO: forward to the next wizard step
+          if ((isFromTemplate && !sd.isOk) || createdFromInvalidTemplate) {
+            const { storageDomains, dataCenterId } = props
+            const storageDomainList = createStorageDomainList(storageDomains, dataCenterId, true)
+            if (!sd.isOk) {
+              storageDomainList.unshift({ id: '_', value: `-- ${msg.createVmStorageSelectStorageDomain()} --` })
+            }
+            return <SelectBox
+              id={`${idPrefix}-${rowData.id}-storage-domain-edit`}
+              items={storageDomainList}
+              selected={sd.isOk ? rowData.storageDomainId : '_'}
+              validationState={!sd.isOk && 'error'}
+              onChange={value => {
+                this.props.onUpdate({ valid: true, update: { ...rowData, storageDomainId: value } })
+              }}
+            />
+          }
           return <React.Fragment>
             { id === '_'
               ? `-- ${msg.createVmStorageSelectStorageDomain()} --`
@@ -305,7 +326,7 @@ class Storage extends React.Component {
           const storageDomainList = createStorageDomainList(storageDomains, dataCenterId, true)
           const row = this.state.editing[rowData.id]
 
-          if (storageDomainList.length > 1 || row.storageDomain === '_') {
+          if (storageDomainList.length > 1 || row.storageDomainId === '_') {
             storageDomainList.unshift({ id: '_', value: `-- ${msg.createVmStorageSelectStorageDomain()} --` })
           }
 
@@ -315,6 +336,7 @@ class Storage extends React.Component {
               items={storageDomainList}
               selected={row.storageDomainId}
               onChange={value => this.handleCellChange(rowData, 'storageDomainId', value)}
+              validationState={row.storageDomainId === '_' && 'error'}
             />
           )
         },
@@ -529,7 +551,7 @@ class Storage extends React.Component {
     const actionCreate = !!this.state.creating && this.state.creating === rowData.id
     const editedRow = this.state.editing[rowData.id]
 
-    if (editedRow.invalidSizeValue) return
+    if (editedRow.invalidSizeValue || editedRow.storageDomainId === '_') return
 
     // if the edited disk is set bootable, make sure to remove bootable from the other disks
     if (editedRow.bootable) {
--- a/src/components/SelectBox.js
+++ b/src/components/SelectBox.js
@@ -27,6 +27,7 @@ class SelectBox extends React.Component {
       items: getItems(props),
     }
     this.handleChange = this.handleChange.bind(this)
+    this.getValidationClass = this.getValidationClass.bind(this)
   }
 
   componentWillReceiveProps (nextProps) {
@@ -44,15 +45,26 @@ class SelectBox extends React.Component {
     }
   }
 
+  getValidationClass () {
+    const { validationState } = this.props
+    switch (validationState) {
+      case 'error':
+        return style['selectBox-has-error']
+      default:
+        return ''
+    }
+  }
+
   render () {
     const { id } = this.props
 
     const selectedItem = this.state.items.find(item => item.id === this.state.selected)
+    const validationClass = this.getValidationClass()
 
     return (
       <div style={{ width: '100%' }} id={id}>
         <div className='dropdown'>
-          <button className={`btn btn-default dropdown-toggle ${style['dropdown-button']}`} type='button' data-toggle='dropdown' id={`${id}-button-toggle`}>
+          <button className={`btn btn-default dropdown-toggle ${style['dropdown-button']} ${validationClass}`} type='button' data-toggle='dropdown' id={`${id}-button-toggle`}>
             <span className={style['dropdown-button-text']} id={`${id}-button-text`} title={selectedItem && selectedItem.value}>
               {selectedItem ? selectedItem.value : NOBREAK_SPACE}
             </span>
@@ -83,6 +95,7 @@ SelectBox.propTypes = {
   /* eslint-enable react/no-unused-prop-types */
   onChange: PropTypes.func.isRequired, // (selectedId: string) => any
   id: PropTypes.string,
+  validationState: PropTypes.oneOf([ false, 'default', 'error' ]),
 }
 
 export default SelectBox
--- a/src/components/sharedStyle.css
+++ b/src/components/sharedStyle.css
@@ -84,3 +84,12 @@
 .text-align-left {
   text-align: left;
 }
+/*
+ * SelectBox validationState classes
+ * compatible with PF3 FormControl component
+ */
+
+.selectBox-has-error,.selectBox-has-error:hover,.selectBox-has-error:focus {
+    /*pf-red-100*/
+    border-color: #cc0000;
+}


ovirt-infra:

hstastna:

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -386,13 +398,16 @@ class CreateVmWizard extends React.Component {
     this.props.onCreate(basic, nics, disks, correlationId)
   }
 
-  wizardGoToStep (newStepIndex) {
+  wizardGoToStep (newStepIndex, moveBackward = false) {
```
rszwajko, line 401: instead of dedicated flag we can compare indices i.e. `moveBackward = nextStepIndex < activeStepIndex`

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -162,21 +163,27 @@ class CreateVmWizard extends React.Component {
...
+              id='create-vm-wizards-storage'
+              vmName={this.state.steps.basic.name}
+              clusterId={this.state.steps.basic.clusterId}
+              dataCenterId={dataCenterId}
+              optimizedFor={this.state.steps.basic.optimizedFor}
+              disks={this.state.steps.storage.disks}
+
+              onUpdate={({ valid = true, ...updatePayload }) => {
+                this.handleListOnUpdate('storage', 'disks', updatePayload)
+                this.wizardStepsMap.storage.preventExit = !valid
```
rszwajko, line 181: `valid` seems always `true` here

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -226,11 +226,24 @@ class Storage extends React.Component {
...
+            const storageDomainList = createStorageDomainList(storageDomains, dataCenterId, true)
+            if (!sd.isOk) {
+              storageDomainList.unshift({ id: '', value: `-- ${msg.createVmStorageSelectStorageDomain()} --` })
+            }
+            return <SelectBox
+              id={`${idPrefix}-${rowData.id}-storage-domain-edit`}
+              items={storageDomainList}
+              selected={sd.isOk ? rowData.storageDomainId : ''}
+              onChange={value => {
+                this.props.onUpdate({ 'update': { ...rowData, storageDomainId: value, sdErrFlag: true } })
```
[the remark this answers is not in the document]
bamsalem: Hi Radek,
let's take a look on the storage domain `valueView` function:
https://github.com/oVirt/ovirt-web-ui/blob/3a5e6bf5ca0a8c76ac961fcb67435efe68e13512/src/components/CreateVmWizard/steps/Storage.js#L225-L255
I want to focus on the use of the `sdErrFlag` - line 232, the if statement:
if we have case of invalid storage domain the `sd.isOk` will be `undefined` otherwise it will be true even it was invalid first and then the user changed it, if we don't use the `sdErrFlag` in the if statement the user could change it once and then it will be unchangeable. so to avoid this I used a dedicated flag to signify that the template defined disk was invalid for the current user, so this flag should be always true if the disk's SD was invalid at first, otherwise it will change once.
that's the purpose of using always `true` for this flag :)

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -162,21 +163,27 @@ class CreateVmWizard extends React.Component {
...
-            onUpdate={({ valid = true, ...updatePayload }) => {
-              this.handleListOnUpdate('storage', 'disks', updatePayload)
-              this.wizardStepsMap.storage.preventExit = !valid
-            }}
-          />
-        ),
+        render: () => {
+          const dataCenterId = this.state.steps.basic.dataCenterId
+          const invalidDisk = this.shouldPreventStorageExit()
+          this.wizardStepsMap.storage.preventExit = !!invalidDisk
```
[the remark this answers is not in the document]
bamsalem: I don't know why it's not part of the state, maybe we should consider to change it.
@sgratch, @sjd78 what do you think? should we change the `preventExit` to be part of the component's state?

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -227,7 +234,12 @@ class CreateVmWizard extends React.Component {
       this.wizardStepsMap.review,
     ]
   }
-
+  shouldPreventStorageExit () {
+    const dataCenterId = this.state.steps.basic.dataCenterId
```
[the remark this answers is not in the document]
bamsalem: Done. Thanks :)

src/components/CreateVmWizard/CreateVmWizard.js
bamsalem: It's the default value.
maybe I'm not understanding what's the problem?

rszwajko: Sorry for not being clear enough - this comment is linked with my comment [above](https://github.com/oVirt/ovirt-web-ui/pull/1233#discussion_r449618665)

> It seems we should trigger the validations and set here also the valid flag (otherwise it will be always true).

It seems that we should explicitly set `valid` flag there.

src/components/CreateVmWizard/steps/Storage.js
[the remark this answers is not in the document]
rszwajko: Thanks for explanation! I would suggest renaming it to sth like `createdFromInvalidTemplate` and adding some comments in the code to explain this use case (that user should always be able to change it).
How about remaining points?

> It seems we should trigger the validations and set here also the valid flag (otherwise it will be always true).
> It seems we don't need to pass update prop name as a string.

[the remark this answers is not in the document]
hstastna: > ...the user could change it once and then it will be unchangeable...

I wonder why it was implemented like that, originally. I think we should solve this the other way than adding an extra flag which will always be set to the same value. But I feel I still don't understand all the details so maybe I just don't see the reason why you've chosen to solve the problem of changing SD by adding such a flag :)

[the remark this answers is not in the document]
bamsalem: Hi Radek,
the validations are in the CreateVmWizard component and the update prop is not string anymore.
Thanks.

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -406,13 +418,16 @@ class CreateVmWizard extends React.Component {
     this.props.onCreate(basic, nics, disks, correlationId)
   }
 
-  wizardGoToStep (newStepIndex) {
+  wizardGoToStep (newStepIndex, moveBackward = false) {
```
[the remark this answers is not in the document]
bamsalem: thanks, I don't know why we still can see that , but it's something I already changed

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -230,11 +230,26 @@ class Storage extends React.Component {
           const {
             storageDomainId: id,
             storageDomain: sd,
+            createdFromInvalidTemplate,
```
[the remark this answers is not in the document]
bamsalem: I'm not sure what you want me to do here, can you explain me please, thanks :)

hstastna:
@bamsalem There are some conflicts in the _CreateVmWizard.js_ that need to be resolved. Thanks! :)

src/components/SelectBox.js
```diff
@@ -43,16 +44,30 @@ class SelectBox extends React.Component {
       this.props.onChange(id)
     }
   }
+  getValidationState () {
+    const { validationState } = this.props
+    switch (validationState) {
+      case 'error':
+        return style['selectBox-has-error']
+      case 'warning':
+        return style['selectBox-has-warning']
```
hstastna, line 53: I think we don't need this case right now. I would avoid adding unnecessary code which can be added later anytime if needed. Otherwise, LGTM.

hstastna (approved):
Except some unnecessary code from my point of view, LGTM.

src/components/SelectBox.js
bamsalem: Hi Hilda, I agree that is not necessary right now, but IMHO it's just component's functionality, so if we already edited the component, it will be better to add the functionality to create consistency with project's UI library - PF3 WDYT?

src/components/CreateVmWizard/steps/Storage.js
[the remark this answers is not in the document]
rszwajko: My understanding is that @sjd78  suggests calculating `createdFromInvalidTemplate` upfront in the same way `isFromTemplate` is calculated.

rszwajko (changes requested):

src/components/CreateVmWizard/steps/Storage.js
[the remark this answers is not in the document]
bamsalem: thanks @rszwajko ,I did it.

src/components/SelectBox.js
hstastna: I still think we should avoid adding unnecessary code. We don't know if we will use it in the future.

rszwajko (approved):

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -286,11 +292,28 @@ class Storage extends React.Component {
...
+            if (!sd.isOk) {
+              storageDomainList.unshift({ id: '_', value: `-- ${msg.createVmStorageSelectStorageDomain()} --` })
+            }
+            return <SelectBox
+              id={`${idPrefix}-${rowData.id}-storage-domain-edit`}
+              items={storageDomainList}
+              selected={sd.isOk ? rowData.storageDomainId : '_'}
+              validationState={!sd.isOk && 'error'}
+              onChange={value => {
+                // createdFromInvalidTemplate - flag which marks that the disk created in unallowed sd for the current user
```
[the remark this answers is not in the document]
sjd78: Or just remove the comment since I think the condition on line 229 is sufficiently descriptive.

src/components/SelectBox.js
```diff
@@ -83,6 +94,10 @@ SelectBox.propTypes = {
   /* eslint-enable react/no-unused-prop-types */
   onChange: PropTypes.func.isRequired, // (selectedId: string) => any
   id: PropTypes.string,
+  validationState: PropTypes.oneOfType([
```
sjd78, line 97: Looks like `validationState` should just be a list of possible values, and have it be optional (default 'default').  So like:

```jsx
validationState: PropTypes.oneOf([ 'default', 'error' ]),
```

Then you can add a default value like this (add below the propTypes):

```jsx
SelectBox.defaultProps = {
  validationState: 'default',
}
```

src/components/SelectBox.js
```diff
@@ -43,16 +44,26 @@ class SelectBox extends React.Component {
       this.props.onChange(id)
     }
   }
+  getValidationState () {
```
sjd78, line 47: Add a blank line between component methods.

I'd just call it `getValidationClass`.

src/components/CreateVmWizard/steps/Storage.js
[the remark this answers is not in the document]
sjd78: Or just remove the comment since I think the condition on line 229 is sufficiently descriptive.

sjd78 (changes requested):
A few minor clean up comments and it'll be all set.

src/components/SelectBox.js
bamsalem: @sjd78 it seems like it doesn't work in that case:
image
In case of using `OR`/`AND` statements `false` is possible returned value, so we need to consider it in our `PropTypes`.
WDYT about using one of the options below:
```
validationState: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.oneOf([ false ]),
  ]),
```
```
  validationState: PropTypes.oneOf([ false, 'default', 'error' ]),
```
Do you have another suggestion?

rszwajko: I understand you refer to following usage pattern:
`validationState={!sd.isOk && 'error'}`
It's an universal problem of dynamic optional attribute. There are several ways to solve that (i.e. [this thread])(https://stackoverflow.com/questions/31163693/how-do-i-conditionally-add-attributes-to-react-components).
Actually I would rather avoid embedding that in type definition (sorry for not spotting that during first review). 
We could i.e. just fall-back to default: `validationState={!sd.isOk && 'error' || 'default'}`

bamsalem: I prefer the trinary statement too, in my first PRs @sjd78 ask me every time to change trinary statements to that pattern and in the project most of the time we are using that pattern:
> validationState={!sd.isOk && 'error'}

So IMHO we need to support using this pattern or change it everywhere in the project to keep consistency.
WDYT?

hstastna: I think both options are okay if t's just about PropTypes. It doesn't have any impact on functionality, right?

sjd78: Ok, so....

While it is true that PropTypes don't break things at run time, we should not be seeing any errors when running in dev mode.  In prod mode, PropTypes warnings are not added.  Runtime type errors tend to indicate that something will go very wrong, very quietly.

Yes I usually like to see a `condition && something` rather than `condition ? something : undefined` but it is more context sensitive.  When doing components that are optionally rendered, the `condition && <Component>` is my favorite.  When doing attributes, could go either way, but really depends on what the attribute values need to be.  With an attribute that needs to be `undefined` (PropTypes not required) or one of a set of possible values (PropTypes.oneOf), the ternary makes sense.  So here, it would look like `validationState={sd.isOk ? 'default' : 'error'}` if the PropTypes doesn't allow a false value that maps to 'default'.

If `SelectBox` will only have 2 states - normal and error - then the attribute can just be a boolean `valid`.  If we want to have >2 states, then a enum of strings makes sense.  That just means the optional attribute must be present and a string or must be undefined and then get a default value.

Finally, using: `PropTypes.oneOf([ false, 'default', 'error' ])` and handling `false` the same as `undefined` / use the default value, should be ok.

sjd78 (approved):
LGTM

sgratch:
@bamsalem When verifying this PR, I'm not forced to choose a valid storage domain for a user when creating a VM.
For both template's and new disks there is no verification that the disk’s storage domain is a valid domain for the current user. It's just failed to create on backend due to permissions issue ("Cannot add VM. The user doesn't have permissions to attach Disk Profile to the Disk").
Am I missing something?

sgratch:
When verifying this PR, I'm not forced to choose a valid storage domain for a user when creating a VM.
> For both template's and new disks there is no verification that the disk’s storage domain is a valid domain for the current user. It's just failed to create on backend due to permissions issue ("Cannot add VM. The user doesn't have permissions to attach Disk Profile to the Disk").

This is actually divided into 2 sub issues:

- **For creating a disk based on a template** - this is blocked by #1237 since currently the option to create a disk based on a template is not fully supported.  Once #1237 is done, we'll need to enable changing the SD and disk type for template's disks for supporting this issue. Or we can display an error and avoid the user from creating such a VM. 

- **For creating a disk not based on a template**, when creating a disk for a user that has no "_diskprofileuser_" permissions for the SD, the ui allows me to choose that SD and the disk creation failed only on backend due to permissions issue.

  It seems that the logged user should have permissions for the SD and also should be allowed to manage disks on that SD.
Do we verify both with the API query?

sjd78:
> When verifying this PR, I'm not forced to choose a valid storage domain for a user when creating a VM.
> 
> > For both template's and new disks there is no verification that the disk’s storage domain is a valid domain for the current user. It's just failed to create on backend due to permissions issue ("Cannot add VM. The user doesn't have permissions to attach Disk Profile to the Disk").
> 
> @bamsalem @sjd78 This is actually divided into 2 sub issues, both are not working:
> 
>   * **For creating a disk based on a template** - this is blocked by #1237 since currently the option to create a disk based on a template is not fully supported.  Once #1237 is done, we'll need to enable changing the SD and disk type for template's disks for supporting this issue. Or we can display an error and avoid the user from creating such a VM.

This PR adds the SD selection, but the create VM saga was not adjusted to handle the change.  My PR #1316, which fixes the types and updates the VM create saga, hopefully handles the saga changes needed in this PR.  I will need to do some extra testing with both PRs applied to verify.

> 
>   * **For creating a disk not based on a template**, when creating a disk for a user that has no "_diskprofileuser_" permissions for the SD, the ui allows me to choose that SD and the disk creation failed only on backend due to permissions issue.
>       It seems that the logged user should have permissions for the SD and also should be allowed to manage disks on that SD.
>       Do we verify both with the API query?

The list of storage domains shown to the user is based on `canUserUseDomain` calculated in the saga that fetches the storage domain: https://github.com/oVirt/ovirt-web-ui/blob/124d2858c7f8a7867d27d9a6fcd5bcd8d9a03660/src/sagas/storageDomains.js#L67-L85

Remember, we check based on permit and not on permission, and `canUserUseDomain` is checking if the user has the permit `create_disk`.  It is quite possible a user picks up the necessary permit via another permission/role.  See: https://github.com/oVirt/ovirt-web-ui/blob/124d2858c7f8a7867d27d9a6fcd5bcd8d9a03660/src/utils/permissions.js#L33-L35

Sounds like we'll need to adjust the permit checking and add a `canUserCreateDisks` type of check and use that in various places.  Or at least update `canUserUseDomain`'s check.

sjd78:
This PR is to be replaced with #1323.  It is this PR merged with #1316.

sjd78:
> > * **For creating a disk based on a template** - this is blocked by #1237 since currently the option to create a disk based on a template is not fully supported.  Once #1237 is done, we'll need to enable changing the SD and disk type for template's disks for supporting this issue. Or we can display an error and avoid the user from creating such a VM.
> 
> This PR adds the SD selection, but the create VM saga was not adjusted to handle the change. My PR #1316, which fixes the types and updates the VM create saga, hopefully handles the saga changes needed in this PR. I will need to do some extra testing with both PRs applied to verify.

@sgratch , @bamsalem - This is now working in every case that I tested.  There may be some other edge cases, but at this point I'm not sure it is worth finding and handling them.  We should be well covered as is over in #1323.

> 
> > * **For creating a disk not based on a template**, when creating a disk for a user that has no "_diskprofileuser_" permissions for the SD, the ui allows me to choose that SD and the disk creation failed only on backend due to permissions issue.
> >   It seems that the logged user should have permissions for the SD and also should be allowed to manage disks on that SD.
> >   Do we verify both with the API query?
> 
> Remember, we check based on permit and not on permission, and `canUserUseDomain` is checking if the user has the permit `create_disk`. It is quite possible a user picks up the necessary permit via another permission/role. See:
> 
> Sounds like we'll need to adjust the permit checking and add a `canUserCreateDisks` type of check and use that in various places. Or at least update `canUserUseDomain`'s check.

@sgratch , @bamsalem - The storage domain drop down was displaying the storage domains that I expected to see when unit testing #1323.  It was not a comprehensive test, more of an average path set of tests.  I would not hold up #1323 (which will take the place of this PR) with additional permit checking concerns.

sjd78:
Moved the last comment to #1323, closing this PR.

Issue #1206, opened 2020-05-11
GET https://engine0.lab/favicon.ico 404 (Not Found)


sandrobonazzola:
Description of problem: 
console shows "GET https://engine0.lab/favicon.ico 404 (Not Found)"

VM Portal version number: ovirt-web-ui-1.6.1-1.el8.noarch

oVirt version number: 4.4.0 RC1

Browser and version: Google Chrome 81.0.4044.138

Steps to Reproduce:
1. Open the web ui
2. look at the console
3.

Actual results:
 "GET https://engine0.lab/favicon.ico 404 (Not Found)"

Expected results:
favicon.ico is found

Additional info:

sgratch:
@sjd78 Maybe this is due to removing the following line from branding's `Makefile`?:
"ln -sf "$(PKG_BRANDING_DIR)/images" "$(DESTDIR)$(WEB_UI_BRANDING_DIR)/images"

sjd78:
> @sjd78 Maybe this is due to removing the following line from branding's `Makefile`?:
> "ln -sf "$(PKG_BRANDING_DIR)/images" "$(DESTDIR)$(WEB_UI_BRANDING_DIR)/images"

Unlikely.  This issue happens on both upstream and downstream.

In absence of a `<link rel="shortcut icon" href="/.../favicon.ico">` style tag in the index html file, the browser automatically attempts to fetch one from `server-root/favicon.ico`.  The error described in this issue seems to have always been this way.

A proper fix will be to put the correct favicon path in a `link` tag on the server side (in prod the index.jsp, in dev something similar).  Once the react app is loaded, a `link` tag with the correct favicon.ico file path is injected on the page.  So the user visible side of the issue is only briefly noticeable.

sgratch:
> Unlikely. This issue happens on both upstream and downstream.

@sjd78 In addition to the mentioned 404 error, the tab icon is not shown for 4.4 d/s VM portal (while it is shown for 4.4 u/s and for 4.3.9 as well). So I guess we did broke that for d/s only.
If it's turn out to be a branding issue after all as well, then I'll open a separate branding bug for that.

sjd78:
> > Unlikely. This issue happens on both upstream and downstream.
> 
> @sjd78 In addition to the mentioned 404 error, the tab icon is not shown for 4.4 d/s VM portal (while it is shown for 4.4 u/s and for 4.3.9 as well). So I guess we did broke that for d/s only.
[...]

PR #1234, opened 2020-06-15
Add initial favicon to index.hbs, branding and config updates as needed

sjd78:
- Branding on web-ui is simpler than on webadmin.  All of the branding
    files need to follow the same path names.  This lets us pre-populate
    the favicon with an expected path.

  - branding code (in `src/index.js) has been updated so it only creates
    new link elements if they don't already exist.  This will prevent
    a second load of favicon.ico if the expected path is the real path.

  - Webpack configs for HtmlWebpackPlugin no contain the "public path"
    of the app.  The template prefixes this path on the favicon path
    so it works on both the dev server and in a production deploy.

 Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1206

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -216,6 +216,7 @@ module.exports = {
       filename: 'index.html',
       inject: true,
       template: `!!handlebars!${paths.appHtml}`,
+      publicPath: '/',
       jspSSO: false,
     }),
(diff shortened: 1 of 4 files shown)


sjd78:
ci test

isaranova:
LGTM

Issue #1227, opened 2020-06-02
Suspended VM cannot be shut down 


isaranova:
Description of problem:
If a VM is suspended (run VM -> suspend -> wait until the status is Suspended), the only action available for the VM is to run it unlike in Admin Portal where you can also shut it down.

VM Portal version number:
ovirt-web-ui-1.6.2-1.el8ev.noarch

oVirt version number:
ovirt-engine-4.4.1-0.1.el8ev.noarch

Browser and version:
firefox-76.0.1-1.fc31.x86_64

Steps to Reproduce:
1. Run VM, suspend it and wait until the state is Suspended
2. Check available actions in main grid
3. Check available actions in VM detail

Actual results:
Only run actions is available.

Expected results:
Both shutdown and run actions are available

Additional info:
According to the tests, the shutdown action used to be working for suspended VMs in previous versions.

hstastna:
Just a note that this is a UI issue. I think I have a fix, I just need to make sure about it.

PR #1236, opened 2020-06-17
Enable Shutdown button for a suspended VM

hstastna:
**Fixes:** https://github.com/oVirt/ovirt-web-ui/issues/1227

If a VM is suspended (run VM -> suspend it -> wait until the status is suspended), the only action available for the VM was to run it unlike in Admin Portal where you can also shut it down. In this PR, I am fixing this by adding missing `'suspended'` in the array of VM statuses for _Shutdown_ button being enabled in VM details page. With this change, we achieve the shutdown action being available for a suspended VM again.

**Related comment** (and PR): https://github.com/oVirt/ovirt-web-ui/pull/581#issuecomment-384895425
_"The shutdown and force stop should be available in all this VM states: up, powering up, powering down, reboot in progress, migrating, **suspended**, paused, not responding."_

---

**Before:**
suspend_before

**After:**
suspend_after

### DIFF
--- a/src/vm-status.js
+++ b/src/vm-status.js
@@ -3,7 +3,7 @@ export function canStart (state) {
 }
 
 export function canShutdown (state) {
-  return ['up', 'migrating', 'reboot_in_progress', 'paused', 'powering_up', 'powering_down', 'not_responding'].includes(state)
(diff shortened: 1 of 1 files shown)


pnovotny:
ci build please

pnovotny:
LGTM

PR #1238, opened 2020-06-18
Update fetch page size, refactor background refresh sagas

sjd78:
Resolves: #1240 
This change is related to [BZ1795457](https://bugzilla.redhat.com/show_bug.cgi?id=1795457).

As first steps to help mitigate performance issues on the
ovirt-engine's DB due to web-ui polling expensive queries
at a high frequency, a few change have been made.

  - The default VM and Pool fetch page size is now 20, closer to a reasonable non-Admin user number of VM.  This can help reduce the number of calls made on login.

  - Refactored the background data refresh sagas to `background-refresh.js` to keep them logically isolated.

  - The refresh scheduler is now controlled by the `changePage()` saga instead of the `login()` saga.  This will by keeping a consistent time between when data is fetched for a page and when the refresh will happen.

  - The refresh now throttles calls to run only 1 refresh at the start and end of a 5 second interval (instead of a 1 second interval).  This helps prevent a manual refresh from running too often or running at the same time as a scheduled refresh.

### DIFF
--- a/src/sagas/background-refresh.js
+++ b/src/sagas/background-refresh.js
@@ -1,0 +1,213 @@
+import {
+  all,
+  call,
+  put,
+  race,
+  select,
+  take,
+  takeEvery,
+  takeLatest,
+  throttle,
+} from 'redux-saga/effects'
+
+import * as Actions from '_/actions'
+import * as C from '_/constants'
+
+import AppConfiguration from '_/config'
+import { isNumber } from '_/utils'
+import { delay } from './utils'
+
+import {
+  fetchPoolsByCount,
+  fetchSinglePool,
+  fetchSingleVm,
+  fetchVmsByCount,
+  selectVmDetail,
+} from './index'
+import { getConsoleOptions } from './console'
+import { fetchIsoFiles } from './storageDomains'
+
+/**
+ * Change the background refresh type based on the page type, and force a refresh.
+ *
+ * This should be done at the time of navigation to the page, typically by the page router.
+ */
+function* changePage (action) {
+  yield put(Actions.stopSchedulerFixedDelay())
+  yield put(Actions.setCurrentPage(action.payload))
+  yield put(Actions.refresh({
+    onNavigation: true,
+    shallowFetch: true,
+    page: yield select(state => state.vms.get('page')),
+  }))
+  yield put(Actions.startSchedulerFixedDelay())
+}
+
+/**
+ * Invoke the correct refresh function based on the app's current page type.
+ */
+function* refreshData (action) {
+  const currentPage = yield select(state => state.config.get('currentPage'))
+  const refreshType =
+    currentPage.type === C.NO_REFRESH_TYPE ? null
+      : currentPage.type === undefined ? C.MAIN_PAGE_TYPE
+        : currentPage.type
+
+  console.info('refreshData() 🡒 payload:', action.payload, 'currentPage:', currentPage, 'refreshType:', refreshType)
+
+  // TODO: We could keep track of the last time the timed refresh starts vs the last
+  // TODO: time a refresh for the same page type was run.  If it is closer than a
+  // TODO: given interval, push the timed refresh back out (kinda like a debounce).
+  // TODO: There is no need to refresh data that was loaded a few seconds ago and may
+  // TODO: not even be full rendered yet.
+
+  if (refreshType) {
+    yield pagesRefreshers[refreshType](Object.assign({ id: currentPage.id }, action.payload))
+  }
+  console.info('refreshData() 🡒 finished')
+}
+
+const pagesRefreshers = {
+  [C.MAIN_PAGE_TYPE]: refreshMainPage,
+  [C.DETAIL_PAGE_TYPE]: refreshDetailPage,
+  [C.DIALOG_PAGE_TYPE]: refreshDialogPage,
+  [C.CONSOLE_PAGE_TYPE]: refreshConsolePage,
+}
+
+function* getIdsByType (type) {
+  const ids = yield select(state =>
+    state.vms
+      .get(type)
+      .reduce((entityIds, entity, entityId) => {
+        entityIds.push(entityId)
+        return entityIds
+      }, [])
+  )
+  return ids
+}
+
+function* refreshMainPage ({ shallowFetch, page }) {
+  shallowFetch = !!shallowFetch
+
+  // refresh VMs and remove any that haven't been refreshed
+  const fetchedVmIds = yield fetchVmsByCount(Actions.getVmsByCount({
+    count: page * AppConfiguration.pageLimit,
+    shallowFetch,
+  }))
+
+  const vmsIds = yield getIdsByType('vms')
+  const fetchedDirectlyVmIds =
+    (yield all(
+      vmsIds
+        .filter(vmId => !fetchedVmIds.includes(vmId))
+        .map(vmId => call(fetchSingleVm, Actions.getSingleVm({ vmId, shallowFetch })))
+    ))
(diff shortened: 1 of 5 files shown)


src/sagas/background-refresh.js
```diff
@@ -0,0 +1,213 @@
...
+
+import {
+  fetchPoolsByCount,
+  fetchSinglePool,
+  fetchSingleVm,
+  fetchVmsByCount,
+  selectVmDetail,
+} from './index'
+import { getConsoleOptions } from './console'
+import { fetchIsoFiles } from './disks'
```
sgratch, line 28: Shouldn't it be imported from ./storageDomains instead of ./disks ?

src/sagas/background-refresh.js
```diff
@@ -0,0 +1,213 @@
...
+      onSchedule: true,
+      shallowFetch: true,
+      page: yield select(state => state.vms.get('page')),
+    }))
+  }
+}
+
+export default [
+  takeEvery(C.START_SCHEDULER_FIXED_DELAY, startSchedulerWithFixedDelay),
+  throttle(5000, C.REFRESH_DATA, refreshData),
```
sgratch, line 211: This was changed from 1000 to 5000, please add that to patch description as well.

src/sagas/background-refresh.js
sjd78: Yep you're quite right.  Good catch!

src/sagas/background-refresh.js
sjd78: Will do.  Really it just help to cut down the number of refreshes that get called if a user gets click happy on the refresh button.

src/sagas/background-refresh.js
```diff
@@ -0,0 +1,213 @@
...
+import { getConsoleOptions } from './console'
+import { fetchIsoFiles } from './storageDomains'
+
+/**
+ * Change the background refresh type based on the page type, and force a refresh.
+ *
+ * This should be done at the time of navigation to the page, typically by the page router.
+ */
+function* changePage (action) {
+  yield put(Actions.stopSchedulerFixedDelay())
```
rszwajko, line 36: Is it necessary to tear down the scheduler? Perhaps it would be enough to force refresh?
With next cycle the scheduler should trigger the refresh for the correct page anyway.
Tearing down the scheduler should not disrupt any logic (it should be prepared for that) but it
makes the scheduler look  one-purpose-only (for refreshing) and  one-page-only.
It seems however that this service will grow in the future into kind of event loop which handles various actions i.e.
in [Add user settings PR](https://github.com/oVirt/ovirt-web-ui/pull/1088) we need to toggle notifications globaly on a time basis.
The proposed implementation re-uses scheduler to accomplish that goal.

rszwajko:
I see the following error after switching to details view and back
```
utils.js:240  error  uncaught at refreshData TypeError: state.getIn(...).toJS is not a function
    at http://localhost:3000/static/js/bundle.js:85724:93
    at Array.forEach (<anonymous>)
    at Object.UPDATE_VMS (http://localhost:3000/static/js/bundle.js:85704:8)
    at http://localhost:3000/static/js/bundle.js:85412:36
    at combination (http://localhost:3000/static/js/bundle.js:46935:30)
    at http://localhost:3000/static/js/bundle.js:75349:31
    at p (<anonymous>:1:36402)
    at v (<anonymous>:1:36684)
    at <anonymous>:1:40069
    at Object.dispatch (http://localhost:3000/static/js/bundle.js:46713:23)
```

hstastna:
> I see the following error after switching to details view and back
> 
> ```
> utils.js:240  error  uncaught at refreshData TypeError: state.getIn(...).toJS is not a function
>     at http://localhost:3000/static/js/bundle.js:85724:93
>     at Array.forEach (<anonymous>)
>     at Object.UPDATE_VMS (http://localhost:3000/static/js/bundle.js:85704:8)
>     at http://localhost:3000/static/js/bundle.js:85412:36
>     at combination (http://localhost:3000/static/js/bundle.js:46935:30)
>     at http://localhost:3000/static/js/bundle.js:75349:31
>     at p (<anonymous>:1:36402)
>     at v (<anonymous>:1:36684)
>     at <anonymous>:1:40069
>     at Object.dispatch (http://localhost:3000/static/js/bundle.js:46713:23)
> ```

Yeah, I can see that, too, but that's in master, too. And I am not sure but I think there's fix for this in some other PR..

sjd78:
> > I see the following error after switching to details view and back
> > ```
> > utils.js:240  error  uncaught at refreshData TypeError: state.getIn(...).toJS is not a function
> >     at http://localhost:3000/static/js/bundle.js:85724:93
> >     at Array.forEach (<anonymous>)
> >     at Object.UPDATE_VMS (http://localhost:3000/static/js/bundle.js:85704:8)
> >     at http://localhost:3000/static/js/bundle.js:85412:36
> >     at combination (http://localhost:3000/static/js/bundle.js:46935:30)
> >     at http://localhost:3000/static/js/bundle.js:75349:31
> >     at p (<anonymous>:1:36402)
> >     at v (<anonymous>:1:36684)
> >     at <anonymous>:1:40069
> >     at Object.dispatch (http://localhost:3000/static/js/bundle.js:46713:23)
> > ```
> 
> Yeah, I can see that, too, but that's in master, too. And I am not sure but I think there's fix for this in some other PR..

That error I logged in #1224 and fixed in #1225.

src/sagas/background-refresh.js
sjd78: The scheduler has become single purpose.  It's job is to use polling to refresh useful data for the currently displayed page.

For the user settings side of things, I have no problem setting up other single use timer based tasks.  But we can chat about it on #1088.

src/sagas/background-refresh.js
```diff
@@ -0,0 +1,213 @@
...
+function* refreshData (action) {
+  const currentPage = yield select(state => state.config.get('currentPage'))
+  const refreshType =
+    currentPage.type === C.NO_REFRESH_TYPE ? null
+      : currentPage.type === undefined ? C.MAIN_PAGE_TYPE
+        : currentPage.type
+
+  console.info('refreshData() 🡒 payload:', action.payload, 'currentPage:', currentPage, 'refreshType:', refreshType)
+
+  // TODO: We could keep track of the last time the timed refresh starts vs the last
```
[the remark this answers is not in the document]
sjd78: I revisited this comment yesterday in my paged refresh work and realized that it is logically cleaner to have the `changePage()` cancel the delayed task, run the proper refresh and then start the scheduler back up once the initial refresh is done.  The redux cost is pretty low to do that, and it makes things much more predictable as to when the next refresh will run (e.g. 1 minute after you enter a page).  That part has always been a bit of a struggle.

sjd78:
> LGTM except the following uncaught error which I see once in a while when using the filters. Not sure what was the exact scenario and if it's related to this fix, but it's not reproduced to me prior to this fix

I read the code and I didn't touch any of that side of things.  It is strange since it implies that a VM or a Pool does not have a "name" to be filtered at that time.  I'll take another deeper look, but I'm not sure how it could even happen.

sgratch:
> > LGTM except the following uncaught error which I see once in a while when using the filters. Not sure what was the exact scenario and if it's related to this fix, but it's not reproduced to me prior to this fix
> 
> I read the code and I didn't touch any of that side of things. It is strange since it implies that a VM or a Pool does not have a "name" to be filtered at that time. I'll take another deeper look, but I'm not sure how it could even happen.

It was reproduced to me quite often when filtering vms+pools and playing with pool's vms status, but I guess it's a separate issue anyway and not relevant to this fix

sgratch:

PR #1241, opened 2020-06-27
Improve Vm Card List pagination and infinite scroll 

sjd78:
This change is related to [BZ1795457](https://bugzilla.redhat.com/show_bug.cgi?id=1795457).  Resolves: #1242

This PR builds on other currently active PRs _(they have all been merged and this PR has been rebased)_:
  - PR #1238, 27e04aa: "Update fetch page size, refactor background refresh sagas" 
  - PR #1222, 7d468a3: "refactor Makefile to prevent unexpected symlinks"
  - PR #1225, 89c38ea: "Simplify UPDATE_VMS reducer handling of 'copySubResources'" 

The paged fetch handling for Vm list and Pool list has been improved:
  - The page tracking for VMs and Pools is now separated and handled independently.
  - The fetch/refresh sagas track the pages themselves instead of relying on other components to pass in the page/count to fetch.
  - The fetch/refresh sagas fetch VMs and Pools in parallel.
  - Since REST API doesn't provide a total record count in a paged response, we need to make an educated guess about if additional pages of data are available.  Sagas `fetchByPage()` and `refreshListPage()` have the logic for determining if more pages of VMs or Pools can be expected.
  - The first call to `refreshListPage()` will cause the first page of VMs and Pools to be fetched.  Since this is triggered by the `changePage()` saga, the list of VMs and Pools are only fetched when the VmCardList will be displayed.  It's a minor efficiency gain for opening or reloading the app directly on a VM's details or console page.

The infinite scroll capabilities of the Card List page has been improved:
  - The card list itself scrolls instead of the entire page.  This allows the filter toolbar and create VM button to always remain visible.
  - Updated to use [@closeio/use-infinite-scroll](https://www.npmjs.com/package/@closeio/use-infinite-scroll) for the scrolling. This configuration uses a sentinel `<div>` to detect when a user scrolls down and more data needs to be fetched.
  - `VmCardList` uses React hooks to handle the scrolling and tracking of page fetched.
  - To handle the edge case of having the sentinel `<div>` still visible after the initial 2 page fetches, an extra `useEffect` is used to fetch additional pages.  Extra pages will be fetched until the sentinel `<div>` is no longer visible in the scroll area.

Basic refactoring:

  - Renamed a few things to better reflect this current use:
    - Page types (constants and refresh functions):
      - DIALOG_PAGE_TYPE to CREATE_PAGE_TYPE
      - MAIN_PAGE_TYPE to LIST_PAGE_TYPE
    - Components:
      - VmsPage to VmsListPage
      - Vms to VmCardList

  - VM fetch sagas:
    - VM fetch additional/follow items: This makes the list of additional items to be followed when a VM is fetched slightly more explicit.

  - VM list and Pool list fetching
    - getVms({ count, page }) replaces both getVmsByCount() and getVmsByPage()
    - getPools({ count, page }) replaces both getPoolsByCount() and getPoolsByPage()

  - Now we have flexibility to fetch VMs and Pools by:
    - all (unfiltered and unsorted)
    - a maximum (by count) sorted by name
    - by page (count + specific page) sorted by name

  - OvirtApi renamed 'getVmsNics' to 'getVmNics'

  - Rearranged some actions to more appropriate files

  - UI updates:
    - Better VmsList empty state component by adopting the PF3-react
      components from CreateVmWizard


sjd78:
Empty state:
VM Portal - Google Chrome_023

Demo of paged loading with page size set to 4:
ezgif com-video-to-gif

sjd78:
NOTE: std-ci needs pre-seed https://gerrit.ovirt.org/#/c/109922/

sjd78:
ci test

sgratch:
@sjd78 please resolve the current conflicts which exist after merging dependent PRs.

src/components/VmsList/VmCardList.js
```diff
@@ -0,0 +1,102 @@
...
+    const scrollVisibleTop = scrollRect.y
+    const scrollVisibleBottom = scrollRect.y + scrollRect.height
+
+    const sentinelRect = sentinelRef.current.getBoundingClientRect()
+    const sentinelTop = sentinelRect.y
+    const sentinelBottom = sentinelRect.y + sentinelRect.height
+
+    const sentinelStillInView = sentinelBottom >= scrollVisibleTop && sentinelTop <= scrollVisibleBottom
+    if (sentinelStillInView) {
+      fetchMoreVmsAndPools()
```
rszwajko, line 69: It seems that we might end up loading all pages one-after-another.
This may happen if filtering is used i.e. 
1. filter removes all vm's from newly loaded page so sentinel was not moved. 
2. vms has changed so hook is triggered and fetches new page
3. new page is also filtered out so again sentinel was not moved  ....

Perhaps filtering can be moved to saga-layer and merged with paging logic? It could fix some corner cases

src/reducers/vms.js
```diff
@@ -28,66 +27,58 @@ import {
...
+  // vms come in as Internal transformed JS objects that will be pushed to ImmutableJS
+  //     objects after an optional merge with existing vm data
+  [UPDATE_VMS] (state, { payload: { vms, copySubResources } }) {
     const updates = {}
 
     vms.forEach(vm => {
-      if (!state.getIn(['vms', vm.id])) {
-        state = state.set('notAllPagesLoaded', true)
-      }
+      const existingVm = state.hasIn(['vms', vm.id]) ? state.getIn(['vms', vm.id]).toJS() : false
```
rszwajko, line 55: Paging logic might get confused if new page contains existing vm's. In worst case no new vm's can be fetched i.e. if vm's were added to pages before newly fetched page or some vm's were renamed(vm's are fetched sorted by name).
Example of such logic:
```
function* currentVmsIds ({ payload: { count, page } }) {
    const start = count * page
    const end = start + count
```

src/mock/ovirtapi.mock.js
```diff
@@ -40,9 +40,9 @@ OvirtApi = {
     }
     return Promise.reject('')
   },
-  getVmsByCount ({ count }) {
-    return Promise.resolve(vms.slice(0, count))
-  },
+  // getVmsByCount ({ count }) {
```
rszwajko, line 43: please remove or uncomment

src/reducers/vms.js
```diff
@@ -28,66 +27,58 @@ import {
...
-  page: 1,
-
   missedVms: Immutable.Set(),
 
-  /**
-   * true ~ we need to fetch further vms and pools
-   * false ~ all visible entities already fetched
-   */
-  notAllPagesLoaded: true,
+  vmsPage: 0,
```
rszwajko, line 38: consider storing also `pageCount` in the store

src/ovirtapi/index.js
```diff
@@ -131,24 +131,22 @@ const OvirtApi = {
...
+    const max = count ? `;max=${count}` : ''
+    const params =
+      [
+        page ? 'search=' + encodeURIComponent(`SORTBY NAME ASC page ${page}`) : '',
+        additional && additional.length > 0 ? 'follow=' + encodeURIComponent(`${additional.join(',')}`) : '',
+      ]
+        .filter(p => !!p)
+        .join('&')
+
+    const url = `${AppConfiguration.applicationContext}/api/vms${max}${params ? '?' : ''}${params}`
```
rszwajko, line 145: consider extracting URL building to a method that could be used by both vms and pools

src/reducers/vms.js
```diff
@@ -254,14 +245,22 @@ const vms = actionReducer(initialState, {
     return state
   },
 
-  [SET_PAGE] (state, { payload: { page } }) {
-    return state.set('page', page)
-  },
-  [SET_CHANGED] (state, { payload: { value } }) {
-    return state.set('notAllPagesLoaded', value)
+  [UPDATE_PAGING_DATA] (state, { payload: { vmsPage, vmsExpectMorePages, poolsPage, poolsExpectMorePages } }) {
```
rszwajko, line 248: how about moving this functionality to UPDATE_VMS reducer? This provide synchronization with the actual vm list.
This might help to solve some corner cases after refresh.

src/sagas/background-refresh.js
```diff
@@ -0,0 +1,230 @@
...
+        const vmsIds = yield getIdsByType('vms')
+        const fetchedDirectlyVmIds =
+          (yield all(
+            vmsIds
+              .filter(vmId => !fetchedVmIds.includes(vmId))
+              .map(vmId => call(fetchSingleVm, Actions.getSingleVm({ vmId, shallowFetch })))
+          ))
+            .reduce((vmIds, vm) => { if (vm) vmIds.push(vm.id); return vmIds }, [])
+
+        yield put(Actions.removeMissingVms({ vmIdsToPreserve: [ ...fetchedVmIds, ...fetchedDirectlyVmIds ] }))
```
rszwajko, line 102: Should we update paging data here?
The actual number of vm's might be more or less then `vmsPage * AppConfiguration.pageLimit`.

src/reducers/vms.js
sjd78: This is the page count for VMs.

`poolsPage` is the page count for Pools.

src/sagas/background-refresh.js
sjd78: Yes, but it happen below the `all()` call.  The next line here `return yield select` is what this sub-saga returns.  Same thing happens in the pool sub-saga.

Lines 137-140 does the paging update to VMs and Pool at the same time after they've both been fully updated.

src/components/VmsList/VmCardList.js
sjd78: Filtering on the front end, paging on the back end and infinite scrolling UX don't work so well together.  

For a reference interaction set: your page size is set to 4, and you have 26 VMs named Alpha to Zulu, and you see 4x3 VMs on the screen.   A total of 3 pages are fetched up front to fill the screen (leaving 4 unfetched pages).  Now the user types in a filter of "ul".  Some cards disappear and the sentinel moves up, triggering more page fetches.  Eventually 12 cards are in view and the extra fetching stops.  If there are still pages remaining (I didn't check to see how many phonetic alphabet words have "ul" in them), then they're accessible on user scroll.  There could be a weird no-op scroll delay if a new page is fetched and no new cards match the filter.

So maybe there isn't a problem?

I'm trying to reason out when it could go bad, but really I think more pages need to be fetched to be able to fill up the screen with visible things.  Worst part will be when the user has to scroll and wait for multiple pages to be fetched until enough visible cards exist or they're hit the end of the list.

I'm just trying to think of the sentinel div coming into view as the same thing as a user clicking "load more".

src/reducers/vms.js
sjd78: Refresh pull ALL pages in one call.  That would cover the edge case of "page size is 20, I fetch my 5 VMs and someone then creates VM number 6".

Only scrolling the infinite scroll does pages fetching.

src/reducers/vms.js
sjd78: `UPDATE_VMS` doesn't know, and doesn't need to know anything about paging.  It just needs to take the VMs and merge them into redux state.

> This might help to solve some corner cases after refresh.

What do you mean "after refresh"?

src/mock/ovirtapi.mock.js
sjd78: Done.

src/ovirtapi/index.js
sjd78: It probably the only place where the URL building code matches that much.  I'd much rather give the entire API file a refactor to address things like this.

sjd78:
> Looks good however I was not able to confirm that all the corner cases are OK (see comments above).
> I would wait with merge until we are sure (unless the real risk is minimal)

We can chat out the corner cases, but I think they're mostly covered.  Worst case is that all pages need to be fetched to find a VM/Pool that matches the user's filter request.  That isn't really a problem since it needs to be done anyway.

src/components/VmsList/VmCardList.js
rszwajko: I was able to test the scenario described above and actually loading-until-full feels the right approach. 
The part that could be improved is triggering the effect on every change of `vms` object. This includes property changes.
Fetch should be probably triggered only if IDs change since only this will trigger reloading cards.

src/reducers/vms.js
rszwajko: the next line in the example above is:
```
return Array.from(yield select(state => state.vms.get('pools').keys())).slice(start, end)
```
I was concerned about going past the end of array which i.e. Java would result in ArrayIndexOutOfBoundsException.
JS seems to be very relaxed in this aspect and the above will work also for invalid start/end so in this place we are safe.
However in general similar logic should rather be based on actual number of vms.

src/sagas/background-refresh.js
rszwajko: Thanks for clarification!

src/reducers/vms.js
rszwajko: Sorry for not being clear enough - `pageCount` is just bad name. My proposal was about adding a new field that would store the number of items per page so `pageSize` is the correct name. Currently we are using value from configuration i.e..
```
count: vmsPage * AppConfiguration.pageLimit,
```
This is a good candidate to be added to user settings (at some point)

src/reducers/vms.js
rszwajko: After updating vms we should also check if paging data needs to be updated.  So both actions seem connected to me.

In `refreshListPage()` this is handled - thanks for clarification.

src/components/VmsList/VmCardList.js
sjd78: I'm not quite sure what your concern is here.

Paged fetching runs ok and new pages of VM/Pools are pushed into redux state.  The new data causes react to render everything again, which is required.  On a refresh cycle, the new data still needs to be evaluated and differences rendered.  Something like a VM starting and now the status icon and action buttons need to be rendered differently.  All of that works as appropriate.

src/reducers/vms.js
sjd78: The only thing the subset of VM or Pool ids is being used for, in the `fetchByPage` saga, is to count how many things were fetched.  If enough things were fetched to fill a single page of things, we then expect more pages.

The logic is simply based on counts of things in a fetch page.  Contents of the fetch pages don't matter.  We only assume there are more pages if the current page is full.

And yeah, in JavaScript, this is perfectly fine and will return `[]`:

```jsx
   [ 1, 3, 5, 7, 9 ].slice(100, 200)
```

Aside, I did always get annoyed at fun little features of dynamic languages things like:
```jsx
    const a = []
    a.length // === 0
    a[100] = 'b'
    a.length // === 101
```

src/reducers/vms.js
sjd78: Ah ok - that would be better placed in the `config` reducers. So, the
 ```jsx
  count: vmsPage * AppConfiguration.pageLimit
```
would just turn into something like:
```jsx
  count: vmcPage * yield select(state => state.config.get('fetchPageSize'))
```

but we can cross that bridge when we get to it.

rhv-qe-github:
ci build please

isaranova:
VM/pools grid seems to be broken for me:
Screenshot_2020-07-23 VM Portal(1)
Screenshot_2020-07-23 VM Portal

If you filter VMs/pools and then clear the filter, the whole grid "freezes" on the filtered VMs/pools. Does not happen always.

Filtering for pools seem to be broken, not sure if related to this issue.
Screenshot_2020-07-23 VM Portal(2)

sjd78:
>
> VM/pools grid seems to be broken for me:
>
Strange!  You're the first person to try it with that result.  Nothing about the card layout was changed, just the scrolling around it.  Which browser were you using?

>
> If you filter VMs/pools and then clear the filter, the whole grid "freezes" on the filtered VMs/pools. Does not happen always.
>

Do you mean the filter is still applied?

> 
> Filtering for pools seem to be broken, not sure if related to this issue.
>

Filtering problems would not be related to this fix.

isaranova:
I checked again after rebooting my environment after kernel update and it seems like I had badly updated firefox (I found the same issue with weirdly sorted card layout elsewhere) - not seeing the problem anymore, filtering also works as expected.
I apologize for the inconvenience.

However, now I can't see the infinite scroller:
Screenshot from 2020-07-23 20-38-37

firefox-78.0.2-1.fc31.x86_64

sjd78:
Layout with Pools is fixed:
screenshot-ghost local dickerson-web com_3000-2020 07

isaranova:
LGTM

PR #1243, opened 2020-07-02
Allow opening console.vv in virt-viewer when clicking on VM console in Firefox

hstastna:
**Fixes:** https://bugzilla.redhat.com/show_bug.cgi?id=1638217

Do not prompt for how to open the _console.vv_ file if the user selects _Remote_ or _virt-viewer_ as a default application in Firefox.

Browsers act differently when setting the `download` attribute of the HTML `<a>` element. For FF, we don't want to set this attribute, but in Chrome/Chromium we do want - to achieve the expected behavior.
Some more info about the `download` attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

**Before:**
open_before

**After:**
open_after

---

**Note:**
Don't forget to set the Remote or virt-viewer as a default app in Firefox general preferences, to allow opening _console.vv_ file automatically:
always-ask

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -49,7 +49,12 @@ export function fileDownload ({ data, fileName = 'myFile.dat', mimeType = 'appli
       return navigator.msSaveBlob(new Blob([data], { type: mimeType }), fileName)
     } else if ('download' in a) { // html5 A[download]
       a.href = `data:${mimeType},${encodeURIComponent(data)}`
-      a.setAttribute('download', fileName)
+
+      // set the 'download' attribute for <a> element; we don't want to set it for FF
+      if (navigator.userAgent.indexOf('Firefox') === -1) {
+        a.setAttribute('download', fileName)
+      }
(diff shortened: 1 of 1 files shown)


sjd78:
Looks good.  With this approach, we can't control the name of the file that is downloaded, the browser generates one based on the file type.  I don't think that is a problem.

If we do need to control the file name, I think the only workable way forward is to use the reflect servlet in webadmin.

I'm going to verify functionality on FF on windows and Safari on MacOS to see how the file gets downloaded.

sjd78:
For firefox, reading the docs on the `download` attribute is interesting (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes):

> Prompts the user to save the linked URL instead of navigating to it.

So one reading of that doc has FF doing exactly what it was asked to do before ... prompt the download!

isaranova:
LGTM

Issue #1244, opened 2020-07-02
Update translations for oVirt 4.4 (July-2020 update)


sjd78:
Update the translations for the project from the July 2020 updated 1.6 version in Zanata.

This correlates to the version of ovirt-web-ui that is included in ovirt-4.4.

sjd78:
July-2020 translation updates
  - ovirt-web-ui: https://github.com/oVirt/ovirt-web-ui/issues/1244
  - ovirt-engine-ui-extensions: https://bugzilla.redhat.com/1853427
  - rhv-branding: https://bugzilla.redhat.com/1853444
  - ovirt-engine: https://bugzilla.redhat.com/1853495

PR #1245, opened 2020-07-02
Translation update 2020-July-02

sjd78:
Pulled from: https://zanata.phx.ovirt.org/project/view/ovirt-web-ui
Version: 1.6

Resolves: #1244


sgratch:
> @rszwajko @sjd78 Looks like the same stuff is translated to different languages but I am not sure what exactly I should review here or how to test it. Any suggestions? Thanks! :)

@hstastna  No need to test it, just making sure that no required translations messages were removed and that added translations seems reasonable and no predefined terms (like VM) or parameters are translated by mistake

sjd78:
> @rszwajko @sjd78 Looks like the same stuff is translated to different languages but I am not sure what exactly I should review here or how to test it. Any suggestions? Thanks! :)

Unit tests take care of making sure that translations don't have keys that are not part of the base English and that all message param placeholders (stuff like "Nic {name} is broken") are NOT translated.  So making sure the build works, and seeing if different languages actually come up when you change the `?locale=it` param on the URL is more than enough.

PR #1253, opened 2020-07-16
Add Timezone dropdown in the sysprep options of Create VM Wizard

hstastna:
**Related comment:** https://github.com/oVirt/ovirt-web-ui/pull/1220#issuecomment-656896190

In this PR, I'm adding the drop down to handle the Timezone in the sysprep options, in the Basic settings of the Create VM Wizard. Also I am removing the old TZ field.

After creating a VM with sysprep enabled and some chosen TZ, the chosen TZ is already displayed as expected, when viewing the VM. Looks like nothing else needs to be done. However, I'm not sure about the expected behavior in some special cases, for example if the user chooses some TZ in the drop down and then they decide to leave the drop down empty. For now, this can be solved for example by choosing a different template and then choosing the expected one and enabling the sysprep options again. But this resets all the values of sysprep options. Not sure if this is an expected behavior. 

**Before:**
Timezone field, we were able to write there anything:
tz_before1
Review step of the Create VM Wizard:
tz_before2

**After:**
New Timezone dropdown, empty drop down is allowed:
tz_after1
tz_after2
Review step of the Create VM Wizard:
tz_after3
After creating the VM, we can see the same Timezone already set, when viewing the VM's options:
tz_after4

### DIFF
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -518,13 +518,12 @@ class BasicSettings extends React.Component {
               />
             </FieldRow>
 
-            {/* TODO: Swap out for a proper timezone drop down once PR #1004 is merged */}
             <FieldRow label={msg.sysPrepTimezone()} id={`${idPrefix}-sysPrepTimezone`} vertical>
-              <FormControl
-                id={`${idPrefix}-sysPrepTimezone-edit`}
-                type='text'
-                defaultValue={data.initTimezone}
-                onChange={e => this.handleChange('initTimezone', e.target.value)}
+              <SelectBox
+                id={`${idPrefix}-sysprep-timezone-select`}
+                items={timezones}
+                selected={data.initTimezone}
+                onChange={selectedId => this.handleChange('initTimezone', selectedId)}
               />
             </FieldRow>
 


sgratch:
@hstastna it seems not working so well for creating a new Windows VM based on the blank template (when sysprep timezone is not set on template and not set by the create VM as well). The default value is not set properly.

hstastna:
Maybe we need also this https://github.com/oVirt/ovirt-web-ui/pull/1220 to have the default value set properly.

sgratch:
> I think we need also this #1220 to have the default value set properly.

I verified with this with #1220.  After re-investigating [my previous comment](https://github.com/oVirt/ovirt-web-ui/pull/1253#issuecomment-675833457) then it seems that this can be fixed on a separate issue:
there is no default value consistency on UI while comparing to webadmin for sysprep timezone null value displayed on vm details card -> advanced options ->timezone value.
When creating a new VM based on a blank template and avoid choosing a specific value for the SysprepTimeZone then the following code is running on webadmin frontend:

  getWindowsSysprepTimeZone().setItems(windowsTimezones.entrySet());
   getWindowsSysprepTimeZone().setSelectedItem(Linq.firstOrNull(windowsTimezones.entrySet(),
                item -> item.getValue().startsWith("(GMT) Greenwich")));

So in case no value is set (null),  the default value is displayed as not set (greyed out) with a "(GMT) Greenwich Standard Time" value.

On web-ui, in case no value is set by the user (null on DB), the default value is displayed as set with value "(GMT-12:00) Dateline Standard Time".
For fixing, we should set this value as not set (greyed out) with the correct default value, the same as on webadmin.

sgratch:
@pnovotny after verifying #1220 then can you please verify this issue as well? It's based on that. issue Thanks!

pnovotny:
> On web-ui, in case no value is set by the user (null on DB), the default value is displayed as set with value "(GMT-12:00) Dateline Standard Time".
> For fixing, we should set this value as not set (greyed out) with the correct default value, the same as on webadmin.

This PR build behaves as you wrote:
- in web-ui, the null/empty TZ value in Sysprep is then displayed in VM detail as "(GMT-12:00)..."
- in webadmin Edit VM dialog, the Sysprep TZ is disabled/greyed out with "(GMT)..." value

hstastna:
@pnovotny @sgratch The problem mentioned above isn't caused by adding a timezone drop down. It just shows us inconsistency we already have here. I can reproduce the problem without this PR perfectly (on master branch). The problem will be fixed by another PR.

hstastna:
Issue regarding the problem mentioned above: https://github.com/oVirt/ovirt-web-ui/issues/1276

hstastna:
@pnovotny @sgratch @sjd78 Is there anything I should make regarding this PR? I think it's ready to merge. Thanks! :)

sjd78:
> @pnovotny @sgratch @sjd78 Is there anything I should make regarding this PR? I think it's ready to merge. Thanks! :)

Looks all set to me.  All followup work is captured in #1276 / PR #1288.

PR #1260, opened 2020-07-23
Create VM - Allow valid only calls to onUpdate functions in CreateVmWizard

sjd78:
In order to better support input validation in the wizard steps, the `onUpdate` functions have been updated to allow calls that only
update the step's valid state.  Now, a step can flag itself as invalid thereby preventing navigation off the step until the input is fixed or canceled.

This change will enable consistent validation updates on #1219, #1221, #1233

### DIFF
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -4,6 +4,7 @@ import { connect } from 'react-redux'
 import { Wizard, Button, Icon } from 'patternfly-react'
 import NavigationConfirmationModal from '../NavigationConfirmationModal'
 import merge from 'lodash/merge'
+import isEmpty from 'lodash/isEmpty'
 import { List } from 'immutable'
 
 import * as Actions from '_/actions'
@@ -177,7 +178,12 @@ class CreateVmWizard extends React.Component {
 
             onUpdate={({ valid = true, ...updatePayload }) => {
               this.handleListOnUpdate('network', 'nics', updatePayload)
+
+              const isPreventExitChanged = this.wizardStepsMap.network.preventExit === valid
               this.wizardStepsMap.network.preventExit = !valid
+              if (isEmpty(updatePayload) && isPreventExitChanged) {
+                this.forceUpdate()
+              }
             }}
           />
         ),
@@ -199,7 +205,12 @@ class CreateVmWizard extends React.Component {
 
             onUpdate={({ valid = true, ...updatePayload }) => {
               this.handleListOnUpdate('storage', 'disks', updatePayload)
+
+              const isPreventExitChanged = this.wizardStepsMap.storage.preventExit === valid
               this.wizardStepsMap.storage.preventExit = !valid
+              if (isEmpty(updatePayload) && isPreventExitChanged) {
+                this.forceUpdate()
+              }
             }}
           />
         ),
@@ -385,6 +396,10 @@ class CreateVmWizard extends React.Component {
   }
 
   handleListOnUpdate (stepName, listName, { remove, update, create }, setStateCallback) {
+    if (!remove && !update && !create) {
+      return
+    }
+
     this.setState(state => {
       let list = this.state.steps[stepName][listName].slice(0)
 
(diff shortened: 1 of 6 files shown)


src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -171,7 +172,12 @@ class CreateVmWizard extends React.Component {
 
             onUpdate={({ valid = true, ...updatePayload }) => {
               this.handleListOnUpdate('network', 'nics', updatePayload)
+
+              const isPreventExitChanged = this.wizardStepsMap.network.preventExit === valid
               this.wizardStepsMap.network.preventExit = !valid
+              if (isEmpty(updatePayload) && isPreventExitChanged) {
+                this.forceUpdate()
+              }
```
bamsalem, line 180: Looks good , but I think that the `===` condition is a little bit confusing and not intuitive
and maybe it can be something like :
`
const newPreventExit = !valid
const isPreventExitChanged = this.wizardStepsMap.network.preventExit !== newPreventExit
this.wizardStepsMap.network.preventExit = newPreventExit
`
IMHO it will improve the readability...
everything else looks good

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -171,7 +172,12 @@ class CreateVmWizard extends React.Component {
 
             onUpdate={({ valid = true, ...updatePayload }) => {
               this.handleListOnUpdate('network', 'nics', updatePayload)
+
+              const isPreventExitChanged = this.wizardStepsMap.network.preventExit === valid
               this.wizardStepsMap.network.preventExit = !valid
+              if (isEmpty(updatePayload) && isPreventExitChanged) {
+                this.forceUpdate()
```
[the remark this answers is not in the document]
sjd78: Reasonable.  I'm actually in process of converting the component state of CreateVmWizard to use immer, and to have the `preventExit` and `preventEnter` be part of state.  It'll be better that way.

sjd78:
ci test

src/components/CreateVmWizard/CreateVmWizard.js
```diff
@@ -358,68 +348,61 @@ class CreateVmWizard extends React.Component {
...
+    if (stepNavigation[activeStep.id].preventExit || stepNavigation[newStep.id].preventEnter) return
 
     // run and the current step's `onExit()`
     if (activeStep.onExit) {
       activeStep.onExit()
     }
 
-    this.setState({ activeStepIndex: newStepIndex })
+    this.setState(produce(draft => { draft.activeStepIndex = newStepIndex }))
   }
```
bamsalem, line 406: what about move to the previous step?
in this case prevent exit will prevent to move backward.

sjd78: Yes, that is correct and has always been that way.  Prevent exit means, "don't let the user leave this step".  Prevent enter means, "don't let the user move to this step either by direct jumping or moving forward/back".

sjd78:
> steps to reproduce:
> 
>   1. Start the create vm wizard dialog.
> 
>   2. Navigate to storage/networking step.
> 
>   3. Add new disk/nic and save the new row.

Oh that one is INTERESTING!

So, **immer** creates objects that are not mutable (really this is accomplished by making the object frozen).  The list of nics being passed down to the `Networking` component is immutable.  The component SORTS the array before rendering it.  `Array.sort()` in JS is an in-place operation.  Can't sort an array that is immutable!

The fix, in this case, is to move the `sort()` below the `map()`.  `Array.map()` creates a new array.  The new array is mutable / is not read-only.  Now the sort is applied to the array created by `map()` and it's all working.

I didn't hit the error during unit testing because the array sort didn't need to do anything for my set of nics.  Some lucky combo of default names and not creating more than 1 or 2 nics.

I'll post the fix shortly.

src/components/CreateVmWizard/CreateVmWizard.js
rszwajko: @sjd78 
in [1233](https://github.com/oVirt/ovirt-web-ui/pull/1233) Ben added possibility to go back regardless of `preventExit`
If this is a bad idea please comment there.

sjd78:
> Except the "move backward" behavior, which we should consider to change to allow users move to the previous step, but you raised that on the stand-up so I assume we will change that, or at least discuss that...

With this PR I do not want to change the allowed navigation.

As originally intended, a step must be valid before moving off of it.  Right now an active add/edit for a Nic or Disk will allow you to navigate off the step.  What is done with the active edit isn't clear to the user.  That was always intended to be changed to while there is an active add/edit, keep the user on the step.  The user needs to explicitly commit the change (and deal with any input validation issues at that time) or cancel it.  We'll deal with that in the validation PRs.

Issue #1248, opened 2020-07-09
Create VM Wizard - cannot tab through edit fields on Networking and Storage steps


sjd78:
>
> 5. you can only reach first input in the row by Tab key
>

Yep - that's partially due to the way the table uses `onBlur` to update values.  Also, that would be easier solved in another issue as it applies in both Storage and Network tabs.

_Originally posted by @sjd78 in https://github.com/oVirt/ovirt-web-ui/pull/1221#issuecomment-656342168_

The Networking and Storage steps should allow users to tab between the active input fields when creating or editing a Nic or Disk.

Since `onBlur` is used to update the field values, the change event handlers would probably need to be updated to allow navigation.

bamsalem:
Hi guys,
@sjd78 @sgratch @hstastna @rszwajko 
I'm working on this issue and I have few things that I want to rase:

1. Double click issue: if one of the fields is focused to focus another field the user need to double click it:
![doubleClickCreateVmWizardField-2020-07-22_12 02 33](https://user-images.githubusercontent.com/64131213/88158257-7d310c00-cc14-11ea-849f-4e60e427dde0.gif)
it happens because we using onBlur event, There have been many discussions on that on #1221 issue, and this is Scott's  [explanation](https://github.com/oVirt/ovirt-web-ui/pull/1221#issuecomment-660606958)

2. Move between dropdown options using keyboard: if one of the dropdown fields focused and the user hit the "space key" the drop down options appear as expected but when he tries to move between the options nothing happened:
![dropDownOptionsSwitch-2020-07-22_12 30 44](https://user-images.githubusercontent.com/64131213/88172831-42d26980-cc2a-11ea-89c3-b938fd4c3050.gif)

I'm not sure that those issues included in this issue or if they should be handled in another issues.
WDYT?

sjd78:
> I'm working on this issue and I have few things that I want to raise:
> 
>     1. Double click issue: if one of the fields is focused to focus another field the user need to double click it:
>        it happens because we using `onBlur` event, There have been many discussions on that on #1221 issue, and this is Scott's  [explanation](https://github.com/oVirt/ovirt-web-ui/pull/1221#issuecomment-660606958)
> 

So it's caused by the same problem as trying to Tab through the fields - onBlur() causes the table to re-render and the input field gets reset/unfocused.  If there is a reasonable solution to the tabbing issue, this problem would probably be resolved also.

>
[...]

Issue #585, opened 2018-04-30
Operating System drop down - can't scroll using keyboard shortcut


mykaul:
When clicking on the OS drop down list, I expect clicking 'R' to hop over to the first OS that begins with the letter 'R'.

sgratch:
Opened for more than 2 years and no one else complained or fixed so closing as WONTFIX.
Please re-open if you think it's important.

PR #1261, opened 2020-07-28
Create VM Wizard - add ability to tab through edit fields in Storage and Networking steps 

bamsalem:
This PR added support in navigation between edit fields using keyboard:
1.Tab - move to the next field.
2. Shift+Tab - move to the previous field.

I used React refs to control the fields' focus.

tabThroughExample-2020-07-28_12 01 37

Fixes: #1248

### DIFF
--- a/src/components/CreateVmWizard/steps/Networking.js
+++ b/src/components/CreateVmWizard/steps/Networking.js
@@ -384,20 +384,20 @@ class Networking extends React.Component {
     const vnicList = createVNicProfileList(vnicProfiles, { dataCenterId, cluster })
     const enableCreate = vnicList.length > 0 && Object.keys(this.state.editing).length === 0
 
-    const nicList = nics
+    const nicList = [...nics]
+      .sort((a, b) => // template based then by name.
+        a.isFromTemplate && !b.isFromTemplate ? -1
+          : !a.isFromTemplate && b.isFromTemplate ? 1
+            : localeCompare(a.name, b.name)
+      )
       .concat(this.state.creating ? [ this.state.editing[this.state.creating] ] : [])
       .map(nic => ({
-        ...nic,
+        ...(this.state.editing[nic.id] ? this.state.editing[nic.id] : nic),
         vnic: vnicList.find(vnic => vnic.id === nic.vnicProfileId)
           ? vnicList.find(vnic => vnic.id === nic.vnicProfileId).value
           : msg.createVmNetUnknownVnicProfile(),
         device: enumMsg('NicInterface', nic.deviceType),
       }))
-      .sort((a, b) =>
-        a.isFromTemplate && !b.isFromTemplate ? -1
-          : !a.isFromTemplate && b.isFromTemplate ? 1
-            : localeCompare(a.name, b.name)
-      )
     const components = {
       body: {
         row: _TableInlineEditRow,
--- a/src/components/CreateVmWizard/steps/Networking.js
+++ b/src/components/CreateVmWizard/steps/Networking.js
@@ -345,6 +345,7 @@ class Networking extends React.Component {
 
   // Cancel the creation or editing of a row by throwing out edit state
   handleRowCancelChange (rowData) {
+    this.components = undefined // remove the current reference to make the table re-render
     this.setState(state => {
       const editing = state.editing
       delete editing[rowData.id]
@@ -397,6 +398,12 @@ class Networking extends React.Component {
           : !a.isFromTemplate && b.isFromTemplate ? 1
             : localeCompare(a.name, b.name)
       )
+    const components = {
+      body: {
+        row: _TableInlineEditRow,
+      },
+    }
+    this.components = this.components || components // if the table should (re)render the value of this.components should be undefined
 
     return <div className={style['settings-container']} id={idPrefix}>
       { nicList.length === 0 && <React.Fragment>
@@ -426,11 +433,7 @@ class Networking extends React.Component {
             dataTable
             inlineEdit
             columns={this.columns}
-            components={{
-              body: {
-                row: _TableInlineEditRow,
-              },
-            }}
+            components={this.components}
           >
             <Table.Header />
             <Table.Body
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -559,9 +559,15 @@ class Storage extends React.Component {
     const filteredStorageDomainList = createStorageDomainList(storageDomains, dataCenterId)
     const enableCreate = storageDomainList.length > 0 && !this.isEditingMode()
 
-    const diskList = disks
+    const diskList = [...disks]
+      .sort((a, b) => // template based then by name.
+        a.isFromTemplate && !b.isFromTemplate ? -1
+          : !a.isFromTemplate && b.isFromTemplate ? 1
+            : localeCompare(a.name, b.name)
+      )
       .concat(this.state.creating ? [ this.state.editing[this.state.creating] ] : [])
       .map(disk => {
+        disk = this.state.editing[disk.id] ? this.state.editing[disk.id] : disk // update editing changes from state after sorting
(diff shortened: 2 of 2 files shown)


bamsalem:
Thanks Radek,
I checked your suggestion and I found that it not render the table and we can tab through the fields, and it's the purpose of this PR,
but it creates another issue, it don't render the table even after saving:
image
in the screenshot above you can see that it render the "Create Disk" button and we can see that it's enabled.
It's simpler solution, but it create another issue that we can't accept,
I tried to change your solution and solve the new issue but I didn't succeed to find a way to control when it will render and when not, so I'm not sure but I think that is not the solution.
screenshot from the dev tools:
image
WDYT?

bamsalem:
@hstastna , I found why it happens:
https://github.com/oVirt/ovirt-web-ui/blob/ed40eb69a9fce4f1655d2db21e81af36eeaaf12d/src/components/CreateVmWizard/steps/Storage.js#L445-L462
In line 447 there is an edge case that zero is passed as a string so in the next render the 0 value stored as a string and that is what makes the warning,
It does not matter so much because I dealt with it in issue #1221 and it's not related to this issue, so I didn't include that changes here...

hstastna:
Thank you very much, for the explanation, Ben! :)

rszwajko:
> It's simpler solution, but it create another issue that we can't accept,

@bamsalem 
Save action should have an effect on the props used  to render the table. Probably sth similar to the way "Create disk" works.
IMHO lack of tabbing is a bug and should be fixed as such (no new framework for tabbing is required)

bamsalem:
Hi Guys,
Thanks to @rszwajko  suggestion, I tried to handle this issue as a bug and not to create a new framework as it was handled in my prev fix.
So I found that the table re rendered because in every render() the <Table.PfProvider> component receive as a component prop a different reference and that what makes the re-rendering. So I find a way to control when the table will re-render and when it won't.
Now we can control the rendering of the table what enable us to tab through the fields.
After this PR will be approved we can make all the validations behave like in editing existing VM because we can use onChange events instead onBlur (I checked it in one field).
WDYT?
@sjd78  @sgratch @hstastna

hstastna:
I think the problem with drop downs and changing selection was already mentioned in some other PR (or issue?) and as I remember it was not easy to fix it. The problem with Save button may be related. So I haven't mentioned those problems here. But yeah, we shouldn't forget about them :)

The problem with bootable flag, well... I haven't found this problem until Radek has mentioned it. I was not able to reproduce it. But now I found it's tricky. The error doesn't occur when pressing space key, but later - when clicking on Save button, and only when there's already some other (and not a template) bootable Disk. Also it does not matter if using keyboard or mouse. In addition, this problem is present also on master branch. I am pretty sure that this worked fine when merging bootable PR so some other very recent change had to cause it.

sjd78:
> Problems I've encountered:
> 
> 1. dropdowns can be opened using keyboard (Enter key) but I was not able to change selection

See #585

bamsalem:
>I think we can use shorter
this.components = this.components || components
instead of
this.components = this.components ? this.components : components

Otherwise LGTM :)


@hstastna done.

@sjd78 can we move the PR to QE?

hstastna (approved):

bamsalem:
@hstastna @sgratch @sjd78 @rszwajko 
Hi guys,
I found  that this PR create another issue , it allows us using onChange event is better for validations (the validation is "on the go" and not on submit or onBlur), but it creates another issues like the issue I'll present here, important to note that the issue is in both of the steps: "Networking" & "Storage":
nicsResortingIssue-2020-08-25_09 35 27
like you can see the disks re-sorted during the editing and it could be annoying.
I'm not sure if it's related to this PR, or it should be dealt in another PR.
this PR's changes included in those PRs: #1219 #1230  #1221 
WDYT?

hstastna:
@yfrimanm @matthewcarleton Is it a problem? https://github.com/oVirt/ovirt-web-ui/pull/1261#issuecomment-679844631

yfrimanm:
Yes https://github.com/hstastna I totally agree it is a problem.

hstastna:
IMHO, I would place the new item at the bottom of the table, as it was before, when creating it, and after the creation is done (after the item is saved) to sort all the items in the table normally.

yfrimanm:
It does make sense to me as well https://github.com/hstastna

sjd78:
@bamsalem, @hstastna - A simple way to handle things is to not update the fields that are used to sort until the save checkbox is checked.  So, keep state for the edit fields and then when the edit is done, run the doUpdate().

bamsalem:
@hstastna @sjd78 @rszwajko @sgratch I pushed commit which resolves all sorting issues:
creating disks: always sorted to the bottom.
editing disks: sorted them according to the name before editing, and updateing them after saving and sorts them accordingly.
WDYT?

hstastna (approved):
LGTM.

src/components/CreateVmWizard/steps/Networking.js
```diff
@@ -383,20 +384,26 @@ class Networking extends React.Component {
     const vnicList = createVNicProfileList(vnicProfiles, { dataCenterId, cluster })
     const enableCreate = vnicList.length > 0 && Object.keys(this.state.editing).length === 0
 
-    const nicList = nics
+    const nicList = [...nics]
```
sjd78, line 387:
```suggestion
    const nicList = nics.slice(0)
```

It is interesting to compare the various ways to create a shallow copy of an array.  Looks like slice(0) is the fastest across browsers.  I ran this in FF and Chrome:
  - https://measurethat.net/Benchmarks/Show/9356/0/array-slice-vs-slice0-vs-spread-operator

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -558,9 +559,15 @@ class Storage extends React.Component {
     const filteredStorageDomainList = createStorageDomainList(storageDomains, dataCenterId)
     const enableCreate = storageDomainList.length > 0 && !this.isEditingMode()
 
-    const diskList = disks
+    const diskList = [...disks]
```
sjd78, line 562:
```suggestion
    const diskList = disks.slice(0)
```

src/components/CreateVmWizard/steps/Networking.js
sjd78: On further thought, the spread operator is ok in this scenario.  @rszwajko is correct, the spread does look better than `.slice(0)`.

Change is not needed.

src/components/CreateVmWizard/steps/Storage.js
sjd78: Sticking with the spread is ok.

sjd78:
Note: This PR should merge before #1219, #1221, and maybe #1233 since the onChange handlers now work instead of relying on onBlur.

src/components/CreateVmWizard/steps/Networking.js
hstastna: Now it feels like looking better is more than being faster. Could you provide some more explanation why we prefer the spread operator?

rszwajko: > being faster

.... in given test scenario on given browser version. Performance tweaks are tricky ;/

>  we prefer

well... I prefer:)

> some more explanation 

IMHO this syntax is very concise(short) and better shows the intention of the the programmer(copy of the whole array). 
As for more objective reasons - first `slice` can be easily switched by mistake with `splice`. 
Second `slice` main purpose is to copy portion of an array - making a copy of the whole array is supported but may be confusing.

hstastna: Okay, Thanks very much for the explanation, Radek! :)

pnovotny:
LGTM

Tab and Shift+Tab through the fields for NICs and disks works properly. 

Side note: having #585 working would make this feature more awesome. Currently, only values in text inputs and checkboxes can be set with keyboard.

Issue #1256, opened 2020-07-16
VM details - CPU Utilization chart details are invalid in case of more than one vCPU.


sgratch:
If a VM is configured with number of virtual CPUs > 1 (regardless of the separation into number of sockets and cores) then the CPU utilization available/used percentages values, displayed in VM details utilization chart, are wrong.

For example, a VM configured with 1 socket, 8 cores (total vCPUs is 8), and it’s using 100% of the CPU, it is displayed on VM details page as using 800% of the CPU and -700% is available:
![image](https://user-images.githubusercontent.com/18169498/87693880-26e84700-c796-11ea-8c4f-071a07e7ae7a.png)

This is reproduced on oVirt 4.3.9  and 4.4, so no regression.

PR #1262, opened 2020-07-30
Display proper VM CPU Load considering the number of vCPUs

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1256

In the actual VM Portal UI, we were displaying `cpu.current.guest` which represents how much cpu time VM's process uses regarding all of the CPUs. This value can be more than 100% so the numbers displayed in the UI were OK. The problem is that the users don't want to see this value. They want to see VM CPU Load, the average value considering the number of VM CPUs, the same as it is in Admin Portal.

To achieve this, I had to make a small calculation and to use the number of sockets and cores of a VM.

**More info** (you have to be logged in to see all the necessary details): https://access.redhat.com/solutions/337013

---

**In the Admin Portal (proper values):**
admin_portal

**VM Portal, before:**
before

**VM Portal, after:**
after

### DIFF
--- a/src/components/VmDetails/cards/UtilizationCard/CpuCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/CpuCharts.js
@@ -25,8 +25,8 @@ import NoLiveData from './NoLiveData'
  * as a sparkline. Sparkline vales to go from oldest far left to most current on far
  * right.
  */
-const CpuCharts = ({ cpuStats, isRunning, id }) => {
-  const cpuUsed = cpuStats['current.guest'].datum
+const CpuCharts = ({ cpuStats, isRunning, id, vcpus }) => {
+  const cpuUsed = cpuStats['current.total'].datum / vcpus // the average value considering the number of VM CPUs, same as in Admin Portal
   const cpuAvailable = 100 - cpuUsed
 
   // NOTE: CPU history comes sorted from newest to oldest
@@ -83,6 +83,7 @@ CpuCharts.propTypes = {
   id: PropTypes.string.isRequired,
   cpuStats: PropTypes.object.isRequired,
   isRunning: PropTypes.bool,
+  vcpus: PropTypes.number.isRequired,
 }
 
 export default CpuCharts
--- a/src/components/VmDetails/cards/UtilizationCard/index.js
+++ b/src/components/VmDetails/cards/UtilizationCard/index.js
@@ -19,6 +19,8 @@ const UtilizationCard = ({ vm }) => {
   const stats = vm.has('statistics') ? vm.get('statistics').toJS() : undefined
   const isRunning = [ 'up' ].includes(vm.get('status'))
 
+  const vCpus = vm.getIn(['cpu', 'vCPUs'])
+
   const idPrefix = 'vmdetail-utilization'
 
   return (
@@ -33,7 +35,7 @@ const UtilizationCard = ({ vm }) => {
           <Col className={style['row-col-charts-box']}>
             <Col className={style['col-charts-box']}>
               { stats.cpu
-                ? <CpuCharts cpuStats={stats.cpu} isRunning={isRunning} id={`${idPrefix}-cpu`} />
+                ? <CpuCharts cpuStats={stats.cpu} isRunning={isRunning} id={`${idPrefix}-cpu`} vcpus={vCpus} />
                 : <NoLiveData message={msg.loadingTripleDot()} id={`${idPrefix}-cpu-no-data`} />
               }
             </Col>


src/components/VmDetails/cards/UtilizationCard/CpuCharts.js
```diff
@@ -25,8 +25,8 @@ import NoLiveData from './NoLiveData'
  * as a sparkline. Sparkline vales to go from oldest far left to most current on far
  * right.
  */
-const CpuCharts = ({ cpuStats, isRunning, id }) => {
-  const cpuUsed = cpuStats['current.guest'].datum
+const CpuCharts = ({ cpuStats, isRunning, id, sockets, cores }) => {
+  const cpuUsed = cpuStats['current.total'].datum / sockets / cores // the average value considering the number of VM CPUs, same as in Admin Portal
```
sjd78, line 29: Have you verified that "current.total" is the VM CPU use from the VM's perspective?

I'm not sure "current.guest" and "current.total" are the same number.

As long as the correct cpu stat is used, the calc to normalize against vCPUs makes sense.

hstastna: I'm not sure what is meant by "verifying rom the VM's perspective".

According to the info I've found, "current.guest" and "current.total" are not the same number:
`cpu.current.total: cpu.current.guest + cpu.current.hypervisor`

And the number displayed in Admin Portal is calculated like this:
`cpu.current.total / ( sockets * cores )`

And as I properly understand this issue, we want to see the same values in the VM Portal, so it seems that we don't need the "current.guest" value but "current.total" instead. Also I've checked that now with this PR, the values in VM Portal are the same as in Admin Portal so the issue should be fixed and we should be consistent in both of the UIs.

src/components/VmDetails/cards/UtilizationCard/index.js
```diff
@@ -19,6 +19,10 @@ const UtilizationCard = ({ vm }) => {
   const stats = vm.has('statistics') ? vm.get('statistics').toJS() : undefined
   const isRunning = [ 'up' ].includes(vm.get('status'))
 
+  const vCpuTopology = vm.getIn(['cpu', 'topology'])
```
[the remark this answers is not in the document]
hstastna: Hmm, I can use the vm `vCPUs` but with this change I would include the number of `threads` in the calculations, too, but it's not clear to me why I should do it. Maybe you can make it clear to me :) Thanks!

sjd78:
> It seems to me that the cpu usage visible in Web Admin is taken from the DB from `vm_statistics.usage_cpu_percent`.
> At runtime the engine accumulates those values into a list used to draw a chart. Last value (the most current) is displayed next to the chart. For details see `VmStatistics.addCpuUsageHistory()`.
> It seems that this data is already available in REST response we use to populate VM details (search for key `cpu.usage.history`)

Did you see any calcs to normalize the usage to a 0% to 100% scale?

src/components/VmDetails/cards/UtilizationCard/CpuCharts.js
sjd78: I don't remember exactly where each number gets collected from.  I believe one number is CPU usage as reported by the guest agent, while another is the CPU usage as reported by vdsm on the VM's host.

I'm not even sure the CPU usage needs to match between webadmin and web-ui.  Webadmin usage is from the admin perspective...how much resources is this VM taking up WRT the host.  Web-ui usage is from the use perspective...how much resources is the OS using WRT the VM.  Subtle difference for sure.

I'll have to look to see if I can find the original discussions about which statistics to use.

sjd78: https://github.com/oVirt/ovirt-engine/blob/a026144f551027629c47a3ccd3f06b57e207dc25/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/VmStatisticalQuery.java#L20-L22

rszwajko:
> Did you see any calcs to normalize the usage to a 0% to 100% scale?

The unit is percentage and it seems it's 0-100 (to be verified). If I understand this PR correctly then 
we want to see the same value as in Web Admin - even if it's >100% then.

hstastna:
The github issue does not inform about the expected result clearly but users were complaining about values higher than 100% but haven't mentioned what exactly they expect to see instead. So I assume they don't want to see the values higher than 100% so they clearly don't want to see `cpu.current.guest` value which is displayed in the VM Portal and is related to the all CPUs. And this value really can be higher than 100% if there are at least 2 vCPUs. According to this it's clear for me that this is not what users want to see. Plus the value is not consistent with the Admin Portal. The users were complaining only regarding the VM Portal, not Admin Portal. I think they want to see the average value and same as in Admin Portal, so with this PR we shouldn't see the value higher than 100% anymore and be consistent with the Admin Portal.

We should be consistent with the Admin Portal also because I think the same VM's info ("CPU") with the same values in both of the UIs is expected. And if not, then we should distinguish between different things in the UIs (and not to mark both of them simply as "CPU").

I think the best is to stick to the quite clear and straightforward explanation and calculations [here(https://access.redhat.com/solutions/337013) and to be consistent with Admin Portal.

@rszwajko I've checked the `cpu.usage.history` in the VM Portal and it seems that you are right, the last value from the history seems to be the same value as I've calculated, but less precise. I would better stick to the actual solution, as we display decimal values also for the memory etc, too, in the VM Portal. And I think I use the proper values for the calculation.

rszwajko:
> We should be consistent with the Admin Portal also because I think the same VM's info ("CPU") with the same values in both of the UIs is expected

IMHO the best way to achieve this is to display the same data and let the engine do magical calculations. Looking at  `VmStatistics.updateRuntimeData()` I see at least one special case (and any time someone can add more)
```
      if ((getCpuSys() != null) && (getCpuUser() != null)) {
            Double percent = (getCpuSys() + getCpuUser()) / numOfCpus;
            setUsageCpuPercent(percent.intValue());
            if (getUsageCpuPercent() != null && getUsageCpuPercent() > 100) {
                setUsageCpuPercent(100);
            }
        }
```

> we display decimal values also for the memory etc, too, in the VM's details, in the VM Portal

We need to decide if we want to display the same values then Web Admin or *better*.
Web Admin seems to display usage for both cpu and memory without fractional part.
BTW the REST we are using contains also `memory.usage.history`.

hstastna:
I think we want to be consistent with Admin Portal in the way that we display the same information. And to be consistent with VM Portal to display the decimal values if we display decimal values for other stuff, in the same screen.

sjd78:
> BTW the REST we are using contains also `memory.usage.history`.

FWIW, the spark line charts under the donuts handle displaying the history for CPU and memory.

src/components/VmDetails/cards/UtilizationCard/index.js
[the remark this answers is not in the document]
sjd78: I looked through some of the stats related code on webadmin and the threads are included by default when calculating vCPUs.  So...my suggestion is to use the vCPU that is already available (threads are mostly 1 AFAIK).

src/components/VmDetails/cards/UtilizationCard/CpuCharts.js
sjd78: Ok, I think you're ok to use the `cpu.current.total`. 

But, change the prop to just `vCPUs` so the chart component only needs to divide by one value.

hstastna:
Now `vCPUs` is used instead of the number of sockets and cores from the `topology`, so the number of threads is considered in the calculation, too. I think the solution is now consistent with both webadmin's calcs and the RH solution article.

sjd78:
> It seems to me that the cpu usage visible in Web Admin is taken from the DB from `vm_statistics.usage_cpu_percent`.
> At runtime the engine accumulates those values into a list used to draw a chart. Last value (the most current) is displayed next to the chart. For details see `VmStatistics.addCpuUsageHistory()`.
> It seems that this data is already available in REST response we use to populate VM details (search for key `cpu.usage.history`)

@rszwajko, I'd rather see the source of the usage data stay the same and just add the division by vCPUs.  We can discuss it further if you feel it is absolutely necessary, but I think this PR is good as-is.

pnovotny:
LGTM

ovirt-web-ui-1.6.5-0.20201005.git0a79224.el8.noarch

The 'CPU % used' shows the same number as Webadmin does.

Issue #1251, opened 2020-07-14
Create VM Wizard - add indication of which field is invalid/missing on "Basic Settings" step for enablng the next step


sgratch:
On the "Create Virtual Machine" wizard ->  "Basic Settings" step: when a user tries to fill in the fields and the "Next" button is still disabled, need to highlight the fields that are mandatory/invalid so that the user will know why the next button is still disabled.

PR #1265, opened 2020-08-10
Create VM - BasicSettings, select automatically single option menu

bamsalem:
Fixes: #1252
**Depends on**: https://github.com/oVirt/ovirt-web-ui/pull/1291 (MERGED)

In the CreateVMWizard dialog -> "Basic settings" step the user had to select an option from list menu even if there is only one option allowed to the user.
It could be annoying and unnecessary, in this PR lists with single option will be automatically selected.

There are few changes which not related to the auto select and used to avoid the dialog close confirmation if the user not changed anything and the only changes were done are from the "auto select" feature, those changes made to maintain the behavior which created in #1231 .

PR #1266, opened 2020-08-11
Create VM Wizard - add indicators to missing/invalid fields on BasicSettings step

bamsalem:
Due to users' complains that is not clear enough why the next button is still disabled, I made the basic setting step bit more interactive and for any field after setting VM's name the dialog will indicate which field is the next missing/invalid field.

To support the interactivity I edited the `SelectBox` component and I added validation capability to the component, I tried to stick to PF3 `<FormControl/>` component's appearance and interface, you can see it in commit 0daaf5b.

I would like to hear your opinion, about the new behavior, and if you have any UX suggestions like "make them all blink together" or suggestions about the `selectBox` validations' appearance.

The new  dialog:
BasicSettingsIndicators-2020-08-10_20 20 00 (1)

Fixes: #1251

### DIFF
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -221,7 +221,7 @@ class BasicSettings extends React.Component {
 
   validateVmName () {
     const name = this.props.data.name
-    return name === undefined || name.length === 0 || isVmNameValid(name) ? null : 'error'
+    return name === undefined || name.length === 0 || !isVmNameValid(name) ? 'error' : null
   }
 
   // Calculate the proper number of virtual sockets, cores per virtual socket, threads per core according to the number of virtual CPUs
@@ -354,6 +354,10 @@ class BasicSettings extends React.Component {
   render () {
     const idPrefix = this.props.id || 'create-vm-wizard-basic'
     const { data, clusters, maxNumOfSockets, maxNumOfCores, maxNumOfThreads } = this.props
+    const indicators = {
+      name: this.validateVmName(),
+    }
+
     const clusterList =
       createClusterList(clusters)
         .map(cluster => ({
@@ -362,6 +366,9 @@ class BasicSettings extends React.Component {
         }))
     if (!isValidUid(data.clusterId)) {
       clusterList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectCluster()} --` })
+      indicators.cluster = !indicators.name && 'error'
+    } else {
+      delete indicators.cluster
     }
 
     const provisionSourceList =
@@ -371,6 +378,9 @@ class BasicSettings extends React.Component {
       ]
     if (!isValidSource(data.provisionSource)) {
       provisionSourceList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectProvisionSource()} --` })
+      indicators.provisionSource = !indicators.cluster && 'error'
+    } else {
+      delete indicators.provisionSource
     }
 
     const enableOsSelect = isValidUid(data.clusterId) && [ 'iso', 'template' ].includes(data.provisionSource)
@@ -389,6 +399,9 @@ class BasicSettings extends React.Component {
       : [ { id: '_', value: `-- ${msg.createVmWizardSelectClusterBeforeISO()} --` } ]
     if (enableIsoSelect && !isValidUid(data.isoImage)) {
       isoList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectISO()} --` })
+      indicators.isoList = !indicators.provisionSource && !indicators.name && 'error'
+    } else {
+      delete indicators.isoList
     }
 
     const enableTemplateSelect = data.provisionSource === 'template'
@@ -397,6 +410,9 @@ class BasicSettings extends React.Component {
       : [ { id: '_', value: `-- ${msg.createVmWizardSelectClusterBeforeTemplate()} --` } ]
     if (enableTemplateSelect && isValidUid(data.clusterId) && !isValidUid(data.templateId)) {
       templateList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectTemplate()} --` })
+      indicators.template = !indicators.provisionSource && !indicators.name && 'error'
+    } else {
+      delete indicators.template
     }
 
     const enableCloudInit = data.cloudInitEnabled && isOsLinux(data.operatingSystemId, this.props.operatingSystems)
@@ -420,7 +436,7 @@ class BasicSettings extends React.Component {
     return <Form horizontal id={idPrefix}>
       <Grid className={style['settings-container']}>
         {/* -- VM name and where it will live -- */}
-        <FieldRow label={msg.name()} id={`${idPrefix}-name`} required validationState={this.validateVmName()}>
+        <FieldRow label={msg.name()} id={`${idPrefix}-name`} required validationState={indicators.name}>
           <FormControl
             id={`${idPrefix}-name-edit`}
             autoComplete='off'
@@ -445,6 +461,7 @@ class BasicSettings extends React.Component {
             items={clusterList}
             selected={data.clusterId || '_'}
             onChange={selectedId => this.handleChange('clusterId', selectedId)}
+            validationState={indicators.cluster}
           />
         </FieldRow>
 
@@ -466,6 +483,7 @@ class BasicSettings extends React.Component {
               items={isoList}
               selected={data.isoImage || '_'}
               onChange={selectedId => this.handleChange('isoImage', selectedId)}
+              validationState={indicators.isoList}
             />
           </FieldRow>
         }
@@ -478,6 +496,7 @@ class BasicSettings extends React.Component {
               items={templateList}
               selected={data.templateId || '_'}
               onChange={selectedId => this.handleChange('templateId', selectedId)}
+              validationState={indicators.template}
             />
           </FieldRow>
         }
--- a/src/components/SelectBox.js
+++ b/src/components/SelectBox.js
@@ -27,6 +27,7 @@ class SelectBox extends React.Component {
       items: getItems(props),
     }
     this.handleChange = this.handleChange.bind(this)
+    this.getValidationClass = this.getValidationClass.bind(this)
   }
 
   componentWillReceiveProps (nextProps) {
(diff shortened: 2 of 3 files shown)


ovirt-infra:

src/components/SelectBox.js
```diff
@@ -43,16 +44,32 @@ class SelectBox extends React.Component {
...
+    const { validationState } = this.props
+    switch (validationState) {
+      case 'error':
+        return style['selectBox-has-error']
+      case 'warning':
+        return style['selectBox-has-warning']
+      case 'success':
+        return style['selectBox-has-success']
+      case 'blinking':
+        return style['selectBox-should-blink']
```
hstastna, line 57: I think we don't need this 'blinking' case anymore. Also I'm not sure if we need other than error case here.

bamsalem: I partially agree,
I think the blinking option is not necessary any more, but the others maybe will be necessary in the future,
it's just  component's functionality,
better keep it to create consistency with PF3's input components.

bamsalem:
@hstastna I changed according to your review:
image

src/components/SelectBox.js
hstastna: > ... but the others maybe will be necessary in the future, ...

I understand that you want to support consistency but I am not sure... I would better avoid adding any unnecessary code, a.k.a any code which is not useful/necessary right now. I think we are not sure if this will really be used in the future. And if yes, we still can add later what we will need.

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -258,6 +258,10 @@ class BasicSettings extends React.Component {
   render () {
     const idPrefix = this.props.id || 'create-vm-wizard-basic'
     const data = this.props.data
+    const nameValidation = this.validateVmName()
```
hstastna, line 261: It's not clear to me why we need this change. Ben, could you, please, explain? Thanks :)

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -336,6 +344,7 @@ class BasicSettings extends React.Component {
             items={clusterList}
             selected={data.clusterId || '_'}
             onChange={selectedId => this.handleChange('clusterId', selectedId)}
+            validationState={indicators.cluster}
```
hstastna, line 347: What's `indicators.cluster` right after opening Create VM Wizard?

`indicators.cluster = !indicators.name && 'error'` really returns `false` (boolean value) when I open a Create VM Wizard. I would suggest to check other "indicators" as well.

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -276,6 +281,7 @@ class BasicSettings extends React.Component {
       ]
     if (!isValidSource(data.provisionSource)) {
       provisionSourceList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectProvisionSource()} --` })
+      indicators.provisionSource = !indicators.cluster && 'error'
```
hstastna, line 284: `indicators.provisionSource ` will be a boolean value, too.

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -357,6 +366,7 @@ class BasicSettings extends React.Component {
               items={isoList}
               selected={data.isoImage || '_'}
               onChange={selectedId => this.handleChange('isoImage', selectedId)}
+              validationState={indicators.isoList}
```
hstastna, line 369: If `indicators.isoList` will be a boolean value, too, we get a proptype error, too. Same with using `indicators.template` [here](https://github.com/oVirt/ovirt-web-ui/pull/1266/files#diff-cef40fc8d8b38f14df16a31106f8d660R382).

src/components/CreateVmWizard/steps/BasicSettings.js
bamsalem: Thanks, I missed it, now it's been fixed

src/components/CreateVmWizard/steps/BasicSettings.js
bamsalem: I changed it a little bit.
Is it still not clear?

src/components/SelectBox.js
```diff
@@ -43,16 +44,30 @@ class SelectBox extends React.Component {
       this.props.onChange(id)
     }
   }
+  getValidationState () {
+    const { validationState } = this.props
+    switch (validationState) {
+      case 'error':
+        return style['selectBox-has-error']
+      case 'warning':
+        return style['selectBox-has-warning']
```
hstastna, line 53: I still think we should avoid adding unnecessary code. We have no any use for this right now and we don't know if we will in the future.

src/components/sharedStyle.css
```diff
@@ -84,3 +84,16 @@
 .text-align-left {
   text-align: left;
 }
+
+.selectBox-has-error {
+    border-color: #cc0000;
+}
```
hstastna, line 90: Maybe we could use already existing PF styles https://www.patternfly.org/v3/styles/color-palette/
Or is it better to add the appropriate border colors like this? I'm not sure. WDYT?

hstastna (commented):
LGTM, the indicators to missing fields work as expected. Just some minor details as I've already mentioned earlier.

src/components/sharedStyle.css
bamsalem: we chatted about that on slack, so I'll mark it as resolved

hstastna (commented):
There are some details as I've mentioned earlier but otherwise LGTM.

src/components/SelectBox.js
bamsalem: @sgratch @sjd78 @rszwajko ,
@hstastna and I have disagreements about if we need or don't need the functionality which are not used right now but we creating the same interface as PF3 [`FormControl`](https://5c507d49471426000887a6a7--react-bootstrap.netlify.app/components/forms/#forms-validation) component
WDYT guys?

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -294,6 +299,7 @@ class BasicSettings extends React.Component {
       : [ { id: '_', value: `-- ${msg.createVmWizardSelectClusterBeforeISO()} --` } ]
     if (enableIsoSelect && !isValidUid(data.isoImage)) {
       isoList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectISO()} --` })
+      indicators.isoList = !indicators.provisionSource && !indicators.name && 'error'
     }
```
rszwajko, line 303: Perhaps we should reset the indicator in the `else` block?
Just to be on the safe side - the code seems to work without that.

src/components/SelectBox.js
rszwajko: > we should avoid adding unnecessary code

> we creating the same interface as PF3 FormControl component

Both views make sense. It's up to you @bamsalem

src/components/SelectBox.js
```diff
@@ -43,16 +44,30 @@ class SelectBox extends React.Component {
       this.props.onChange(id)
     }
   }
+  getValidationState () {
+    const { validationState } = this.props
```
rszwajko, line 48: minor one: consider passing validation state as parameter. This would make `getValidationState` a pure function.
It brings some nice features (although we don't use them now).

src/components/sharedStyle.css
rszwajko: > we chatted about that on slack

A comment in the code with summary would be nice:)

Anyway ...this rule gets overridden by on-hover rule. Built-in styles should support all possible variations i.e. error and on-hover, error and selected, error and disabled etc

rszwajko (changes requested):
Generally it looks very good! Some minor comments below:

src/components/SelectBox.js
bamsalem: It won't be fully pure function, because we are using external resource, so IMO it's not matter,WDYT?

src/components/SelectBox.js
bamsalem: I thought about that, and we don't want to maintain unnecessary code so I'll remove that code :)

hstastna (approved):
LGTM.

rszwajko (approved):

bamsalem:
I pushed changes to  commit 6e55db8 :
according to #1233 [comment](https://github.com/oVirt/ovirt-web-ui/pull/1233#discussion_r495027669)

isaranova:
ci build please

isaranova:
After entering correct name, the cluster field is not in red frame as expected as it is required to be reselected to continue.
Screenshot_2020-10-12 VM Portal

bamsalem:
@isaranova I can't reproduced that in my machine.
Can you please write the steps to reproduce the issue?
I selected all the drop down fields (cluster, provision source and template) then I filled the name and it's worked as expected:
image

image

Thanks

isaranova:
@bamsalem I thought this PR should tell the user which fields are still missing as a whole including the cluster and provisioning source fields. For example, I get a missing template field after picking the provisioning source (pic), but cluster and provisioning source fields are not marked and if I won't reselect them, I'm not able to continue. If this is not part of the fix in this PR, then please let me know, because the name field indicator is working as expected.

Screenshot_2020-10-14 VM Portal

bamsalem:
@isaranova 
This is the order that the fields indicators should work.
1. Name
2. Cluster
3.  if (provsionSource == template) then: Tempale.
indicators-2020-10-14_17 54 45
     if (provisionSource == iso ) then: CD.
indicators-2020-10-14_17 59 02

did you filled the fields in this order and the indication is not work as expected?

I added 2 gifs with those scenarios in my machine and it's works perfectly.
Did you tested it with brq-dev?
I'm trying to figure out why it works for me and not for you...

BTW this PR is not dealing with the next button at all...
The PR is dealing only with indication about missing/invalid fields.
But we can't accept that regression.

isaranova:
So I tested the issue again to find out that the reason I don't see the cluster field as required is that I had only one cluster in my environment. After adding another cluster, the functionality is correct.

sgratch:
> So I tested the issue again to find out that the reason I don't see the cluster field as required is that I had only one cluster in my environment. After adding another cluster, the functionality is correct.

@isaranova  the issue of only one cluster is fixed by #1265 so that the user won't need to pre-select a cluster in case there is only one in the environment. Therefore this issue seems ok and I'm merging it after your verification.

isaranova:
LGTM

Issue #1267, opened 2020-08-12
Regression in Create VM, can't move bootable flag between disks


sjd78:
When reviewing/testing #1261, a regression on the create vm storage step was found:

> 3. selecting bootable flag (Space key) results in
> 
> ```
> Storage.js:511 Uncaught TypeError: Cannot assign to read only property 'bootable' of object '#<Object>'
>     at Storage.handleRowConfirmChange (Storage.js:511)
>     at Object.onConfirm (Storage.js:119)
> ```

That is now causing an error because of immer and making state immutable.  Pretty much any error that is "ready only property" based will be something trying to update an object created by immer.

_Originally posted by @sjd78 in https://github.com/oVirt/ovirt-web-ui/pull/1261#issuecomment-673047605_

PR #1268, opened 2020-08-12
Fix create vm storage step moving bootable flag regression

sjd78:
Resolves: #1267

A regression was introduced with making the VM state immutable. The storage step was directly updating one of the disks to remove the bootable flag before sending that object back through the `onUpdate` callback.  With immutable state via immer, the objects are set to read only and any updates to the object will throw a runtime error typically including the phrase "read only property".

The fix is simple.  Do not reuse the object.  Now `onUpdate` is called with a simple object created just to hold the update work.


sgratch:
LGTM. Merging prior to QE verification so that it will be included on coming build.

PR #1269, opened 2020-08-13
Update dependencies of jquery and lodash

sgratch:
Update jquery to 3.5.1
Update lodash to 4.17.19


sgratch:
i did a basic verification to make sure that there are no regressions and it seems OK. So Merging.

Issue #1183, opened 2020-04-07
The VM's filter always returns the Pools entities regarldess to the query result


sgratch:
The Pool entities are always displayed on VMs dashboard, regardless to the filter type/value query used. It seems that the Pool entity is not included on filter calculations.

For example:
![image](https://user-images.githubusercontent.com/18169498/78682454-aabca880-78f6-11ea-8ca0-cdebca97b66f.png)

This is reproduced on ovirt-web-ui 1version 1.6.1-1

hstastna:
Started looking into this and after I apply some filtering, I would expect ALL of the Pool entities to be displayed on VMs dashboard if they are not included in filter calculations. But that's not true, at least for me.

For me only some of them are displayed. And the number of them (when filtering) varies - depending on filter type/value query used, for example:
- originally (no filtering) I can see 5 Pool entities
- with filtering by _Name_ I can see 2 of them - as on the screenshot above
- with filtering by _Status: Running_ I can see 4 Pool entities

So the problem may be more complicated, I will continue investigating.

hstastna:
@sgratch Hi Sharon, could you, please, specify if the scenario is related to admin/non admin user and maybe also some details about user roles? I think this could help us to figure out what should be displayed. Thanks :)

sgratch:
> @sgratch Hi Sharon, could you, please, specify if the scenario is related to admin/non admin user and maybe also some details about user roles? I think this could help us to figure out what should be displayed. Thanks :)

@hstastna it was reproduced to me with admin user, didn't verify that with non-admin.

hstastna:
@sgratch, what I've found also with @sjd78 is that for admin user, Pool entities are **always** displayed, no matter filtering or number of available VMs from a Pool. It's obviously implemented like this in the code. It would be good to check the behavior also for non-admin user: we expect only those Pool entities with at least 1 available VMs to be displayed so the user could take a VM from that Pool.

Another thing I've found is that **Pool entities are included in filter calculations**, for admin user:
![pools](https://user-images.githubusercontent.com/13417815/79875581-271cb480-83ea-11ea-9f80-ca54a1e7bdf6.png)

But if we filter VMs and Pools the way that the query does not match with some Pool entity - such a Pool entity is just displayed (for admin, together with the result of filtering) and not included in the number of _Results_ displayed in the toolbar, so it just seems that Pool entities are excluded in filter calculations:
[...]

PR #1271, opened 2020-08-18
Display Pool entities containing search term when filtering

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1183

In the VM list, display the only those Pool entities which contain the current search term (not all of them as it was before for admins), treat VMs and Pool entities the same way for all user types when filtering. Always show Pool cards only when logged in as an admin user and not filtering.

---

**Before:**
before

**After:**
after

---

This PR fixes regression introduced in https://github.com/oVirt/ovirt-web-ui/pull/1241.

### DIFF
--- a/src/components/VmsList/VmCardList.js
+++ b/src/components/VmsList/VmCardList.js
@@ -28,7 +28,10 @@ const VmCardList = ({ vms, alwaysShowPoolCard, fetchMoreVmsAndPools }) => {
 
   // Filter the Pools (only show a Pool card if the user can currently 'Take' a VM from it)
   const filteredPools = vms.get('pools')
-    .filter(pool => alwaysShowPoolCard || (pool.get('vmsCount') < pool.get('maxUserVms') && pool.get('size') > 0 && filterVms(pool, filters)))
+    .filter(pool =>
+      (alwaysShowPoolCard || (pool.get('vmsCount') < pool.get('maxUserVms') && pool.get('size') > 0)) &&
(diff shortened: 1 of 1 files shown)


src/components/VmsList/VmCardList.js
```diff
@@ -28,7 +28,7 @@ const VmCardList = ({ vms, alwaysShowPoolCard, fetchMoreVmsAndPools }) => {
 
   // Filter the Pools (only show a Pool card if the user can currently 'Take' a VM from it)
   const filteredPools = vms.get('pools')
-    .filter(pool => alwaysShowPoolCard || (pool.get('vmsCount') < pool.get('maxUserVms') && pool.get('size') > 0 && filterVms(pool, filters)))
+    .filter(pool => (alwaysShowPoolCard || (pool.get('vmsCount') < pool.get('maxUserVms') && pool.get('size') > 0)) && filterVms(pool, filters))
```
sjd78, line 31:
```suggestion
    .filter(pool => 
      (alwaysShowPoolCard || (pool.get('vmsCount') < pool.get('maxUserVms') && pool.get('size') > 0)) 
      && filterVms(pool, filters)
    )
```

Good catch, nothing like parens causing problems!

I'm only suggesting an alternate layout so it's a bit easier to eye-ball the filter condition bits.  I'm ok with leaving it on one line, but my eye/brain/code-reading coordination prefers the multiline style. WDYT?

sgratch:
LGTM

PR #1272, opened 2020-08-18
Fix Create VM Storage step using the wrong iface for a new disk

sjd78:
New disks should default to "virtio_scsi" in most cases.  However, when editing a disk created in the storage step, the base "virtio_scsi" constant would be converted to the user facing string "VirtIO-SCSI" by mistake.  This has been removed.


sgratch (approved):
LGTM

Issue #1101, opened 2019-08-22
[ja_JP] The VM status and up-time text needs to be localized


sgratch:
Originally opened by Bhushan Barve.

Description of problem:The VM status and the up-time text needs to be localized for ja_JP locale. I see that the VM status is localized for other locales. Please refer attached screenshot.

Version-Release number of selected component (if applicable):
RHV 4.3

How reproducible:
always

Steps to Reproduce:
1. Login to VM portal with jp locale
2. Observe the VM status and up-time for any running VM.

Actual results:
The VM status and uptime are not localized.

Expected results:
The VM status and up-time shou;d be localized.

Additional info:
https://bugzilla.redhat.com/attachment.cgi?id=1571257

sjd78:
The VM status is translatable, and the status keys and messages exist in the **ja** translation.  They are:
```json
    "enum_VmStatus_down": "Off",
    "enum_VmStatus_image_locked": "Image locked",
    "enum_VmStatus_migrating": "Migrating",
    "enum_VmStatus_not_responding": "Not Responding",
    "enum_VmStatus_paused": "Paused",
    "enum_VmStatus_powering_down": "Powering down",
    "enum_VmStatus_powering_up": "Powering up",
    "enum_VmStatus_reboot_in_progress": "Reboot in progress",
    "enum_VmStatus_restoring_state": "Restoring state",
    "enum_VmStatus_saving_state": "Saving state",
    "enum_VmStatus_suspended": "サスペンド中",
    "enum_VmStatus_unassigned": "未割り当て",
    "enum_VmStatus_unknown": "不明",
    "enum_VmStatus_up": "Running",
    "enum_VmStatus_wait_for_launch": "Wait For Launch",
```

Screen shot of the Overview card with the dummy locale that puts arrows around all text:
![screenshot-ghost local dickerson-web com_3000-2020 08 20-22_15_08](https://user-images.githubusercontent.com/3985964/90844773-ae474e00-e332-11ea-93b3-620319c1f75a.png)

The uptime is not yet localized.

Uptime needs to handled in the following sports:
https://github.com/oVirt/ovirt-web-ui/blob/c4f55006ea7d686408ce22b87fb23d2d46188995/src/components/VmDetails/cards/OverviewCard/index.js#L276-L278

and
https://github.com/oVirt/ovirt-web-ui/blob/c4f55006ea7d686408ce22b87fb23d2d46188995/src/utils/format.js#L7-L37

[...]

PR #1282, opened 2020-08-24
Externalize duration formatting, refactor intl scripts

sjd78:
Fixes: #1101 

Duration formatting for VM uptime is now localizable:

  - Bump the version of moment and moment-duration-format

  - Add support and app initialization code to localize moment and
    moment-duration-format

  - Add `src/intl/time-durations.js`:
    - define short and standard duration labels
    - apply translations of the duration labels at app-init

  - Refactor the intl scripts to have yarn call shell scripts
    and have the scripts handle messages.js AND time-durations.js

  - Adjust zanata.xml to handle pulls of multiple documents


sjd78:
This is how the Overview card with the dummy/test locale turned on (translatable parts are surrounded by arrows):
VM Portal - Google Chrome_028

sjd78:
ci test

hstastna (commented):
The code changes look quite ok to me, however I'm not sure how to properly test the changes. Any suggestion? Thanks in advance! :)

rszwajko:
> how to properly test the changes?
```
yarn intl:dummy
```
and then enter http://localhost:3000//?locale=aa

hstastna:
Tested, LGTM.

Just a small question, maybe not related to the duration formatting but to other translations: is it intentional not to translate "Blank" or "Other OS"? Thanks! :)

src/intl/time-durations.js
```diff
@@ -0,0 +1,235 @@
...
+  //
+  // Extend the moment locale object for moment-duration-format:
+  //   https://github.com/jsmreese/moment-duration-format#extending-moments-locale-object
+  //
+  const translations = require('./translated-time-durations.json')
+  if (translations[locale]) {
+    const t:{ [messageId: string]: string } = translations[locale]
+
+    moment.updateLocale(locale, {
+      durationLabelsStandard: {
```
rszwajko, line 33: Are we sure that our supported languages have only 1 plural form? If not then we should be ready for more then one plural form. This is the case i.e. in Polish where different plural is used for 2-4 and for >= 5.
For more reading  check: https://en.wikipedia.org/wiki/Dual_(grammatical_number)

rszwajko (changes requested):

sjd78:
> Tested, LGTM.
> 
> Just a small question, maybe not related to the duration formatting but to other translations: is it intentional not to translate "Blank" or "Other OS"? Thanks! :)

"Blank" ... where is that string?  If it is part of data fetched from rest, we can't translate it.

"Other OS" is the label as defined in the OS info fetched from rest.  Not translatable.

src/intl/time-durations.js
sjd78: So basically the solution implemented isn't perfect but should be good enough.

The perfect solution would be to have moment-duration-format defined localizations the same way as moment does it. Moment embeds both the strings and functions on how to render plurals properly per locale.  Since I don't know that for different languages, and since we use zanata for translations, bridging them how I did I think is reasonable.

Also note: In Japanese, I'm fairly sure, they don't use plurals in this context at all.  So their translations for 1 second and >1 second will be the same.

hstastna:
> "Blank" ... where is that string? If it is part of data fetched from rest, we can't translate it.
It's a chosen Template in the VM Details card.

sjd78:
> > "Blank" ... where is that string? If it is part of data fetched from rest, we can't translate it.
> 
> It's a chosen Template in the VM Details card.

Ah ok, that makes sense.  It is data!

src/intl/time-durations.js
```diff
@@ -0,0 +1,235 @@
...
+        S: t.durationLabelStandard_S || timeDurations.durationLabelStandard_S.message,
+        SS: t.durationLabelStandard_SS || timeDurations.durationLabelStandard_SS.message,
+        s: t.durationLabelStandard__s || timeDurations.durationLabelStandard_s.message,
```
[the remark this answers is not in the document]
sjd78: Good catch, thanks!



> Typo in key name (double underscore).

Good catch, thanks!

> Manual mapping is quite error prone - it would be nice to add more validation here (either runtime or via type system).

Yeah, but stuff does need to be mapped, and validations aren't automatic either, and types also aren't automatic....I don't really know of an easy way to do this kind of thing in javascript.  We can chat about ideas for these kinds of improvements in future.

rszwajko (approved):

hstastna (approved):

PR #1287, opened 2020-09-03
Translation pull from zanata (2020-Sep-2)

sjd78:
Translation pull from zanata (2020-Sep-2)

Pulled from: https://zanata.phx.ovirt.org/project/view/ovirt-web-ui
Version: 1.6

Bug-Url: https://bugzilla.redhat.com/1358295


src/intl/translated-messages.json
```diff
@@ -305,8 +306,8 @@
     "diskEditorDiskTypeCreateHelp": "Sobald Sie einen Disk-Typ ausgewählt haben, können Sie ihn nicht mehr ändern.",
     "diskEditorDiskTypeLabel": "Disk-Typ",
     "diskEditorDiskTypeNotAvailable": "n. z.",
-    "diskEditorDiskTypeOptionThin": "Schlanke Speicherzuweisung",
     "diskEditorDiskTypeOptionPre": "Vorab zugeordnet",
+    "diskEditorDiskTypeOptionThin": "Schlanke Speicherzuweisung",
```
rszwajko, line 310: it seems that in the new version the keys are now sorted correctly. How it's possible that previously 
the sort order was wrong? do we have some bug in our zanata scripts?

sgratch (approved):
LGTM (but I wonder regarding https://github.com/oVirt/ovirt-web-ui/pull/1287#discussion_r484880497 as well)

hstastna (approved):
LGTM.

rszwajko:
Looks OK. 

Regarding sorting - it seems that we don't do that for web UI. Instead we are explicitly using stable to/from JSON conversion.
Can you confirm @sjd78 ?

rszwajko (approved):

sjd78:
> Looks OK.
> 
> Regarding sorting - it seems that we don't do that for web UI. Instead we are explicitly using stable to/from JSON conversion.
> Can you confirm @sjd78 ?

There is implied sorting in the normalize script.  Specifically in this bit of library code:

https://github.com/samn/json-stable-stringify/blob/4df7f6b08a702473173d0ae1968f994dfa179b8c/index.js#L51

rszwajko:
> 
> There is implied sorting in the normalize script. Specifically in this bit of library code:
> 

The sorting seems optional - you need to provide comparator in the option object. 
This would explain why key `diskEditorDiskTypeOptionThin` changed location.
Perhaps we should add sorting?

sjd78:
> > There is implied sorting in the normalize script. Specifically in this bit of library code:
> 
> The sorting seems optional - you need to provide comparator in the option object.
> This would explain why key `diskEditorDiskTypeOptionThin` changed location.
> Perhaps we should add sorting?

Providing a key sort function `cmp` is optional, but the way the library builds the object keys into an array and then does `keys.sort(cmp && cmp(node))` will always sort the array.  With no `cmp` provided, it's the same as calling `[ .. ].sort(undefined)` which then uses the default sort order (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort):

> The default sort order is ascending, built upon converting the elements into strings, then comparing their sequences of UTF-16 code units values.

So why would the sort order would be different? I don't know.  But putting the explicit sort, probably just copying from ui-extensions, in there is easy enough in another PR.

PR #1289, opened 2020-09-06
Tooltip refactoring - use one type of tooltip to create consistency in VM portal

### LINKED ISSUE #1094: Different Tooltip Types

bamsalem:
Fixes #1094 .
This PR includes changes according to [Tooltips Document](https://docs.google.com/document/d/1ouQ7b4tKH1ksmB54hnPz69rkHcynv7qhG3hDXf342XA/edit#).
I found that the document is not up to date, there are few components which not included, I changed them too but I ceparated them from the document's changes those changes included in 6a3c1cd.
After consultation with Sharon we decided to use `<OverlayTooltip />` component instead PF4 `<Tooltip />` component.
WDYT?


ovirt-infra:

src/components/OverlayTooltip/InfoTooltip/index.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import OverlayTooltip from '../'
+import { InfoCircleIcon } from '@patternfly/react-icons'
+import style from './style.css'
+
+const InfoToolTip = ({ id, tooltip, placement, ...rest }) => (
```
rszwajko, line 7: We loose some functionality when replacing FieldLevelHelp:
1. keyboard support
2. accessibility support

It seems that PF4 Tool tip seems a better replacement in those aspects (see [1]) 
Alternatively we can follow the guideline (see [2]) and re-implement this ourselves. 

[1] https://www.patternfly.org/v4/documentation/react/components/tooltip#on-icon
[2] https://www.patternfly.org/v4/get-started/accessibility-guide

src/components/OverlayTooltip/index.js
```diff
@@ -17,8 +17,8 @@ const OverlayTooltip = ({ id, tooltip, placement = 'left', children, ...rest })
 }
 OverlayTooltip.propTypes = {
   id: PropTypes.string.isRequired,
-  tooltip: PropTypes.string.isRequired,
-  placement: PropTypes.string,
+  tooltip: PropTypes.oneOfType([ PropTypes.string, PropTypes.node ]).isRequired,
+  placement: PropTypes.oneOf(['top', 'right', 'bottom', 'left', false]),
```
rszwajko, line 21: This prop type is repeated many times. Perhaps it's worth to create a tool-tip-alignment-prop-type and re-use it.

src/components/VmActions/index.js
```diff
@@ -132,26 +132,12 @@ class VmActions extends React.Component {
...
-    // TODO: On the card list page, the VM's consoles would not have been fetched yet,
-    // TODO: so the tooltip on the console button will be blank.
-    let consoleProtocol = ''
-    if (!vm.get('consoles').isEmpty()) {
-      const vConsole = vm.get('consoles').find(c => c.get('protocol') === 'spice') || vm.getIn(['consoles', 0])
-      const protocol = vConsole.get('protocol').toUpperCase()
-      consoleProtocol = msg.openProtocolConsole({ protocol })
-    }
-
-    if (vm.get('consoleInUse')) {
```
rszwajko, line 144: "console in use" functionality was not marked for removal

rszwajko (dismissed):
Consistent look feels good but we loose functionality.

hstastna (commented):
Maybe you could add little more info in the description of this PR, to sum up the changes in this PR. For example that this PR includes some changes also in messages, tooltips on the console etc.
And to add some screenshot(s) because this PR is mainly about visual changes - about making tooltips look consistent in the VM Portal.

As Radek has mentioned, we should check the functionality regarding replacing _FieldLevelHelp_ :) Otherwise it looks good to me :)

src/components/VmActions/index.js
bamsalem: Hi @rszwajko 
The quoted text from the tooltips document.

>  Similar comment goes for these infotips, I would get rid of them unless they are there for accessibility.

since `consoleProtocol` variable used only for infotip which removed, then I removed all the lines that deals with that variable, include "console in use.
the console infotip is not mentioned explicitly but IMHO if I wouldn't remove that inotip it would create inconsistency.
WDYT?

src/components/OverlayTooltip/InfoTooltip/index.js
rszwajko: second thought on that: re-implementing is not worth the effort.
I would rather postpone this PR and implement it when we will start to use PF4. Then we could just use Tooltip component.
Alternatively if we are OK with losing functionality then let's make our wrapper easy to be replaced by PF4 component (i.e. the same props, similar usage pattern). The wrapper then will be just a temporary solution.

src/components/VmActions/index.js
rszwajko: OK. I took a second look at this code and it seems that we have a mix of features and bugs here. 
Features:
1. when console is closed the tool tip is `Open {protocol} console`  (where protocol is SPICE or first available)
2. when console is open the tool tip is `Console in use`

Bugs:
1. it doesn't work in 'on card' scenario - the tool tip is set on main action whereas `VmDropdownActions` uses `action.items`. This seems the only scenario where it makes sense - the primary action is triggered by the dropdown itself.
2. it doesn't work for VM details scenario - `ActionButtonWraper` does not pass tool tip to `DropdownButton` (which by the way seems not to support it anyway). Btw I'm not sure if this bug or feature because in this use case the main button triggers no action.
3. `Console in use` is not translated
4. tool tips for individual console actions seems not to work (all scenarios)

I vote for removing this feature: it is broken anyway (and none complained) .

src/components/OverlayTooltip/index.js
bamsalem: I agree, I changed that according to your review

src/components/VmDetails/CardEditButton.js
```diff
@@ -40,30 +38,29 @@ class CardEditButton extends React.Component {
...
-      </OverlayTrigger>
+      </OverlayTooltip>)
     }
     if (!editable) {
       return null
     }
 
     return (
-      <OverlayTooltip id={`${id}-tooltip`} tooltip={tooltip} placement='bottom'>
+      <OverlayTooltip id={`${id}-tooltip`} tooltip={tooltip} placement={placement}>
```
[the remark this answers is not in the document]
bamsalem: We are not changing that, we are setting the value to be setted dynamically, instead using always 'bottom', we are using default value which is 'top' like `FieldLevelHelp` but if we wants it to be other value we just need to set the value explicitly.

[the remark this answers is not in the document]
hstastna: OK. Just a note that choosing the 'top' placement as the default is not clear to me. I don't see a reason to use it as the default because it's " like `FieldLevelHelp`". But I am okay with that because we probably want the placement in the top in most of the cases, in general :)

src/components/OverlayTooltip/index.js
```diff
@@ -1,14 +1,15 @@
...
 
-const OverlayTooltip = ({ id, tooltip, placement = 'left', children, ...rest }) => {
+const OverlayTooltip = ({ id, tooltip, placement, children, ...rest }) => {
   return (
     <OverlayTrigger
       overlay={
         <Tooltip id={id}>{tooltip}</Tooltip>
       }
-      placement={placement}
+      placement={placement || 'left'}
```
rszwajko, line 12: I see that in many places we need to force position `top` - perhaps it makes sense to change the default here?
How often we need placement `left`?

rszwajko (changes requested):
Some minor issues left (see below).
@sgratch  can you confirm that (hopefully temporarily) lost functionality is not a problem for us?

src/propTypeShapes.js
```diff
@@ -27,3 +27,8 @@ export const RouterPropTypeShapes = {
     goForward: PropTypes.func,
   }),
 }
+export const tooltipPropType = PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.oneOf([false])])
+
+export const requiredTooltipPropType = PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired
```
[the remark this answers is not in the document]
bamsalem: I agree that is overkill if the difference is only the isRequired, but in this case there is one more difference and
```
export const tooltipPropType = PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.oneOf([false])])
```
the difference is for not required tooltip false should be allowed if we want to use AND/OR statements : 
`prop={condition1 && value}`
otherwise it will throw an error like in this [case](https://github.com/oVirt/ovirt-web-ui/pull/1233#discussion_r496452236)

src/intl/messages.js
```diff
@@ -57,12 +57,11 @@ export const messages: { [messageId: string]: MessageType } = {
   bootMenuWarning: 'All changes will take effect after reboot only.',
   bootOrder: 'Boot Order',
   bootSequence: 'Boot Sequence',
-  bootSequenceTooltip: 'Device order in which VM will search OS for running.',
+  SelectTheBootableDeviceTooltip: 'Select the bootable devices in the order you would like them to boot up in.',
```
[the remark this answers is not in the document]
bamsalem: It changed according to the document.
[link to the section in the document.](https://docs.google.com/document/d/1ouQ7b4tKH1ksmB54hnPz69rkHcynv7qhG3hDXf342XA/edit?disco=AAAAKMl-SfI)

src/components/OverlayTooltip/index.js
bamsalem: Even not once, I didn't want to change the default value to avoid unnecessary errors, but after I checked that it seems like there is no even one place in the project that using the default value, I don't know what were the considerations to choose `left` but now the best option is absolutely `top`.
So I changed it to `top`.
Thanks :)

src/propTypeShapes.js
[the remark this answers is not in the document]
bamsalem: @rszwajko ?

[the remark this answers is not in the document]
rszwajko: Ok. Let's go this path and treat `false` | `undefined` | `'default'` the same.

src/components/OverlayTooltip/InfoTooltip/index.js
hstastna: @rszwajko @sgratch @sjd78 So are we OK with (hopefully temporarily) losing functionality using this approach?

src/propTypeShapes.js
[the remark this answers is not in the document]
hstastna: IMO creating these special types is unnecessary. We can specify these types directly in appropriate files, without importing anything. I would better use this approach in other cases, for more complicated proptypes specification.

src/components/VmDetails/cards/DisksCard/index.js
```diff
@@ -130,7 +130,7 @@ class DisksCard extends React.Component {
         idPrefix={idPrefix}
         icon={{ type: 'pf', name: 'storage-domain' }}
         title={msg.disks()}
-        editTooltip={msg.disksCardEditTooltip({ vmName: vm.get('name') })}
+        editTooltip={msg.edit()}
```
hstastna, line 133: Do we really want to see only "Edit" instead of 'Edit Disks for {vmName}'? I think this change is unnecessary.

src/components/VmDetails/cards/NicsCard/index.js
```diff
@@ -123,7 +123,7 @@ class NicsCard extends React.Component {
       <BaseCard
         icon={{ type: 'pf', name: 'network' }}
         title={msg.nic()}
-        editTooltip={msg.nicCardEditTooltip({ vmName: vm.get('name') })}
+        editTooltip={msg.edit()}
```
hstastna, line 126: Same as for `disksCardEditTooltip`. I think it's better to see in the tooltip what exactly we are going to edit.

src/components/VmDetails/cards/DisksCard/index.js
hstastna: Another thing is that you are removing `disksCardEditTooltip` message here and that makes `disksCardEditTooltip` useless, but I still can see it in the _messages.js_ [file](https://github.com/oVirt/ovirt-web-ui/pull/1289/files#diff-c7c4c4da7b77958d3e42b7c152709bffR220). Is there any reason for that?

hstastna (commented):
There are a few details and questions I have (see also [this comment](https://github.com/oVirt/ovirt-web-ui/pull/1289#discussion_r499503401)) but otherwise it looks ok to me :)

src/components/VmDetails/cards/DisksCard/index.js
bamsalem: > Do we really want to see only "Edit" instead of 'Edit Disks for {vmName}'? I think this change is unnecessary.
I stuck to the document:
https://docs.google.com/document/d/1ouQ7b4tKH1ksmB54hnPz69rkHcynv7qhG3hDXf342XA/edit?disco=AAAAKMHErgg
If there are any other instructions I'll change it accordingly

Regarding:
> Another thing is that you are removing `disksCardEditTooltip` message here and that makes `disksCardEditTooltip` useless, but I still can see it in the _messages.js_ [file](https://github.com/oVirt/ovirt-web-ui/pull/1289/files#diff-c7c4c4da7b77958d3e42b7c152709bffR220). Is there any reason for that?

I'll remove it thanks :)

src/components/VmDetails/cards/NicsCard/index.js
bamsalem: Like I mentioned [here](https://github.com/oVirt/ovirt-web-ui/pull/1289#discussion_r499566560), that were the document's instructions,
if we want to change anything let's make a meeting and discuss that.

sgratch: Changing disksCardEditTooltip and nicCardEditTooltip text was suggested by @lwrigh (yellow sections within  [Tooltips Document](https://docs.google.com/document/d/1ouQ7b4tKH1ksmB54hnPz69rkHcynv7qhG3hDXf342XA/edit#)) and I think it's ok since it's pretty obvious that we are editing those entities for the VM. Shorter info text is better.

sgratch:
> @sgratch can you confirm that (hopefully temporarily) lost functionality is not a problem for us?

@bamsalem @rszwajko @hstastna  Are we using PF4 Tooltip functionality on ui-extensions? it seems that this functionality (keyboard & accessibility support) is used only on Dashboard and not working on other PF4 dialogs. 
I guess we can wait with that here til it will be replaced to PF4, but I do agree that making our wrapper easy to be replaced by PF4 component in the future will be better. How complicated is to change the wrapper to be more easily replaced?

sgratch:
@bamsalem Few comments (a long list but easy to fix :):

1.  **details card "ip adresss" tooltip**:  why we added this tooltip? I'm not sure it's correct. In addiiton, the tooltip info icon alignment is wrong since it makes the field title to split into 2 lines.
2.  **details card "Total Virtual CPUs**" information icon - the tooltip info icon is not alligned with field title (should be like for "boot order" field). 
3.  **details card "Data Center**"  information icon - the tooltip info icon is not alligned with field title.
4.  **details card "Threads per Core**"  information icon - the tooltip info icon is not alligned with field title.
5. **details card "Cluster"** information icon - should be next to the text field and not next to the value field
6.  **details card "Change CD"** information icon - should be next to the text field and not next to the value field. in addition, the tooltip text is missing a dot at the end of the sentence. 
7.  **Utilization card: donut Chart** tooltips are ok as it is now, I would just replace the dash within the text with a colon (:). e.g. "used - 50%"  replaced to  "used: 50%"
8.  **pool entity card**: the tooltip for the pool name should be removed (has a wrong  BoxShadow shape anyway)
9.  **create vm-> networking step - create a new nic**: tooltips for the "select vnic profile" and for "device type" fields should be removed and anyway have a wrong BoxShadow shape.
10.  **create vm-> networking step - Edit a NIC** ("nic actions") tooltip has a wrong  BoxShadow shape
11.  **create vm->storage step - create a new disk** - tooltips for the "select storage domain" and for "disk type" fields should be removed and anyway have a wrong BoxShadow shap.
12. **create vm->storage step -Edit a disk** ("disk actions") tooltip has a wrong  BoxShadow shape

rszwajko:
> Are we using PF4 Tooltip functionality on ui-extensions? 

@sgratch  Yes. Check VM Migrate dialog: you can tab into the tooltip (it's focusable and opens automatically).
We also still use PF3 tooltips i.e. Cluster Upgrade dialog (upgrade options).

bamsalem:
> 1. **details card "ip adresss" tooltip**:  why we added this tooltip? I'm not sure it's correct. In addiiton, the tooltip info icon alignment is wrong since it makes the field title to split into 2 lines.

The `tooltip` have been there, the only thing that I added is the icon I thought that it will be more consistent with all the other info tooltips, but I'll remove the icon and it will look as it looks before the changes.
> 2. **details card "Total Virtual CPUs**" information icon - the tooltip info icon is not alligned with field title (should be like for "boot order" field).

Done.
> 3. **details card "Data Center**"  information icon - the tooltip info icon is not alligned with field title.
> 4. **details card "Threads per Core**"  information icon - the tooltip info icon is not alligned with field title.

I'm not sure if I understood you but if the meaning is like section 2 so it's done.
> 5. **details card "Cluster"** information icon - should be next to the text field and not next to the value field

I'm assuming that you meant to "Data Center" and if so, so it's done.
> 6. **details card "Change CD"** information icon - should be next to the text field and not next to the value field. in addition, the tooltip text is missing a dot at the end of the sentence.

image
That's how it was in the document, if you want me to change it I will.
Please let me know what you decided.
> 7. **Utilization card: donut Chart** tooltips are ok as it is now, I would just replace the dash within the text with a colon (:). e.g. "used - 50%"  replaced to  "used: 50%"

image
Done.
> 8. **pool entity card**: the tooltip for the pool name should be removed (has a wrong  BoxShadow shape anyway)

Done.
> 9. **create vm-> networking step - create a new nic**: tooltips for the "select vnic profile" and for "device type" fields should be removed and anyway have a wrong BoxShadow shape.
> 10. **create vm-> networking step - Edit a NIC** ("nic actions") tooltip has a wrong  BoxShadow shape
> 11. **create vm->storage step - create a new disk** - tooltips for the "select storage domain" and for "disk type" fields should be removed and anyway have a wrong BoxShadow shap.
> 12. **create vm->storage step -Edit a disk** ("disk actions") tooltip has a wrong  BoxShadow shape

Done.

@sgratch did I understood you correctly in sections 3-5?
And please let me know what you decided to do regarding section 6?

sgratch:
> > Are we using PF4 Tooltip functionality on ui-extensions?
> 
> @sgratch Yes. Check VM Migrate dialog: you can tab into the tooltip (it's focusable and opens automatically).
> We also still use PF3 tooltips i.e. Cluster Upgrade dialog (upgrade options).

it works for me on Chrome but on FF (ESR) the tab into tooltips for PF4 dialogs (like migrate VM) is not working.

sgratch:
> @bamsalem 
... I do agree that making our wrapper easy to be replaced by PF4 component in the future will be better. How complicated is to change the wrapper to be more easily replaced?
?

bamsalem:
The wrapper use all PF4 [`<Tooltip />`](https://www.patternfly.org/v4/components/tooltip) props and can be replaced easily:
https://github.com/oVirt/ovirt-web-ui/blob/26bb1bf7b9d80dd355f6372932757d96d4996e24/src/components/OverlayTooltip/InfoTooltip/index.js#L8-L17
IMO it can be easily replaced.

sgratch:
> > 1. **details card "ip adresss" tooltip**:  why we added this tooltip? I'm not sure it's correct. In addiiton, the tooltip info icon alignment is wrong since it makes the field title to split into 2 lines.
> 
> The `tooltip` have been there, the only thing that I added is the icon I thought that it will be more consistent with all the other info tooltips, but I'll remove the icon and it will look as it looks before the changes.

+1 Right, because this tooltip should appear only when ip value is N/A and in previous fix it always appear regardless to field value.

> 
> > 1. **details card "Total Virtual CPUs**" information icon - the tooltip info icon is not alligned with field title (should be like for "boot order" field).
> 
> Done.

+1

> 
> > 1. **details card "Data Center**"  information icon - the tooltip info icon is not alligned with field title.

+1

> > 2. **details card "Threads per Core**"  information icon - the tooltip info icon is not alligned with field title.
> 
> I'm not sure if I understood you but if the meaning is like section 2 so it's done.
> 
+1

> > 1. **details card "Cluster"** information icon - should be next to the text field and not next to the value field
> 
> I'm assuming that you meant to "Data Center" and if so, so it's done.

No, I meant cluster (for a running VM a tooltip appears).
> 
> > 1. **details card "Change CD"** information icon - should be next to the text field and not next to the value field. in addition, the tooltip text is missing a dot at the end of the sentence.
> 
> image
> That's how it was in the document, if you want me to change it I will.
> Please let me know what you decided.

Yes please. We decided that there will be no tooltips next to values (mentioned on issue) and a dot is required as well.
> 
> > 1. **Utilization card: donut Chart** tooltips are ok as it is now, I would just replace the dash within the text with a colon (:). e.g. "used - 50%"  replaced to  "used: 50%"
> 
> image
> Done.

+1

> 
> > 1. **pool entity card**: the tooltip for the pool name should be removed (has a wrong  BoxShadow shape anyway)
> 
> Done.

+1

> 
> > 1. **create vm-> networking step - create a new nic**: tooltips for the "select vnic profile" and for "device type" fields should be removed and anyway have a wrong BoxShadow shape.

+1

> > 2. **create vm-> networking step - Edit a NIC** ("nic actions") tooltip has a wrong  BoxShadow shape

-1 (wasn't fixed)

> > 3. **create vm->storage step - create a new disk** - tooltips for the "select storage domain" and for "disk type" fields should be removed and anyway have a wrong BoxShadow shap.

+1

> > 4. **create vm->storage step -Edit a disk** ("disk actions") tooltip has a wrong  BoxShadow shape
> 

-1 (wasn't fixed)

> Done.
> 
> @sgratch did I understood you correctly in sections 3-5?
> And please let me know what you decided to do regarding section 6?

@bamsalem please see inline comments above.

rszwajko:
> it works for me on Chrome but on FF (ESR) the tab into tooltips for PF4 dialogs (like migrate VM) is not working.

@sgratch  yep. The icon is not focusable on FF (even 80.0). We probably should fix that.

rszwajko:
@bamsalem   
Regarding "Change CD" label: it's now split into 2 lines which looks strange

@sgratch  
why we change the label "CD" -> "Change CD"?  If a "Name " field becomes editable we are not changing it to "Change name".
IMHO this could be confusing for the user.

rszwajko (approved):

sgratch:
@bamsalem please resolve conflicts

sgratch:
> > it works for me on Chrome but on FF (ESR) the tab into tooltips for PF4 dialogs (like migrate VM) is not working.
> 
> @sgratch yep. The icon is not focusable on FF (even 80.0). We probably should fix that.

A bug was opened for this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1889817

sgratch:
@bamsalem 
> > > 1. **create vm-> networking step - Edit a NIC** ("nic actions") tooltip has a wrong  BoxShadow shape
> 
> -1 (wasn't fixed)
>

Why did you remove the  createVmNetEditActions instead of changiing the BoxShadow shape?
> 
> > > 1. **create vm->storage step -Edit a disk** ("disk actions") tooltip has a wrong  BoxShadow shape
> 
> -1 (wasn't fixed)
> 

Same here?

sgratch:
@bamsalem @rszwajko 
> Regarding "Change CD" label: it's now split into 2 lines which looks strange
> 
It is still reproduced for me.

> why we change the label "CD" -> "Change CD"? If a "Name " field becomes editable we are not changing it to "Change name".
> IMHO this could be confusing for the user.

I don't know why it was like that and I totally agree. Changing that to "CD" on both view/edit modes will look better and will solve the split lines problem

sgratch (changes requested):
Please see latest comments.

hstastna (changes requested):
Regarding "Change CD" label, I suggest to have the same label ("CD") for both view and edit modes. It does not look good when it "jumps" from "CD" to "Change CD" after I click on edit the details of a VM (and back again, after saving the changes). It seems to me that this change comes from the Tooltips Document. If yes, then I suggest not to follow this document strictly and to leave the label as "CD". If you have a different opinion, we still can ask our actual UX designers for their opinion.

Another issue: I can see wrong aligning where the icon is added, in the VM Details Card edit mode (FF 80, 81, ESR, Chromium 85):
tooltip
 
Another aligning issue in the VM Details card:
 - This is on master and also in the Tooltips document:
a
 - This occurs with changes in this PR, the Total vCPUs number seems to be aligned wrong even in the view mode:
a2

Other tooltips seem ok to me :)

bamsalem:
@sgratch 
> @bamsalem
> 
> > > > 1. **create vm-> networking step - Edit a NIC** ("nic actions") tooltip has a wrong  BoxShadow shape
> > 
> > 
> > -1 (wasn't fixed)
> 
> Why did you totally remove the createVmNetEditActions tooltip instead of just changing the BoxShadow shape?
> 
> > > > 1. **create vm->storage step -Edit a disk** ("disk actions") tooltip has a wrong  BoxShadow shape
> > 
> > 
> > -1 (wasn't fixed)
> 
> Same here?

It wasn't appear in the documents and it seems to me like byproduct, so I removed it, if you want I can move it back
image

WDYT?

bamsalem:
@hstastna @sgratch 
I changed the message from "Change CD" to CD like @rszwajko suggested.
Regarding the issue @hstastna found I have a question how to fix it.
WDYT will be better solution?
The first option is:
Align to the same position but then long messages will wrap.
runningVm
offVm

The second option:
Using `white-space: nowrap` attributes, then long messages will wrap.
image

The third and the last (that I thought about) is:
Align the icons to the right
image
offVm

If you have any other suggestions feel free to share with me :)

hstastna:
**I am for the third option** because:
- the labels look nice and same as originally (_Total Virtual CPUs_ is on two lines - same as originally, for example),
- the icons are nicely aligned so I can find the relevant icon with the info I need very easily (see the pictures above)
- the whole Details card looks nice (labels, fields nicely aligned)

icon1
icon2

---

---

_Why I am NOT for the 2nd option:_
 - icons are not nicely aligned, and they could be
 - _Total Virtual CPUs_ label looks like it does not have enough space to breathe
second_option

_Why I am NOT for the 1st option:_
- long labels can take even three lines and it creates unnecessary unused extra space which does not look good
first_option

hstastna (commented):
"Change CD" changed to "CD" 👍🏽 

---

Regarding the other issue: 
**I am for the third option** because:
- the labels look nice and same as originally (_Total Virtual CPUs_ is on two lines - same as originally, for example),
- the icons are nicely aligned so I can find the relevant icon with the info I need very easily (see the pictures above)
- the whole Details card looks nice (labels, fields nicely aligned)

icon1
icon2

---

_Why I am NOT for the 2nd option:_
 - icons are not nicely aligned, and they could be
 - _Total Virtual CPUs_ label looks like it does not have enough space to breathe
second_option

_Why I am NOT for the 1st option:_
- long labels can take even three lines and it creates unnecessary unused extra space which does not look good
first_option

sgratch:
> @sgratch
> 
> > @bamsalem
> > > > > 1. **create vm-> networking step - Edit a NIC** ("nic actions") tooltip has a wrong  BoxShadow shape
> > > 
> > > 
> > > -1 (wasn't fixed)
> > 
> > 
> > Why did you totally remove the createVmNetEditActions tooltip instead of just changing the BoxShadow shape?
> > > > > 1. **create vm->storage step -Edit a disk** ("disk actions") tooltip has a wrong  BoxShadow shape
> > > 
> > > 
> > > -1 (wasn't fixed)
> > 
> > 
> > Same here?
> 
> It wasn't appear in the documents and it seems to me like byproduct, so I removed it, if you want I can move it back

@bamsalem I was referring on both for the "nic actions" and "disk actions" tooltips:
image

sgratch:
> "Change CD" changed to "CD" 👍🏽
> 
BTW, I'm not sure if this tooltip info text and appearance are correct at all.
I think that "CD can only be changed when the VM is running" should be replaced with "CD can only be changed when the VM is not running" and should be displayed for running vms only. Since changing a bootable CD can be applied only on a non running VM. For running VMs a restart is required.
@sjd78 any reason that you can remember for having it like this?

sgratch:
> 
> Regarding [another question/issue](https://github.com/oVirt/ovirt-web-ui/pull/1289#issuecomment-713515835):
> **I am for the third option** 
> 

I don't like this third option since the tooltips are closer to the field value instead of being located next to the field title and it looks strange. Usually the tooltip icon should be tie to the field label. 
BTW, which one of those options looks better after windows resizing and zooming in/out? Is the layout kept?

bamsalem:
@sgratch 
Regarding this [comment](https://github.com/oVirt/ovirt-web-ui/pull/1289#issuecomment-713702251):
I didn't understood you correctly, now I changed it and it's consist with the other tooltips.


> > "Change CD" changed to "CD" 👍🏽
> 
> BTW, I'm not sure if this tooltip info text and appearance are correct at all.
> I think that "CD can only be changed when the VM is running" should be replaced with "CD can only be changed when the VM is not running" and should be displayed for running vms only. Since changing a bootable CD can be applied only on a non running VM. For running VMs a restart is required.
> @sjd78 any reason that you can remember for having it like this?

From what I saw in the code and in the documentation users can change only for running VMs:
https://github.com/oVirt/ovirt-web-ui/blob/master/src/vm-status.js#L29-L31


> > Regarding [another question/issue](https://github.com/oVirt/ovirt-web-ui/pull/1289#issuecomment-713515835):
> > **I am for the third option**
> 
> I don't like this third option since the tooltips are closer to the field value instead of being located next to the field title and it looks strange. Usually the tooltip icon should be tie to the field label.
> BTW, which one of those options looks better after windows resizing and zooming in/out? Is the layout kept?

I agree with that but options one and three have advantages and disadvantages and we need to decide what we are willing to give up in the appearance of the card, @hstastna summed it up great in her [comment](https://github.com/oVirt/ovirt-web-ui/pull/1289#pullrequestreview-513624754).
IMO the 1st option is the best one.
Regarding the zoom in, when resizing the window the first and the third options keep their layout more than the second option which the label covers the value:
Second option:
second option
second option
Third option:
image

sgratch:
@bamsalem 

> Regarding this [comment](https://github.com/oVirt/ovirt-web-ui/pull/1289#issuecomment-713702251):

> I didn't understood you correctly, now I changed it and it's consist with the other tooltips.

LGTM


**Another thing**: Based on what we discussed offline, we should still change all current existed select menu tooltips :
image

to use the same box shadow shape. We decided not to remove them, just update the shadow shape to PF4 style:
image


> From what I saw in the code and in the documentation users can change only for running VMs:
> https://github.com/oVirt/ovirt-web-ui/blob/master/src/vm-status.js#L29-L31

Where did you find this on docs?

> IMO the 1st option is the best one.
> Regarding the zoom in, when resizing the window the first and the third options keep their layout more than the second option which the label covers the value.

Since we need to be consistent with other dialogs tooltips layout (e.g. create disk), let's stay with option 1 for now. 
We can open a separate issue for tooltips layout and discuss that with UX experts.

Anyway, due to this PR changes, fields layout is not aligned anymore when sizing is changed - so this needed to be fixed.
image
image

bamsalem:
> **Another thing**: Based on what we discussed offline, we should still change all current existed select menu tooltips :
> image
> 
> to use the same box shadow shape. We decided not to remove them, just update the shadow shape to PF4 style:
> image
> 
> >


> Anyway, due to this PR changes, fields layout is not aligned anymore when sizing is changed - so this needed to be fixed.
> image
> image

Fixed.
Now `SelectBox` tooltip is consistent with the other tooltips.
Regarding the misaligned fields, I wanted to discuss with you how to fix it, now it's aligned.

Screenshot:
image

sgratch:
@bamsalem LGTM except two remaining issues to handle:

1.  "The VM is down" tooltip box shadow should be updated:
image

2. The "threads per core" new tooltip on basic details step should be updated to PF4 style:
image

bamsalem:
I fixed those 2 issues:
image
image

The same issue was occurred in the VM list:
image
It's changed as well.

sgratch (approved):
LGTM

src/components/OverlayTooltip/index.js
```diff
@@ -17,8 +18,8 @@ const OverlayTooltip = ({ id, tooltip, placement = 'left', children, ...rest })
 }
 OverlayTooltip.propTypes = {
   id: PropTypes.string.isRequired,
-  tooltip: PropTypes.string.isRequired,
-  placement: PropTypes.string,
+  tooltip: requiredTooltipPropType,
```
sjd78, line 21: This should be inlined and referenced by other components as `Tooltip.propTypes.tooltip`

src/components/OverlayTooltip/index.js
```diff
@@ -17,8 +18,8 @@ const OverlayTooltip = ({ id, tooltip, placement = 'left', children, ...rest })
...
+  placement: tooltipPositionPropType,
```
sjd78, line 22: This should be inlined and referenced by other components as `Tooltip.propTypes.placement`

src/components/SelectBox.js
```diff
@@ -64,12 +65,14 @@ class SelectBox extends React.Component {
     return (
       <div style={{ width: '100%' }} id={id}>
         <div className='dropdown'>
-          <button className={`btn btn-default dropdown-toggle ${style['dropdown-button']} ${validationClass}`} type='button' data-toggle='dropdown' id={`${id}-button-toggle`}>
-            <span className={style['dropdown-button-text']} id={`${id}-button-text`} title={selectedItem && selectedItem.value}>
-              {selectedItem ? selectedItem.value : NOBREAK_SPACE}
-            </span>
-            <span className='caret' id={`${id}-button-caret`} />
-          </button>
+          <OverlayTooltip id={`${id}-selectbox-tooltip`} placement={'bottom'} tooltip={selectedItem ? selectedItem.value : ''}>
```
sjd78, line 68: I just really really want to drop this SelectBox and replace it with a PF-react version.

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -210,11 +210,12 @@ class Storage extends React.Component {
                 title='Bootable flag'
               />
             }
-            <FieldLevelHelp
-              content={this.bootableInfo(rowData.bootable)}
-              inline
-              className={style['disk-size-edit-label']}
-            />
+            <div className={style['bootable-info-tooltip']}>
```
sjd78, line 213: I'd rather see this on the `className` on `InfoToolTip` itself.  That would require `InfoToolTip` to accept an extra set of classNames it will append to its internal list.  But it would make this bit of code look a lot better.

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -260,10 +261,9 @@ class Storage extends React.Component {
               />
             </FormGroup>
             <span className={style['disk-size-edit-label']}>GiB</span>
-            <FieldLevelHelp
-              inline
-              content={msg.diskEditorSizeCreateHelp()}
-            />
+            <div>
```
sjd78, line 264: Is the wrapping `div` necessary?

src/components/VmActions/index.js
sjd78: The whole console loading button on the VM List and on VM Details is slated to be updated in the near future anyway.  No need to worry about it here.

src/components/VmDetails/BaseCard.js
```diff
@@ -160,8 +164,9 @@ BaseCard.propTypes = {
   editMode: PropTypes.bool,
   editable: PropTypes.bool,
   disableSaveButton: PropTypes.bool,
-  editTooltip: PropTypes.string,
-  disableTooltip: PropTypes.string,
+  editTooltip: tooltipPropType,
```
sjd78, line 167: After other changes:
```suggestion
  editTooltip: Tooltip.propTypes.tooltip,
```

src/components/VmDetails/cards/DetailsCard/FieldRow.js
```diff
@@ -1,24 +1,34 @@
...
+      }
     </Col>
     <Col cols={7} className={style['col-data']} id={id}>{children}</Col>
   </Row>
 )
 FieldRow.propTypes = {
   id: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
-  tooltip: PropTypes.string,
+  tooltip: tooltipPropType,
```
sjd78, line 29: Reference propType from the tooltip components

src/components/VmDetails/cards/DetailsCard/FieldRow.js
```diff
@@ -1,24 +1,34 @@
...
   children: PropTypes.node.isRequired,
+  tooltipPosition: tooltipPositionPropType,
```
sjd78, line 31: Reference propType from the tooltip components

src/components/VmDetails/CardEditButton.js
```diff
@@ -72,12 +70,13 @@ class CardEditButton extends React.Component {
   }
 }
 CardEditButton.propTypes = {
-  tooltip: PropTypes.string,
+  tooltip: tooltipPropType,
```
sjd78, line 73: Reference propType from the tooltip components

src/components/VmDetails/CardEditButton.js
```diff
@@ -72,12 +70,13 @@ class CardEditButton extends React.Component {
...
   editEnabled: PropTypes.bool,
   editable: PropTypes.bool,
-  disableTooltip: PropTypes.string,
+  disableTooltip: tooltipPropType,
```
sjd78, line 76: Reference propType from the tooltip components

src/components/VmDetails/CardEditButton.js
```diff
@@ -72,12 +70,13 @@ class CardEditButton extends React.Component {
...
   id: PropTypes.string.isRequired,
   onClick: PropTypes.func,
+  placement: tooltipPositionPropType,
```
sjd78, line 79: Reference propType from the tooltip components

src/components/VmDetails/BaseCard.js
```diff
@@ -160,8 +164,9 @@ BaseCard.propTypes = {
...
+  disableTooltip: tooltipPropType,
```
sjd78, line 168: Reference propType from the tooltip components

src/components/VmDetails/BaseCard.js
```diff
@@ -160,8 +164,9 @@ BaseCard.propTypes = {
...
+  editTooltipPlacement: tooltipPositionPropType,
```
sjd78, line 169: Reference propType from the tooltip components

src/components/VmDetails/cards/DetailsCard/FieldRow.js
```diff
@@ -1,24 +1,34 @@
...
+import { tooltipPropType, tooltipPositionPropType } from '_/propTypeShapes'
 
-const FieldRow = ({ label, children, id, tooltip }) => (
+const FieldRow = ({ label, children, id, tooltip, tooltipPosition }) => (
   <Row className={style['field-row']}>
     <Col cols={5} className={style['col-label']}>
       <span>{label}</span>
-      {tooltip && <FieldLevelHelp buttonClass={style['field-level-help']} disabled={false} content={tooltip} inline />}
+      {tooltip &&
+        <span>
```
sjd78, line 14: Span required?

src/components/VmDetails/cards/DisksCard/DiskStateIcon.js
```diff
@@ -12,19 +12,19 @@ const diskStateSettings = {
     type: 'fa',
     name: 'arrow-circle-o-up',
     className: style['state-icon-active'],
-    tooltip: msg.diskStateActiveTooltip(),
+    tooltip: `${msg.status()}: ${msg.diskStateActiveTooltip()}`,
```
sjd78, line 15: Nit-pick:  Not all languages will naturally go "A: B" so it is better form (but not required) to embed the tooltip value in the status string:

```js
  `${msg.status({ statusInfo: msg.diskStateActiveTooltip() })}`
```

Then the string in messages.js would look like:
```
  status: "Status: {statusInfo}"
```

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
```diff
@@ -121,9 +122,10 @@ class SnapshotItem extends React.Component {
           key='info'
         >
           <a id={`${this.props.id}-info`}>
-            <OverlayTooltip placement={this.state.isMobile ? 'right' : 'left'} id={`${this.props.id}-info-tt`} tooltip={msg.details()}>
-              <Icon type='pf' name='info' />
-            </OverlayTooltip>
+            <InfoToolTip
```
sjd78, line 125: Ok, with all the placements removed, does the component auto-align based on available space?  Have you tested it with a narrow/phone screen?

sjd78 (changes requested):
I would like to see the tooltip components refactored such that:

  1. They are both located in `src/components/tooltips`, specifically:
  * `src/components/tooltips/index.js` <-- reference and export the "custom" components we're using (see https://github.com/patternfly/patternfly-react/blob/patternfly-3/packages/patternfly-react/src/components/Dropdown/index.js for an example of how patternfly-react3 does it)
  * `src/components/tooltips/Tooltip.js` <-- renamed/moved `OverlayTooltip/index.js`
  * `src/components/tooltips/InfoTooltip.js` <-- renamed/moved `InfoTooltip/index.js`
  * `src/components/tooltips/style.css` <-- renamed/moved `InfoTooltip/style.css`


  2. They are imported like how patternfly-react components are imported:
```jsx
import { Tooltip, InfoTooltip } from '_/components/tooltips'
```


  3. The `propTypes` are baked in to the components, and if they need to be reused, the component's propTypes are referenced directly.

All of the other tooltip text work looks good.  Mostly just code changes from me.  Other comments inline.

src/intl/messages.js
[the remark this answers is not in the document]
sjd78: Also, the message name is wrong and sorted in the wrong place - it shouldn't be capitalized.

src/intl/messages.js
```diff
@@ -669,14 +662,16 @@ export const messages: { [messageId: string]: MessageType } = {
...
   vncConsole: 'VNC Console',
-  vncConsoleOpen: 'Open VNC Console',
   vncConsoleBrowser: 'VNC Console (Browser)',
-  vncConsoleBrowserOpen: 'Open VNC Console (Browser)',
   vnicProfile: 'VNIC Profile',
   vnicProfileEmpty: '<Empty>',
   yearsShort: 'y',
   yes: 'Yes',
   youHaveNoAllowedVnicProfiles: 'You cannot create or edit NICs because you do not have permission to use any vNIC Profiles in the VM\'s Data Center.',
+  sortAToZ: 'Sort from A to Z',
```
sjd78, line 671: Let's keep the keys in this object sorted A-Z properly please.

src/intl/messages.js
```diff
@@ -206,7 +204,7 @@ export const messages: { [messageId: string]: MessageType } = {
   },
   diskEditorResizeLabel: 'Increase Size By (GiB)',
   diskEditorSizeCantChangeHelp: 'Disk size cannot be extended for this type of disk.',
-  diskEditorSizeCreateHelp: 'After you create a disk, you may only extend its size if you make edits.',
+  diskEditorSizeCreateInfoTooltip: 'After creating a disk, you can only extend the size of it. The size of it will not be able to be decreased.',
```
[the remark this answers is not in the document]
bamsalem: My assumption was the document is verified by UX experts including messages, next time I won't assume this kind assumptions.

src/components/SelectBox.js
bamsalem: Let's create separate issue for that,
would you like to open an issue for that and attach link to the PF component that you wants to replace this component?

src/components/CreateVmWizard/steps/Storage.js
bamsalem: I found collisions between the css attributes of the cell and  the PF icon so the shortest way without create any regressions was to wrap i with div, if you have any better suggestion please let me know.

src/components/VmDetails/cards/DetailsCard/FieldRow.js
bamsalem: The PF icon need to be wrap in `<div>` or `<span>` in some cases (like the `<FieldLevelHelp />` component)

src/components/VmDetails/cards/DisksCard/DiskStateIcon.js
bamsalem: Thanks :)

src/intl/messages.js
bamsalem: Can you you explain what you meant here?
do you mean to change the message to: "Sorted A-Z"?

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
bamsalem: no the component isn't auto align based on available space, like PF3 component:
image
do we want to edit the PF3 components?

src/intl/messages.js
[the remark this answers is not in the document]
sjd78: ...even they get things wrong and use awkward phrasing sometimes!

src/components/SelectBox.js
sjd78: It would need to be a separate issue for sure.  I was just complaining.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
sjd78: Narrow screens are more like best effort.  The Snapshot item list tooltips had handling for narrow screens that was removed.  So, if specifically those tooltips look ok with that code removed on a narrow screen, it's ok.  We just don't want to loose functionality.

src/intl/messages.js
sjd78: No.  The keys in the messages object are sorted alpha (i.e. A-Z) in the file.  `sortAToZ` should be in the right place, not below the "y" keys.

src/components/VmDetails/cards/SnapshotsCard/SnapshotItem.js
bamsalem: It looks OK:
image

src/intl/messages.js
bamsalem: Oh I've never noticed that, Done.

sjd78 (changes requested):
Good work!  Only a handful of minor tweaks remain.

src/components/VmDetails/CardEditButton.js
```diff
@@ -40,44 +38,44 @@ class CardEditButton extends React.Component {
...
           <Icon type='pf' name='edit' />
         </a>
-      </OverlayTooltip>
+      </Tooltip>
     )
   }
 }
 CardEditButton.propTypes = {
-  tooltip: PropTypes.string,
+  tooltip: PropTypes.oneOfType(Tooltip.propTypes.tooltip),
```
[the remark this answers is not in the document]
bamsalem: Yes the tooltip is required for the `Tooltip` component.

sjd78:
@bamsalem - I think you forgot to push your last set of changes.

bamsalem:
> @bamsalem - I think you forgot to push your last set of changes.

@sjd78 You are right, I forgot to push, please take a look now.

sjd78 (approved):
LGTM

isaranova:
LGTM

Issue #1264, opened 2020-08-05
Create VM Wizard - support advanced CPU topology setting on "Basic Settings" step


sgratch:
Currently the user can only update advanced CPU topology (i.e. Virtual Sockets, Virtual Cores, Virtual Threads) for an existing VM via the Details VM card.
This should be also supported for a new VM creation.

We can consider adding that as a hidden "Advanced Options" section within the  "Basic Settings" step, the same as appears on the Details VM card.

hstastna:
From the design point of view, I can see more possibilities, how to implement this feature in the Create VM Wizard. Two of them look like this:

1)
![adv_options1a](https://user-images.githubusercontent.com/13417815/89775357-e9ae7500-db07-11ea-9f02-8c678ce54eb5.png)
![adv_options1b](https://user-images.githubusercontent.com/13417815/89775363-ec10cf00-db07-11ea-997b-bf62b6fd0562.png)
With this design above we would achieve consistency with the appropriate options already in the VM details page. However, I think it does not look very well in the wizard.

2)
![adv_options2a](https://user-images.githubusercontent.com/13417815/89775367-f03cec80-db07-11ea-8694-f86f06500130.png)
![adv_options2b](https://user-images.githubusercontent.com/13417815/89775373-f29f4680-db07-11ea-90ea-699d8d2e4c81.png)

@yfrimanm WDYT? IMHO, the second of my suggestions is more consistent with other extra options in the _Basic step_ of the wizard. But maybe we want some combination of those two options above.. ? Thanks!

yfrimanm:
Hi @hstastna@redhat.com,
IMHO the 1st option is much better because 'Advanced options' are
usually shown in a collapse/expand mode as you suggest there.
The 2nd option of showing it in a checkbox is kind of odd because a
checkbox is not a CTA (call to action|).
Yifat

hstastna:
@yfrimanm Thanks very much, Yifat, for the reply. 

However the 1st option is better and a CTA, I think it's not very nice. Maybe some changes regarding aligning text or the size of the drop downs? To make it look more consistent with the rest of the settings. I would like to know your opinion. Thanks a lot!

yfrimanm:
Hi  @Hilda Stastna <hstastna@redhat.com>
I think that if you will keep the text alignment to the right, as shown in
the first part of the form, it will make it look better.
Yifat

matthewcarleton:
[...]

PR #1290, opened 2020-09-11
Increase spacing between Boot Order and VCPU Topology sections in VM Details card

hstastna:
Related comment: https://github.com/oVirt/ovirt-web-ui/issues/1264#issuecomment-676308058

In this PR, I've increased the space between the two sections which are part of the _Advanced Options_ of a VM _Details_ card. I've chosen a simple solution by using `margin-top` and `margin-bottom` css properties, applied on the element with the name of the second section of the second column in Advanced Options (_VCPU Topology_).

This change improves the visual feeling, separates the two sections, so the user can easily distinguish which options belong to which section.

**Before:**
space_before

**After:**
space_after

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -911,7 +911,7 @@ class DetailsCard extends React.Component {
                         />
                       </FieldRow>
                       {/* VCPU Topology */}
-                      <Row className={style['field-row']}>
+                      <Row className={style['field-row-divide']}>
                         <Col cols={12} className={style['col-label']}>
                           <div>
                             <span>{msg.vcpuTopology()}</span>
--- a/src/components/VmDetails/cards/DetailsCard/style.css
+++ b/src/components/VmDetails/cards/DetailsCard/style.css
@@ -17,6 +17,11 @@
   margin-top: 5px;
 }
 
+.field-row-divide {
+  margin-top: 20px;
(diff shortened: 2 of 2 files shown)


bamsalem (approved):
I'm not sure if it's the best solution, but LGTM.
great work :)

hstastna:
@bamsalem You always can suggest a better solution ;)

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -911,7 +911,7 @@ class DetailsCard extends React.Component {
                         />
                       </FieldRow>
                       {/* VCPU Topology */}
-                      <Row className={style['field-row']}>
+                      <Row className={style['field-row-divide']}>
```
sjd78, line 914: I suggest a second class to make the margin adjustments without loosing the `field-row` styles.

Something like `className={`${style['field-row']} ${style['group-divider']}`}`.

Maybe it looks more complex, but I think it's a bit more flexible.

Thoughts?

sjd78 (changes requested):
See comment for suggestion.

src/components/VmDetails/cards/DetailsCard/index.js
hstastna: I think the original and also this solution - both are okay.

sjd78 (approved):
LGTM

PR #1291, opened 2020-09-14
Extract Create VM Wizard BasicSettings changes handlers

bamsalem:
Hi guys, due to @michalskrivanek [comment](https://github.com/oVirt/ovirt-web-ui/pull/1265#issuecomment-690296071)

> In general you should separate the actual fix and refactoring to separate PRs. We don't mix unrelated changes

I separated the refactoring and the actual fix to dseperated PRs.
in this PR I extracted the changes handlers to external file.
PR #1265 depended on this PR and should merge after this PR.

### DIFF
--- a/src/components/CreateVmWizard/helpers.js
+++ b/src/components/CreateVmWizard/helpers.js
@@ -1,0 +1,117 @@
+import { createTemplateList } from '_/components/utils'
+import timezones from '_/components/utils/timezones.json'
+import { isWindows } from '_/helpers'
+
+const handleClusterIdChange = (clusterId, blankTemplateId, { templates, clusters, defaultValues, operatingSystems, defaultGeneralTimezone, defaultWindowsTimezone }) => {
+  let changes = {}
+  const templateList = clusterId ? createTemplateList(templates, clusterId) : []
+  changes.dataCenterId = clusterId ? clusters.getIn([clusterId, 'dataCenterId']) : undefined
+  changes.clusterId = clusterId
+  changes.provisionSource =
+    (templateList.length === 1 && templateList[0].id === blankTemplateId)
+      ? 'iso'
+      : 'template'
+  changes.templateId = templateList.length === 1 ? blankTemplateId : undefined
+  changes = { ...changes, ...handleProvisionSourceChange(changes.provisionSource, defaultValues, { defaultGeneralTimezone, defaultWindowsTimezone, templates, operatingSystems }) }
+  return changes
+}
+
+const handleProvisionSourceChange = (provisionSource, defaultValues, { defaultGeneralTimezone, defaultWindowsTimezone, templates, operatingSystems }) => {
+  const changes = {}
+  changes.provisionSource = provisionSource
+  changes.isoImage = undefined
+  changes.templateId = undefined
+  changes.operatingSystemId = defaultValues.operatingSystemId
+  changes.memory = defaultValues.memory
+  changes.cpus = defaultValues.cpus
+  changes.topology = defaultValues.topology
+  changes.optimizedFor = defaultValues.optimizedFor
+  changes.cloudInitEnabled = defaultValues.initEnabled
+
+  // when changing Provision type to ISO, set the default time zone according to the OS
+  if (changes.provisionSource === 'iso') {
+    changes.timeZone = checkTimeZone(changes.operatingSystemId, changes.templateId, defaultGeneralTimezone, defaultWindowsTimezone, { templates, operatingSystems })
+    changes.initTimezone = '' // reset the sysprep timezone value, we don't support cloud-init TZ yet
+  }
+  return changes
+}
+
+const handleTemplateIdChange = (templateId, optimizedFor, { templates, operatingSystems, defaultGeneralTimezone, defaultWindowsTimezone }) => {
+  const changes = {}
+  const template = templates.find(t => t.get('id') === templateId)
+  changes.templateId = template.get('id')
+
+  changes.memory = template.get('memory') / (1024 ** 2) // stored in bytes, input in Mib
+  changes.cpus = template.getIn([ 'cpu', 'vCPUs' ])
+  changes.topology = template.getIn([ 'cpu', 'topology' ]).toJS()
+  changes.optimizedFor = template.get('type', optimizedFor)
+  changes.operatingSystemId = operatingSystems
+    .find(os => os.get('name') === template.getIn([ 'os', 'type' ]))
+    .get('id')
+
+  // Check template's timezone compatibility with the template's OS, set the timezone corresponding to the template's OS
+  changes.timeZone = checkTimeZone(changes.operatingSystemId, changes.templateId, defaultGeneralTimezone, defaultWindowsTimezone, { templates, operatingSystems })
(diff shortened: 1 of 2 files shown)


ovirt-infra:

hstastna (commented):
> I separated the refactoring and the actual fix to dseperated PRs.

IMO it is not separated. I can only see a new PR which moves some logic in some other place (which I consider unnecessary). I don't see a PR which would contain only the fix for selecting automatically single option menu.

I suggest to separate the fix and refactoring part the way that we will able to merge the fix first because the bug is annoying, the fix is important and I assume just a few code changes. And then we have more time for refactoring. It takes more time to review and test the refactoring changes you suggest which touch many scenarios, timezone related changes etc. We have to be careful to prevent creating new bugs when refactoring. Mixing refactoring stuff together with some fix is risky.

bamsalem:
Hi @hstastna ,
I just pushed changes to #1265 .
Now you can see the differences. first in this PR I extracted the exact code to external file and now we can use it in another components without code duplicates, for each change in `BassicSettings` there is a logic which fields should change and what should be the new values of the fields according to that change.
We wants that the next change will behave in the same way in every component (`CreateVmWizard`, `BasicSettings`...) without change it in many places.
So IMHO it requires create one "root" that all the component should use that root, I agree that there are no logic changes in this PR but that what I meant to do, separate the fix (the logic which included in #1265) from the refactoring (the extraction of code to external file without any change).
After this PR will be merged we can use the extracted handlers in CreateVmWizard too.
maybe I didn't understand @michalskrivanek 's comment?

hstastna:
What I can see here in this PR is the refactoring change. Okay, we can discuss this proposal, but let's fix the issue first and then we will have more time dealing with refactoring stuff. It's much easier and faster to review the fix comparing to not so small refactoring changes with the impact on lots of functionality in the whole _Basic settings_ step of the wizard. And it's better to separate those things to prevent creating new bugs and to simplify reviewing the changes and testing. I think that separating fix from the refactoring was meant to create a PR containing only the fix without any (necessary/significant) refactoring related changes + another separate follow-up PR with the refactoring stuff. What I can see in the https://github.com/oVirt/ovirt-web-ui/pull/1265 is both refactoring and the fix.

bamsalem:
Hi @hstastna actually it's not a refactoring it's a code extraction, I extracted the code from BassicSettings.js so we can use it from another components, you can see there are no changes in the code.
The fix should use this code, to change cluster ,provision source and iso or template there are procedures, those procedures are already implemented in BasicSetting.js in the function `handleChange`.
If the procedures are in method inside `BassicSettings`  those procedures can be used only from `BasicSettings` component, to trigger the auto select feature (the fix) we need to identify that there is only one cluster and you can do it in 2 places (if you think there are more options feel free to suggest them to me)
1. in BasicSettings -> render - it's a bad practice you don't update state/props from render.
2. getInitialState - when we creating  the CVMW state we can identify that there is only one cluster and select him.

how we select cluster? it's already implemented in BasicSettings so I have two options copy or extract it to another file so we can access it from both components.
So I separated the extraction from the actual fix, I can't fix it in reasonable way without extract the code first, WDYT?

rszwajko:
> another separate follow-up PR with the refactoring stuff

In this case the refactoring is needed **before** we can do the fix.

> What I can see in the #1265 is both refactoring and the fix.

I think it's github limitation (@sjd78 can you confirm?). In gerrit you can create patches that depend on other patches without including the commits from those patches.  However it's just minor inconvenience - the problem will be solved after merging this patch. Then we could rebase  #1265 .

>  it's not a refactoring it's a code extraction,

For me it's both:)

rszwajko (approved):

src/components/CreateVmWizard/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+  changes.cloudInitEnabled = defaultValues.initEnabled
+
+  // when changing Provision type to ISO, set the default time zone according to the OS
+  if (changes.provisionSource === 'iso') {
+    changes.timeZone = checkTimeZone(changes.operatingSystemId, changes.templateId)
+  }
+  return changes
+}
+
+const handleTemplateIdChange = (templateId, optimizedFor, { templates, operatingSystems, defaultGeneralTimezone, defaultWindowsTimezone }) => {
```
hstastna, line 46: Do we have to get the `defaultGeneralTimezone` and `defaultWindowsTimezone` like this, as function parameters? Those values are always the same. It would be nice if it was possible and if we had those values available in the whole helper, so we could simplify the code (not only) here.

hstastna (changes requested):
If we want to move this way regarding the "selecting automatically single option menu" issue, it would be good to add some more explanation. Could you, please, explain for example the reason why we want to extract only specific parts of the code and not moving the whole `handleChange`? In other words, let's make sure that this is exactly what we want. Thanks! :)

src/components/CreateVmWizard/helpers.js
rszwajko: Both values originate from `props` and in theory can change with every render.
We could use some tricks to make the param list look shorted:
1.  continue destructuring in  the param list but pack unused params using rest pattern i.e. ` {templates, ...otherProps}`
     Then we can restore the object using spread operator if needed i.e.  `const restoredProps = { templates, ...otherProps}`
2.  don't do destructuring in the param list. Move it to the body and take only what you need

rszwajko: Both approaches assume that we don't care about getting more properties then needed - destructuring and restoring has a (nice) side effect of filtering out alien properties.

bamsalem:
> If we want to move this way regarding the "selecting automatically single option menu" issue, it would be good to add some more explanation. Could you, please, explain for example the reason why we want to extract only specific parts of the code and not moving the whole `handleChange`? In other words, let's make sure that this is exactly what we want. Thanks! :)

@hstastna the `handleChange` function includes more than just handlers, there are few checks in the beginning and in the end of the function:
https://github.com/oVirt/ovirt-web-ui/blob/eece1da218a66cbce05670e09b9918d8f2fa5299/src/components/CreateVmWizard/steps/BasicSettings.js#L183-L191
https://github.com/oVirt/ovirt-web-ui/blob/eece1da218a66cbce05670e09b9918d8f2fa5299/src/components/CreateVmWizard/steps/BasicSettings.js#L231-L234

those checks and validations are irrelevant to the other scenario.
I wanted to extract only the "recipe" how to handle clusterId change and etc...
WDYT? IMHO extract the  whole function is unnecessary.

hstastna:
Ok, so let's move on, resolve the conflicts. Regarding so many function parameters, really, `defaultGeneralTimezone`, `defaultWindowsTimezone` are always the same, maybe we could make some constants from them. @rszwajko @sjd78 WDYT?

bamsalem:
> Ok, so let's move on, resolve the conflicts. Regarding so many function parameters, really, `defaultGeneralTimezone`, `defaultWindowsTimezone` are always the same, maybe we could make some constants from them. @rszwajko @sjd78 WDYT?

@hstastna after I discussed with @sjd78 about your suggestion,
I understand that the best way to use state's values in helpers to pass them,
I agree that we are using a lot of parameters but we are destructing them from props.

PR #1292, opened 2020-09-16
Remove react-intl and related code

sjd78:
The react-intl package was used and initialized has never been
used directly.  Removing it will save complexity and bundle space.

The react-intl-po package is used to interface the js/json files to
zanata compatible po/pot files.


rszwajko (dismissed):
PR #1088 depends on it to provide hot-reloading of locales - see https://github.com/oVirt/ovirt-web-ui/pull/1088/files#diff-d3ad4834e3d3b4cc20c0e78de755b73f
Of course this could be re-implemented (again) however this requires some time.

My second concern is about the direction we are heading with translation: some recognized framework or our own solutions. In the long run I would rather choose react-intl (or similar framework).

hstastna (commented):
I am for removing react-intl, too. It makes sense to me because using it would require more changes, incl. changes for Admin portal, as we already discussed with @sgratch.

@sjd78 I still can see `"react-intl": "^2.1.2",` in _node_modules/babel-loader/package.json_ in `devDependencies`. I know that packages under devDependencies are packages that are only needed for local development and testing. So it seems to be ok. Just wanted to make sure. WDYT? Otherwise LGTM. Thanks! :)

sjd78:
> @sjd78 I still can see `"react-intl": "^2.1.2",` in _node_modules/babel-loader/package.json_ in `devDependencies`. I know that packages under devDependencies are packages that are only needed for local development and testing. So it seems to be ok. Just wanted to make sure. WDYT? Otherwise LGTM. Thanks! :)

TL;DR - It's not a problem.

Dependencies of dependencies don't really matter.  Doubly so for devDependencies of dependencies.  What matters most is what is in `yarn.lock`.  If you remove, or upgrade, or add a library in package.json, a `yarn install` will update `yarn.lock` with the necessary packages and versions.  `react-intl` got removed after I took it off the dependency list.  We're good to go.

hstastna (approved):

sjd78:
> PR #1088 depends on it to provide hot-reloading of locales - see https://github.com/oVirt/ovirt-web-ui/pull/1088/files#diff-d3ad4834e3d3b4cc20c0e78de755b73f
> Of course this could be re-implemented (again) however this requires some time.
>

WIth refactoring the init side of things, bridging that particular gap shouldn't be difficult.

>
> My second concern is about the direction we are heading with translation: some recognized framework or our own solutions. In the long run I would rather choose react-intl (or similar framework).
>

I don't foresee us investing time in converting from our current message infra to full react-intl.  We're not doing anything particularly complicated.  What we have is adequate, and is consistent with ovirt-engine-ui-extensions and similar to GWT webadmin i18n techniques.

hstastna (approved):
LGTM.

rszwajko (approved):

PR #1295, opened 2020-09-16
Refactor DetailsCard component to use explicit config values

hstastna:
Small refactoring in the _DetailsCard_ react component to use the explicit config values accessed in the `connect` function
to help keep them in one place and to simplify the `handleChange` function a little bit.


sjd78 (approved):
LGTM

Issue #1258, opened 2020-07-20
Buttons "Save", "Cancel" in the Create VM Wizard move when clicking on them repeatedly


hstastna:
"Save" (blue-white tick) and "Cancel" buttons "jump" in the _Networking_ step when clicking on them repeatedly while the name remains invalid.

This issue is visible in the UI due to adding NIC names validation in https://github.com/oVirt/ovirt-web-ui/pull/1219 but it was already present before this PR was created. This issue is present also in the _Storage_ step of the Create VM Wizard with a slightly different scenario (see the second variant of the steps to reproduce). Note that there are some more and very similar scenarios to reproduce this issue.

This issue was also mentioned [here](https://github.com/oVirt/ovirt-web-ui/pull/1219#pullrequestreview-451562249).

---

**Steps to reproduce 1:** (require https://github.com/oVirt/ovirt-web-ui/pull/1219)
1. Start creating a new VM using _Create VM Wizard_, fill in the _Basic Settings_, move to the next step 
2. In the _Networking_ step, create or edit some NIC
3. Change the NIC's name to something invalid (for example add spaces or use more than 50 chars)
4. Click on "Save" (tick) button
=> the error message appears, all the buttons seem to be ok
5. Try some another NIC's name, but choose something invalid again
6. Click on "Save" (tick) button again
=> the error message remains displayed, but the buttons "Save", "Cancel" have moved down from their original position:
![2](https://user-images.githubusercontent.com/13417815/87937542-38329b80-ca95-11ea-82f5-7c1068b610dd.png)

**Steps to reproduce 2:** (second variant, already present on master branch)
1. Start creating a new VM using _Create VM Wizard_, move to the _Storage_ step
2. Create some Disk and save it
3. Try to edit the disk, click on _Edit_ in the kebab menu, see where exatly the "Save" and "Cancel" buttons are placed:
![edit_1](https://user-images.githubusercontent.com/13417815/87964739-f3b9f680-caba-11ea-8466-8ea865a839c8.png)
4. Click into the _Disk Name_ or _Size_ or field
5. Click anywhere out of the fields
=> the buttons moved up!
![edit_2](https://user-images.githubusercontent.com/13417815/87965083-7fcc1e00-cabb-11ea-8383-2cfd8f0282d4.png)

hstastna:
**Update:**

[...]

PR #1300, opened 2020-09-30
CreateVmWizard Networking step - Fix saving new nic error

bamsalem:
Fixes: #1299 .
The previous syntax includes bug.
I changed Iterator's syntax, now vnic saved successfully

PR #1301, opened 2020-09-30
TableInlineEditRow -  set button position using row position and fixed row height

bamsalem:
**Related issue:** #1258
**Depends on**: https://github.com/oVirt/ovirt-web-ui/pull/1300 (MERGED)

>"Save" (blue-white tick) and "Cancel" buttons "jump" in the Networking step when clicking on them repeatedly while the name remains invalid.

few aliases for the explanation:
Origin: position (x = 0, y = 0)

Before the change we used `rowDimensions.bottom` as the height position (which is the distance from the origin to the Table's row bottom, in Y axis) that value might change in case of error message appeared or in cases we need to include more components in Table's  row.
In this PR we are using the `rowDimensions.y + rowHeight`.
`rowDimensions.y ` is the distance from the origin to the top of the Table's row, in Y axis that probably won't change, `rowHeight` is a fixed value which represents the edited row's height, in case there are no errors etc.. .

I Also changed the "interface" of the TableInlineEditRow component to create a way to set `rowHeight` value dynamically, if you think that is unnecessary tell me and I'll remove it.

Before:
Before

After:
After

please ignore the fact that we can't save vnics it will be fixed in #1300 .

### DIFF
--- a/src/components/CreateVmWizard/steps/_TableInlineEditRow.js
+++ b/src/components/CreateVmWizard/steps/_TableInlineEditRow.js
@@ -1,25 +1,30 @@
-import React from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import TableConfirmButtonsRow from 'patternfly-react/dist/js/components/Table/TableConfirmButtonsRow'
-
-const TableInlineEditRow = props => {
-  const buttonsPosition = (window, rowDimensions) => {
+class TableInlineEditRow extends Component {
+  constructor (props) {
+    super(props)
+    this.buttonsPosition = this.buttonsPosition.bind(this)
+    this.rowHeight = undefined
+  }
+  buttonsPosition (window, rowDimensions) {
     const position = {}
+    const modalDomElement = document.querySelectorAll('div.fade.in.modal')
+    const scrolledInPx = modalDomElement.length === 1 ? modalDomElement[0].scrollTop : 0
 
-    if (props.last) {
-      position.bottom = window.height - rowDimensions.top - 1
-    } else {
-      position.top = rowDimensions.bottom
+    if (!this.rowHeight || Math.abs(this.rowHeight - rowDimensions.height) < 10) {
+      this.rowHeight = rowDimensions.height
     }
+
+    position.top = rowDimensions.y + this.rowHeight + scrolledInPx
     position.right = 75 // window.width - rowDimensions.right + 10
 
     console.info('button position', position)
     return position
   }
-
-  const buttonsClassName = props.last ? 'top' : 'bottom'
-
-  return <TableConfirmButtonsRow {...props} buttonsPosition={buttonsPosition} buttonsClassName={buttonsClassName} />
+  render () {
+    return <TableConfirmButtonsRow {...this.props} buttonsPosition={this.buttonsPosition} buttonsClassName={'bottom'} />
+  }
 }
 
 TableInlineEditRow.shouldComponentUpdate = true


ovirt-infra:

hstastna:
This PR:

Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1258
**Depends on**: https://github.com/oVirt/ovirt-web-ui/pull/1300

hstastna:
@bamsalem I've updated the issue you are thying to fix: https://github.com/oVirt/ovirt-web-ui/issues/1258#issuecomment-702686343

hstastna (dismissed):
The scenario 1 from the issue successfully fixed, LGTM. However the [another scenario](https://github.com/oVirt/ovirt-web-ui/issues/1258#issuecomment-702686343) isn't fixed yet.

Regarding the `TableInlineEditRow` component, I think that setting `rowHeight` dynamically is unnecessary. I wouldn't add this until we will have a use for it.

bamsalem:
> This PR:
> 
> Fixes: #1258
> **Depends on**: #1300

@hstastna Thanks for that.
> Regarding the TableInlineEditRow component, I think that setting rowHeight dynamically is unnecessary. I wouldn't add this until we will have a use for it.

I removed the unnecessary code now the rowHeight value cannot be set dynamically.

I fixed the new scenario that you updated.
image

hstastna (changes requested):
I think the buttons should be below the editing row. Not like this:
not

Another thing I've found is that it does not work well when changing the size of the window:
size_wind
I think there's plenty of space for the buttons below the editing row, but they 'jump' up, to an improper position.

bamsalem:
@hstastna .

> I think the buttons should be below the editing row. Not like this:

There are two classes/types:
1. top which his frame not includes the upper border
2. bottom which his  frame not includes the lower border.

https://github.com/oVirt/ovirt-web-ui/blob/3d5c0c9d5ca3dbe03f77890ae9aaf6af4eacf6b0/src/components/CreateVmWizard/steps/_TableInlineEditRow.js#L21
IMO probably the "missing" boarder should be next to the row.
@sjd78 WDYT?

Regarding this:
> Another thing I've found is that it does not work well when changing the size of the window:
> I think there's plenty of space for the buttons below the editing row, but they 'jump' up, to an improper position.

In my machine it works great.which browser do you use? FF/chromium?

sgratch:
> I think the buttons should be below the editing row. Not like this:
> not
> 
> Another thing I've found is that it does not work well when changing the size of the window:

> I think there's plenty of space for the buttons below the editing row, but they 'jump' up, to an improper position.

@bamsalem both issues are reproduced for me as well on both FF and Chrome. 
For the second issue, just try to zoom in till the buttons positioned is misaligned and then reset zoom level and see that buttons stayed misaligned,

bamsalem:
> > I think the buttons should be below the editing row. Not like this:
> > not
> > Another thing I've found is that it does not work well when changing the size of the window:
> 
> > I think there's plenty of space for the buttons below the editing row, but they 'jump' up, to an improper position.
> 
> @bamsalem both issues are reproduced for me as well on both FF and Chrome.
> For the second issue, just try to zoom in till the buttons positioned is misaligned and then reset zoom level and see that buttons stayed misaligned,

OK, I investigated the described issue,
IMO we should deal with this issue in another PR or PRs:
The issue appears in 2 scenarios:
The first scenario is when the `Wizard` scrollbar appears:
in my screen is when the user adding the 16th disc/nic and this is for sure an issue:
jumpingSaveCancel16th-2020-10-11_16 49 07
If you want I'll tell you about possible fix for the issue.

The second scenario is when the user zoom in to 200% and over:
jumpingSaveCancelbottom-2020-10-11_15 03 51
jumpingSaveCancel-2020-10-11_15 01 28
IMHO the fix should be to make the create vm wizard responsive and change the design for each scale, but the question is if it's required? if we want to support this change? WDYT @sgratch @hstastna

hstastna:
@bamsalem Why should we deal with this issue in another PR? Is it complicated to fix? I think that's no problem, this PR is actually very small and it's the same thing: to place the buttons to their proper positionin the Create VM Wizard. You can add more commits. But before any huge complicated refactoring changes, please discuss this or explain shortly what you plan to do.

IMO, I wouldn't deal with the the zooming issues right now - I think this moves away a little bit from the original issue. I would focus more on [this](https://github.com/oVirt/ovirt-web-ui/pull/1301#pullrequestreview-501917545) first and then fixed the zooming issues in anther PR if possible. However, I don't know how important is to have the buttons in the right position when zooming. @michalskrivanek @sgratch?

bamsalem:
> @bamsalem Why should we deal with this issue in another PR? Is it complicated to fix? I think that's no problem, this PR is actually very small and it's the same thing: to place the buttons to their proper positionin the Create VM Wizard. You can add more commits. But before any huge complicated refactoring changes, please discuss this or explain shortly what you plan to do.
>
> IMO, I wouldn't deal with the the zooming issues right now - I think this moves away a little bit from the original issue. I would focus more on [this](https://github.com/oVirt/ovirt-web-ui/pull/1301#pullrequestreview-501917545) first and then fixed the zooming issues in anther PR if possible. However, I don't know how important is to have the buttons in the right position when zooming. @michalskrivanek @sgratch?

The fix is not so complicated but it includes pass props from CraeteVmWizard to its children somehow, IMO it's something that require more attention and not be part of another fix.

To fix it we should call CreateVmWizard->render on scroll, but the render itself won't change the position of the buttons because nothing will be changed in the Storage / Networking component (if we call forceUpdate from there it won't change anything in the children component, their state/ props stayed as before).

The jumping button is an issue that caused by #1219 which is already merged, this fix is small fix that solves most of the cases and can be released quickly.

Most of the users don't add 16 discs/nics, and don't zoom in to 200%.

IMO the 16 rows issue is uncommon issue and not worth to delay this fix, and the zoom issue requires UI design, there is a range that the code should work and looks good and if users exeeding this range so it should be responsive to the change and to look different, and the quiestion is if the VM portal should be designed to small screen devices like mobile and tablet?
@hstastna WDYT?

bamsalem:
@sjd78 is there any reason that the button should be above the row? like @hstastna mentioned in her [comment](https://github.com/oVirt/ovirt-web-ui/pull/1301#pullrequestreview-501917545)
It seems like we don't need it because if there are more rows than we can show in the screen so there is a scroll.
I would like to hear if we need it or it should be removed, Thanks :)

hstastna:
> @sjd78 is there any reason that the button should be above the row? like @hstastna mentioned in her comment

@bamsalem No, I haven't mentioned that the buttons should be above the row. See the comment: I think the buttons should be **below** the editing row.

> It seems like we don't need it because if there are more rows than we can show in the screen so there is a scroll.
I would like to hear if we need it or it should be removed.

What exactly do you mean by "removing it"? We do need the buttons, we cannot remove them. We should only change their position in the screen.

bamsalem:
@hstastna the fact it's above the row is not something that I decided...
Take a look at the code and see that is already implemented.
I just changed the height.
so I would like to ask Scott why it's implemented as it is, Sharon and I tried to figure out why there are 2 different types of button positions..
Before we decide to remove that option I want to be sure it's OK and part of it is know what were the considerations to use different positions...
I mentioned your comment because you asked to change it.
Am I clear now?

hstastna:
> @hstastna the fact it's above the row is not something that I decided...

@bamsalem Thanks for the explanation, Ben. Regarding "why it's implemented as it is", well, this is why we need to comment the code :) And I've checked the https://github.com/oVirt/ovirt-web-ui/pull/1158/files and haven't figured it out, too. But I think there's simply a bug and if you haven't found any use case for one of those options, and if the problem will be fixed by removing it (and if the buttons position will be proper also in other scenarios we need), it can be very probably removed. I just wouldn't spend too much time figuring out "why" :)

sjd78:
ci test

sjd78:
ci add to whitelist

sjd78:
@bamsalem, @hstastna, @rszwajko, @sgratch - Let me answer a few questions and add my thoughts...

  - `_TableInlineEditRow.js` is a copy of the one in patternfly-react but modified to get the display more correct.  See: https://github.com/patternfly/patternfly-react/blob/patternfly-3/packages/patternfly-react/src/components/Table/TableInlineEditHeaderRow.js

  - You can see a demo of how it was designed to look at the pf3-react storybooks: https://patternfly-react-pf3.surge.sh/?path=/story/patternfly-react-content-views-table-view--table-with-inline-edit-row

  - I don't mind if the buttons are always displayed below the row.  However, it'll mean that if the table gets large (unlikely as Ben mentioned) users will need to scroll down to see the action buttons.

  - Some of the `rowHeight` calcs are dangerous unless you can verify that the `rowHeight` will be recalculated and passed down in the props if the row actually changes height.  In our code, that should only happen if the error messages are dynamically added.

  - The root of all evil here, and the reason we need to deal with this at all, is because the table is in a modal.  With the positioning changes that happen in parent components/dom, trying to accurately calc locations of things doesn't always work the way you'd expect.  The demo page (see above) works great with the pf3-react default positioning of the buttons.  But, if you launch the [Table in Modal With Inline Edit Row](https://patternfly-react-pf3.surge.sh/iframe.html?id=patternfly-react-content-views-table-view--table-in-modal-with-inline-edit-row) demo, you can see the basic problem that our local component is solving.

  - A complete solution would be to ensure that `reactabular-table` (the basis of the pf3-react Table component) is recalculating row dimensions on browser window resize, and that the new sizes are flowing down the Table props all the way to our `TableInlineEditRow` component.  If that isn't achievable in a reasonable time (how much longer are you around Ben? :wink:), then a simple change to cover a bit more render errors is acceptable.  

  - Going forward, if the buttons jump around less than before, I'm cool with that change.  But, I'm not worried about the extreme edge case of a user zooming in and out while editing a row.  Let them report the problem at that point.

I think that's about it for now.  I'll try out the change as it sits now.  If it works better than now, I'd be good with merging the PR.  Ideally the positioning code would just work and zooming wouldn't cause any problems.  We can split off further edge cases to another issue.  Maybe one like "In CreateNewVm, when editing or creating a Nic or Disk, the action buttons become misaligned if the browser window is resized or zoomed".

src/components/CreateVmWizard/steps/_TableInlineEditRow.js
```diff
@@ -4,12 +4,13 @@ import TableConfirmButtonsRow from 'patternfly-react/dist/js/components/Table/Ta
 
 const TableInlineEditRow = props => {
   const buttonsPosition = (window, rowDimensions) => {
+    const { rowHeight } = props
```
sjd78, line 7: As far as I can tell, `rowHeight` is not provided as a prop to the component, it is only being defined in the `defaultProps` as the constant value 31.  This will force the buttons to render in one spot no matter the height of the row.

Screen shot of the component tree in master (i.e. without this change):
VM Portal - Google Chrome_034

sjd78 (changes requested):
I expect the buttons would be stuck to the bottom of the row and move if the row height changes.  On master they don't do that, they are delayed in their movement.  With this PR, the buttons are locked in position.

The "proper" solution is probably to better handle row height changes within `TableConfirmButtonsRow` itself.  You'd have to copy the component from pf3-react and modify it locally.  The stock version of the component will detect scroll and browser resize changes, but won't detect if the `tr` dom element changes size.  There _should_ be an observer around that can do that.

https://github.com/patternfly/patternfly-react/blob/patternfly-3/packages/patternfly-react/src/components/Table/TableConfirmButtonsRow.js

bamsalem:
OK, after an investigation I found what is the "root of all evil here".
The dimensions was given to the `buttonPosition` (rowDimensions) function were measured from the top left corner of the viewed window and the returned position should be from the top left corner of the while DOM element, the problem starts when we can scroll the window.
When there is no option to scroll the dimensions are equal but when we scroll there is a missing "part" between the Y distance from the element's corner to view's corner  (the part which scrolled out from the view).
I added few changes in the implementation I'll explain them here:

1. `scrolledInPx`:
https://github.com/oVirt/ovirt-web-ui/blob/8a4a986e9eb08b651e285fd9da74e8d7d8eacc44/src/components/CreateVmWizard/steps/_TableInlineEditRow.js#L13
To deal with the scroll issue that described above I added new variable that represents how many pixels are missing between those two dimensions. 

2. rowHeight & state:
https://github.com/oVirt/ovirt-web-ui/blob/8a4a986e9eb08b651e285fd9da74e8d7d8eacc44/src/components/CreateVmWizard/steps/_TableInlineEditRow.js#L4
The `TableInlineEditRow` is class component to track changes in rowHeight:
https://github.com/oVirt/ovirt-web-ui/blob/8a4a986e9eb08b651e285fd9da74e8d7d8eacc44/src/components/CreateVmWizard/steps/_TableInlineEditRow.js#L15-L17
I found that the changes in case of scaling changes (zoom in/out) are less then `10px` for the  whole range (30%-300%) and any change that more than `10px` is abnormal and represents error message or another change that shouldn't reflect in button position so the component will ignore that change.

3. I removed the option to locate the button above the row.
WDYT??

hstastna (changes requested):
Now the buttons position seems to be fine, but only with a few rows present in the _Networking/Storage_ step. For more rows, it's still problematic - see the scenarios below. Those scenarios were reproduced without any browser window size changes, without zooming. However there are many items created to make the problems visible, but when browser window size changes or when zooming in, those scenarios are reproduced with much less items/rows than in the gifs I've posted in this comment. So I think we really need to fix that.

**Scenario 1:**
The buttons are stuck in one place and it doesn't look good - problematic when displaying/selecting options from the drop downs and scrolling:
output1

---

**Scenario 2:**
The buttons "jump" again, once making any change of the row in the edit mode:
output2
The buttons position should be the same (and below the row) when editing the row. Also I think the buttons shouldn't overlay other butons (Cancel, Back, Next) because it looks confusing. Maybe we could try to add some extra space below the last row (space for the buttons in the edit mode when editing the last row) - similarly as when clicking on the kebab menu of the row:
extra_space

@bamsalem @sjd78 @sgratch WDYT?

bamsalem:
@hstastna 


> Now the buttons position seems to be fine, but only with a few rows present in the _Networking/Storage_ step. For more rows, it's still problematic - see the scenarios below. Those scenarios were reproduced without any browser window size changes, without zooming. However there are many items created to make the problems visible, but when browser window size changes or when zooming in, those scenarios are easily reproduced with much less items/rows. So I think we do need to fix that.
> 
> **Scenario 1:**
> The buttons are stuck in one place and it doesn't look good - problematic when displaying/selecting options from the drop downs and scrolling:
> output1
> 
> **Scenario 2:**
> The buttons "jump" again, once making any change of the row in the edit mode:
> output2
> The buttons position should be the same (and below the row) when editing the row. Also I think the buttons shouldn't overlay the other buttons (Cancel, Back, Next) because it looks confusing. Maybe we could try to add some extra space for the buttons below the last row - similarly as when clicking on the kebab menu of the row:
> extra_space
> 
> @bamsalem @sjd78 @sgratch WDYT?

I tested your claiming in 300% zoom and it happens in my browser only in the 16th row, I'm using FF 82.0.
Regarding the 16th row issue you described above, I'm aware to that. @sgratch and I discussed on that and Sharon said  we don't need to spend time on that because it requires a lot of efforts and it's not a reasonable scenario,
it's not reasonable scenario that user create vm with 16 nics or 16 disks,
so we decided that it's unnecessary to handle that.

hstastna:
@bamsalem @sjd78 @sgratch, I think we do need to fix the scenarios I've mentioned in my previous comment because... see the gifs below:

- 150% and only 6 rows + no any change of the browser window size (1st scenario):
output1

- 150% and only 9 rows + no any change of the browser window size (2nd scenario):
output2
 
When I include browser window size changes in the scenarios or when zooming in more, those scenarios are reproduced with much less items/rows than in the gis above. Reproduced in FF 78 ESR, FF 80, FF 82, Chromium.

sgratch:
@bamsalem @hstastna @sjd78 as mentioned few times within this PR, all latest issues mentioned by @hstastna above are triggered once vertical scrollbar appears. 
When scrollbar doesn't appear, everything works great.
We can't avoid scrolling since resizing/zooming should be supported and customers might commonly use small screens (we know of some customers that require us to support that) and therefore it should reasonably work for all rations and window sizes. 

> 
> Scenario 1:
> The buttons are stuck in one place and it doesn't look good - problematic when displaying/selecting options from the drop downs and scrolling

It's ugly but I think we can leave with that. If the user scroll up/down while editing a row then buttons won't move with scrolling. I'm not sure it worth the effort for now and if customers will complain then we will consider fixing it.

> Scenario 2:
> The buttons "jump" again, once making any change of the row in the edit mode

This worth fixing so that at least buttons won't jump in case there is no space and scrollbars should appear. 

> the buttons shouldn't overlay the other buttons (Cancel, Back, Next) because it looks confusing. Maybe we could try to add some extra space for the buttons below the last row - similarly as when clicking on the kebab menu of the row

If it's easy to fix then OK. Otherwise, let's keep it as is for now.

bamsalem:
> > Scenario 2:
> > The buttons "jump" again, once making any change of the row in the edit mode
> 
> This worth fixing so that at least buttons won't jump in case there is no space and scrollbars should appear.

@sgratch @hstastna @sjd78 ,
To Fix the issue quoted above we need to change the events of the field from onChange to onBlur, due to that change the validation won't while user type it will occur only once when the user "leaves" the filed,
IMO lose the `onChange` validation doesn't worth that the button won't be aligned,
The buttons issue occurs in the 9th (I'm not sure) row, but the lose of functionality of validations is from the first row.
WDYT?

hstastna:
Well, this problem seems not to be an easy one. We need to be careful and to think more before we decide what the next steps will be. Maybe bigger changes will be needed. @bamsalem The buttons jump with less than 9 rows, it is just not so visible as in the gif I've posted + the more zooming, the earlier the jumping occurs. So I still think we do need to fix it. But we need to move forward, so I suggest to finish this PR without the fix for this problem, so the github issue should remain opened (or the new one should be opened specifically for this problem) and we can continue with discussions in that issue, then we can deal with that bug in the follow up PR.

sgratch:
> To Fix the issue quoted above we need to change the events of the field from onChange to onBlur, due to that change the validation won't while user typing the name, it will validated only once when the user "leaves" the filed,

The buttons position is "jumping" also when selecting a value from one of the pull down menus ("vnic profile", "device type" etc)  and not just when typing the name. 
I just don't understand: if you know to calculate the "correct" position for the buttons to jump to, so why not locating it on the 1st place in the right position?

> Well, this problem seems not to be an easy one. We need to be careful and to think more before we decide what the next steps will be. Maybe bigger changes will be needed. @bamsalem The buttons jump with less than 9 rows, it is just not so visible as in the gif I've posted + the more zooming, the earlier the jumping occurs. So I still think we do need to fix it. But we need to move forward, so I suggest to finish this PR without the fix for this problem, so the github issue should remain opened (or the new one should be opened specifically for this problem) and we can continue with discussions in that issue, then we can deal with that bug in the follow up PR.

+1

bamsalem:
> The buttons position is "jumping" also when selecting a value from one of the pull down menus ("vnic profile", "device type" etc) and not just when typing the name.
> I just don't understand: if we know to calculate the "correct" position for the buttons to jump to, so why not locating it in the right position for the 1st time before triggering any event?

We know how to calculate the correct position but the given values are not up to date to the position after rendering, I tested that and I found  that the PF component call the `buttonsPosition` with the wrong values, the values are correct to the moment before rendering and sometimes it's incorrect for the position after the `render()` call.

As already mentioned by @sjd78 , PF3 inline editor is not perfect, especially if it used in modal.

sgratch:
> We know how to calculate the correct position but the given values are not up to date to the position after rendering, I tested that and I found that the PF component call the `buttonsPosition` with the wrong values, the values are correct to the moment before rendering and sometimes it's incorrect for the position after the `render()` call.
> 
> As already mentioned by @sjd78 , PF3 inline editor is not perfect, especially if it used in modal.

+1
So let's stay with the current fix since it solves most of the original problems. 

The remaining issues mentioned above, related to the vertical scrollbar, are more complicated to fix and less urgent for now.
@bamsalem  Please open a new issue for handling those remaining problems.

sgratch (approved):
Thanks Ben.
The fix LGTM as it is based on assumptions detailed on https://github.com/oVirt/ovirt-web-ui/pull/1301#issuecomment-720369598

All remaining problems that are not part of this fix, are referred by a separate issue  #1322

isaranova:
ci build please

isaranova:
LGTM

PR #1303, opened 2020-10-06
Update minimum ovirt version to 4.4

sjd78:
With adding user settings, the minimum rest api level required to
run all of web-ui will be 4.4.  There have been a few other rest
api bugs since 4.2 that solved a few web-ui bugs as well.

See: #1088


hstastna (approved):
LGTM.

PR #1304, opened 2020-10-09
Shorten the recomendedValuesForThreads message for the Threads tooltip

hstastna:
**Related comment:** https://github.com/oVirt/ovirt-web-ui/pull/1279#issuecomment-705454171

In this PR, I've shortened the message by removing the 'Other recommended values...' part not to be long and misleading anymore, according to the Michal's suggestion in another PR.

**Before:** (VM Details Card)
before_tooltip

**After:**
after_tooltip


michalskrivanek:
LGTM

sjd78 (changes requested):
The shorter message reads ok.

CI is failing because the message params are not matched up between the English text and the translations.  So two things to do to clean up the rest of things:

  1. Make sure where the `msg.recomendedValuesForThreads()` is invoked, that the `threads` param is dropped.

  2. Remove the translations for that key since they'll also be too long and will require re-translation.  Note: After committing this PR, doing a Zanata push will invalidate the translations and then an immediate Zanata pull should drop them from the translation file.  We just can't wait for that since we still want CI to pass.

sgratch (approved):
LGTM

PR #1193, opened 2020-04-27
Update VM name filter to match anywhere in the VM name

bamsalem:
hi I just change the filter methods from startsWith to includes.

Fixes: #1127

PR #1309, opened 2020-10-14
VM filter - Makes the filter be case insensitive

bamsalem:
Fixes: [BZ1879655](https://bugzilla.redhat.com/show_bug.cgi?id=1879655)
> Description of problem:
At present, it is not possible to search VM names using case sensitive keys or using partial VM name in VM Portal.
>
>For example, in VM portal, if VM name is,
>
>rhel-python-affinity-1, then it is not possible to search for the VM using simply affinity or Rhel(R is capital, testing for case sensitive)

In this PR I fixed the issues described above.
The issue is divided into 2 sub issues:
1. The filter is not includes vms which the filter string is not in the beginning of their names.
This sub issue is already resolved by #1193 .

2. The filter is case sensitive.
To fix this sub issue we need to make the filter be case insensitive.

To achieve the case insensitivity I tests the filter string and the vm name as upper case strings.
Few screenshots:
**The tested VMs:**
VMs

**Before:**
Lower case search:
beforeLower
Middle sub string search:
beforeMiddle
Case sensitive search:
beforeUpper

**After:**
Middle sub string case sensitive search: 
afterPython
Case sensitive:
afterRhel

### DIFF
--- a/src/utils/vms-filters.js
+++ b/src/utils/vms-filters.js
@@ -2,7 +2,7 @@ import { getOsHumanName } from '../components/utils'
 import { enumMsg } from '_/intl'
 
 const compareMap = {
-  name: (item, filter) => !!filter.find(n => item.get('name').includes(n)),
+  name: (item, filter) => !!filter.find(n => item.get('name').toUpperCase().includes(n.toUpperCase())),
   os: (item, filter) => getOsHumanName(item.getIn(['os', 'type'])) === filter,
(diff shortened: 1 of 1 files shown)


hstastna (approved):
LGTM. Another solution would be using regular expression.

sjd78 (approved):
LGTM

sgratch (approved):
LGTM

Issue #1308, opened 2020-10-13
CreateVmWizard: Error after trying to move to the Review step in specific scenario


hstastna:
Error occurs after trying to move to the _Review_ step, for _Provision source_ ISO and without any change of the OS in the _Basic settings_ step of the Create VM Wizard.

The core of the problem is displaying the timezone name in the _Review_ step but the timezone is undefined so it won't work, obviously. In the specific scenario, the timezone is unfortunately not set, in the Basic settings step. This is probably caused by https://github.com/oVirt/ovirt-web-ui/pull/1220.

**Steps to Reproduce:**
1. Open the Create VM Wizard, fill in the _Name_ of a new VM
2. For _Cluster_ field, choose the cluster for which no templates (or only _Blank_ template) are available
=> _Provision source_: ISO should be automatically set
3. Choose some _CD_ from the drop down
4. Click _Next_ three times to reach the _Review step_

**Actual results:**
`TypeError: basic.timeZone is undefined`

![sorry](https://user-images.githubusercontent.com/13417815/95912426-21a19500-0da3-11eb-9d44-103977824428.png)

**Expected results:**
Displaying the _Review_ step with the proper info:
![after](https://user-images.githubusercontent.com/13417815/96194591-f8bc0400-0f4a-11eb-9f5d-19305f94a260.png)

**Additional info:**
Don't touch the _Operating system_ drop down, don't play with the _Provision source_ field, otherwise the issue will not be reproducible.
Also If you choose some cluster in the step 2 with some templates available and then if you change cluster to some other, the issue will not be reproducible, too. You need to close the wizard and start again.

PR #1310, opened 2020-10-15
CreateVmWizard: Fix error in Review step and set timezone in the Basic settings correctly

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1308

Error has occurred after trying to move to the _Review_ step of the Create VM Wizard in a specific scenario when _Provision source_ was set to _ISO_ (see the exact steps to reproduce [here](https://github.com/oVirt/ovirt-web-ui/issues/1308#issue-720876530)). This PR fixes it.

The core of the problem was displaying the [timezone name here](https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/CreateVmWizard/steps/SummaryReview.js#L61) in the _Review_ step, but the `timeZone` was `undefined`. Why? Well, actually the only possible places, where the timezone is set, are in some of the cases of the `handleChange` function of the `BasicSettings` component ([for example here](https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/CreateVmWizard/steps/BasicSettings.js#L304)). But if those cases are not reached in the scenario we perform (according to the fields we change in the Basic settings), then the timezone is simply not defined.

This could be fixed by more ways, for example by handling the timezone undefined value in the `ReviewBasic` component for displaying the Review step of the wizard. But I've better chosen to add the missing logic into the `handleChange` function. Before that, I've refactored some cases as there's some code which repeats unnecessary. We wanted to refactor/move this logic anyway, as it's getting too complicated and confused.

First, I've extracted the provision source changes related code in `handleChange` to `handleProvisionSourceChange` function. Added [missing change](https://github.com/oVirt/ovirt-web-ui/pull/1310/files#diff-ecc9f96259bb1bc0ee42cc7ff4de2121caf4ac2bfecc640b2cc798f976565489R259) to reset the sysprep timezone value when changing the provision source to ISO in that function. Simplified the cluster id changes related code and added the `handleProvisionSourceChange`, because it was missing there and causing the error. The provision source is (re)set together with the cluster id change so we need to handle the provision source changes in that situation, too. This will cause setting proper value of the `timeZone` and others.

**Before:** Not possible to reach the Review step, error after clicking on _Next_ in the Storage step:
before

**After:** The Review step is properly displayed together with the timezone properly set:
after

### DIFF
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -276,24 +276,15 @@ class BasicSettings extends React.Component {
     let changes = {}
     switch (field) {
       case 'clusterId':
-        if (value === undefined) {
-          changes.dataCenterId = undefined
-          changes.clusterId = undefined
-          changes.provisionSource = 'template'
-          changes.isoImage = undefined
-          changes.templateId = undefined
-        } else {
-          const templateList = createTemplateList(this.props.templates, value)
-
-          changes.dataCenterId = this.props.clusters.getIn([value, 'dataCenterId'])
-          changes.clusterId = value
-          changes.provisionSource =
+        const templateList = value ? createTemplateList(this.props.templates, value) : []
+        changes.dataCenterId = value ? this.props.clusters.getIn([value, 'dataCenterId']) : undefined
+        changes.clusterId = value
+        changes.provisionSource =
             (templateList.length === 1 && templateList[0].id === this.props.blankTemplateId)
               ? 'iso'
               : 'template'
-          changes.isoImage = undefined
-          changes.templateId = templateList.length === 1 ? this.props.blankTemplateId : undefined
-        }
+        changes.templateId = templateList.length === 1 ? this.props.blankTemplateId : undefined
+        changes.isoImage = undefined
         break
 
       case 'provisionSource':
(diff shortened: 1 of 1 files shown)


bamsalem (approved):
LGTM Great job!!!

bamsalem:
This PR has a lot in common with PRs #1265 #1291 and the described issue is resolved by  #1265 too.

rszwajko:
The change looks OK to me however it seems a subset of #1265 .  If we want to have the fix quickly then OK but I would rather merge #1265 instead.

hstastna:
@rszwajko IMO it's more reasonable to merge this PR first, because of more reasons: there are more changes in #1265, it's more safe to make smaller changes in the Basic settings change handlers to prevent creating new bugs. Also I can see that in #1265, `handleClusterIdChange` seems to be much more complicated comparing to the proposed cluster id changes handling in this PR. It may be because of handling more possible edge cases (I don't know about yet) but I already can see some unnecessary code repeating there, for example setting the `isoImage` to be `undefined`. It could be refactored a little bit so I think it's more reasonable to merge this PR first.

sgratch (approved):
@hstastna @rszwajko @bamsalem
LGTM. since it seems that #1265 is still under review and since @bamsalem approved this as well, let's merge this and rebase #1265 on top of it with all relevant required changes (move handleProvisionSourceChange out to helpers.js etc)
@bamsalem is it ok?

sgratch (approved):
LGTM

Issue #1313, opened 2020-10-21
Create VM Wizard - on basic settings step the cluster pull down should include only clusters with at least one assigned host


sgratch:
When creating a new VM on basic settings step, the options for the Cluster menu field are only those clusters that include at least one host (host status is not relevant). The VM can't run on a cluster without any host assigned, so those clusters should be blocked for vms creation..

Currently all existed clusters are displayed.

hstastna:
I think this issue is part of the https://github.com/oVirt/ovirt-web-ui/issues/1307

Issue #1307, opened 2020-10-13
CreateVmWizard: Blank OS in Basic settings after choosing some clusters


hstastna:
There's an issue in the _Basic settings_ step of the wizard: after choosing a _Cluster_, blank/none of the OS is chosen for the _Operating system_ field/drop down. This does not happen always, only with some of the clusters available in the appropriate drop down.

I've found that this issue is related to the **CPU type** chosen in webadmin when creating a cluster. Looks like choosing/selecting the proper OS in the drop down works fine in the Basic settings only when selecting clusters created with any _CPU Type_ available for _CPU architecture_ _x86_64_.

**Steps to Reproduce:**
1. In webadmin, create a cluster in one of the two ways:
A) _CPU Architecture_ `undefined` and _CPU Type_ `Auto Detect`:
![webadmin](https://user-images.githubusercontent.com/13417815/95909377-accc5c00-0d9e-11eb-8237-223d0a7f33a9.png)
B) `ppc64` and any of the two available CPU types:
![c2](https://user-images.githubusercontent.com/13417815/95901545-cf0cac80-0d93-11eb-9b24-1a534650f950.png)
2. In web UI, open the Create VM Wizard, select the created cluster in the drop down, in _Basic settings_ step
=> look at the _Operating System_ drop down

---

**Actual results:** 
Nothing is selected as the _Operating System_:
![n1](https://user-images.githubusercontent.com/13417815/95902957-cf0dac00-0d95-11eb-8320-29582d80d4db.png)
A) the drop down is empty!
![a](https://user-images.githubusercontent.com/13417815/95909185-5a8b3b00-0d9e-11eb-805a-c949a61e058b.png)

B) at least, the drop down is not empty and contains the proper items
![b](https://user-images.githubusercontent.com/13417815/95909190-5ced9500-0d9e-11eb-9f7d-d1b7cd076448.png)

In addition, after adding a _Name_ for a VM and choosing a _CD_ (or _Blank_ template), we are able to click on the Next button, even when no OS is selected, which is required field!
![power](https://user-images.githubusercontent.com/13417815/95904482-df268b00-0d97-11eb-957b-1f488057a6ca.png)
Then the _Other OS_ is displayed in the Review step, which is inconsistent with what we've seen in the Basic settings step:
![power2](https://user-images.githubusercontent.com/13417815/95904820-52c89800-0d98-11eb-978d-7444aa092aaa.png)

---

**Expected results:**
A) ???
[...]

PR #1318, opened 2020-10-28
Filter the clusters with no architecture and cpu-type attached - CreateVmWizard and Details card

bamsalem:
Fixes: #1313 .

Before, in CreateVMWizard on basic settings step, Cluster's selectBox options included clusters which are not includes any host:

> When creating a new VM on basic settings step, the options for the Cluster menu field are only those clusters that include at least one host (host status is not relevant). The VM can't run on a cluster without any host assigned, so those clusters should be blocked for vms creation..

The issue mentioned only for CreateVmWizard but I found that the same issue can be occurred in edit existing VM dialog.
Now, The Cluster's list includes clusters that include at least one host, I made a little improvement as well, in the Details card the call to `createClusterList` function includes the dataCenterId parameters so it won't filter it in any `render()`.

Before:
CreateVm

After:
CreateVm

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -113,7 +113,7 @@ class DetailsCard extends React.Component {
       promptNextRunChanges: false,
 
       isoList: createIsoList(props.storageDomains, vmDataCenterId),
-      clusterList: createClusterList(props.clusters),
+      clusterList: createClusterList(props.clusters, vmDataCenterId),
       osList: createOsList(vmClusterId, props.clusters, props.operatingSystems),
     }
     this.trackUpdates = {}
@@ -170,13 +170,14 @@ class DetailsCard extends React.Component {
     //       __storageDomains__, and __operatingSystems__ don't need to be kept in state for
     //       change comparison
     const vmClusterId = this.props.vm.getIn(['cluster', 'id'])
+    const vmDataCenterId = this.props.clusters.getIn([vmClusterId, 'dataCenterId'])
 
     if (prevProps.clusters !== this.props.clusters) {
-      this.setState({ clusterList: createClusterList(this.props.clusters) }) // eslint-disable-line react/no-did-update-set-state
+      const { clusters } = this.props
+      this.setState({ clusterList: createClusterList(clusters, vmDataCenterId) }) // eslint-disable-line react/no-did-update-set-state
     }
 
     if (prevProps.storageDomains !== this.props.storageDomains) {
-      const vmDataCenterId = this.props.clusters.getIn([vmClusterId, 'dataCenterId'])
       this.setState({ isoList: createIsoList(this.props.storageDomains, vmDataCenterId) }) // eslint-disable-line react/no-did-update-set-state
     }
 
@@ -734,7 +735,7 @@ class DetailsCard extends React.Component {
                       { isFullEdit && canChangeCluster &&
                         <SelectBox
                           id={`${idPrefix}-cluster-edit`}
-                          items={clusterList.filter(cluster => cluster.datacenter === dataCenterId)}
+                          items={clusterList}
                           selected={clusterId}
                           onChange={(selectedId) => { this.handleChange('cluster', selectedId) }}
                         />
--- a/src/components/utils/build-select-box-lists.js
+++ b/src/components/utils/build-select-box-lists.js
@@ -17,6 +17,9 @@ function createClusterList (clusters, dataCenterId = null) {
       .toList()
       .filter(cluster =>
         cluster.get('canUserUseCluster') &&
+        !!cluster.get('architecture') &&
+        cluster.get('architecture') !== 'undefined' &&
+        !!cluster.get('cpuType') &&
         (dataCenterId === null
           ? true
           : cluster.get('dataCenterId') === dataCenterId)
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -446,6 +446,8 @@ export const messages: { [messageId: string]: MessageType } = {
   nicNoVnicAssigned: 'N/A',
   nicsTooltip: 'Connected VM network interfaces.',
   noActiveStorageDomainInDataCenter: 'There is no active data storage domain in data center "{dataCenterName}"',
+  noCdsAvailable: 'No CDs available',
+  noClustersAvailable: 'No Clusters available',
   noDisks: 'no disks',
   noError: 'No error',
   noMessages: 'There are no notifications to display.',
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -367,7 +367,7 @@ class BasicSettings extends React.Component {
           value: `${cluster.value} (${this.props.dataCenters.find(dc => dc.id === cluster.datacenter).name})`,
         }))
     if (!isValidUid(data.clusterId)) {
-      clusterList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectCluster()} --` })
+      clusterList.unshift({ id: '_', value: clusterList.length === 0 ? `-- ${msg.noClustersAvailable()} --` : `-- ${msg.createVmWizardSelectCluster()} --` })
       indicators.cluster = !indicators.name && 'error'
     } else {
       delete indicators.cluster
@@ -400,7 +400,7 @@ class BasicSettings extends React.Component {
         .sort((a, b) => localeCompare(a.value, b.value))
       : [ { id: '_', value: `-- ${msg.createVmWizardSelectClusterBeforeISO()} --` } ]
     if (enableIsoSelect && !isValidUid(data.isoImage)) {
-      isoList.unshift({ id: '_', value: `-- ${msg.createVmWizardSelectISO()} --` })
+      isoList.unshift({ id: '_', value: isoList.length === 0 ? `-- ${msg.noCdsAvailable()} --` : `-- ${msg.createVmWizardSelectISO()} --` })
       indicators.isoList = !indicators.provisionSource && !indicators.name && 'error'
     } else {
       delete indicators.isoList
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -658,6 +658,7 @@ const Cluster = {
       name: cluster.name,
       dataCenterId: cluster.data_center && cluster.data_center.id,
       architecture: cluster.cpu && cluster.cpu.architecture,
+      cpuType: cluster.cpu && cluster.cpu.type,
 
       memoryPolicy: {
         overCommitPercent:


ovirt-infra:

sjd78:
ci add to whitelist

sgratch (commented):
- This issue doesn't reproduce to me when editing an existing VM even prior to this fix. I found this problem relevant only to create vm wizard. What is the exact scenario for reproducing? 

- In case no cluster fits this requirement of at least one host, the "Cluster" menu includes no options (the "select cluster" option is missing) as well as marking this field with red. This will prevent the user from understanding why the "next" button is still disabled.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -172,7 +172,10 @@ class DetailsCard extends React.Component {
     const vmClusterId = this.props.vm.getIn(['cluster', 'id'])
 
     if (prevProps.clusters !== this.props.clusters) {
-      this.setState({ clusterList: createClusterList(this.props.clusters) }) // eslint-disable-line react/no-did-update-set-state
+      const vmClusterId = this.props.vm.getIn(['cluster', 'id'])
+      const vmDataCenterId = this.props.clusters.getIn([vmClusterId, 'dataCenterId'])
+      const { clusters, hosts } = this.props
```
hstastna, line 177: Ben, is there any purpose for the order of the two lines above? I think you could use
```javascript
      const { clusters, hosts } = this.props
      const vmDataCenterId = clusters.getIn([vmClusterId, 'dataCenterId'])
```
instead.

Also I think you don't need the line 175 `const vmClusterId = this.props.vm.getIn(['cluster', 'id'])`
See the line 172 above the "if" block.

hstastna (changes requested):
LGTM, however there are some little improvements that could be done ;)

Note that the error occurs in the VM Details card when trying to change the Cluster. However I am able to reproduce this error also on the actual master branch so it's probably unrelated to this PR.

bamsalem:
> * This issue doesn't reproduce to me when editing an existing VM even prior to this fix. I found this problem relevant only to create vm wizard. What is the exact scenario for reproducing?

@sgratch .
I created a cluster in "Default" data center on brq-dev, there are no hosts attached to the new cluster:
image
But we still can select this cluster, even there are no hosts attached to it.

Take a look on my screenshots:

**before:**

openDropDown
closedDropDown

**after:**

image


> * In case no cluster fits this requirement of at least one host, the "Cluster" menu includes no options (the "select cluster" option is missing) as well as marking this field with red. This will prevent the user from understanding why the "next" button is still disabled.

Regarding no cluster fits, I can't reproduce this issue for CreateVm dialog:
image

hstastna (commented):
LGTM, it works well for me. Tested also the case when no clusters are available (clusters with no hosts). _-- Select Cluster --_ with the red border together with the _Next_ button disabled is displayed properly.

However, I think it may not be clear to the user why they see no clusters in the drop down. In addition, IMO it's not correct to see a message which says "Select Cluster" and not being able to do so in the same time. It's conflicting. So I suggest to add a tooltip next to the _Cluster_ label to make it clear. The message in the tooltip could say that there aren't any clusters available for creating a VM and that we need to create some new cluster and/or attach a host to it/to some existing cluster without hosts.
vm

sgratch:
> This issue doesn't reproduce to me when editing an existing VM even prior to this fix. I found this problem relevant only to create vm wizard. What is the exact scenario for reproducing? 

It seems a bit more complicated than only checking if hosts are currently assigned to the cluster and that's why it sometimes works as expected and sometimes doesn't.
On backend you can still assign a VM for a cluster with no hosts but as long as the following conditions are kept:
1. CPU architecture, cpu name and emulated machine are set for the cluster (either manual or by assigning a host to the cluster which can be removed later on). 
2. Cluster's storage (storage_pool_id) is set to the same storage as set for the data center and is valid and active. i.e. If there is a data center with no active storage domains then you shouldn't choose it.

Condition 1 is auto set once you assign a host to a cluster, but it's kept even after removing hosts from the cluster.

To sum it up, I would change the fix to verify that the cluster has a defined architecture and cpu type instead of verifying that there are current hosts assigned (condition 1). I won't verify the storage domain and let it fail on backend since it seems too complicated and not worth it (condition 2). 
Note that issue #1307 is caused by condition 1 as well. 

> In case no cluster fits this requirement of at least one host, the "Cluster" menu includes no options (the "select cluster" option is missing) as well as marking this field with red. This will prevent the user from understanding why the "next" button is still disabled.

It can happen in case no clusters exists, but no need to handle that. This use case probably won't happen anyway,

> However, I think it may not be clear to the user why they see no clusters in the drop down. In addition, IMO it's not correct to see a message which says "Select Cluster" and not being able to do so in the same time. It's conflicting. So I suggest to add a tooltip next to the Cluster label to make it clear. The message in the tooltip could say that there aren't any clusters available for creating a VM and that we need to create some new cluster and/or attach a host to it/to some existing cluster without hosts.

We can add that but I won't specify all reasons for why there are no clusters to select since there are many more like permission issue for example (cluster is not permitted for the user). So it's up to the admin to check that. I would show something like: "No clusters are available. Contact your administrator for checking that" or something like that...

hstastna:
> We can add that but I won't specify all reasons for why there are no clusters to select since there are many more like permission issue for example (cluster is not permitted for the user). So it's up to the admin to check that. I would show something like: "No clusters are available. Contact your administrator for checking that" or something like that...

Sharon, I do agree. Let's just add there something more useful for the user than having there totally nothing ;)

bamsalem:
@sgratch @hstastna ,
I'll explain how I'm going to implement the requested changes and please tell me WDYT :)
Regrading the clusters filtering:
I'm going to remove the condition "at least one cluster attached" and check if the attribute architecture is set instead:
image
Regarding the no clusters in the list issue I will change the message from "Select Cluster " to "There Are No Available Clusters"
WDYT?

sgratch:
> @sgratch @hstastna ,
> I'll explain how I'm going to implement the requested changes and please tell me WDYT :)
> Regrading the clusters filtering:
> I'm going to remove the condition "at least one cluster attached" and check if the attribute architecture is set instead:

You need to check that both cpu architecture and cpu type are set.

> Regarding the no clusters in the list issue I will change the message from "Select Cluster " to "There Are No Available Clusters"
> WDYT?

This is not so consistent since for other pull down menus in basic settings we still use the "select X" approach, even when there is nothing to select. We just mark the field as red. e.g in case there is no ISO image.

I prefer leaving it as is or just add a tooltip which will appear only for this cluster case when there are no clusters to select, like @hstastna suggested.

bamsalem:
@sgratch 
>This is not so consistent since for other pull down menus in basic settings we still use the "select X" approach, even when there is nothing to select. We just mark the field as red. e.g in case there is no ISO image.
>
>I prefer leaving it as is or just add a tooltip which will appear only for this cluster case when there are no clusters to select, like @hstastna suggested


IMHO tooltip is unnecessary in that case because it will look strange and it's not consistent with other pull down menus.
If you think it's necessary I will add it.

>You need to check that both cpu architecture and cpu type are set.

I changed the filtering method, WDYT?

src/components/utils/build-select-box-lists.js
```diff
@@ -17,6 +17,8 @@ function createClusterList (clusters, dataCenterId = null) {
       .toList()
       .filter(cluster =>
         cluster.get('canUserUseCluster') &&
+        cluster.get('architecture') &&
```
sgratch, line 20: It won't filter clusters with undefined architecture since in that case I think that the cluster is set to "undefined" and not null. But worth checking.

sgratch (dismissed):

sgratch:
> IMHO tooltip is unnecessary in that case because it will look strange and it's not consistent with other pull down menus.
> If you think it's necessary I will add it.

@hstastna @bamsalem 
So either change both pull down menus for cluster and ISO to "There Are No Available .."  instead of "Select ..." or just leave it as is now. After re-thinking, it's an edge case and anyway in case of no available clusters for the user then there is nothing that the user can do except approaching the admin (which he will do anyway if he can't create vms).
So I'm not sure this tooltip worth the effort in that case. 
I suggest leaving that as is and in case users will complain then we'll need to think about a more informative message than just "there are no available clusters to select".

hstastna:
> So either change both pull down menus for cluster and ISO to "There Are No Available .." instead of "Select ..." or just leave it as is now.

IMO it's better to do something about this edge case, not just leaving it as it is. Adding a tooltip or changing some string is really not a lot of work. I think it worth the effort. 

> I suggest leaving that as is and in case users will complain...

I think waiting for users to complain is not the best strategy. Especially if we can so easily do something for them like in this case - to inform them  by adding some extra info for them in the UI or by changing some string in some case. I can imagine that seeing "Select Cluster" and not being able to do so could be very frustrating for some users. We shouldn't rely on the users - that they would immediately know that everything they do is ok and that it's not a bug.

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -172,7 +172,9 @@ class DetailsCard extends React.Component {
     const vmClusterId = this.props.vm.getIn(['cluster', 'id'])
 
     if (prevProps.clusters !== this.props.clusters) {
-      this.setState({ clusterList: createClusterList(this.props.clusters) }) // eslint-disable-line react/no-did-update-set-state
+      const { clusters } = this.props
+      const vmDataCenterId = clusters.getIn([vmClusterId, 'dataCenterId'])
+      this.setState({ clusterList: createClusterList(clusters, vmDataCenterId) }) // eslint-disable-line react/no-did-update-set-state
```
sjd78, line 177: Just move the `vmDataCenterId` accessing code from the storageDomains check block (line 181) up to under line 172 and have `vmClusterId` and `vmDataCenterId`. That way the clusters and storageDomains can reuse the same lookup

src/components/utils/build-select-box-lists.js
sjd78: If a cluster has it's architecture set to "undefined" in webadmin, it just won't have a `cpu`  data element in its JSON at all.  The transform+reducer will leave `cluster.architecture` undefined.

So...

Undefined in webadmin = missing in JSON = undefined after internal transform = missing in redux = truthy false/

I update the check to force conversion to an actual boolean :
```suggestion
        !!cluster.get('architecture') &&
```

If the value is undefined, 0, false, '', or NaN, the double !! will convert it to `false`.  Any other value will be `true`.

src/components/utils/build-select-box-lists.js
```diff
@@ -17,6 +17,8 @@ function createClusterList (clusters, dataCenterId = null) {
...
+        cluster.get('cpuType') &&
```
sjd78, line 21:
```suggestion
        !!cluster.get('cpuType') &&
```

sjd78 (dismissed):
Ok, so now, if the PR is filtering clusters such that only clusters with a defined architecture and a defined cpuType are shown, the title of the PR, the initial description and probably the base issue itself, need to be updated to reflect that.

So now, if a cluster can be displayed if it has no hosts, that should be recorded somewhere prominent.

Last thought - if a cluster with no hosts can be selected, and if that cluster IS selected, then the "Start virtual machine on creation" check box should be at least disabled.  No hosts = can't run = can't start on creation.

bamsalem:
@hstastna 
I changed the message for no clusters/CDs as you suggested:
image
image

sgratch:
> You need to check that both cpu architecture and cpu type are set.

@bamsalem it still displays clusters to select with no architecture defined - it failed on backend with error: "CREATE_VM failed [Cannot add VM. The cluster does not have a defined architecture.]"

> Last thought - if a cluster with no hosts can be selected, and if that cluster IS selected, then the "Start virtual machine on creation" check box should be at least disabled. No hosts = can't run = can't start on creation.

It think it's ok to leave it like that and let it fail on backend since no hosts are available.

src/intl/messages.js
```diff
@@ -584,6 +584,8 @@ export const messages: { [messageId: string]: MessageType } = {
   templateDefined: 'Template Defined',
   timeAgo: '{time} ago',
   timezone: 'Timezone',
+  thereAreNoAvailableClusters: 'There Are No Available Clusters',
+  thereAreNoAvailableCDs: 'There Are No Available CDs',
```
hstastna, line 588: What about making the messages shorter? We don't need to see the whole sentence in the drop down :)
My suggestion:
```
noCdsAvailable: 'No CDs available',
noClustersAvailable: 'No Clusters available',
```
Check how it is already done for a VM, for example: https://github.com/oVirt/ovirt-web-ui/blob/master/src/intl/messages.js#L470

hstastna (dismissed):
There's one little suggestion for shortening the messages regarding the case of zero items available from my side. Otherwise LGTM 👍🏽

src/intl/messages.js
sjd78: I second the suggested text.

Side note: Only the first word of the sentence and the actual thing being referenced would be capitalized.  So, if the message would stay the same, it should read "There are no Clusters available".

sjd78 (commented):
The rest of the code LGTM, just fix the messages as @hstastna suggests.

sjd78 (approved):
LGTM

sgratch:
> @bamsalem it still displays clusters to select with no architecture defined - it failed on backend with error: "CREATE_VM failed [Cannot add VM. The cluster does not have a defined architecture.]"

You didn't refer to this issue. 
Here is the detailed problem:
In case the cluster's architecture is not set yet (i.e. set as "undefined") regardless to cpu type value:
image

then the following error occurs:
image

Both architecture and cpu type should be set for allowing VMs to run on this cluster. When a host is attached to a cluster then both are auto set but if you set those manually then it may occur that only cpu type is set while the architecture is still undefined.

sgratch (commented):
Please check my comment above.

bamsalem:
@sgratch  

> > @bamsalem it still displays clusters to select with no architecture defined - it failed on backend with error: "CREATE_VM failed [Cannot add VM. The cluster does not have a defined architecture.]"
> 
> You didn't refer to this issue.
> Here is the detailed problem:
> In case the cluster's architecture is not set yet (i.e. set as "undefined") regardless to cpu type value:
> image
> 
> then the following error occurs:
> image
> 
> Both architecture and cpu type should be set for allowing VMs to run on this cluster. When a host is attached to a cluster then both are auto set but if you set those manually then it may occur that only cpu type is set while the architecture is still undefined.

I can't reproduce the issue you raised above. can you provide me more details how you produced that?
When I set cpu type via WebAdmin the architecture set automatically even when I set it to undefined.
Thanks :)

sgratch:
> I can't reproduce the issue you raised above. can you provide me more details how you produced that?
> When I set cpu type via WebAdmin the architecture set automatically even when I set it to undefined.

@bamsalem I updated that on DB just to make sure both are checked properly. It might work with rest api or maybe when upgrading cluster compatibility level after upgrading. My suggestion is just to change the code to check if architecture is not set to "undefined" to make the code more comprehensive since changing the cluster cpu properties are involved on few use cases.

bamsalem:
@sgratch I changed the code as you suggested.

sgratch (approved):
LGTM

src/intl/messages.js
```diff
@@ -446,6 +446,8 @@ export const messages: { [messageId: string]: MessageType } = {
   nicNoVnicAssigned: 'N/A',
   nicsTooltip: 'Connected VM network interfaces.',
   noActiveStorageDomainInDataCenter: 'There is no active data storage domain in data center "{dataCenterName}"',
+  noClustersAvailable: 'No Clusters available',
+  noCdsAvailable: 'No CDs available',
```
hstastna, line 450: Could you, please, swap these two messages? To have them sorted nicely alphabetically ;) Thanks!

hstastna (commented):
There's one little thing to improve, otherwise LGTM! :)

src/intl/messages.js
bamsalem: Thanks, Done.

hstastna (approved):
LGTM.

isaranova:
Could you please provide verification steps? I'm sorry, but I'm not sure which of the original requirements stayed and which changed and what was added. Thanky you very much

sgratch:
> Could you please provide verification steps? I'm sorry, but I'm not sure which of the original requirements stayed and which changed and what was added. Thanky you very much

@isaranova Sure, sorry for the confusion:
1. open create vm wizard and try to select a cluster out of the list.
2. make sure that all clusters within the list are already set with architecture and CPU type (i.e. the user can't attach VMs to clusters with empty values for cpu type and architecture). 
Clusters that are not set with architecture and cpu type should not appear in the list.
3. Now try to edit an existing non running VM via details card and try to change its cluster.
4. make sure that issue 2 enforced for edit vm as well.

isaranova:
LGTM

PR #1319, opened 2020-10-29
Enable DUMMY_LOCALE via "?locale=aa" on URL

sjd78:
The locale coercion function needs to be aware of the dummy locale
("aa") to allow testing using the dummy locale.  So now the following
should work for testing with the dummy locale:

    > yarn intl:dummy
    > ENGINE_URL=https://localhost:8443/ovit-engine yarn start

 Open VM Portal URL with the dummy locale:
 * http://localhost:3000?locale=aa

### DIFF
--- a/src/intl/initialize.js
+++ b/src/intl/initialize.js
@@ -69,7 +69,7 @@ function coerceToSupportedLocale (locale: string): ?string {
     return 'en'
   }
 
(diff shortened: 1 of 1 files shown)


hstastna (approved):
Tested, LGTM.

Issue #1237, opened 2020-06-17
Create VM Wizard - For creating new disks, the 'disk type' and 'disk format' default values should be fixed


sgratch:
The disk type and disk format for new disks created by the create vm wizard should be fixed.

- For new disks not based on a template, the default disk type should be "thin provision" and not reallocated (the same as for edit vm->create disk).

- For new disks based on a template, the provisioning is based on the "optimized for" field in a way that:
If if it's Server or High Performance then provisioning=true (i.e. disks from the template are cloned)
If it's Desktop then provisioning=false (i.e. disks from template are thin provisioned).

This issue was raised and discussed in the following comments: 
https://github.com/oVirt/ovirt-web-ui/pull/1218#issuecomment-640753874
https://github.com/oVirt/ovirt-web-ui/pull/1218#issuecomment-641049650
https://github.com/oVirt/ovirt-web-ui/pull/1232#issuecomment-676371674

PR #1323, opened 2020-11-05
Create VM - Force template disk storage domain selection and change default diskType for Template and ISO provisioned VMs

sjd78:
### Overview
Merge diskType PR #1316 and create vm template disk storage domain selection PR #1233 and fix the the VM creation saga.

Now a new VM can be created where a user has access to a template but not to the storage domain of the template's disks.  The saga modifications for diskType handles this scenario properly.

### Disk Type changes
Fixes: #1237

Following the description in issue #1237, change the diskType as per the template VM type and where disks will be created.

Summary of changes:

  - **Storage**: When creating a new disk, default `diskType` to 'thin'.  This matches behavior in VM Details Disk card.

  - **CreateVmWizard**: 
    - **diskType**: When preparing Template Disks for the Storage card, default the disk's `diskType` to the template disk's sparse flag when the template is `optimizedFor` (VM type of) "desktop" ~~and the user can use the disk's storage domain~~.  When `optimizedFor` is not "desktop", ~~or if a "desktop" disk resides in a storage domain the user does not have access to use,~~ set disk type to preallocated.
    - **clone disks**: When `optimizedFor` is Server or if a disk needs to change storage domains, the VM's creation is set to clone the disks.

  - **composeAndCreateVm()** saga: To be able to have a template disk created in a different target storage domain, or to change its `diskType`/sparsity on creation, disks need to specified on the VM's creation JSON object.  The compose sagas and supporting functions now support this kind of creation.

### Storage domain selection for template disks in a storage domain that the creating user does not have disk create permits
Fixes: #1230 

Added validation to storage domain for template defined disks.

Previously, in the case the template's disk is in a storage domain for which the user is not permitted to create a disk the Create Vm Wizard will be displayed with 'UNKNOWN' storage domain.  The user can create this vm with no validation or Error handling.  For example:
image

The new feature will force the user to choose new a storage domain for which they can create a disk.  The Storage step will disable the next button as long as the user has not selected a valid storage domain for the template disk.

Initially opening the Storage step, this is the new storage domain dropdown with the disabled "Next" button:
image

Storage domain dropdown expanded:
image

After choosing a storage domain, the "Next" button is enabled:
image

### DIFF
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -356,7 +356,7 @@ class CreateVmWizard extends React.Component {
               : template.get('disks', List())
                 .map(disk => {
                   const canUserUseStorageDomain =
-                    storageDomains.getIn([ disk.get('storageDomainId'), 'canUserUseDomain' ], false)
+                    !!filteredStorageDomainList.find(sd => sd.id === disk.get('storageDomainId'))
 
                   const diskType = // constrain to values from createDiskTypeList()
                     template.get('type') === 'desktop' // optimizedFor
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -348,7 +348,10 @@ class CreateVmWizard extends React.Component {
         }
 
         if (resetDisks) {
-          const { storageDomains } = this.props
+          const dataCenterStorageDomainsList = createStorageDomainList(
+            this.props.storageDomains,
+            this.state.steps.basic.dataCenterId)
+
           draft.steps.storage = {
             updated: (draft.steps.storage.updated + 1),
             disks: !template
@@ -356,7 +359,7 @@ class CreateVmWizard extends React.Component {
               : template.get('disks', List())
                 .map(disk => {
                   const canUserUseStorageDomain =
-                    !!filteredStorageDomainList.find(sd => sd.id === disk.get('storageDomainId'))
+                    !!dataCenterStorageDomainsList.find(sd => sd.id === disk.get('storageDomainId'))
 
                   const diskType = // constrain to values from createDiskTypeList()
                     template.get('type') === 'desktop' // optimizedFor
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -348,28 +348,39 @@ class CreateVmWizard extends React.Component {
         }
 
         if (resetDisks) {
-          const filteredStorageDomainList = createStorageDomainList(this.props.storageDomains, this.state.steps.basic.dataCenterId)
+          const { storageDomains } = this.props
           draft.steps.storage = {
             updated: (draft.steps.storage.updated + 1),
             disks: !template
               ? []
               : template.get('disks', List())
-                .map(disk => ({
-                  id: disk.get('attachmentId'),
-                  name: disk.get('name'),
+                .map(disk => {
+                  const canUserUseStorageDomain =
+                    storageDomains.getIn([ disk.get('storageDomainId'), 'canUserUseDomain' ], false)
 
-                  diskId: disk.get('id'),
-                  storageDomainId: disk.get('storageDomainId'),
-                  canUserUseStorageDomain: this.props.storageDomains.getIn([ disk.get('storageDomainId'), 'canUserUseDomain' ], false),
-                  createdFromInvalidTemplate: !filteredStorageDomainList.find(sd => sd.id === disk.get('storageDomainId')),
+                  const diskType = // constrain to values from createDiskTypeList()
+                    template.get('type') === 'desktop' // optimizedFor
+                      ? canUserUseStorageDomain
+                        ? disk.get('sparse') ? 'thin' : 'pre'
+                        : 'pre'
+                      : 'pre'
 
-                  bootable: disk.get('bootable'),
-                  iface: disk.get('iface'),
-                  type: disk.get('type'), // [ image | lun | cinder ]
-                  diskType: disk.get('sparse') ? 'thin' : 'pre', // constrain to values from createDiskTypeList()
-                  size: disk.get('provisionedSize'), // bytes
-                  isFromTemplate: true,
-                }))
+                  return {
+                    id: disk.get('attachmentId'),
+                    name: disk.get('name'),
+
+                    diskId: disk.get('id'),
+                    storageDomainId: disk.get('storageDomainId'),
+                    canUserUseStorageDomain,
+
+                    bootable: disk.get('bootable'),
+                    iface: disk.get('iface'),
+                    type: disk.get('type'), // [ image | lun | cinder ]
+                    diskType,
+                    size: disk.get('provisionedSize'), // bytes
+                    isFromTemplate: true,
+                  }
+                })
                 .toJS(),
           }
         }
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -9,6 +9,7 @@ import * as Actions from '_/actions'
 import { generateUnique } from '_/helpers'
 import { msg } from '_/intl'
 import { handleClusterIdChange } from './helpers'
+import { createStorageDomainList } from '_/components/utils'
 
 import NavigationConfirmationModal from '../NavigationConfirmationModal'
 import BasicSettings from './steps/BasicSettings'
@@ -347,6 +348,7 @@ class CreateVmWizard extends React.Component {
         }
 
         if (resetDisks) {
+          const filteredStorageDomainList = createStorageDomainList(this.props.storageDomains, this.state.steps.basic.dataCenterId)
           draft.steps.storage = {
             updated: (draft.steps.storage.updated + 1),
             disks: !template
@@ -359,6 +361,7 @@ class CreateVmWizard extends React.Component {
                   diskId: disk.get('id'),
                   storageDomainId: disk.get('storageDomainId'),
                   canUserUseStorageDomain: this.props.storageDomains.getIn([ disk.get('storageDomainId'), 'canUserUseDomain' ], false),
+                  createdFromInvalidTemplate: !filteredStorageDomainList.find(sd => sd.id === disk.get('storageDomainId')),
 
                   bootable: disk.get('bootable'),
                   iface: disk.get('iface'),
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -362,10 +362,8 @@ class CreateVmWizard extends React.Component {
                     !!dataCenterStorageDomainsList.find(sd => sd.id === disk.get('storageDomainId'))
 
                   const diskType = // constrain to values from createDiskTypeList()
-                    template.get('type') === 'desktop' // optimizedFor
-                      ? canUserUseStorageDomain
-                        ? disk.get('sparse') ? 'thin' : 'pre'
-                        : 'pre'
+                    this.state.steps.basic.optimizedFor === 'desktop'
+                      ? disk.get('sparse') ? 'thin' : 'pre'
                       : 'pre'
 
                   return {
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -104,6 +104,12 @@ class Storage extends React.Component {
     this.isBootableDiskTemplate = this.isBootableDiskTemplate.bind(this)
     this.isEditingMode = this.isEditingMode.bind(this)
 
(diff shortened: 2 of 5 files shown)


sjd78:
copy of comment https://github.com/oVirt/ovirt-web-ui/pull/1233#issuecomment-722565975:

> > * **For creating a disk based on a template** - this is blocked by #1237 since currently the option to create a disk based on a template is not fully supported.  Once #1237 is done, we'll need to enable changing the SD and disk type for template's disks for supporting this issue. Or we can display an error and avoid the user from creating such a VM.
> 
> This PR adds the SD selection, but the create VM saga was not adjusted to handle the change. My PR #1316, which fixes the types and updates the VM create saga, hopefully handles the saga changes needed in this PR. I will need to do some extra testing with both PRs applied to verify.

@sgratch , @bamsalem - This is now working in every case that I tested.  There may be some other edge cases, but at this point I'm not sure it is worth finding and handling them.  We should be well covered as is over in #1323.

> 
> > * **For creating a disk not based on a template**, when creating a disk for a user that has no "_diskprofileuser_" permissions for the SD, the ui allows me to choose that SD and the disk creation failed only on backend due to permissions issue.
> >   It seems that the logged user should have permissions for the SD and also should be allowed to manage disks on that SD.
> >   Do we verify both with the API query?
> 
> Remember, we check based on permit and not on permission, and `canUserUseDomain` is checking if the user has the permit `create_disk`. It is quite possible a user picks up the necessary permit via another permission/role. See:
> 
> Sounds like we'll need to adjust the permit checking and add a `canUserCreateDisks` type of check and use that in various places. Or at least update `canUserUseDomain`'s check.

@sgratch , @bamsalem - The storage domain drop down was displaying the storage domains that I expected to see when unit testing #1323.  It was not a comprehensive test, more of an average path set of tests.  I would not hold up #1323 (which will take the place of this PR) with additional permit checking concerns.

sjd78:
copy of comment https://github.com/oVirt/ovirt-web-ui/pull/1316#issuecomment-716574907:

> The disk default values look ok to me. However, in the Create VM Wizard, when choosing the template with "Optimized For" field set to "Desktop", I can see the new disks based on the template with thin provision, but if I change the "Optimized For" to "Server" field, nothing changes regarding the disks. The default is still thin provision, for the new disks. Is this behavior expected? I'm not sure.

@hstastna - Ah, I'm looking at the template's `optimizedFor` value, not the one on basic settings.  So switching it up on basic would have no effect on the template's `diskType`.

@sgratch - I think it is safer to consider the template's `optimizedFor` (VM type) instead of the basic details value.  WDYT?  It is an easy change to use user selected, but I'm not sure it would be correct.

bamsalem (dismissed):
@sjd78 ,
I found few bugs/issues in this PR:
1. User can't create VM from template with disk inside SD which he doesn't have permissions to it: when I tried to create VM from template that the user has no permissions to one of the disks' SD I couldn't change it in the Storage step and I got "CREATE_VM failed [Cannot add VM. The user doesn't have permissions to attach Disk Profile to the Disk.]" error later.
StorageStep
reviewStepError
It seems like it's not identify that user don't have permissions to create disk in this SD,
maybe we need to create new parameter or edit the `canUserUseStorageDomain` parameter like you mentioned in this [comment](https://github.com/oVirt/ovirt-web-ui/pull/1323#issuecomment-723113216).
2. I found a regression in the Storage step when user creates disk, after saving the disk the storage domain appears as UNKNOWN and not with the storageDomain name:
image

sjd78:
>     2. I found a regression in the Storage step when user creates disk, after saving the disk the storage domain appears as UNKNOWN and not with the storageDomain name:

Fixed with the latest commit d88149f

sjd78:
> 1. User can't create VM from template with disk inside SD which he doesn't have permissions to it: when I tried to create VM from template that the user has no permissions to one of the disks' SD I couldn't change it in the Storage step and I got "CREATE_VM failed [Cannot add VM. The user doesn't have permissions to attach Disk Profile to the Disk.]" error later.
>
>  It seems like it's not identify that user don't have permissions to create disk in this SD, maybe we need to create new parameter or edit the `canUserUseStorageDomain` parameter like you mentioned in this [comment](https://github.com/oVirt/ovirt-web-ui/pull/1323#issuecomment-723113216).
> 

After checking user permissions (and therefore permits) against your test case, turns out that to create a VM from a template and clone its disk to a new storage domain, the user needs roles "create_disk", "attach_disk", and "attach_disk_profile".  I'll update the permission check function with that info.

hstastna:
Isn't this issue related to this PR? https://github.com/oVirt/ovirt-web-ui/issues/1329

bamsalem (commented):
I'm not sure if it's should take so much time but the VM (from template, the template based disk created on different SD) is still on status "image locked" (10 minutes after creation).
Probably it's not related to this PR but it should be raised (not sure if it's a bug or if it should takes so long).
This PR LGTM.
image

sgratch:
@sjd78 Overall looks good. Few comments:

1. in case no SD can be selected when creating a disk (either due to permissions issues or due to other reasons) then the "select storage domain" menu option is marked in red but there is no option to select and next button is disabled. I guess we just need to change the menu option text to "no SD available" in such a case.

2. > I think it is safer to consider the template's optimizedFor (VM type) instead of the basic details value. WDYT? It is an easy change to use user selected, but I'm not sure it would be correct.

   The correct behavior should be to consider the current value set within the basic details step (if it's taken from the template or overridden by the user - the actual type of the created VM will be the one set on basic settings and the logic for provisioning the disks should be based on that type 
(e.g. if the template's type is "Desktop" but the user tries to create a VM based on that template and only changes the "optimized for" field to be "Server" on basic settings, then the VM will be created as "Server" type and therefore the template's disks should be cloned regardless to original tempate's type).

3. When trying to create a VM based on a "server" template then the VM creation failed on backend with the following error:
image

4.  Buttons enable/disable issue - when creating disks based on a template with no permissions for original SD then the next button is enabled even if not all SD is chosen for the disks:
image


5.  Buttons enable/disable issue - when creating disks based on a template with no permissions for original SD then after clicking next and then previous, the next/previous buttons are disabled without any real reason:
image

sgratch (changes requested):
Please check my long comment with 6 issues mentioned above.

sjd78:
> I'm not sure if it's should take so much time but the VM (from template, the template based disk created on different SD) is still on status "image locked" (10 minutes after creation).
> Probably it's not related to this PR but it should be raised (not sure if it's a bug or if it should takes so long).
> This PR LGTM.

It is not a bug, just a function of how long it takes to copy the disks from one storage domain to another.  I'm not sure how optimized that part is, but worst case is the entire image is copied.  So, worst case, a 40GB image will still need to copy 40GB from one place to another.  Depending on lots of factors(system load, SD disk speed, network speed, etc), that can take quite a lot of time.

Side-note: if the disk is a sparse file and the copy is sparse aware, then only the actual data stored needs to be copied.  `rsync --sparse` does something similar.  So if you ever need to copy VM disk images from place to place, for sure use `rsync --sparse` and save yourself some time.

sjd78:
> @sjd78 Overall looks good. Few comments:
> 
>    1. in case no SD can be selected when creating a disk (either due to permissions issues or due to other reasons) then the "select storage domain" menu option is marked in red but there is no option to select and next button is disabled. I guess we just need to change the menu option text to "no SD available" in such a case.
>

Done:
VM Portal - Google Chrome_050


> 
>    2. > I think it is safer to consider the template's optimizedFor (VM type) instead of the basic details value. WDYT? It is an easy change to use user selected, but I'm not sure it would be correct.
>        
>        
>        The correct behavior should be to consider the current value set within the basic details step (if it's taken from the template or overridden by the user - the actual type of the created VM will be the one set on basic settings and the logic for provisioning the disks should be based on that type
>        (e.g. if the template's type is "Desktop" but the user tries to create a VM based on that template and only changes the "optimized for" field to be "Server" on basic settings, then the VM will be created as "Server" type and therefore the template's disks should be cloned regardless to original tempate's type).
>

So there are 2 different things going on.  First thing is, what `diskType` to use: Thin or Preallocated.  For new disks, we default to Thin.  For template disks, if `optimizedFor` is Desktop, use the template disk's diskType.  If `optimizedFor` is Server, default to Preallocated.  We don't allow `optimizedFor` High Performance so nothing to do there.

...link to the code block, CreateVmWizard 364...

Second things is if the template disks will be cloned (by setting clone=true on the API call) at VM creation.  The `clone` parameter is set true if creating from a template and `optimizedFor` is not desktop or if a template's disks/disk attachments need to be modified at creation.  For example, if a disk's storage domain is changed, the disks will be cloned.

...link to the code block, vmChanges 84-86, 189-230...


> 
>    3. When trying to create a VM based on a "server" template with **two** disks OR when trying to create a VM based on a template with no permissions for the template's **two** disks SD, then for both cases the VM creation failed on backend with the following error (I guess the issue is due to the two disks that needed to be cloned at once, while image is locked or something like that):
> 

Fixed!  There was an error in the way multiple disk changes were being merged to the new VM request.  Creation of VMs from multi-disk templates cloning the disk to a new storage domain works just fine now.

>    4. Buttons enable/disable issue - when creating disks based on a template with no permissions for original SD then the next button is enabled even if not all SD is chosen for the disks:
>

Storage tab validation has been updated to handle this case.

> 
>    5. Buttons enable/disable issue - when creating disks based on a template with no permissions for original SD then after clicking next and then previous, the next/previous buttons are disabled without any real reason:
>

Storage tab validation has been updated to handle this case.

> 
>    6. > When optimizedFor is not "desktop", or if a "desktop" disk resides in a storage domain the user does not have access to use, set disk type to preallocated.
>        
>        
>        This is not correct. When cloning  a disk, you should keep the same disk type as for the original copied disk.
>        Check this comment as well:[#1218 (comment)](https://github.com/oVirt/ovirt-web-ui/pull/1218#issuecomment-640753874) and specifically this:
>        "if I create a VM based on a template with "optimized for" field set to Server/High Performance then the VM's disk should be
>        created as Clone/Independent and uses the exact same Allocation Policy as the disk's template.."

This is very similar to point 2 above.

For an `optimizedFor` Desktop VM, template disks are only cloned if they need to change storage domain.  The diskType will use the same diskType as the Template disks.

For an `optimizedFor` Server VM, template disks are always cloned.  The diskType will always be Preallocated.

sjd78:
CI will work after #1346 is merged and this PR is rebased.

rszwajko:
I was able to test some scenarios (not all) and encountered no problems except of one case (see below).

Please confirm if this is the expected behaviour:
1. choose a template with 2 disks (i.e. eee1-desktop)
2. one of the disks is marked as bootable
3. add a 3rd custom disk
    Expected result: it's not possible to move bootable flag to 3rd disk
   Actual result: it's possible to change and it is respected during VM creation
4. delete the newly added 3rd disk
   Expected result: bootable flag is moved to some other disk
   Actual result: no disk is marked bootable (but after creation one of the disks is bootable)

no_bootable

sjd78:
> I was able to test some scenarios (not all) and encountered no problems except of one case (see below).
> 
> Please confirm if this is the expected behaviour:
> 
>    1. choose a template with 2 disks (i.e. eee1-desktop)
> 
>    2. one of the disks is marked as bootable
> 
>    3. add a 3rd custom disk
>        Expected result: it's not possible to move bootable flag to 3rd disk
>        Actual result: it's possible to change and it is respected during VM creation
> 
>    4. delete the newly added 3rd disk
>        Expected result: bootable flag is moved to some other disk
>        Actual result: no disk is marked bootable (but after creation one of the disks is bootable)
> 

I verified that it works the same way on master currently.  So while it is strange behavior, it isn't a regression by this PR.

sjd78:
@rszwajko  - See #1347 for the two disk template bootable fag issue.

rszwajko (approved):

hstastna (approved):
I've tested some of the scenarios, worked as expected, LGTM.

sgratch:
@sjd78 I found out that there are still few issue with the creation of a VM with disks and I'll open a new follow up issue for it: 

**1.** When creating a VM based on a template while the template includes more than 1 disk and is optimized to "server" or a clone is required due to SD permissions then the following error still occurs:
"CREATE_VM failed [Cannot add VM. One or more provided storage domains are either not in active status or of an illegal type for the requested operation.]"

This was already mentioned above on issue 3 of https://github.com/oVirt/ovirt-web-ui/pull/1323#issuecomment-728239827 and claimed to be fixed:
>  3.       When trying to create a VM based on a "server" template with two disks OR when trying to create a VM based on a template with no permissions for the template's two disks SD, then for both cases the VM creation failed on backend with the following error (I guess the issue is due to the two disks that needed to be cloned at once, while image is locked or something like that):
> 
> Fixed! There was an error in the way multiple disk changes were being merged to the new VM request. Creation of VMs from multi-disk templates cloning the disk to a new storage domain works just fine now.


**2.** when there are no permissions for all SDs , both previous and next buttons are disabled:
image


**3.** when there are no permissions for a template disks SD and the user can choose a different SD, the previous button is disabled till all SD are selected. This is wrong:
image


**4.** when the user tries to create a disk not based on  a template and there are no permissions for current SDs, the SD field is marked as red and buttons are disabled, but no message is displayed:
image

isaranova:
LGTM

PR #1324, opened 2020-11-05
Add 'yarn intl:check-use' script

sjd78:
This script will use `jq` and `grep` to do a simple source search
to check if each message, defined in `src/intl/messages.js`, is
actually used in a source file.  It is helpful to be able to remove
intl message keys that are not being used anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,8 @@
     "intl:pull": "scripts/intl/intl-zanata-pull.sh",
     "intl:apply": "scripts/intl/intl-apply.sh",
     "intl:report": "babel-node scripts/intl/report.js --presets env,flow",
-    "intl:sync": "babel-node scripts/intl/sync-messages.js --presets env,flow"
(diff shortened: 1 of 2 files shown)


hstastna (commented):
Awesome! I mean, I do consider this script as very useful. And it's working well.

I've created some unused messages in _messages.js_ and tested. They were found successfully, UNUSED KEY. Also they weren't found as unused when I used them, which is correct. Also tested without them, many unused message keys were found in our project, I've chosen a few of them, checked if we don't use them, and we really don't.

@sjd78 and what about getting the number of all unused keys and displaying the number at the end? To prevent unnecessary scrolling in the terminal if there are no unused keys ;)

sjd78:
@hstastna - I added key counts to the output.  Good suggestion.

Intersting note, I ran in to this issue: [I set variables in a loop that's in a pipeline. Why do they disappear after the loop terminates?](http://mywiki.wooledge.org/BashFAQ/024).  I used the process substitution solution.

rszwajko (approved):

PR #1332, opened 2020-11-17
Use DELETE method to remove SSH Public key

rszwajko:
Removing the key using PUT(with empty content) is an undocumented
[REST API](http://ovirt.github.io/ovirt-engine-api-model/4.4/#services/ssh_public_key) feature and will be dropped in the future versions of the API.

### DIFF
--- a/src/ovirtapi/index.js
+++ b/src/ovirtapi/index.js
@@ -481,12 +481,20 @@ const OvirtApi = {
   saveSSHKey ({ key, userId, sshId }: { key: string, userId: string, sshId: ?string }): Promise<Object> {
     assertLogin({ methodName: 'saveSSHKey' })
     const input = JSON.stringify({ content: key })
-    if (sshId !== undefined && sshId !== null) {
+    if (sshId && key) {
+      // update existing key
       return httpPut({
         url: `${AppConfiguration.applicationContext}/api/users/${userId}/sshpublickeys/${sshId}`,
         input,
       })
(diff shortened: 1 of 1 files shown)


sgratch (commented):
@rszwajko looks ok except one issue:
it seems to me that current fix will more likely fail sequential ssh keys updates between refresh intervals after deleting the ssh key.
The problematic scenario is:
1. enter user settings dialog
2. remove the ssh key (set value to empty).
3. set ssh key to a non empty val - succeed.
4. set ssh key to a non empty val before next sync occurs - fail with error: "Cannot add User Profile. User profile already created."

The reason for that is that removing the ssh key in step 2 triggers now a DELETE http method and therefore from now on every update to the ssh key will trigger a POST request instead of a PUT that was triggered before this fix (since ssh_id is still empty). Only after the next refresh interval then the new ssh_id is fetched.

I suggest triggering a getSSHKey call immediatlly after adding one (POST) for preventing this.

rszwajko:
> it seems to me that current fix will more likely fail sequential ssh keys updates between refresh intervals after deleting the ssh key.

Yes. The patch #1088 needs to be updated to prevent that.
Current code is not affected because we fetch SSH key on dialog open.

sgratch (approved):
> Yes. The patch #1088 needs to be updated to prevent that.
> Current code is not affected because we fetch SSH key on dialog open.

Agreed. Therefore this fix LGTM

PR #1333, opened 2020-11-18
Fix tooltip propType errors

sjd78:
Fixed the use of `PropTypes.oneOfType` to wrap the tooltip
type in an array.  Since the base tooltip propType is required,
this form allows the tooltip to be optional on other components.


sgratch (approved):
LGTM

hstastna:
LGTM to me, too, and it fixes the errors I've found recently:
err1

PR #1334, opened 2020-11-25
Prevent duplicate refresh on page change

rszwajko:
Add following safeguards:
    1. merge 3 possible paths to refresh() into one: via scheduler.
       Both page change and manual refresh will restart the scheduler.
    2. refresh only the page that was used to start the scheduler.
       Subsequent page changes will have no effect on a running scheduler.
    3. check the scheduler ID against the current scheduler count and do a
       self-shutdown if a newer scheduler has been detected

### DIFF
--- a/src/sagas/background-refresh.js
+++ b/src/sagas/background-refresh.js
@@ -34,28 +34,27 @@ import { fetchIsoFiles } from './storageDomains'
  * This should be done at the time of navigation to the page, typically by the page router.
  */
 function* changePage (action) {
-  yield put(Actions.stopSchedulerFixedDelay())
   yield put(Actions.setCurrentPage(action.payload))
-  yield put(Actions.refresh({
-    onNavigation: true,
-    shallowFetch: true,
-  }))
-  yield put(Actions.startSchedulerFixedDelay())
+  yield put(Actions.startSchedulerFixedDelay({ pageRouterRefresh: true, targetPage: action.payload, startDelayInSeconds: 0 }))
+}
+
+function* refreshManually () {
+  const currentPage = yield select(state => state.config.get('currentPage'))
+  yield put(Actions.startSchedulerFixedDelay({ manualRefresh: true, targetPage: currentPage, startDelayInSeconds: 0 }))
 }
 
 /**
  * Invoke the correct refresh function based on the app's current page type.
  */
-function* refreshData (action) {
-  const currentPage = yield select(state => state.config.get('currentPage'))
+function* refreshData ({ payload: { targetPage, ...otherPayload } }) {
   const refreshType =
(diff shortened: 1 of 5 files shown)


rszwajko:
Note that this patch tries to solve race conditions only between schedulers.
There are also conflicts between infinite scroller and scheduler which manifest on slow network and/or fast refresh rate.
Example scenario:
1. enable slow network in dev tools i.e. "Slow 3G"
2. decrease scheduler delay to 5 sec
3. on the main page start (fast and continuous) scrolling down

Expected result: one refresh request at time is active
Actual result: multiple request are launched (both for individual VMs and gouped)

src/sagas/background-refresh.js
```diff
@@ -34,28 +34,27 @@ import { fetchIsoFiles } from './storageDomains'
...
   const refreshType =
-    currentPage.type === C.NO_REFRESH_TYPE ? null
-      : currentPage.type === undefined ? C.LIST_PAGE_TYPE
-        : currentPage.type
+    targetPage.type === C.NO_REFRESH_TYPE ? null
+      : targetPage.type === undefined ? C.LIST_PAGE_TYPE
+        : targetPage.type
 
-  console.info('refreshData() 🡒', 'refreshType:', refreshType, 'currentPage:', currentPage, 'payload:', action.payload)
+  console.info(`refreshData() 🡒 refreshType: ${refreshType} currentPage: ${targetPage} payload: ${otherPayload}`)
```
sjd78, line 55: Either swap back to the separated arguments or use `JSON.stringify()` to output the values.  As-is the output will always be:

```
refreshData() 🡒 refreshType: listPage currentPage: [object Object] payload: [object Object]
```

sjd78 (commented):
It does fix the multiple refreshes on a page change.  Only thing is to fix one `console.log` statement so it actually logs some values.  Good to go after that fix.

sjd78:
> Note that this patch tries to solve race conditions only between schedulers.
> There are also conflicts between infinite scroller and scheduler which manifest on slow network and/or fast refresh rate.
> Example scenario:
> 
> 1. enable slow network in dev tools i.e. "Slow 3G"
> 2. decrease scheduler delay to 5 sec
> 3. on the main page start (fast and continuous) scrolling down
> 
> Expected result: one refresh request at time is active
> Actual result: multiple request are launched (both for individual VMs and gouped)

Opened #1340 to capture this.

ovirt-infra:

sjd78:
ci add to whitelist

src/sagas/background-refresh.js
rszwajko: @sjd78  I was not aware of this difference! Good to know!

sgratch:
LGTM

Issue #1330, opened 2020-11-11
Refreshing the VM grid shows all the VMs as if the user was an administrator


isaranova:
Description of problem:
When entering the VM Portal, the VM grid contains only the VMs that user is supposed to see (permissions) and VM pools only are shown if the user did not allocate any VMs yet. However, after pushing the refresh button all the VMs including possible VM Pool VMs appear as if the user was an admin.

VM Portal version number:
ovirt-web-ui-1.6.6-0.20201030.gitf464801.el8.noarch

oVirt version number:
ovirt-engine-4.4.4-0.0.master.20201109161351.git5dac13e7e22.el8.noarch
ovirt-engine-4.4.3.10-0.1.el8ev.noarch

Browser and version:
firefox-82.0.2-1.fc31.x86_64

Steps to Reproduce:
1. Login as UserRole user
2. Click on refresh button
3. Check that you can still see only the VMs you have permissions for

Actual results:
User sees all the VMs and VM Pool VMs.

Expected results:
User sees only the VMs he has permissions for

isaranova:
After investigating the issue a bit more, I got some more reproduction steps:
- The issue seems to be involving only VM Pools.
- The issue seems to be present only if the user has global UserRole, directly applied UserRole to a specific VM or VM Pool does not reproduce the issue.
- Refresh done by the VM Portal itself does not always reproduce the issue.
- Hard refresh (F5) does "fix" the issue.
- VM Portal refresh does reproduce the issue always even after hard refresh.
- User can see all the VM Pool VMs and run all of them exceeding the limit of VM Pool VMs per user.
- The VM Pool card behaves correctly and after running more VM Pool VMs than the user is limited to, the VM Pool card disappears but all of the VM Pool VMs are still present and able to be run.

sjd78:
@isaranova , @sgratch - If a user has a global `UserRole`, why wouldn't they see all VMs in the system?  Wouldn't the role inherit to everything?

hstastna:
**Steps to reproduce:**
1. As an admin, create some VM pool in Admin portal ("The issue seems to be involving only VM Pools.")
2. Log in as a user with the _UserRole_ permission to VM Portal
=> you can see the VM Pool (step 1) (with the proper number of "Available VMs from this pool" for the user according to the limit, in the pool card), but not the VM Pool VMs by themselves
3. Click on the _Refresh_ button up right in the VM Portal
[...]

PR #1335, opened 2020-11-30
Display only VMs that user is supposed to see after manual refresh

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1330

For the regular fetch (which works correctly and is managed through the infinite scroller), `fetchVms` is called in `fetchByPage`, with the `page` set to `1`, at least. For the refresh fetch which doesn't work as expected (manual refresh, clicking on the _Refresh button_), `fetchVms` is called in `refreshListPage` and the `vmsPage` variable is set to `1` there, but although it is set, `page` [is missing](https://github.com/oVirt/ovirt-web-ui/blob/master/src/sagas/background-refresh.js#L88) when calling `fetchVms`, because of the `getVmsByCount` [VM action](https://github.com/oVirt/ovirt-web-ui/blob/master/src/actions/vm.js#L99). So I made a change in this action, set the page to `1` to fix the query which wasn't set correctly and caused displaying the disallowed VMs. Setting the `page` to `1` is ok because we get VMs by count and not by page (`getVmsByPage`) in this case.

**Before:**
pool_before

**After:**
pool_after

### DIFF
--- a/src/actions/vm.js
+++ b/src/actions/vm.js
@@ -101,6 +101,7 @@ export function getVmsByCount ({ count, shallowFetch = true }) {
     type: GET_VMS,
     payload: {
       shallowFetch,
+      page: 1,
       count,
     },
   }


sjd78 (approved):
LGTM

sgratch (changes requested):
Most issues on basic original problem mentioned on https://github.com/oVirt/ovirt-web-ui/issues/1330#issuecomment-731200784 are solved now by this fix, but there is still a refresh problem regarding the following (I'm not sure if it's related to latest changes or was there from start after pools redesign):
- User can see all the VM Pool VMs and run all of them exceeding the limit of VM Pool VMs per user.
- The VM Pool card behaves correctly and after running more VM Pool VMs than the user is limited to, the VM Pool card disappears but all of the VM Pool VMs are still present and able to be run.

For reproducing the problem:
- create a pool and run web-ui as a specific user A
- let user A grab a vm out of the pool.
- let user A manually stop that pool's vm via web-ui 

The expected result:
This stopped pool's VM is taken back to the pool and dismissed from user A dashboard and is not permitted to him anymore, as mentioned on stopped vm's warning notification and as should be:
image

The actual result:
This stopped vm is back to the pool on backend as it should, but it's still displayed on user A dashboard and is not dismissed, even after using manual refresh (pushing the refresh button). Only hard (F5) refresh dismissed that VM from user A view.

(So user A can still view/start/stop this VM even though it was back to the pool and in case other user takes that VM, user A can still view it and stopped it for him as if it is still owned by him).

hstastna:
@sgratch I was able to reproduce the problem mentioned above and to debug the code, and I think it is a different problem. I can see `fetchVms` returns the two VMs (which is correct), the query seems to be correct, too, but in `VmCardList` component, there's still data for three VMs in the `vms` prop of the component. I suggest to fix this problem separately as I'm not sure how long would it take to fix it.

sgratch:
> @sgratch I was able to reproduce the problem mentioned above and to debug the code a little bit, and I think it is a different problem. I can see `fetchVms` returns the two VMs (which is correct), the query seems to be correct, too, but in `VmCardList` React component, there's still data for three VMs in the `vms` prop of the component. I suggest to fix this problem separately as I'm not sure how long would it take to fix it, and as it is a different problem.

Sure, you can open a separate PR for handling this.
This PR looks ok so I'm moving it to QE.

Issue #1336, opened 2020-12-02
VM Portal crash when refreshing data or performing a VM action (SyntaxError: JSON.parse)


pnovotny:
Description of problem
VM Portal crashes when trying to refresh the VM data manually or when trying to perform a VM action (run/suspend/shutdown or go to VM detail).

VM Portal version number:
ovirt-web-ui-1.6.6-0.20201126.git909add4.el8.noarch

oVirt version number:
ovirt-engine-4.4.4.2-0.1.el8ev.noarch

Browser and version:
Firefox 83
Chromium 87

Steps to Reproduce:
1. Log into  VM Portal (tried with UserRole and SuperUser).
2. Wait for VMs to load.
3. Click the refresh icon in VM Portal 
  or go to a VM detail. 
  or try to run/suspend/shutdown a VM.

Actual results:
VM Portal crash:
[Firefox] `Uncaught saga error (store.js):  SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data` 
[Chromium] `Uncaught saga error (store.js):  SyntaxError: Unexpected token u in JSON at position 0`

Expected results:
No crash.

Additional info:
screenshot (Firefox)
![web-ui-refresh-crash](https://user-images.githubusercontent.com/314377/100857149-a435ff00-348c-11eb-835a-563718d4c24d.png)

hstastna:
I tried to reproduce this issue but it works as expected.

PR #1337, opened 2020-12-02
Update parsing of VM statistics (fetched with VM details)

sjd78:
This commit changes the check for the `disks.usage` statistic
to exclude any falsey datum value from further parsing when
processing a VM's statistics.

With changes on the master branch ovirt-engine from at least 4.4.4.2,
VM statistics for file system usage is responding slightly different.
As a result, the VmStatistics transform code was returning an
`undefined` for the `disks.usage` named statistic instead of the
anticipated `null`. Since the code was explicitly checking for a `null`
value, the `undefined` datum was not excluded from further attempted
parsing.  `undefined` can't be further parsed and the errors happen.

Fixes: #1336

### DIFF
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -326,12 +326,12 @@ const VmStatistics = {
         stat.values &&
         stat.values.value &&
         (stat.name === 'disks.usage'
-          ? stat.values.value[0].detail || null
+          ? stat.values.value[0] && stat.values.value[0].detail
           : stat.values.value.length === 1
             ? stat.values.value[0].datum
             : stat.values.value.map(value => value.datum))
 
-      if (stat.name === 'disks.usage' && datum !== null) {
+      if (stat.name === 'disks.usage' && !!datum) {
         datum = JSON.parse(datum)
         datum = datum.map(data => {
           data.total = convertInt(data.total)


hstastna (commented):
The changes LGTM. And how exactly did you reproduce the issue? Thanks!

sgratch (approved):
LGTM

sgratch:
@pnovotny can you please verify that crashing is not reproduced.

sjd78:
> The changes LGTM. And how exactly did you reproduce the issue? Thanks!

@sgratch found the patch on ovirt-engine that caused the problem: https://gerrit.ovirt.org/112146

So a most recent engine is required.  Then you need a VM, and you probably need to run the VM to get some statistics.  I tend to create a simple VM via webadmin.  Import a fedora cloud image from the "ovirt-image-repository" to your storage domain and then create a VM with it to quickly create a "real" VM.  If you want to use it/login to generate some real statistical data, you need to remember to add a user with cloud-init (webadmin > New VM > Initial Run tab > Cloud-init > Authentication section > User Name and Password fields).  Once that VM exists, the VM dashboard can show some stuff, or parsing can fail if this PR isn't applied.

sgratch:
@sjd78 @hstastna 

> So a most recent engine is required. Then you need a VM, and you probably need to run the VM to get some statistics. 

It was reproduced to me on a stopped VM as well...

pnovotny:
LGTM

Verified with Firefox 83 and Chromium 87 in the latest nightly build ovirt-web-ui-1.6.6-0.20201202.git896c042.el8.noarch and ovirt-engine-4.4.4.2-0.1.el8ev.noarch. 

VM Portal no longer crashes when refreshing the VM list manually or when performing a VM action, like run, suspend, shutdown. Going to VM detail page also works fine.

PR #1341, opened 2020-12-07
Refactor initial VM/Pool page loading and background refresh

sjd78:
Refactor initial VM/Pool page loading and background refresh

The major change is the `VmsList` will not display `VmCardList` until the `refreshListPage()` completes the initial data fetch.  This allows the fetch saga to complete BEFORE the infinite scroll component is displayed and can prematurily request additional pages.

Covers the most common case described in oVirt#1340.

Specific changes made:

  - The `refreshListPage()` saga is now responsible for loading the first page of VMs and Pools for `VmCardList`

  - `VmsList` will wait for the first page of VMs and Pools to be available before showing the `VmCardList`.  This will prevent the initial `getByPage()` saga from being called twice (once by the first `changePage()` and once by the infinite scroll component on `VmCardList`).

  - Refactored how `schedulerWithFixedDelay()` calls `refreshData()`. The call is now synchronous.  Now, the refresh saga will complete BEFORE the next run is scheduled.  This prevents any problems where the refresh saga takes longer to run than the refresh interval.

  - Saga `refreshData()` now explicitly updates the `lastRefresh` config value since the `REFRESH_DATA` action has been removed

  - Extract refresh interval change handling out of `PageRouter` into `RefreshIntervalChangeHandler`.  `PageRouter` now is only concerned with rendering the app's routes and keeping the `currentPage` in redux in sync with the route.

  - Documented the data flow for fetching the initial page's data in the `login` saga


rszwajko:
> 
>       Refactor the initial VM & Pool page loading, moving it from the `refreshListPage` saga to the `VmsList` component.  This
>       lets the React lifecycle start the fetching with a `useEffect` hooked called based on the initial state of the `vms` reducer.
>

This seems equivalent to fetching data on app start. In both cases we have a request fired once with no way of recovering in case of error (only page reload). I think that previously it worked differently - refresh is a periodic task so any failures could be corrected in the next run.  User could also trigger manual refresh.  
 
>
>       Refactor `refreshData` such that it only runs a background refresh if the app is quiet.  If the App is waiting on an active
>       request (i.e. it is waiting for the response to a REST call), skip the current data refresh.  For example, this will prevent a
>       refresh from running if the App is waiting on a page of VMs or waiting on a VM creation.
>

First some small improvement - it would be safer to postpone next refresh instead of skipping it i.e. by using some small fixed `startDelayInSeconds`.
Avoiding the refresh is for sure the right thing when active requests and next refresh are working on the same data. However consider a scenario where user changes the page. The refresh triggered by page change works on different data then main VM list((i.e. to account settings or to some degree to VM details) so they can work in parallel.  This is the idea behind tying scheduler to one page only ( #1334 ). By relying on `activeRequests` we say that all requests are conflicting which is a very strong limitation.
 
>       Refactor `refreshListPage` to only do a refresh operation of existing pages.  If the App is on page 0 for VMs or Pools, do
>       not do any refreshing.  React UI components will take care of requesting the initial page as needed.
> 

I like the approach of separating "existing" from "pending" VMs!  Initial fetch and infinite scroller could work on a different array("pending" VMs) then refresh("exisiting") so no conflicts would occur. Only after successful fetch that "pending" VMs could be appended to "existing" (atomic operation). On each refresh (also manual) we could check if initialization was successful and kick it again. @sjd78  What do you think ?

sjd78:
> >
> >   Refactor the initial VM & Pool page loading, moving it from the `refreshListPage` saga to the `VmsList` component.  This
> >   lets the React lifecycle start the fetching with a `useEffect` hooked called based on the initial state of the `vms` reducer.
> >
> 
> This seems equivalent to fetching data on app start. In both cases we have a request fired once with no way of recovering in case of error (only page reload). I think that previously it worked differently - refresh is a periodic task so any failures could be corrected in the next run. User could also trigger manual refresh.
> 

Maybe probably not right the way I changed it around.  I was trying to control where the page fetching would be fired from.  I tried it again this morning and the double fetch still happened because of infinite scrolling code.  In my specific case, the `fetchByPage` fetches 20 VMs in parallel to 2 Pools.  The Pools come back and get pushed to the redux store, and cause the page to render 2 cards before the VMs come back.  The 2 cards are not enough to push the infinite scroller's `sentinelRef` out of view.  While the sentinel is in view the `fetchByPage` is called.  That is how I end up with multiple pages fetched....

> >
> >   Refactor `refreshData` such that it only runs a background refresh if the app is quiet.  If the App is waiting on an active
> >   request (i.e. it is waiting for the response to a REST call), skip the current data refresh.  For example, this will prevent a
> >   refresh from running if the App is waiting on a page of VMs or waiting on a VM creation.
> >
> 
> First some small improvement - it would be safer to postpone next refresh instead of skipping it i.e. by using some small fixed `startDelayInSeconds`.

I don't think the subsequent refresh is important enough to push like that.

> Avoiding the refresh is for sure the right thing when active requests and next refresh are working on the same data. However consider a scenario where user changes the page. The refresh triggered by page change works on different data then main VM list((i.e. to account settings or to some degree to VM details) so they can work in parallel. This is the idea behind tying scheduler to one page only ( #1334 ). By relying on `activeRequests` we say that all requests are conflicting which is a very strong limitation.
> 

When a user changes the page the initial fetch used to be done outside of the scheduler and then the scheduler would start.  Now the scheduler gates all and is more complicated.

Yeah `activeRequests` catches everything but is convenient and available.  I guess it needs to be more complicated.  Make an explicit `changePageFetch` saga corresponding to a `refreshPage` saga.  Then `changePage` can run the `changePageFetch`, which has a "page_fetch_is_running" indicator in redux, and then the scheduler can run the `refreshPage` saga in future.  Fetching is different than refreshing, at least for the list page.

> > 
> >   Refactor `refreshListPage` to only do a refresh operation of existing pages.  If the App is on page 0 for VMs or Pools, do
> >   not do any refreshing.  React UI components will take care of requesting the initial page as needed.
> > 
> 
> I like the approach of separating "existing" from "pending" VMs! Initial fetch and infinite scroller could work on a different array("pending" VMs) then refresh("exisiting") so no conflicts would occur. Only after successful fetch that "pending" VMs could be appended to "existing" (atomic operation). On each refresh (also manual) we could check if initialization was successful and kick it again. @sjd78 What do you think ?

Seems like a bit of overkill.  A "fetch is running, don't fetch again" flag should be ok.  Failed fetches mostly will be handled just fine by the error boundary and force the user to reload the page anyway.

TL;DR - My approach is too coarse.  I'll push some things a bit lower and try to make fetch/refresh cycles for a page more explicit and serialized as necessary.

rszwajko:
>
>  A "fetch is running, don't fetch again" flag should be ok.
>

That's synchronization approach. I think that at our scale avoiding conflicts would be simpler. Using a global binary semaphore (like `activeRequests.isEmpty()`)  as mentioned above is a strong limitation. Introducing more specialized semaphores (i.e. "fetch of type A is running, don't do fetch of type A again") will complicate the flow.

>
>  I guess it needs to be more complicated. 
>

+1

bamsalem (commented):
I'm not sure that I totally understood, but it LGTM
+1

hstastna (commented):
The code changes look ok to me, I don't have any suggestions for improvements. +1 for using the `useEffect`. I like it. 

Anyway, @sjd78 do you have any plan to add more changes in this PR or leave it as it is now?

sgratch:
> TL;DR - My approach is too coarse. I'll push some things a bit lower and try to make fetch/refresh cycles for a page more explicit and serialized as necessary.

@sjd78 any progress with this PR?

sjd78:
@rszwajko, @hstastna, @sgratch - I reworked this PR quite a bit from the original solution.  It should be pretty solid now and have a much safer initial data loading path for the VMs and Pools card list.

@rszwajko - There are a few changes to the refresh scheduling, but I think they work well.  They shouldn't impact any other schedule tweaking functionality in place.  The previous work to try to synchronize the refresh fetching was thrown out in favor of being more controlled on the initial fetch side.

src/sagas/background-refresh.js
```diff
@@ -317,7 +336,6 @@ function* logoutAndCancelScheduler () {
 export default [
   takeEvery(C.START_SCHEDULER_FIXED_DELAY, startSchedulerWithFixedDelay),
   throttle(5000, C.MANUAL_REFRESH, refreshManually),
-  throttle(5000, C.REFRESH_DATA, refreshData),
```
rszwajko, line 320: This was a (suboptimal) way to merge 3 kinds of refresh sources: manual, time-based and page changes.
Try following scenario:
1. with one page of VMs loaded ( ~ 20 VMS) go to Account Settings
2.  toggle "Slow 3G" network
3. go back and forth quickly using keyboard (few times). Result: each change triggers a fetch. Previously only change that was slow enough triggered the fetch.
4.  stop at one screen. Result: queued fetches are still executed (even for the page that is not currently displayed). Previously only few "late" calls were still executed.

rszwajko (commented):
>  * The `refreshListPage()` saga is now responsible for loading the first page of VMs and Pools for `VmCardList`
> 
>  * `VmsList` will wait for the first page of VMs and Pools to be available before showing the `VmCardList`.  This will prevent the initial `getByPage()` saga from being called twice (once by the first `changePage()` and once by the infinite scroll component on `VmCardList`).
>

This worked as described for me. Note that this does not mean that infinite scroller is synchronized with page refresh in general - only initial load is handled (which is definitely an improvement). 
 
>  * Refactored how `schedulerWithFixedDelay()` calls `refreshData()`. The call is now synchronous.  Now, the refresh saga will complete BEFORE the next run is scheduled.  This prevents any problems where the refresh saga takes longer to run than the refresh interval.
> 

The saga is synchronous but the underlying call is not. See my comment about removing  `throttle(5000, C.REFRESH_DATA,...`.

Some thoughts (for follow-up PRs probably):
1. we do refresh on every page change - this is not really needed most of the time i.e. consider use case where user is entering VM Details one-by-one looking for some info. Each time when he is back at Dashboard a refresh is triggered although normally we would wait full refresh interval. Perhaps we should track refresh interval for each page separately ?
2. error handling of `getByPage` action - if the initial fetch fails there is no auto-recovery (however manual refresh works) and fetching guards (`vmsExpectMorePages`) seem incorrect. Under some error conditions (i.e. multiple calls failing) VM's are refreshed one-by-one which makes things even more messy.
3. synchronization between infinite scroller and page refresh - without this calls will overlap.

sjd78:
> > * The `refreshListPage()` saga is now responsible for loading the first page of VMs and Pools for `VmCardList`
> > * `VmsList` will wait for the first page of VMs and Pools to be available before showing the `VmCardList`.  This will prevent the initial `getByPage()` saga from being called twice (once by the first `changePage()` and once by the infinite scroll component on `VmCardList`).
> 
> This worked as described for me. Note that this does not mean that infinite scroller is synchronized with page refresh in general - only initial load is handled (which is definitely an improvement).

I mean that is a good point.  It is possible that `getByPage()` starts due to user scroll and while it is running `refreshListPage()` starts.  Then if `getByPage()` ends, pushes the page count up and then `refreshListPage()` ends resetting the page count back down by 1, the next refresh will include a bunch of single calls.

And ideas on how to synchronize the sagas?  I don't think there are really any underlying sync constructs in saga to rely on...

> 
> > * Refactored how `schedulerWithFixedDelay()` calls `refreshData()`. The call is now synchronous.  Now, the refresh saga will complete BEFORE the next run is scheduled.  This prevents any problems where the refresh saga takes longer to run than the refresh interval.
> 
> The saga is synchronous but the underlying call is not. See my comment about removing `throttle(5000, C.REFRESH_DATA,...`.
> 

That is also an interesting thing --- going quickly from page to page can queue up a bunch of `START_SCHEDULER_FIXED_DELAY` calls and the user could get ahead of the call queue quite easily, especially on a slow connection.  The problem could probably be solved by using `throttle(5000, ...)` or `takeLatest(...)` instead of `takeEvery(...)` on the scheduler or changePage actions.  That way the scheduled task wouldn't get stacked up itself instead of relying on the actual refresh from getting stacked up.  That one is best in a separate issue.

> Some thoughts (for follow-up PRs probably):
> 
>     1. we do refresh on every page change - this is not really needed most of the time i.e. consider use case where user is entering VM Details one-by-one looking for some info. Each time when he is back at Dashboard a refresh is triggered although normally we would wait full refresh interval. Perhaps we should track refresh interval for each page separately ?
> 

Or just an issue to add a "requireRefreshOnEntry" flag on the page or route.  Also a good followup issue.

>     2. error handling of `getByPage` action - if the initial fetch fails there is no auto-recovery (however manual refresh works) and fetching guards (`vmsExpectMorePages`) seem incorrect. Under some error conditions (i.e. multiple calls failing) VM's are refreshed one-by-one which makes things even more messy.
> 

I'm not really worried about this potentiality as I figure a user will be more apt to click the app refresh or normal browser refresh button if they don't see things "in a reasonable amount of time".  The app should default back to the `<NoVmAvailable />` display if the initial fetch does fail anyway.

>     3. synchronization between infinite scroller and page refresh - without this calls will overlap.

Yeah that one is tricky.  Feels like things get messy when the saga can't update the reducers all in one shot.

sjd78:
@rszwajko - I added debouncing to the scheduler action.  That should cover the prevention of most of the crazy user keyboard navigation quickly between pages causing multiple fetch to stack up.

src/sagas/background-refresh.js
```diff
@@ -312,13 +332,49 @@ function* scheduleResumingNotifications ({ payload: { delayInSeconds } }) {
...
+   *       to refresh the new page data each time, it is possible to get to a point where
+   *       multiple page refresh sagas are running at the same time.  That can cause
+   *       problems and slow down the app.  To help prevent this without rearchitecting
+   *       the scheduler to support canceling running fetches, the action to start the
+   *       scheduler is debounced.  Now quick page changes that happen within 0.5 seconds
+   *       of each other won't generate excess refresh fetches.
+   */
   takeLatest(C.CHANGE_PAGE, changePage),
-  takeEvery(C.LOGOUT, logoutAndCancelScheduler),
+  debounce(500, C.START_SCHEDULER_FIXED_DELAY, startSchedulerWithFixedDelay),
```
rszwajko, line 376: This (and above) `debounce()` solves the problem at the cost of slowing down every refresh. I think that the price is too high - I would give a try with replacing `debounce` with `throttle` or alternatively just ignore the problem and go back to original code here (assuming that we provide a fix in the near future).

sjd78: `throttle()` for the manual refresh makes sense ... work on the first one and ignore subsequent for the next 5 seconds.

`debounce()` still makes sense the start scheduler to make sure proper semantics are followed (i.e. only start the most recent one received in the debouce interval).  I'll turn the interval down to 1/4 second to make the delay less noticeable.

rszwajko (approved):

PR #1342, opened 2020-12-08
Change Consoles data fetching time - fetch Consoles data on demand instead of fetch on login

bamsalem:
This PR should test with engine that include this [gerrit patch](https://gerrit.ovirt.org/#/c/ovirt-engine/+/112536/) and will need to be merged after the patch is merged.
depended on gerrit patch https://gerrit.ovirt.org/112536.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1795457

This change include 4 sub changes:
1. Removes the graphic console from `getVms` request in case of `shallowFetch`, that action should reduce the CPU performance issue of the postgreSQL server. this change is in commit 4e52833 .
2. Adds actions, reducers and sagas.
Those changes allow fetch the consoles separately and included in commit 4e52833.
3. Adds new config value `defaultConsole` to the app config, this value is the system level `defaultConsole` in the future we can add more levels like `userDefaultConsole`, `vmDefaultConsole`, etc..
Those changes are in f60363c .
4. Adds new options to `vmActions` when it is on card 'Console' .
When User clicks on the 'Console' button 2 things happens: app makes an api call that fetches consoles data,
and when the call is finished it selects automatically console from the consoles list.

Error handling and few assumptions:
1. If it cannot select console from the list the app will use `ToastNotifications` to notify the user
2. If The list is empty it will notify the user.
3. In case The list includes only one protocol it will select this one automatically.
4. The default console is 'vnc' since 'spice' won't be supported from RHEL 8.3 .
5. In case of windows it won't select the RDP as the default console it will select the spice.
6. RDP is always shown in the `vmActions` list since it's not require any data from the consoles data.

Screenshots:
Before:
image

After:
image


ovirt-infra:

sjd78:
ci add to whitelist

src/sagas/index.js
```diff
@@ -183,6 +184,16 @@ export function* fetchByPage () {
   }))
 }
 
+export function* fetchConsoles ({ payload: { vmId } }) {
```
rszwajko, line 187: Exporting this function seems not necessary. Consider also removing `return` as this saga is not called directly.

src/components/VmActions/index.js
```diff
@@ -134,15 +136,40 @@ class VmActions extends React.Component {
...
-      priority: 0,
-      shortTitle: msg[c.get('protocol') + 'Console'](),
-      icon: <Icon name='external-link' />,
-      id: `${idPrefix}-button-console-${c.get('protocol')}`,
-      confirmation: (
-        <ConsoleConfirmationModal consoleId={c.get('id')} vm={vm} />
-      ),
-    })).toJS()
+    let consoles = []
+    if (isOnCard) {
```
rszwajko, line 140: `isCard` was already used before to separate card-scenario from toolbar scenario - however the `getDeafaultAction` step was common to both paths. Right know we no longer have a clean point where the scenarios diverge. Perhaps it's time to break this method into smaller ones i.e. `getDefaultActionsForCard()` and `getDefaultActionsForToolbar()` ?

rszwajko (changes requested):

src/components/VmActions/index.js
bamsalem: IMHO most of the logic is shared and both of the options using it, I think that we should not separate it,
but if most of the team thinks different I'll change it.

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,112 @@
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import PropTypes from 'prop-types'
+import { addUserMessage, downloadVmConsoles } from '_/actions'
+import ConsoleConfirmationModal from '_/components/VmActions/ConsoleConfirmationModal'
+import { msg } from '_/intl'
+
+class AutoSelectConsole extends Component {
```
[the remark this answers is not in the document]
bamsalem: > 1. it's a new code so I would stick to the functional component pattern

Done.
> 2. `setTimeout` usually requires some `clearTimeout` or we risk triggering action on unmounted component

we don't use `setTimeout` anymore
> 3. I've seen following error when clicking Console button
> 
> > Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.

`onFailure` is updating the state, I moved the call out from `render()`

src/intl/messages.js
```diff
@@ -61,7 +61,12 @@ export const messages: { [messageId: string]: MessageType } = {
   bootSequence: 'Boot Sequence',
   cancel: 'Cancel',
   cannotUpdateCloudInitHostname: 'Host name cannot be synchronized with VM name because the VM name is not a valid host name.',
+  cantDownloadConsoleEmptyList: 'There are no consoles assigned to this VM.',
+  cantDownloadConsoleGeneral: "Can't download console:",
```
sgratch, line 65: Please rephrase to something more user friendly like: "can't open a graphical console since this is  a headless VM"
Please also add the VM name to the message.

sgratch (dismissed):
Few comments:

- You didn't read the value of the ClientModeVncDefault config val for deciding which VNC mode to use (native or noVNC) as we already discussed.
- In case of a Windows OS, the RDP seems to be neglected as a separate option, not related to other consoles which is confusing. e.g. in case of headless the "no consoles assigned" message is displayed even though there is the RDP option. I suggest renaming the "console" pull-down menu option in case of Windows base to something else and the error message as well - maybe to "non RDP Console"...
- Any reason why not removing the follow=graphics_consoles for manual refresh as well? We are anyway not using the graphics console info after manual refresh on vms list, right?
- After generating the "Can't download console: There are no consoles assigned to this VM." error when trying to open a headless VM console, this notification message re-appear and added to drawer once per each auto refresh. You need to somehow clear this.

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,133 @@
...
+          selectedProtocol = consoles.first()
+        } else {
+          selectedProtocol = consoles.find(c => c.get('protocol') === defaultConsole)
+          failureReason = !selectedProtocol && 'noDefault'
+        }
+      } else {
+        failureReason = 'emptyConsolesList'
+      }
+    }
+    return { selectedProtocol, failureReason }
```
rszwajko, line 82: Function can be made pure by moving `defaultConsole` to parameters. We could then move it outside the component and unit test. This would also allow to use it directly as hook dependency.

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,133 @@
...
+  const {
+    dispatchStatus,
+    selectedProtocol,
+    active,
+    // failureReason,
+    // consoles,
+  } = state
+
+  useEffect(() => {
+    if (show) {
```
rszwajko, line 38: Consider data driven design instead of introducing state flags (`active, dispatchStatus, show, ...`).
Basic flow:
1. No data -> fetch data (to the Redux store)
2. there is data -> render data

Check [VmCardList](https://github.com/oVirt/ovirt-web-ui/blob/8a5a26cb2c3310a67d05673d3482bf1a751cc0a9/src/components/VmsList/VmCardList.js#L44)

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,133 @@
...
+      newState.selectedProtocol = selectedProtocol
+      newState.failureReason = failureReason
+    }
+    if (active && selectedProtocol && dispatchStatus !== 'doNotDispatch') newState.dispatchStatus = 'dispatched' // cannot change from 'doNotDispatch' to 'dispatched'
+    if (active && !selectedProtocol && failureReason && !requestActive && dispatchStatus !== 'doNotDispatch') onFailure(failureReason)
+
+    setState(state => ({ ...state, ...newState }))
+  }, [vm, requestActive]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  const onClose = () => {
```
rszwajko, line 63: why we need `onClose` here?

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,133 @@
...
+  onFailure: PropTypes.func.isRequired,
+  id: PropTypes.string.isRequired,
+}
+
+export default connect(
+  (state) => ({
+    defaultConsole: state.config.get('defaultConsole'),
+    requestActive: !state.activeRequests.isEmpty(),
+  }),
+  (dispatch, { defaultConsole }) => ({
```
rszwajko, line 114: `defaultConsole` here is passed as `ownProps` but is retrieved from the store. Usually `ownProps` and store based are different categories.

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,133 @@
...
+    fetchConsoles: (vmId) => {
+      dispatch(downloadVmConsoles({ vmId }))
+    },
+    onFailure: (reason) => {
+      const generalMsg = msg.cantDownloadConsoleGeneral()
+      switch (reason) {
+        case 'emptyConsolesList':
+          dispatch(addUserMessage({ message: `${generalMsg} ${msg.cantDownloadConsoleEmptyList()}`, type: 'error' }))
```
rszwajko, line 122: I'd rather avoid concatenating messages in the code as this prevents proper internationalization (grammar differences, write order etc)

rszwajko (changes requested):

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,133 @@
...
+  useEffect(() => {
+    if (show) {
+      setState(state => ({
+        ...state,
+        dispatchStatus: 'shouldDispatch',
+        active: true,
+      }))
+      fetchConsoles(vm.get('id'))
+    }
+  }, [show]) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
bamsalem: IMO the vm, the fetchConsole shouldn't trigger the function since nothing happened unless show === true.
so why it should be in the dependency list? vm might change on any refresh but it shouldn't affect the AutoSelectConsole component. WDYT?

src/components/VmActions/AutoSelectConsole.js
bamsalem: @rszwajko  I'm not sure if it's the same case, in `vmCardList` data fetching should be triggered automatically, but in this case it should be triggered by click, so we must use flags to avoid fetching when it's not needed...

src/components/VmActions/AutoSelectConsole.js
bamsalem: To update the state before call the `props.onClose` function, but it's changed so take a look on the latest version :)

src/components/VmActions/AutoSelectConsole.js
bamsalem: Done.

src/components/VmActions/AutoSelectConsole.js
rszwajko: Usually in React/Redux a component just renders data from props. If we need more data then we dispatch an action that fetches missing data and saves it in the store. From there the components are notified via props.
Fetch logic (i.e. error handling) is placed in saga-layer.

In case of AutoSelect the rendered piece of data is `selectedProtocol` - we could place it in the store and populate via some kind of `fetchAndSelectConsole` saga. This would remove all error/state handling logic from the component. Perhaps it would even make the component obsolete since it's more a date provider then a regular component.

src/components/VmActions/AutoSelectConsole.js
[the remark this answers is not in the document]
rszwajko: OK. The we should provide VM ID as dependency(i.e. via props or local variable).
```....
    fetchConsoles(vmId)}}, [show, vmId])
```

src/components/VmActions/index.js
```diff
@@ -134,15 +136,40 @@ class VmActions extends React.Component {
...
+      })).toJS()
+
+      if (vncConsole) {
+        consoles.push({
+          priority: 0,
+          shortTitle: msg.vncConsoleBrowser(),
+          actionDisabled: config.get('websocket') === null,
+          id: `${idPrefix}-button-console-browser`,
+          confirmation: (
+            <ConsoleConfirmationModal isNoVNC consoleId={vncConsole.get('id')} vm={vm} />
```
rszwajko, line 168: Using `isNoVNC=true` for VNC is confusing. Perhaps we should re-name?
Additionally I think the intention here was probably to pass `isNoVnc` property not just `true`.
See [JSX syntax](https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true)

rszwajko (dismissed):

bamsalem:
@sgratch 
> Few comments:
> 
> * You didn't read the value of the ClientModeVncDefault config val for deciding which VNC mode to use (native or noVNC) as we already discussed.

Done.
> * In case of a Windows OS, the RDP seems to be neglected as a separate option, not related to other consoles which is confusing. e.g. in case of headless the "no consoles assigned" message is displayed even though there is the RDP option. I suggest renaming the "console" pull-down menu option in case of Windows base to something else and the error message as well - maybe to "non RDP Console"...

Done: 
image

> * Any reason why not removing the follow=graphics_consoles for manual refresh as well? We are anyway not using the graphics console info after manual refresh on vms list, right?

No reason, removed.
> * After generating the "Can't download console: There are no consoles assigned to this VM." error when trying to open a headless VM console, this notification message re-appear and added to drawer once per each auto refresh. You need to somehow clear this.

Thanks, I didn't notice that ...
It's been handled.

src/components/VmActions/index.js
```diff
@@ -134,15 +136,40 @@ class VmActions extends React.Component {
...
-        <ConsoleConfirmationModal consoleId={c.get('id')} vm={vm} />
-      ),
-    })).toJS()
+    let consoles = []
+    if (isOnCard) {
+      consoles.push({
+        priority: 1,
+        actionDisabled: isPool || !canConsole(status) || vm.getIn(['actionInProgress', 'getConsole']),
+        shortTitle: msg.console(),
+        onClick: (e) => { e.preventDefault() },
```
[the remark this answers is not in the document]
bamsalem: RDP will not launch via auto select, it's a separate option:
image

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,122 @@
...
+  requestActive: PropTypes.bool.isRequired,
+  onClose: PropTypes.func.isRequired,
+  onFailure: PropTypes.func.isRequired,
+  id: PropTypes.string.isRequired,
+}
+
+export default connect(
+  (state) => ({
+    defaultConsole: state.config.get('defaultConsole'),
+    requestActive: !state.activeRequests.isEmpty(),
```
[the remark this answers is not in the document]
bamsalem: It's something that I took from `VmList`, we can make it more specific (pseudo)
```
    requestActive: !!state.activeRequests.find(r => /^\/api\/vms\/([0-9]|[A-Z]|[a-z]|-)*\/graphicsconsoles$/.test(r)),
```
WDYT?

src/components/VmActions/index.js
bamsalem: It should be removed, I forgot to remove it:
>You didn't read the value of the ClientModeVncDefault config val for deciding which VNC mode to use (native or noVNC) as we already discussed.

Thanks

src/intl/messages.js
rszwajko: > "non RDP Console"...

I would rather not name a key functionality using Windows protocol. It looks as RDP was the default and other consoles were our second choice.

> the RDP seems to be neglected as a separate option, not related to other consoles which is confusing

@bamsalem  Can we treat RDP the same as other consoles? This would remove the source of confusion.

src/components/VmActions/AutoSelectConsole.js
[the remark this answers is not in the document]
rszwajko: It fixes the `any active request` part but it is still a hack. I would rather remove it unless there is really no other way.

rszwajko:
@sgratch  @sjd78 
we already have a slice in Redux store dedicated to console [options](https://github.com/oVirt/ovirt-web-ui/blob/8a5a26cb2c3310a67d05673d3482bf1a751cc0a9/src/sagas/console/index.js#L126). Until now it was used exclusively by auto-connect feature which is currently dead (but will probably get resurrected at some point). Auto-selecting protocol seems to fit there.

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,144 @@
...
+        handleFailure(selectedProtocol, vm.get('name'), hasRdp)
+        handleClose()
+      } else {
+        const consoles = vm.get('consoles')
+        newState.selectedProtocol = consoles.find(console => console.get('protocol') === selectedProtocol)
+        newState.dispatchStatus = DISPATCHED
+      }
+      setState(state => ({ ...state, ...newState }))
+    }
+  }, [vm, dispatchStatus, show])
```
rszwajko, line 104: eslint warning about dependencies

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,144 @@
...
+      }))
+      const vmId = vm.get('id')
+      fetchConsoles(vmId)
+    }
+  }, [show, fetchConsoles])
+
+  useEffect(() => {
+    if (show && dispatchStatus === SHOULD_DISPATCH) {
+      const newState = {}
+      const selectedProtocol = vm.get('defaultConsole')
```
rszwajko, line 86: `defaultConsole` is used here in 2 contexts: `config.defaultConsole` and `vm.defaultConsole`.
Perhaps we should rename the second (according to its usage) to `selectedProtocol`?

src/components/VmActions/index.js
```diff
@@ -134,15 +136,40 @@ class VmActions extends React.Component {
...
+      })).toJS()
+
+      if (vncConsole) {
+        consoles.push({
+          priority: 0,
+          shortTitle: msg.vncConsoleBrowser(),
+          actionDisabled: config.get('websocket') === null,
+          id: `${idPrefix}-button-console-browser`,
+          confirmation: (
+            <ConsoleConfirmationModal isNoVNC consoleId={vncConsole.get('id')} vm={vm} />
```
rszwajko, line 168: It's still there:)

rszwajko (changes requested):
@sgratch @sjd78 @hstastna  Take a look at the topics below.  I would like to hear your opinion about the direction we should take:
1.  [isCard ](https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r545258589)
2.  [RDP](https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r555746901)
3.  [console slice](https://github.com/oVirt/ovirt-web-ui/pull/1342#issuecomment-758660205)

src/components/VmActions/AutoSelectConsole.js
bamsalem: IMO use `selectedProtocol` can be confusing because inside `AutoSelectConsole` component it's the selectedConsole but in the context of the vm is the default, it's individual and some people might think different...

I leave it that way unless the rest of the team thinks different...

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,111 @@
...
+      if (!selectedProtocol || selectedProtocol === 'undefined') {
+        return
+      }
+
+      if (defaultConsoleProtocolErrorsList.includes(selectedProtocol)) {
+        handleClose()
+      } else {
+        const consoles = vm.get('consoles')
+        setState({
+          selectedProtocol: consoles.find(console => console.get('protocol') === selectedProtocol),
```
rszwajko, line 69: This seem to duplicate logic from `selectDefaultConsoleProtocol()`. Actually it seems to me that `vm.get('defaultConsole')` could be used directly as `props.selectedProtocol` without the need of storing it in local state (own props already contain `vm` so we can retrieve it in `connect()`).

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,111 @@
...
+          dispatchStatus: DISPATCHED,
+        })
+      }
+    }
+  }, [vm, dispatchStatus, show, handleClose])
+
+  return (
```
rszwajko, line 76: Can we return empty component when `!show || !selectedProtocol`? This would simplify the logic here - probably the dispatch status could be dropped.

rszwajko (changes requested):

sgratch:
> You didn't read the value of the ClientModeVncDefault config val for deciding which VNC mode to use (native or noVNC) as we already discussed.

It still doesn't work to me. The noVnc console is not activated.

>     In case of a Windows OS, the RDP seems to be neglected as a separate option, not related to other consoles which is confusing. e.g. in case of headless the "no consoles assigned" message is displayed even though there is the RDP option. I suggest renaming the "console" pull-down menu option in case of Windows base to something else and the error message as well - maybe to "non RDP Console"...

This looks a bit strange to display those 2 options of no rdp and rdp: 
image

In addition, in case of a headless vm then no console should be supported including RDP.

>     After generating the "Can't download console: There are no consoles assigned to this VM." error when trying to open a headless VM console, this notification message re-appear and added to drawer once per each auto refresh. You need to somehow clear this.

Now there is no error at all.

sgratch:
> You didn't read the value of the ClientModeVncDefault config val for deciding which VNC mode to use (native or noVNC) as we already discussed.

It still doesn't work to me. The noVnc console is not activated.

>     In case of a Windows OS, the RDP seems to be neglected as a separate option, not related to other consoles which is confusing. e.g. in case of headless the "no consoles assigned" message is displayed even though there is the RDP option. I suggest renaming the "console" pull-down menu option in case of Windows base to something else and the error message as well - maybe to "non RDP Console"...

This looks a bit strange to display those 2 options of non rdp and rdp:
image
Maybe chage to: spice/vnc console, rdp console

In addition, in case of a headless vm then no console should be supported including RDP.

>     After generating the "Can't download console: There are no consoles assigned to this VM." error when trying to open a headless VM console, this notification message re-appear and added to drawer once per each auto refresh. You need to somehow clear this.

Now there is no error at all.

sgratch:
> @sgratch @sjd78
> we already have a slice in Redux store dedicated to console [options](https://github.com/oVirt/ovirt-web-ui/blob/8a5a26cb2c3310a67d05673d3482bf1a751cc0a9/src/sagas/console/index.js#L126). Until now it was used exclusively by auto-connect feature which is currently dead (but will probably get resurrected at some point). Auto-selecting protocol seems to fit there.

This might be used when the real "console option settings" as part of account settings feature will be implemented per user with a backend persistently. For now we implement a very basic flow in which for all users we use a backend global settings for console options by querying the config vals.

src/intl/messages.js
sgratch: @rszwajko 

> > "non RDP Console"...
> I would rather not name a key functionality using Windows protocol. It looks as RDP was the default and other consoles were our second choice.
> 
> > the RDP seems to be neglected as a separate option, not related to other consoles which is confusing
> Can we treat RDP the same as other consoles? This would remove the source of confusion.

I agree. I suggested calling this: "spice/vnc console", "rdp console"

The problem with treating RDP the same as other consoles is that RDP is always enabled for windows based OS vms only, regardless to spice/vnc. So we'll need to add another config value for setting global default console for windows based vms and one for non windows based vms instead of one "ClientModeConsoleDefault". This will be required for account settings solution but I wanted to save that extra config parameter. If you think it's better then let's add this.

src/components/VmActions/index.js
sgratch: @bamsalem @rszwajko I don't have a strong opinion on this. The only difference is the consoles menu so not sure it worth separating in current phase.

sgratch:
@rszwajko 
>Take a look at the topics below. I would like to hear your opinion about the direction we should take
>     1. [isCard ](https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r545258589)
>     2. [RDP](https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r555746901)
>     3. [console slice](https://github.com/oVirt/ovirt-web-ui/pull/1342#issuecomment-758660205)

I answered all 3 topics above by replying each separately.

src/components/VmActions/index.js
rszwajko: ok. let's leave it as it is.

bamsalem:
@sgratch @rszwajko  
I made all the change you requested:
Sharon regarding headless VM:
Windows headless vm can use RDP in webAdmin.
The default for windows vm before these change is spice.
Regarding non RDP VM, I change it to "Console" and "Remote Desktop"
image

sgratch:
@bamsalem I didn't verify all  yet but found that 2 issue mentioned above https://github.com/oVirt/ovirt-web-ui/pull/1342#issuecomment-768462866 are still reproduced:

- The noVnc console is not activated.
   use case: 
   1. set ClientModeConsoleDefault=vnc, ClientModeVncDefault=NoVnc
   2. create a VM with vnc supported
   3. open console for the VM
   4. vnc console is activated instead of  noVnc

- In addition, in case of a headless vm, the console option appears as enabled but nothing happens when clicking on it - no error appears.

bamsalem:
> @bamsalem I didn't verify all yet but found that 2 issue mentioned above [#1342 (comment)](https://github.com/oVirt/ovirt-web-ui/pull/1342#issuecomment-768462866) are still reproduced:
> 
> * The noVnc console is not activated.
>   use case:
>   
>   1. set ClientModeConsoleDefault=vnc, ClientModeVncDefault=NoVnc
>   2. create a VM with vnc supported
>   3. open console for the VM
>   4. vnc console is activated instead of  noVnc
> * In addition, in case of a headless vm, the console option appears as enabled but nothing happens when clicking on it - no error appears.

@sgratch 
Some of my changes wasn't staged, please Review it again

src/actions/index.js
```diff
@@ -25,6 +25,8 @@ import {
   STOP_SCHEDULER_FIXED_DELAY,
   STOP_SCHEDULER_FOR_RESUMING_NOTIFICATIONS,
   UPDATE_PAGING_DATA,
+  SET_DEFAULT_CONSOLE,
+  SET_VNC_MODE,
```
sgratch, line 29: Since we read those from config and we know that we'll probably override this by user settings in the future, I suggest to rename default_console and vnc_mode on all places in code to something like global-default-console and global-default-vnc-mode accordingly.

src/actions/vm.js
```diff
@@ -307,6 +306,17 @@ export function updateIcons ({ icons }) {
   }
 }
 
+export function setVmConsoles ({ vmId, consolesList, selectedProtocol }) {
+  return {
+    type: SET_VM_CONSOLES,
+    payload: {
+      vmId,
+      consoles: consolesList,
+      selectedProtocol,
```
sgratch, line 315: Again regarding naming:
Sometimes we use protocol and sometimes console which actually it's the same meaning in our code. So we need to be more persistent.
For example, instead of selectedProtocol here, we can have selectedConsole (on all places in code).

src/intl/messages.js
```diff
@@ -62,6 +62,13 @@ export const messages: { [messageId: string]: MessageType } = {
   cancel: 'Cancel',
   cannotUpdateCloudInitHostname: 'Host name cannot be synchronized with VM name because the VM name is not a valid host name.',
   cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues.',
+  cantOpenConsole: "Can't open console: {message}",
```
sgratch, line 65: when should this message appear? it's not informative and seems redundant.

src/intl/messages.js
```diff
@@ -62,6 +62,13 @@ export const messages: { [messageId: string]: MessageType } = {
...
+  cantOpenGraphicalConsoleEmptyList: "Can't open graphical console since {vmName} is a headless VM.",
```
sgratch, line 66: please rephrase/shorten to "Can't open a console since VM (vmname} is headless."

src/intl/messages.js
```diff
@@ -62,6 +62,13 @@ export const messages: { [messageId: string]: MessageType } = {
...
+  cantOpenGraphicalConsoleMoreOptions: "Can't open graphical console you can find more options in the VM page.",
```
sgratch, line 67: when should this message appear? it's not informative and seems redundant.

src/intl/messages.js
```diff
@@ -62,6 +62,13 @@ export const messages: { [messageId: string]: MessageType } = {
...
+  cantOpenGraphicalConsoleNoDefault: "Can't open graphical console {vmName} has no {defaultConsole} console assigned. You can find more options in the VM page.",
```
sgratch, line 68: Please rephrase to:
Can't open a console since VM {vmName} is not assigned with the default console {defaultConsole}....

src/intl/messages.js
```diff
@@ -62,6 +62,13 @@ export const messages: { [messageId: string]: MessageType } = {
...
+  cantOpenNonRdpGraphicalConsoleEmptyList: "Can't open non Rdp graphical console since {vmName} is a headless VM.",
+  cantOpenNonRdpGraphicalConsoleMoreOptions: "Can't open non Rdp graphical console you can find more options in the VM page.",
+  cantOpenNonRdpGraphicalConsoleNoDefault: "Can't open non Rdp graphical console {vmName} has no {defaultConsole} console assigned. You can find more options in the VM page.",
```
sgratch, line 71: relevant to all new added console messages:
1. "You can find more options in the VM page" - what do you mean by that? Which info on which page?
2.Why do we need those "non RDP" message? Console is the same regardless to RDP.

src/sagas/index.js
```diff
@@ -109,9 +122,7 @@ const VM_FETCH_ADDITIONAL_DEEP = [
   'statistics',
 ]
 
-const VM_FETCH_ADDITIONAL_SHALLOW = [
-  'graphics_consoles',
-]
+const VM_FETCH_ADDITIONAL_SHALLOW = []
```
sgratch, line 125: Do we still need this VM_FETCH_ADDITIONAL_SHALLOW?

src/utils/index.js
```diff
@@ -40,3 +42,25 @@ export function dividers (num: number): Array<number> {
   }
   return divs
 }
+
+export function getNoDefaultConsoleErrorMessages (reason: string, vmName: string, hasRdp: boolean, defaultConsole: string): string {
```
sgratch, line 46: Why do we need  `hasRdp` for distinguishing between non rdp and console on current solution?  Ithink we should have one message referring to console only.

src/utils/index.js
```diff
@@ -40,3 +42,25 @@ export function dividers (num: number): Array<number> {
...
+      message = hasRdp
+        ? msg.cantOpenNonRdpGraphicalConsoleEmptyList({ vmName })
+        : msg.cantOpenGraphicalConsoleEmptyList({ vmName })
+      break
+    case NO_DEFAULT_CONSOLE:
+      message = hasRdp
+        ? msg.cantOpenNonRdpGraphicalConsoleNoDefault({ vmName, defaultConsole })
+        : msg.cantOpenGraphicalConsoleNoDefault({ vmName, defaultConsole })
+      break
+    default:
```
sgratch, line 59: It seems to me that this default case is never reached. Can you explain when we'll get this error and why?

sgratch (changes requested):
@bamsalem 

- In case of a headless VM, RDP shouldn't be available as well - this is the behavior on webadmin and this was the behaviour on current web-ui on both vms dashboard (prior to this fix) and vm details dialogs.

- Please also update first comment by adding a dependency on patch https://gerrit.ovirt.org/112536.

- Check inline comments.

sjd78 (changes requested):
A few gaps to close.  Most of my comments are inline.

A thought about the PR's first comment -- you can keep going back and editing it to match the current state of the PR.  For the work you've done here, I'd recommend an overview of the functional changes made and then as much about the technical as makes sense.

E.g. - This PR changes console handling on the Card List page.  Due to the performance issues detailed in BZ123, the graphics_consoles are now fetched asynchronously.  A single "Console" button on each VM Card will use system config values to auto select what console to display to the user.  (Detail how the auto-select works with the config values later).  The Console buttons on the VM details page remain unchanged.

sjd78:
@rszwajko ...

> @sgratch @sjd78 @hstastna Take a look at the topics below. I would like to hear your opinion about the direction we should take:
> 
>    1. [isCard ](https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r545258589)

I'd rather refactor the VmActions all together and have a real split and distinct components between 1. On Card, 2. On Details but narrow width / kebab, 3. On Details but normal width / button toolbar.  They can still have the action handlers and confirmation popups, but the blended renders really really annoy me.

> 
>    2. [RDP](https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r555746901)
> 

Consoles for a VM are either Spice or VNC (native) or VNC (web).  RDP isn't really a console (even when you tick the connect to console session box in the Remote Desktop app on windows) since it doesn't even become available until the system is running.  You can't RDP to a windows installer etc.  Having the distinct button on the card action drop down or the details page toolbar console drop down is sufficient IMHO.

>    3. [console slice](https://github.com/oVirt/ovirt-web-ui/pull/1342#issuecomment-758660205)

IIRC the slice is more for handling SSO and console access privlidges than anything else.  Just go ahead and try opening a console from and admin user, and then try to open the console for a normal user.  For fun you can get make sure SSO via guest agent is enabled on your VM and see those dialogs (once per app session at least).

bamsalem:
@sgratch 

> @bamsalem
> 
> * In case of a headless VM, RDP shouldn't be available as well - this is the behavior on webadmin and this was the behaviour on current web-ui on both vms dashboard (prior to this fix) and vm details dialogs.

I tested that on brq-dev, and it seems like in case of windows  headless VM RDP is still available:
image
headlessRdp-2021-02-14_13 15 00

So I'm not sure if it should be available or it's a bug...

src/intl/messages.js
bamsalem: It's a fallback message with the latest changes is redundant  I removed that :)

src/intl/messages.js
bamsalem: >1. "You can find more options in the VM page" - what do you mean by that? Which info on which page?

I meant to inform users that in case of no default they can find more options like: RDP, another console in the vm details page.
>2. Why do we need those "non RDP" messages? Console is the same regardless to RDP.

It's just mention that RDP is available, I'll remove that...

src/utils/index.js
bamsalem: > I'd rather the messages be selected in the place they're needed than be in this util function...

I did it thanks :)

src/reducers/utils.js
```diff
@@ -66,3 +70,14 @@ export function removeMissingItems ({ state, subStateName, idsToPreserve }) {
     .asImmutable()
   return state.set(subStateName, newItems)
 }
+export function selectDefaultConsoleProtocol (defaultProtocol, consoles = []) {
```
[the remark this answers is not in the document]
bamsalem: I'm not sure if it's better than notify user that the default is not available and then he can use VM details and select one of the options...

sgratch:
@bamsalem 
> I tested that on brq-dev, and it seems like in case of windows headless VM RDP is still available

Correct, it is enabled on webadmin (my mistake, sorry) and disabled on current web-ui for headless. 
The correct behaviour is to enable RDP for web-ui as well since no graphic protocol is needed to be installed on that VM so headless Windows VMs can be managed using RDP or SSH access and that's the right behaviour.

src/constants/index.js
```diff
@@ -120,7 +120,10 @@ export const SET_VM_SESSIONS = 'SET_VM_SESSIONS'
 export const SET_VM_SORT = 'SET_VM_SORT'
 export const SET_VNIC_PROFILES = 'SET_VNIC_PROFILES'
 export const SET_VM_SNAPSHOTS = 'SET_VM_SNAPSHOTS'
+export const SET_VM_CONSOLES = 'SET_VM_CONSOLES'
 export const SET_WEBSOCKET = 'SET_WEBSOCKET'
+export const SET_GLOBAL_DEFAULT_CONSOLE = 'SET_GLOBAL_DEFAULT_CONSOLE'
+export const SET_GLOBAL_VNC_MODE = 'SET_GLOBAL_VNC_MODE'
```
sgratch, line 126: SET_GLOBAL_DEFAULT_VNC_MODE

src/intl/messages.js
```diff
@@ -62,6 +62,7 @@ export const messages: { [messageId: string]: MessageType } = {
   cancel: 'Cancel',
   cannotUpdateCloudInitHostname: 'Host name cannot be synchronized with VM name because the VM name is not a valid host name.',
   cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues.',
+  cantOpenConsole: 'Can\'t open console: {message}',
```
sgratch, line 65: I still don't understand why we added this message (Based on Scott's question: "why did we touch the the SSO code that would warrant adding an error toast message?") and anyway it's not informative enough.

src/sagas/index.js
```diff
@@ -183,6 +187,42 @@ export function* fetchByPage () {
...
+
+    if (selectedConsole === EMPTY_CONSOLES_LIST) {
+      yield put(addUserMessage({
+        type: 'error',
+        message: msg.consoleNotAvailableHeadless({ vmName }),
+      }))
+    } else if (selectedConsole === NO_DEFAULT_CONSOLE) {
+      yield put(addUserMessage({
+        type: 'error',
+        message: msg.consoleDefaultNotAvailable({ vmName }),
```
sgratch, line 218: What about the defaultConsole as an input?

sgratch (changes requested):
@bamsalem 
- It crashes when trying to open a console for a headless VM:
image

- please fix PR's first comment based on previous review
- check inline comments

src/intl/messages.js
bamsalem: I thought it will be better if we will notify users if any error occurs during console download...
If it not necessary I'll remove that.

bamsalem:
@sgratch 
> @bamsalem
> 
> * It crashes when trying to open a console for a headless VM:
>   image

It have been fixed, I didn't checked enough one of the suggested changes.

> * please fix PR's first comment based on previous review

Alredy Done.

> * check inline comments

I replayed to the relevant comments.

src/components/VmActions/index.js
sjd78: `isNoVNC` is the correct attribute to be there in this case.  It tells the confirmation modal to open VNC in the "noVnc" in-browser mode.  The map through the consoles above will generate the confirmation modal to open VNC in the "console.vv" style.

src/intl/messages.js
sjd78: You should remove it.  If the SSO doesn't work, a modal is already displayed with an awkward message to handle that scenario.

Right now, on the first run of the app, you'll see both the modal and the toast at the same time.  It isn't necessary.

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,104 @@
...
+export default connect(
+  (state, { vm }) => ({
+    defaultConsole: vm.get('defaultConsole'),
+    isNoVnc: state.config.get('defaultVncMode') === 'NoVnc',
+  }),
+  (dispatch) => ({
+    fetchConsoles: (vm) => {
+      dispatch(downloadVmConsoles({ vm }))
+    },
+  })
```
sjd78, line 103:
```suggestion
  (dispatch, { vm }) => ({
    fetchConsoles: () => dispatch(downloadVmConsoles({ vm }),
  })
```

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,104 @@
...
+          if (dispatchStatus === DISPATCHED) {
+            handleClose()
+          }
+        }}
+        id={id}
+      />
+      }
+    </div>
+  )
+}
```
sjd78, line 79: I think you can get rid of component state use completely by just doing:

```jsx
}) => {
  useEffect(() => {
    if (show && defaultConsole === 'undefined') {
      fetchConsoles()
    }
  }, [show, defaultConsole])

  if (!show || defaultConsole === 'undefined' || defaultConsole === EMPTY_CONSOLES_LIST) {
    return null
  }

  return (
    <ConsoleConfirmationModal
      show
      vm={vm}
      consoleId={defaultConsole.id}
      isNoVNC={defaultConsole.protocol === 'vnc' ? isNoVnc : false}
      onClose={onClose}
      id={id}
    />
  )
}
```

Note 1: If the console is already fetched for the VM, it won't be fetched again.

Note 2: The `defaultConsole === 'undefined'` value is set in the base VM transform.  You can change it to whatever constant makes sense.  

In the case of a headless VM, clicking the action button will set `show=true` and then with `defaultConsole === 'undefined'` will hit the effect and call the fetch saga.  But, the saga would then set `deafulConsole` to `EMPTY_CONSOLES_LIST`.  The fetch won't be triggered again.  Only an error toast on the first click will be displayed.  Subsequent clicks will have no effect.

Perhaps if a VM is known to be headless, the console button can be removed from the action card or otherwise disabled to avoid confusion.  @sgratch, @rszwajko - Thoughts on multiple clicks to the console button for a headless VM?

sjd78 (changes requested):
It's close.  `AutoSelectConsole` can be simpler.

src/components/VmActions/AutoSelectConsole.js
bamsalem: @sjd78
I got rid of the component's state but I made few small changes to integrate between the components because `ConsoleConfirmationModal` is class component and `AutoSelectConsole` is functional ConsoleConfirmationModal sometimes triggers the onClose with no reason and then the console wasn't fetched as expected.

src/sagas/index.js
```diff
@@ -183,6 +187,41 @@ export function* fetchByPage () {
...
+    yield put(setVmConsoles({ vmId, consolesList: [], selectedConsole: EMPTY_CONSOLES_LIST }))
+  }
+
+  if (apiConsoles && apiConsoles.graphics_console) {
+    const defaultConsoleProtocol = yield select(state => state.config.get('defaultConsole'))
+    const consoles = Transforms.VmConsoles.toInternal({ consoles: apiConsoles })
+    // select the console to use based on availability and app/user configs
+    const selectedConsole =
+      consoles.length === 0 ? EMPTY_CONSOLES_LIST
+        : consoles.length === 1 ? consoles[0]
```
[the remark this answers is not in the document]
sgratch: @sjd78 I think that we should expect the supported only console to be used in case only one protocol is supported, regardless to default console. The default console is only for cases in which all options are supported and we need to auto choose for the user.

src/components/VmActions/AutoSelectConsole.js
sgratch: @sjd78 @bamsalem 

> Perhaps if a VM is known to be headless, the console button can be removed from the action card or otherwise disabled to avoid confusion. @sgratch, @rszwajko - Thoughts on multiple clicks to the console button for a headless VM?

I think it might be confusing for the user if the second click will disable the console button till next hard refresh or console configuration changes on backend. I think it's better to leave it as is to keep the ui simple and persistent.

src/sagas/index.js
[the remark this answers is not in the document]
bamsalem: @sgratch @sjd78 
I changed the code back to the prev version (now the system will use the only console available).

src/components/VmActions/AutoSelectConsole.js
rszwajko: > I think you can get rid of component state use completely

yep. This is the right direction.

> Perhaps if a VM is known to be headless, the console button can be removed from the action card or otherwise disabled to avoid confusion. @sgratch, @rszwajko - Thoughts on multiple clicks to the console button for a headless VM?

I would disable the action (make the button grey) to clearly indicate that this option is not available. Active buttons encourage users to click & try them i.e. users may try it few times hoping it will finally work.

src/components/VmActions/index.js
rszwajko: @sjd78  @bamsalem 
OK.  I was confused by the `isNoVnc` prop introduced in AutoSelectConsole which was very similar to `isNoVNC` prop used in ConsoleConfirmationModal. Not passing this property to  ConsoleConfirmationModal seemed a typo to me.
I still think that launching VNC console with `isNoVNC == true` is confusing. Perhaps a different name would be better explain the purpose? However this refactoring is out of scope of this patch so let's leave it as it is.

src/components/VmActions/AutoSelectConsole.js
bamsalem: @sjd78 @rszwajko 
image

>I would disable the action (make the button grey) to clearly indicate that this option is not available. Active buttons encourage users to click & try them i.e. users may try it few times hoping it will finally work.

We can't disable the button without disabling the dropdown menu (with the current component, PF component).
and I'm not sure it's worth to create our own component for that, but we must add another notifier, maybe keep showing the error on every click...
WDYT?

sgratch:
@bamsalem please fix the TypeError:
image

src/intl/messages.js
bamsalem: image

bamsalem:
> @bamsalem please fix the TypeError:
> image

@sgratch
I tried to reproduce the error but I couldn't, can you provide more details/steps to reproduce please?
BTW I'm not sure but it seems like the issue is not related to this fix, maybe we should deal with that in separate PR...

sgratch:
@bamsalem 
> I tried to reproduce the error but I couldn't, can you provide more details/steps to reproduce please?
> BTW I'm not sure but it seems like the issue is not related to this fix, maybe we should deal with that in a separate PR...

This error generated for me when app was up but it's not reproduced anymore.

src/components/VmActions/AutoSelectConsole.js
sgratch: @bamsalem for headless:
I suggest reverting back for keeping the headless error per every click.
 We can change the menu items order so that "console" menu item will not be the default till next refresh (the default/first item will be "Remote Desktop" for win based vms and "Suspend" for others).

src/components/VmActions/AutoSelectConsole.js
```diff
@@ -0,0 +1,76 @@
...
+    if (show && defaultConsole === 'undefined') {
+      fetchConsoles(vm)
+    }
+  }, [show, defaultConsole, fetchConsoles])
+
+  useEffect(() => {
+    if (show && DEFAULT_CONSOLE_PROTOCOL_ERRORS_LIST.includes(defaultConsole)) {
+      onClose()
+    }
+  }, [defaultConsole, show, onClose])
```
sjd78, line 35: Why is `onClose` on the deps list?

sjd78 (changes requested):
Tests ok on my dev environment, but I have concerns about the deps list on AutoSelectConsole's `useEffect` hooks.

sjd78:
> @bamsalem
> 
> > I tried to reproduce the error but I couldn't, can you provide more details/steps to reproduce please?
> > BTW I'm not sure but it seems like the issue is not related to this fix, maybe we should deal with that in a separate PR...
> 
> This error generated for me when app was up but it's not reproduced anymore.

line 444 of `transform.js` indicates that a Pool did not have a `vm` property returned from the rest api.  I've not seen that before - the user would have to have access to either (1) an empty pool, or (2) a pool that is visible to the user but ALL vms are not.  There may be other scenarios there, but they all seem very unlikely.

src/components/VmActions/AutoSelectConsole.js
bamsalem: > @bamsalem for headless:
> I suggest reverting back for keeping the headless error per every click.

Done.

> We can change the menu items order so that "console" menu item will not be the default till next refresh (the default/first item will be "Remote Desktop" for win based vms and "Suspend" for others).

Regarding changing the order, it's more complex than it looks, because it's a PF component when we change the order it's change the confirmation modal and than it display non related notification, for example "Are you sure you want to Suspend the VM?", to handle that we need to implement our own component.

sgratch: @bamsalem so we can go back to original behavior of just hidden/dismiss the "console" menu item till next refresh.

sgratch (approved):
LGTM except 2 issues to consider as follow up issues:
1. https://github.com/oVirt/ovirt-web-ui/pull/1342#discussion_r587488486 - consider removing the console menu option for headless - this can be implemented separately. 

2. Please note that If the consoles type list is already fetched for the VM, it is currently not fetched again till next refresh.
Until now, we didn't call fetchConsoles per every console click, but the user could see which types are supported and choose which to open on dashboard view.  If consoles can't be displayed/chosen on dashboard, at least fetching the consoles can be up to date when user tries to open one. It might add a delay for console opening though...
This should be considered again when implementing console settings.

sjd78 (approved):
LGTM

Issue #1347, opened 2020-12-11
Creating VM from Template with multiple defined disks should not allow the bootable disk flag to be changed


sjd78:
When creating a VM from a template that has a defined bootable disk, the user should not be allowed to change the bootable flag when adding new disks.  For a template with a single disk, this is the behavior.  For a template with multiple disks, the bootable flag is allowed to be changed.  The scenario with >1 disk needs to be fixed.

Reference comment from #1323:

I was able to test some scenarios (not all) and encountered no problems except of one case (see below).

Please confirm if this is the expected behaviour:
1. choose a template with 2 disks (i.e. eee1-desktop)
2. one of the disks is marked as bootable
3. add a 3rd custom disk
    Expected result: it's not possible to move bootable flag to 3rd disk
   Actual result: it's possible to change and it is respected during VM creation
4. delete the newly added 3rd disk
   Expected result: bootable flag is moved to some other disk
   Actual result: no disk is marked bootable (but after creation one of the disks is bootable)

![no_bootable](https://user-images.githubusercontent.com/64194103/101931490-ca207980-3bd9-11eb-9938-7930414b062f.gif)

_Originally posted by @rszwajko in https://github.com/oVirt/ovirt-web-ui/issues/1323#issuecomment-743307505_

PR #1348, opened 2020-12-11
Fix create VM from template bootable disk flag check

sjd78:
In the Create VM Storage step, if any of a VM's template defined disks is marked bootable, then the bootable flag cannot be changed when adding or editing a new disk.

Previously, the `isBootableDiskTemplate` check in the `Storage` component worked for a template provisioned new VM with a single template defined disk, but not for templates that have multiple defined disks.  The check has been updated to handle multiple template defined disks, any of which may be bootable.

Fixes: #1347

Example of how it looks now when adding a disk to a single template defined disk:
VM Portal - Google Chrome_051

Example of how it looks now when adding a disk to multiple template defined disks:
VM Portal - Google Chrome_052

### DIFF
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -424,12 +424,12 @@ class Storage extends React.Component {
     return Object.keys(this.state.editing).length > 0
   }
 
-  // return true if there's any disk set as bootable and if it's a template disk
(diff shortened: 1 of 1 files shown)


rszwajko (approved):

bamsalem (approved):
LGTM

hstastna (approved):
Looks good to me, works as expected.

Issue #1320, opened 2020-10-29
Error after changing the Cluster when editing VM Details


hstastna:
Error occurs after trying to change the _Cluster_ in the VM _Details_ card.

**How reproducible:**
Not always, I think it depends on the actual Cluster/Cluster which is being selected and/or the actual OS (and/or the other details) of the selected VM. 

**Steps to Reproduce:**
1. Select a VM in the VM list
2. In the VM _Details_ card, click on "Edit details" for the selected VM
3. Choose another _Cluster_ from the drop down

**Actual results:**
`TypeError: os is undefined`
![os](https://user-images.githubusercontent.com/13417815/97608160-40925f00-1a12-11eb-88b5-1946e6689f1b.png)

**Expected results:**
No error.

**Additional info:**
After we try to select another Cluster in the appropriate drop down, it successfully appears as the selected one (but the error appears).

PR #1349, opened 2020-12-14
Select The Right "Other OS" operatingSystemId in CreateVmWizard -> BasicSettings according to cluster's cpu architecture.

bamsalem:
Fixes: #1307 .
In CreateVmWizrd -> BasicSettings The operating system drop down includes another options according to selected cluster's CPU architecture (even if it's not visible to users, the IDs are different between architectures), But our default value for "Other OS" is not compatible with all the architectures.

To solve this issue I added new helper function that verifies that every time we sets OperatingSystemId property (not by the user) we check if we selected "Other OS" value that is in the list.
If the selected value is Other OS default values or undefined The function will verify the OperatingSystemId if not it will return the original value.

image

PR #1351, opened 2020-12-15
Fix OS error after changing the cluster when editing DetailsCard component

bamsalem:
This change should fix the issue that mentioned in this [comment](https://github.com/oVirt/ovirt-web-ui/issues/1320#issue-732507158).

The cause of the issue:
In `operatingSystems` there is no entry `'type'`:
image
As a result `templateOs` will always be `undefined`,
So the requested fix should be, to change it to `'name'`, but it won't work for ppc 64 architecture because any OS has its own id and name and it might cause errors,
for example Other OS for ppc64 has different id from x86_64.
So I used the `'description'` attribute (like I did in #1349 ).

Fixes: #1320

### DIFF
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -106,8 +106,10 @@ const DEFAULT_GMT_TIMEZONE = timezones.find(timezone => timezone.value.startsWit
 class DetailsCard extends React.Component {
   constructor (props) {
     super(props)
-    const vmClusterId = props.vm.getIn(['cluster', 'id'])
-    const vmDataCenterId = props.clusters.getIn([vmClusterId, 'dataCenterId'])
+    const { vm, clusters } = props
+    const vmClusterId = vm.getIn(['cluster', 'id'])
+    const vmDataCenterId = clusters.getIn([vmClusterId, 'dataCenterId'])
+    const clusterArchitecture = getClusterArchitecture(vmClusterId, clusters)
 
     this.state = {
       vm: props.vm, // ImmutableJS Map
@@ -121,7 +123,7 @@ class DetailsCard extends React.Component {
       promptNextRunChanges: false,
 
       isoList: createIsoList(props.storageDomains, vmDataCenterId),
-      clusterList: createClusterList(props.clusters, vmDataCenterId),
+      clusterList: createClusterList(props.clusters, vmDataCenterId, clusterArchitecture),
       osList: createOsList(vmClusterId, props.clusters, props.operatingSystems),
 
       enableInitTimezone: !!props.vm.getIn(['cloudInit', 'timezone']), // true if sysprep timezone set or Configure Timezone checkbox checked
@@ -180,12 +182,14 @@ class DetailsCard extends React.Component {
     // NOTE: Doing the following here instead of getDerivedStateFromProps so __clusters__,
     //       __storageDomains__, and __operatingSystems__ don't need to be kept in state for
     //       change comparison
-    const vmClusterId = this.props.vm.getIn(['cluster', 'id'])
-    const vmDataCenterId = this.props.clusters.getIn([vmClusterId, 'dataCenterId'])
+    const { clusters, vm } = this.props
+    const vmClusterId = vm.getIn(['cluster', 'id'])
+    const vmDataCenterId = clusters.getIn([vmClusterId, 'dataCenterId'])
+    const clusterArchitecture = getClusterArchitecture(vmClusterId, clusters)
 
     if (prevProps.clusters !== this.props.clusters) {
       const { clusters } = this.props
-      this.setState({ clusterList: createClusterList(clusters, vmDataCenterId) }) // eslint-disable-line react/no-did-update-set-state
+      this.setState({ clusterList: createClusterList(clusters, vmDataCenterId, clusterArchitecture) }) // eslint-disable-line react/no-did-update-set-state
     }
 
     if (prevProps.storageDomains !== this.props.storageDomains) {
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -255,7 +255,6 @@ class DetailsCard extends React.Component {
       maxNumberOfCores,
       maxNumberOfThreads,
       operatingSystems,
-      blankTemplateId,
       clusters,
       templates,
     } = this.props
@@ -269,14 +268,6 @@ class DetailsCard extends React.Component {
         case 'cluster':
           updates = updates.set('cluster', clusters.get(value))
           fieldUpdated = 'cluster'
-
-          // Change the template to 'Blank' if the VM's template isn't in the new cluster
-          {
-            const template = templates.get(updates.getIn(['template', 'id']))
-            if (template && template.get('clusterId') && template.get('clusterId') !== value) {
-              changeQueue.push({ fieldName: 'template', value: blankTemplateId })
-            }
-          }
           break
 
         case 'template':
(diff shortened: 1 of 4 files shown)


ovirt-infra:

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -251,35 +257,37 @@ class DetailsCard extends React.Component {
...
+          updates = updates.set('template', templates.get(value))
           fieldUpdated = 'template'
 
           // Apply settings from the template to the VM (memory, CPUs, OS, cloudInit, bootMenuEnabled)
           {
             const template = this.props.templates.get(value)
             if (template) {
               const templateOsType = template.getIn(['os', 'type'], 'other')
-              const templateOs = this.props.operatingSystems.find(os => os.get('type') === templateOsType)
+              const oldOS = operatingSystems.find(os => os.get('name') === templateOsType)
```
rszwajko, line 281: IMHO this line contains the fix - `os.name` and `template.os.type` seem to be OS unique name.
Check both REST definitions:
1.  `os.name` is set [here](https://github.com/oVirt/ovirt-engine/blob/2314e3b31e42ff1c96395948af7a9a7f996f56ab/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendOperatingSystemResource.java#L36)
2. `template.os.type` is set [here](https://github.com/oVirt/ovirt-engine/blob/2314e3b31e42ff1c96395948af7a9a7f996f56ab/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/types/TemplateMapper.java#L90)

Why not just:
```
const templateOs = operatingSystems.find(os => os.get('name') === templateOsType) || findDefaultForCluster()
```
?
Of course `findDefaultForCluster()` remains the tricky part :)

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -251,35 +257,37 @@ class DetailsCard extends React.Component {
...
+              const clustersOs = createOsList(updates.getIn(['cluster', 'id']), clusters, operatingSystems)
+              const templateOs = clustersOs.find(os => os.value === oldOS.get('description')) || clustersOs.find(os => os.value === 'Other OS')
```
rszwajko, line 283: As mentioned  [here](https://github.com/oVirt/ovirt-web-ui/pull/1349#discussion_r545376619) I think we should base the logic on unique names/IDs.

rszwajko (changes requested):

src/components/VmDetails/cards/DetailsCard/index.js
bamsalem: I've sent you question on slack on this subject, I'll post the conclusions here later

sgratch:
@bamsalem @rszwajko 
Both OS name and OS id are unique and can be used. The default OS indexes are kept as constants on backend:
https://github.com/oVirt/ovirt-engine/blob/af7bb6b91639edf92b977b08484423cd0108d723/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/osinfo/OsRepository.java#L22-L24
https://github.com/oVirt/ovirt-engine/blob/af7bb6b91639edf92b977b08484423cd0108d723/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/OsRepositoryImpl.java#L52-L58

And the default OS per cluster is based on its architecture:
setOSType(new NotChangableForVmInPoolListModel<Integer>() {
 ...
            super.setSelectedItem(AsyncDataProvider.getInstance().getDefaultOs(cluster.getArchitecture()));
 ...

The main important thing here is to fix the error exception of course, but avoid handing a use case of switching between clusters from **different** architectures since this is not supported and will fail on backed anyway, so no need to handle that on web-ui as well.

sgratch:
ci add to whitelist

src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -251,28 +266,32 @@ class DetailsCard extends React.Component {
...
           // Apply settings from the template to the VM (memory, CPUs, OS, cloudInit, bootMenuEnabled)
           {
-            const template = this.props.templates.get(value)
+            const template = templates.get(value)
             if (template) {
-              const templateOsType = template.getIn(['os', 'type'], 'other')
-              const templateOs = this.props.operatingSystems.find(os => os.get('type') === templateOsType)
+              const clusterId = updates.getIn(['cluster', 'id'])
+              const clusterArchitecture = getClusterArchitecture(clusterId, clusters)
+              const templateOsName = template.getIn(['os', 'type'], 'other')
```
sgratch, line 291: Why the const is **templateOsName** and not **templateOsType**?

sgratch (commented):
The original exception is not reproduced and therefore fixed, but saving the VM after changing to a cluster from the same architecture fails, even though it shouldn't, with an error:
"EDIT_VM failed [Cannot edit VM. The requested template doesn't belong to the same base template as the original template.]"

src/components/VmDetails/cards/DetailsCard/index.js
bamsalem: It's templateOsName because it's stored as `name` in the `operatingSystems` objects list:
image
Another reason, we compare it to OS Name in line [293](https://github.com/oVirt/ovirt-web-ui/blob/b63b0737f16f1354d2437560b9b55e7adb9eb92d/src/components/VmDetails/cards/DetailsCard/index.js#L293).
I'm not sure which one is clearer but if you think templateOsType is clearer I'll change that.

src/constants/oparatingSystems.js
```diff
@@ -0,0 +1,5 @@
+export const defaultOperatingSystemIds = [
```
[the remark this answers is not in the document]
bamsalem: Done.

bamsalem:
> The original exception is not reproduced and therefore fixed, but saving the VM after changing to a cluster from the same architecture fails, even though it shouldn't, with an error:
> "EDIT_VM failed [Cannot edit VM. The requested template doesn't belong to the same base template as the original template.]"

I checked that and I got this error too, but I found also that cluster has been changed , after the error is shown click cancel and wait for the next refresh. you can see it's changed...
Do you want me to find out what is the cause and fix it? Or we should deal with that in another PR?

src/components/VmDetails/cards/DetailsCard/index.js
sgratch: The right name should really be templateOsName ("type" term is only used for templates). For clarifications I suggest adding a line comment, specifying that template.os.type and operatingSystems.name are referring to the same value on backend.

sgratch:
> I checked that and I got this error too, but I found also that cluster has been changed , after the error is shown click cancel and wait for the next refresh. you can see it's changed...

Yes, cluster is changed and template is switched to blank but this error shouldn't occur. 
The reason for this error is that if cluster is changed by the user for vm based on non Blank template then auto changing the template is not allowed since this vm was provisioned from that template (template disks are not cloned) so you can't just change the template without cloning the disks.
The right behaviour is therefore to allow changing cluster (if allowed by architecture) and doesn't auto change template and OS type to Blank, "other os". Only the cluster should be changed. This is how it behaves on webadmin as well for vms based on non Blank template.

> Do you want me to find out what is the cause and fix it? Or we should deal with that in another PR?

You can but you'll anyway need to fix this PR as well.

bamsalem:
@sgratch 
I removed the template and os change on cluster change, I also added option to filter the clusterList by architecture to show only clusters with the same architecture (clusters that the vm can migrate to)

sgratch (approved):
LGTM

isaranova:
LGTM

Issue #1329, opened 2020-11-11
CreateVmWizard: Disk name is not checked


isaranova:
Description of problem:
VM Creator does not check that the disk name is valid and allows user to create disks with wrong names (containing invalid characters, according to the AdminPortal) that cause error message to appear and not creating the disk.

Error message: 
CREATE_DISK_FOR_VM failed Invalid UUID string: _

VM Portal version number:
ovirt-web-ui-1.6.6-0.20201030.gitf464801.el8.noarch

oVirt version number:
ovirt-web-ui-1.6.6-0.20201030.gitf464801.el8.noarch

Browser and version:
firefox-82.0.2-1.fc31.x86_64

Steps to Reproduce:
1. Open VM Creator
2. Add disk with wrong name (e.g. '[[];]']];]{}?><')
3. Create the VM

Actual results:
Disk is saved without any error message. VM is created, but the disk is not and an error message appears.

Expected results:
Disk is not allowed to be saved.

PR #1355, opened 2020-12-23
Add validation to disk's name in CreateVMWizard Storage step

bamsalem:
This PR adds validations to disk name such as in NIC name in Network Step.
The rules for the validation taken from [the documentation](https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/administration_guide/sect-virtual_disk_tasks).
In case of error it will show error message, mark the field in red and not let user save the edited disk and move to the next step:

image

If you have any questions please feel free to ask
Fixes: #1329


ovirt-infra:

sgratch:
ci add to whitelist

src/intl/messages.js
```diff
@@ -167,6 +167,7 @@ export const messages: { [messageId: string]: MessageType } = {
   createVmWizardStepTitleNetwork: 'Networking',
   createVmWizardStepTitleReview: 'Review',
   createVmWizardStepTitleStorage: 'Storage',
+  createVmWizardStorageDiskNameRules: 'Disk name must be 40 or less alphanumeric characters or "-_."',
```
sgratch, line 170: This message is not accurate, please check that you can create a disk name longer than 40 chars. Please rephrase it to be the same as on webadmin:
Disk name can contain only alphanumeric, '.', '_' or '-' characters.

src/components/utils/validation.js
```diff
@@ -88,7 +88,7 @@ export function isVmNameValid (nameCandidate: string): boolean {
 }
 
 export function isDiskNameValid (nameCandidate: string): boolean {
-  return isVmNameValid(nameCandidate)
+  return /^[a-zA-Z0-9_\-\\.]{1,40}$/.test(nameCandidate)
```
sgratch, line 91: Please update  this length validation "{1,40}" to a much longer length (e.g. 64 chars)

sgratch (dismissed):
Please see inline comments.

In addition, there is a problem of saving a new disk after an error is displayed:
1. create  a new disk with invalid name - validation failed and an inline error is displayed.
2. fix the disk name to become a valid one.

Expected: saving the disk is possible.
Actual behaviour: no error is displayed but you can't save the disk

src/intl/messages.js
bamsalem: image

src/components/utils/validation.js
bamsalem: @sgratch Are you sure?
I stuck to the [documentation](https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/administration_guide/sect-virtual_disk_tasks#Creating_a_Virtual_Disk)

sgratch: > @sgratch Are you sure?
> I stuck to the [documentation](https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/administration_guide/sect-virtual_disk_tasks#Creating_a_Virtual_Disk)

When comparing webadmin code/behaviour to the documentation, usually the webadmin behaviour is the the enforcing rule (and probably the doc is not up to date) :). 
Specifically here, the limit is 255 chars:
https://github.com/oVirt/ovirt-engine/blob/e48484ccb3c1c8fcaed81023b234295a109ebd8b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/DisksAllocationModel.java#L486-L490

bamsalem: OK, I'll check webAdmin code next time :)

src/components/utils/validation.js
```diff
@@ -88,7 +88,7 @@ export function isVmNameValid (nameCandidate: string): boolean {
 }
 
 export function isDiskNameValid (nameCandidate: string): boolean {
-  return isVmNameValid(nameCandidate)
+  return /^[a-zA-Z0-9_\-\\.]{1,64}$/.test(nameCandidate)
```
sgratch, line 91: Why limiting to 64 chars while on webadmin and backend we now know that the limitation is for 255 chars? It's better to use the same limitation

sgratch (changes requested):

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -503,21 +517,32 @@ class Storage extends React.Component {
...
+        if (value >= minDiskSizeInGiB && value <= maxDiskSizeInGiB) {
+          delete editingRow.invalidSizeValue
+        } else {
+          editingRow.invalidSizeValue = true
+        }
+        value = +value * (1024 ** 3) // GiB to B
+        break
+      case 'name':
+        editingErrors.diskInvalidName = !isDiskNameValid(value)
+        console.log('editingErrors', editingErrors)
```
sgratch, line 535: Don't you need to delete this console.log line? I can't see the benefit of leaving it.

sgratch (commented):
A small inline comment.

src/components/utils/validation.js
```diff
@@ -88,7 +88,7 @@ export function isVmNameValid (nameCandidate: string): boolean {
 }
 
 export function isDiskNameValid (nameCandidate: string): boolean {
-  return isVmNameValid(nameCandidate)
+  return /^[a-zA-Z0-9_\-\\.]{1,255}$/.test(nameCandidate)
```
rszwajko, line 91: Please add an unit test  - previously `isVmNameValid()` was unit tested.
@sgratch  @sjd78 
Validation is now more restrictive. However `isDiskNameValid` is also used to validate alias name in `DiskImageEditor`. Can we break some use case here?

rszwajko (dismissed):

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -503,21 +517,32 @@ class Storage extends React.Component {
...
-      }
-      value = +value * (1024 ** 3) // GiB to B
+    const editingErrors = {}
+    switch (field) {
+      case 'size':
+        if (!isNumber(value)) return
+
+        const { minDiskSizeInGiB, maxDiskSizeInGiB } = this.props
+        if (value >= minDiskSizeInGiB && value <= maxDiskSizeInGiB) {
+          delete editingRow.invalidSizeValue
```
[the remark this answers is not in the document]
bamsalem: it was part of another fix, I'll change that

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -105,6 +107,9 @@ class Storage extends React.Component {
     this.isEditingMode = this.isEditingMode.bind(this)
 
     this.state = {
+      editingErrors: {
```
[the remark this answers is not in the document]
bamsalem: `editing` is an object that includes the data of the edited/created disk, and `creating`  just holds the id of the created disk in case of editing its value is false, so the answer is no, hope I explained it well :)

Creating disk:
creating
Editing disk:
editing

src/components/utils/validation.js
bamsalem: I added unit test :)

sgratch: @bamsalem @rszwajko regarding using isDiskNameValid in both create/edit modes:
we can't break any use case since isVmNameValid and isDiskNameValid actually share the same validation logic except length limitation, so now validating of disk alias name in DiskImageEditor is correct.
But two issues just bumped up:
-  is there any reason for using VM_NAME_REGEX.test(nameCandidate) within isVmNameValid function and  /^[a-zA-Z0-9_\-\\.]{1,255}$/.test(nameCandidate) within isDiskNameValid function? We can check the length limitation separately and share the same regex...

- for both DiskImageEditor and Storage.js we use the same validation but display a different error message (pleaseEnterValidDiskName and createVmWizardStorageDiskNameRules). I suggest using the same message.

sgratch (dismissed):

src/components/utils/validation.js
rszwajko: @sgratch  my understanding is that VM_NAME_REGEXP is **less** strict - check [UNICODE_NON_ASCII_LETTERS](https://github.com/oVirt/ovirt-web-ui/blob/a41b92f1e340480129c2cd5f6c9da58066ef96ce/src/components/utils/validation.js#L64)

bamsalem: @sgratch
I fixed the validations now we are using only the /^[a-zA-Z0-9_-\.]{1,255}$/.test(nameCandidate) validation.

Regarding the message we display different messages because the message in Create VM wizard must be shorter than DiskImageEditor, that more descriptive:

Long message:
image

Short message:
image

If you still think we should use the same message, WDYT? which one is preferred?

src/components/utils/validation.test.js
```diff
@@ -20,6 +20,28 @@ describe('check host names', function () {
   })
 })
 
+describe('check disk names', function () {
+  it('valid disk names', function () {
+    expect(isDiskNameValid('veryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryValidDiskName12')).toEqual(true)
```
sjd78, line 25:
```suggestion
    expect(isDiskNameValid('abcde'.repeat(51))).toEqual(true) // 255 characters
```

sjd78 (dismissed):

sjd78:
@sgratch , @bamsalem , @rszwajko - The VM name regexp needs to be changed back.  VMs can have names that do not contain latin alpha characters.  In fact, the regexp is a copy from webadmin's VM name validation.

I'm not sure if disk names can contain non-latin alpha characters.  I did not test that, or lookup the validation code, yet.

src/components/utils/validation.test.js
```diff
@@ -18,8 +18,30 @@ describe('check host names', function () {
...
+    expect(isDiskNameValid('-abc')).toEqual(true)
+  })
+
+  it('invalid disk names', function () {
+    expect(isDiskNameValid('')).toEqual(false)
+    expect(isDiskNameValid('a b')).toEqual(false)
+    expect(isDiskNameValid('a!@$b298')).toEqual(false)
+    expect(isDiskNameValid('a_name()')).toEqual(false)
+    expect(isDiskNameValid('Кирилиця')).toEqual(false)
+    expect(isDiskNameValid('veryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylongDiskName')).toEqual(false)
```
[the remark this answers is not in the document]
rszwajko: How about this: `'a'.repeat(256)`? we want to check exactly one above the limit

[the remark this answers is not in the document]
sjd78: > How about this: `'a'.repeat(256)`? we want to check exactly one above the limit

Even better

src/components/utils/validation.js
```diff
@@ -1,97 +1,42 @@
 // @flow
 
+const charactersRegExpTest = (nameCandidate: string) => /^[a-zA-Z0-9_\-\\.]*$/.test(nameCandidate)
```
rszwajko, line 3: @sgratch  Actually I'm not sure if we should throw away UTF chars. Check  I18NNameValidation which is used DiskAllocationModel - it [defines letters as all chars including UTF](https://github.com/oVirt/ovirt-engine/blob/e48484ccb3c1c8fcaed81023b234295a109ebd8b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/validation/I18NNameValidation.java#L11)

rszwajko (changes requested):

sgratch:
@bamsalem @sjd78 @rszwajko 

> is there any reason for using VM_NAME_REGEX.test(nameCandidate) within isVmNameValid function and /^[a-zA-Z0-9_-.]{1,255}$/.test(nameCandidate) within isDiskNameValid function? We can check the length limitation separately and share the same regex...

You are all right, based on the fact that disk alias string is auto generated from the vm name ({vmName}_Disk1) then we should support UTF for both disk name and vm name.

And by checking the code: 
disk: https://github.com/oVirt/ovirt-engine/blob/e48484ccb3c1c8fcaed81023b234295a109ebd8b/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/storage/DisksAllocationModel.java#L486-L490

vm: https://github.com/oVirt/ovirt-engine/blob/9059df4d26f5d327c87495ee9e19cb1876ad92b7/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/VmModelBehaviorBase.java#L1665-L1667

Both use _I18NNameValidation_().

So in our web-ui code we should use VM_NAME_REGEX.test(nameCandidate) for both validations


> Regarding the message we display different messages because the message in Create VM wizard must be shorter than DiskImageEditor, that more descriptive

I'm totally ok with having the shorter message version for both dialogs. They both cover the same info and it's better to maintain one message instead of two (if validation login will be ever changed).

bamsalem:
I changed it to use VM_NAME_REGEX again in diskName and vmName.
I changed the code according to your review @sgratch @sjd78 @rszwajko...


> I'm totally ok with having the shorter message version for both dialogs. They both cover the same info and it's better to maintain one message instead of two (if validation login will be ever changed).

Now the same message displayed in CreateVm and editVm:
image

sgratch:
@bamsalem please rebase

sjd78 (approved):
Component code and error messages look good.
Validation checks and tests look good as well.

bamsalem:
@sgratch 
rebased.

isaranova:
LGTM

Issue #1339, opened 2020-12-03
 VM Portal crash after viewing vm details of a suspending vm


sgratch:
When suspending a vm, try to view vm details during the hibernate action (before vm is successfully suspended). That sometimes causes the web-ui to crash with the following error:
![image](https://user-images.githubusercontent.com/18169498/101076552-37198b00-35ac-11eb-855c-da7d58f9ef91.png)

This seems somehow related to https://github.com/oVirt/ovirt-web-ui/pull/1337 (Update parsing of VM statistics)

PR #1358, opened 2021-01-18
Prevent VM Portal crash on viewing suspending/shut down vm details

bamsalem:
Fixes: #1339 .
The issue described in #1339 caused when the vm is starts and the entry `cpuStats['usage.history'].datum` has only one element, its value represented as a number and not as array with one element...
When this is occur we call `reverse()` on number but the function is undefined and that is the issue that caused web UI crash.

To fix it I changed the transform function `VmStatistics.toInternal` to set array in case of single element instead of set the element.
I'm not sure if it's necessary, but I added validation to check the value of `cpuStats['usage.history'].datum` to prevent future failures.

### DIFF
--- a/src/components/VmDetails/cards/UtilizationCard/NetworkingCharts.js
+++ b/src/components/VmDetails/cards/UtilizationCard/NetworkingCharts.js
@@ -31,7 +31,7 @@ import NoLiveData from './NoLiveData'
 const NetworkingCharts = ({ netStats, isRunning, id }) => {
   const haveNetworkStats = !!netStats['current.total']
 
-  const used = (netStats['current.total'] && netStats['current.total'].datum) || 0
+  const used = (netStats['current.total'] && netStats['current.total'].firstDatum) || 0
   const available = 100 - used
 
   // NOTE: Network history comes sorted from newest to oldest
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -305,7 +305,8 @@ const VmStatistics = {
       cpu: {},
       network: {},
       elapsedUptime: {
-        datum: 0,
+        firstDatum: undefined,
+        datum: [ 0 ],
         unit: 'seconds',
         description: 'Elapsed VM runtime (default to 0)',
       },
@@ -314,22 +315,28 @@ const VmStatistics = {
 
     for (const stat: ApiVmStatisticType of statistics) {
       if (stat.name === 'elapsed.time') {
-        base.elapsedUptime.datum = stat.values.value[0].datum
+        base.elapsedUptime.datum = stat.values.value.map((val: any) => val.datum)
+        base.elapsedUptime.firstDatum = base.elapsedUptime.datum[0]
         base.elapsedUptime.description = stat.description
       }
 
       if (stat.kind !== 'gauge') continue
 
-      // no values -> undefined, 1 value -> value.datum, >1 values -> [...values.datum]
-      // ?disks.usage -> {detail...}
-      let datum: any =
-        stat.values &&
-        stat.values.value &&
-        (stat.name === 'disks.usage'
-          ? stat.values.value[0] && stat.values.value[0].detail
-          : stat.values.value.length === 1
-            ? stat.values.value[0].datum
-            : stat.values.value.map(value => value.datum))
+      // no values -> undefined, >0 value -> [...values.datum]
+      let datum: any
+      if (stat.values && stat.values.value) {
+        if (stat.type === 'decimal' || stat.type === 'integer') {
+          datum = Array.isArray(stat.values.value)
+            ? stat.values.value.map((val: any) => val.datum)
+            : [stat.values.value.datum]
+        }
+
+        if (stat.type === 'string') {
+          datum = Array.isArray(stat.values.value)
+            ? stat.values.value.map((val: any) => val.detail)
+            : [stat.values.value.detail]
+        }
+      }
 
       if (stat.name === 'disks.usage' && !!datum) {
         datum = JSON.parse(datum)
@@ -339,9 +346,16 @@ const VmStatistics = {
           return data
         })
       }
+
+      const firstDatum: any =
+        (datum && datum.length > 0)
+          ? datum[0]
+          : undefined
+
       const nameParts = /^(memory|cpu|network|disks)\.(.*)?$/.exec(stat.name)
       if (nameParts) {
         base[nameParts[1]][nameParts[2]] = {
+          firstDatum,
           datum,
           unit: stat.unit,
           description: stat.description,
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -986,6 +986,7 @@ export {
   Icon,
   VmConsoles,
   VmSessions,
+  VmStatistics,
   CloudInit,
   Permissions,
   Event,
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -299,7 +299,7 @@ const VM = {
 //
 //
 const VmStatistics = {
-  toInternal ({ statistics }: { statistics: Array<ApiVmStatisticType> }): VmStatisticsType {
+  toInternal ({ statistics = [] }: { statistics: Array<ApiVmStatisticType> } = {}): VmStatisticsType {
     const base: VmStatisticsType = {
       memory: {},
       cpu: {},
--- a/src/ovirtapi/transform.test.js
+++ b/src/ovirtapi/transform.test.js
@@ -1,0 +1,341 @@
+/* eslint-env jest */
+
+import { VmStatistics } from './transform'
+
+const TESTS_DATA = {
+  title: 'Test transform.js VmStatistics.toInternal ',
+  testCases: [
+    {
+      title: 'test case 0',
+      test: [
+        {
+          id: '0',
+          kind: 'gauge',
+          type: 'integer',
+          unit: 'bytes',
+          values: {
+            value: [
+              { datum: 1073741824 },
+            ],
+          },
+          name: 'memory.installed',
+          description: 'Total memory configured',
+        },
+      ],
+      expect: { memory: { installed: { firstDatum: 1073741824, datum: [1073741824], unit: 'bytes', description: 'Total memory configured' } }, cpu: {}, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: {} },
+    },
+    {
+      title: 'test case 1',
+      test: [
+        {
+          id: '1',
+          kind: 'gauge',
+          type: 'decimal',
+          unit: 'percent',
+          values: {
+            value: [
+              { datum: 0.61 },
+            ],
+          },
+          name: 'cpu.current.total',
+          description: 'Total CPU used',
+        },
+      ],
+      expect: { memory: {}, cpu: { 'current.total': { firstDatum: 0.61, datum: [0.61], unit: 'percent', description: 'Total CPU used' } }, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: {} },
+    },
+    {
+      title: 'test case 2',
+      test: [
+        {
+          id: '2',
+          kind: 'gauge',
+          type: 'string',
+          unit: 'none',
+          values: {
+            value: [
+              {
+                detail: '[{"path":"/boot","total":"952840192","used":"100130816","fs":"ext4"},{"path":"/run/media/sdickers/Fedora-SB-ostree-x86_64-33","total":"2874408960","used":"2874408960","fs":"iso9660"}]',
+              },
+            ],
+          },
+          name: 'disks.usage',
+          description: 'Disk usage, in bytes, per filesystem as JSON (agent)',
+        },
+      ],
+      expect: {
+        memory: {}, cpu: {}, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: { usage: { firstDatum: { path: '/boot', total: 952840192, used: 100130816, fs: 'ext4' }, datum: [{ path: '/boot', total: 952840192, used: 100130816, fs: 'ext4' }, { path: '/run/media/sdickers/Fedora-SB-ostree-x86_64-33', total: 2874408960, used: 2874408960, fs: 'iso9660' }], unit: 'none', description: 'Disk usage, in bytes, per filesystem as JSON (agent)' } } },
+    },
+    {
+      title: 'test case 3',
+      test: [
+        {
+          id: '3',
+          kind: 'counter',
+          type: 'integer',
+          unit: 'seconds',
+          values: {
+            value: [
+              {
+                datum: 1670253,
+              },
+            ],
+          },
+          name: 'elapsed.time',
+          description: 'Elapsed VM runtime',
+        },
+      ],
+      expect: { memory: {}, cpu: {}, network: {}, elapsedUptime: { firstDatum: 1670253, datum: [1670253], unit: 'seconds', description: 'Elapsed VM runtime' }, disks: {} },
+    },
+    {
+      title: 'test case 4',
+      test: [
+        {
+          id: '4',
+          kind: 'gauge',
+          type: 'decimal',
+          unit: 'percent',
+          values: {
+            value: [
+              {
+                datum: 1,
+              },
+              {
+                datum: 2,
+              },
+              {
+                datum: 3,
+              },
+            ],
+          },
+          name: 'memory.usage.history',
+          description: 'List of memory usage history, sorted by date from newest to oldest, at intervals of 30 seconds',
+        },
+      ],
+      expect: { memory: { 'usage.history': { firstDatum: 1, datum: [1, 2, 3], unit: 'percent', description: 'List of memory usage history, sorted by date from newest to oldest, at intervals of 30 seconds' } }, cpu: {}, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: {} },
+    },
+    {
+      title: 'test case 5',
+      test: [
+        {
+          id: '5',
+          kind: 'gauge',
+          type: 'integer',
+          unit: 'bytes',
+          values: {
+            value: [
+              {
+                datum: 268435456,
+              },
+            ],
+          },
+          name: 'memory.used.history',
+          description: 'Memory used (agent)',
+        },
+      ],
+      expect: { memory: { 'used.history': { firstDatum: 268435456, datum: [268435456], unit: 'bytes', description: 'Memory used (agent)' } }, cpu: {}, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: {} },
+    },
+    {
+      title: 'test case 6',
+      test: [
+        {
+          id: '6',
+          kind: 'gauge',
+          type: 'decimal',
+          unit: 'percent',
+          values: {
+            value: [
+              {
+                datum: 0.74,
+              },
+            ],
+          },
+          name: 'cpu.current.guest',
+          description: 'CPU used by guest',
+        },
+      ],
+      expect: { memory: {}, cpu: { 'current.guest': { firstDatum: 0.74, datum: [0.74], unit: 'percent', description: 'CPU used by guest' } }, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: {} },
+    },
+    {
+      title: 'test case 7',
+      test: [
+        {
+          id: '7',
+          kind: 'gauge',
+          type: 'decimal',
+          unit: 'percent',
+          values: {
+            value: [
+              {
+                datum: 0.74,
+              },
+            ],
+          },
+          name: 'cpu.current.hypervisor',
+          description: 'CPU overhead',
+        },
+      ],
+      expect: { memory: {}, cpu: { 'current.hypervisor': { firstDatum: 0.74, datum: [0.74], unit: 'percent', description: 'CPU overhead' } }, network: {}, elapsedUptime: { datum: [0], unit: 'seconds', description: 'Elapsed VM runtime (default to 0)' }, disks: {} },
+    },
+    {
+      title: 'test case 8',
+      test: [
+        {
+          id: '8',
+          kind: 'gauge',
+          type: 'decimal',
+          unit: 'percent',
+          values: {
+            value: [
+              {
+                datum: 0.74,
+              },
+            ],
+          },
+          name: 'cpu.current.hypervisor',
+          description: 'CPU overhead',
+        },
+      ],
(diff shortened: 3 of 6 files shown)


bamsalem:
ci add to whitelist

src/ovirtapi/transform.js
```diff
@@ -327,9 +327,7 @@ const VmStatistics = {
         stat.values.value &&
         (stat.name === 'disks.usage'
           ? stat.values.value[0] && stat.values.value[0].detail
-          : stat.values.value.length === 1
-            ? stat.values.value[0].datum
-            : stat.values.value.map(value => value.datum))
+          : stat.values.value.map(value => value.datum))
```
rszwajko, line 330: It seems the API changed or is inconsistent. According to the comment above the API used to return a single object instead of a single element array.
I like that we treat all cases uniformly (as an array). Perhaps we could also check for the (legacy?) scenario with single element without array and wrap it into array?

bamsalem: I made changes there, now it's array or undefined, please let me know WDYT

src/ovirtapi/transform.js
```diff
@@ -327,9 +327,9 @@ const VmStatistics = {
         stat.values.value &&
         (stat.name === 'disks.usage'
           ? stat.values.value[0] && stat.values.value[0].detail
-          : stat.values.value.length === 1
-            ? stat.values.value[0].datum
-            : stat.values.value.map(value => value.datum))
+          : (Array.isArray(stat.values.value)
+            ? stat.values.value.map(value => value.datum)
+            : [stat.values.value]))
```
rszwajko, line 332: Sorry, I think I've misunderstood the comment (but let's @sjd78 clarify that):
> no values -> undefined, 1 value -> value.datum, >1 values -> [...values.datum]

The left side is input(JSON from backend) and right is our output, right? What are the possible values of the input: `undefined` or an array? Or is it  sometimes a single object (not wrapped in array)?

Anyway, the current version could result in `[undefined]`. We should probably return empty array then.

sgratch: Based on backend data, this is always returned as an array (List<Integer>), so not sure converting  here is required.

I think that the fix for stat.name !== 'disks.usage' should be the same as for 'disks.usage'. i.e.: checking that  stat.values.value[i]  is defined because in case of null values, the datum was not set on backend.
@sjd78 ?

hstastna: @rszwajko @sgratch I checked that and when `stat.values.value` is `undefined`, we end up with `datum` set to `undefined`, not `[undefined]` because of [this](https://github.com/oVirt/ovirt-web-ui/pull/1358/files#diff-83343b99424fb2a62ce7723a77810ebe7d4f3ccf8fd53ad4cd92c244902dc80aR327). And if `stat.values.value` is something but not an array, it will become wrapped in array because of [this change](https://github.com/oVirt/ovirt-web-ui/pull/1358/files#diff-83343b99424fb2a62ce7723a77810ebe7d4f3ccf8fd53ad4cd92c244902dc80aR332). IMO this change is okay.

src/components/VmDetails/cards/UtilizationCard/NetworkingCharts.js
```diff
@@ -31,11 +31,11 @@ import NoLiveData from './NoLiveData'
 const NetworkingCharts = ({ netStats, isRunning, id }) => {
   const haveNetworkStats = !!netStats['current.total']
 
-  const used = (netStats['current.total'] && netStats['current.total'].datum) || 0
+  const used = (netStats['current.total'] && netStats['current.total'].datum && netStats['current.total'].datum[0]) || 0
```
hstastna, line 34: @bamsalem Why don't we need to check here if `netStats['current.total'].datum` is an array, too?

hstastna (approved):
Tested the changes in this PR and it works as expected, the changes prevent VM Portal crashing after viewing VM Details of a suspending VM.

Although there is a little bit of uncertainty regarding some of the changes, I am ok with them.

src/components/VmDetails/cards/UtilizationCard/NetworkingCharts.js
bamsalem: You're right it should be checked, I added that thanks :)

src/components/VmDetails/cards/UtilizationCard/NetworkingCharts.js
```diff
@@ -31,11 +31,11 @@ import NoLiveData from './NoLiveData'
 const NetworkingCharts = ({ netStats, isRunning, id }) => {
   const haveNetworkStats = !!netStats['current.total']
 
-  const used = (netStats['current.total'] && netStats['current.total'].datum) || 0
+  const used = (netStats['current.total'] && Array.isArray(netStats['current.total'].datum) && netStats['current.total'].datum[0]) || 0
```
sgratch, line 34: Any reason to check Array.isArray() here? If so then it should be applied to all places referring to datum[0].

src/ovirtapi/transform.js
sgratch, line 332: not sure but shouldn't it be just: 
[stat.values.value.datum]
instead of: 
[stat.values.value]

sgratch (commented):
@bamsalem please check inline comments

rszwajko:
@bamsalem  so many questions to such small piece of code:) I think this is the best sign we should unit test it and show explicitly what input data is expected and what should be returned.

src/ovirtapi/transform.js
sgratch: // no values -> undefined, 1 value -> value.datum, >1 values -> [...values.datum]
      // ?disks.usage -> {detail...}
      // ?disks.usage -> {detail...}
Please update the comment on 323-324 as well

sjd78:
> @bamsalem so many questions to such small piece of code:) 

Because it's doing "complicated" transforms, and when the source data changes...well...it's fun times.

> I think this is the best sign we should unit test it and show explicitly what input data is expected and what should be returned.

Probably, maybe.  Still wouldn't have caught this particular error.  But it would at least prove what input to output transformation is expected.

@bamsalem, @rszwajko, @hstastna, @sgratch - Some more background info:

Really it was https://gerrit.ovirt.org/c/ovirt-engine/+/112146 that has caused the problems.

The engine code that builds the statistical data for the rest call is here: 
https://github.com/oVirt/ovirt-engine/blob/9059df4d26f5d327c87495ee9e19cb1876ad92b7/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/VmStatisticalQuery.java#L39-L75

Our ovirt-web-ui flow types are nearly accurate (they could be updated to mark "optional" a bunch more stuff):
https://github.com/oVirt/ovirt-web-ui/blob/04977cc8297cb055d32258425ecdf8681267825b/src/ovirtapi/types.js#L8-L37

When there is no value, or the value is empty (this is the change introduces in 112146), the `values` containing object will just be empty or even non-existent in the json response. #1337 changed the transform to accommodate this a bit more.

When there is a singe statistical value, it is still wrapped in an array.  Only history stats have >1 value.  The transform unwraps the single value from the array.  This allows accessing that value without having to do `.datum[0]`.  I mean the whole point of doing the transforms are to make the data easier to use in the app itself.  So the `stat` goes from `stat.values.value[0].datum` to `stat.datum`.

When there are multiple statistical values, the datum values are plucked out of their wrapping objects.  So the `stat` goes from `stat.values.value[x].datum` to `stat.datum[x]`.

The `disks.usage` bit gets special treatment since the value is not "datum" value but a "detail" string and the contents is a JSON string that needs to be further parsed.  So the `stat` goes from `JSON.parse(stat.values.value[0].detail)[x].path` to `stat.datum[x].path`.

src/ovirtapi/transform.js
sjd78: I'm gonna suggest just rewriting this section to be less cryptic.

How does this look:
```js
    for (const stat: ApiVmStatisticType of statistics) {
      if (stat.name === 'elapsed.time') {
        base.elapsedUptime.datum = stat.values.value[0].datum
        base.elapsedUptime.description = stat.description
      }

      if (stat.kind !== 'gauge') continue

      // no values -> undefined
      let datum: any

      if (stat.values && stat.values.value) {
        if (stat.type === 'decimal' || stat.type === 'integer') {
          // 1 value -> value.datum, >1 values -> [...values.datum]
          datum = Array.isArray(stat.values.value) && stat.values.value.length === 1
            ? stat.values.value[0].datum
            : stat.values.value.map(value => value.datum)
        }

        if (stat.type === 'string') {
          // 1 value -> value.detail, >1 values -> [...values.detail]
          datum = Array.isArray(stat.values.value) && stat.values.value.length === 1
            ? stat.values.value[0].detail
            : stat.values.value.map(value => value.detail)
        }
      }

      // further parse the JSON string in `disk.usage` stats
      if (stat.name === 'disks.usage' && !!datum) {
        datum = JSON.parse(datum).map(data => {
          data.total = convertInt(data.total)
          data.used = convertInt(data.used)
          return data
        })
      }

      const nameParts = /^(memory|cpu|network|disks)\.(.*)?$/.exec(stat.name)
      if (nameParts) {
        base[nameParts[1]][nameParts[2]] = {
          datum,
          unit: stat.unit,
          description: stat.description,
        }
      }
    }
```

And then the other users of the stats data can go back to how they were.

sjd78 (changes requested):
I'm suggesting a different change to the transform that will not require any changes to component code.

sgratch:
@sjd78 sounds ok but it raises 3 questions to verify:
1. Don't we still need to solve the original issue of an exception caused by calling `reverse`() on a non array element `cpuStats['usage.history'].datum`? e.g.: 
https://github.com/oVirt/ovirt-web-ui/blob/cca163c868cefdaf27736a1b4447ee23c955ef7f/src/components/VmDetails/cards/UtilizationCard/CpuCharts.js#L33

2. in your suggested solution, you are not based only on `stat.name === 'disks.usage'` for reading the `detail` field but rather on `stat.type === 'string'` so need to make sure that it's always working and will work as expected based on backend changes.

3. in your suggested solution, you are based on the fact that if `stat.type === 'string',` then `stat.values.value.length => 1` which was not the case up til now since we assumed there is only a singe statistical value.

sjd78:
> @sjd78 sounds ok but it raises 3 questions to verify:
> 
> 1. Don't we still need to solve the original issue of an exception caused by calling `reverse`() on a non array element `cpuStats['usage.history'].datum`? e.g.:
>    https://github.com/oVirt/ovirt-web-ui/blob/cca163c868cefdaf27736a1b4447ee23c955ef7f/src/components/VmDetails/cards/UtilizationCard/CpuCharts.js#L33

That check can be more specific sure, but the datum value will either `undefined` if no values are in the response, a single number if 1 value is in the response, or an array of values if multiple values are in the response.  For `undefined`, it will work fine.  For an array of values, it'll work fine.  For a single value, it would need further checks.

Maybe the check should look more like this just in case there is a single history item:
```js
  const datum = cpuStats['usage.history'] && cpuStats['usage.history'].datum
  const history = !!datum ? Array.isArray(datum) ? datum.reverse() : [ datum ] : []
```

> 2. in your suggested solution, you are not based only on `stat.name === 'disks.usage'` for reading the `detail` field but rather on `stat.type === 'string'` so need to make sure that it's always working and will work as expected based on backend changes.

In the backend code, if `stat.type === 'string'` then the data is stored under `detail` instead of `datum`.  The first if blocks handles that bit.  The second if block further parses `disks.usage` values.

See https://github.com/oVirt/ovirt-engine/blob/9059df4d26f5d327c87495ee9e19cb1876ad92b7/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/utils/StatisticResourceUtils.java#L45-L52

> 3. in your suggested solution, you are based on the fact that if `stat.type === 'string',` then `stat.values.value.length => 1` which was not the case up til now since we assumed there is only a singe statistical value.

True.  Handling of multiple datum or multiple detail would be the same.  This doesn't change how things will work for `disks.usage` stats as it only sends a single detail string.

rszwajko:
@sjd78 
Few thoughts on this part:

1. it's good to know the type of data slice that you are using - if it can be array sometimes and sometimes a single object then you need to check each time. If for some usages we only need the first datum then let's create 2 properties: `firstDatum` (object or undefined) and `data` (array). 
2. part of the complexity comes from trying to process all possible properties at once. In the end we have conditions checking for prop names all along the way. How about splitting the processing per prop? We could re-use lower level parsing routines (i.e. the one for `stat.type === 'string'`)
3. unit tests would play a nice role here:)

sjd78:
> @sjd78
> Few thoughts on this part:
> 
>    1. it's good to know the type of data slice that you are using - if it can be array sometimes and sometimes a single object then you need to check each time. If for some usages we only need the first datum then let's create 2 properties: `firstDatum` (object or undefined) and `data` (array).

That seems reasonable.  `datum` always an array reguardless as to the source data.  `firstDatum` as `undefined` or the actual first element in the `datum` array.

> 
>     2. part of the complexity comes from trying to process all possible properties at once. In the end we have conditions checking for prop names all along the way. How about splitting the processing per prop? We could re-use lower level parsing routines (i.e. the one for `stat.type === 'string'`)

In the transform code only 1 stat prop needs special handling, and that is to parse a JSON string to normal JS.  Generic processing of the stats is reasonable to me.

The condition checking in the components is pretty normal across the app so I don't see a problem there.  Test for existence of a piece of data and then render.

> 
>     3. unit tests would play a nice role here:)

Unit tests on the transform function to show how the anticipated api data converts to app data?  Absolutely!

To see how data looks so components can model data tests after it?  Sure!

For catching the source of this PR's issue?  Not at all.  Changes to the source data wouldn't get caught by a unit test.

I wouldn't spend time adding unit tests in this PR.  We'd want more unit tests on api/transform work ultimately, so we can create a new issue to track time and progress adding those.

rszwajko:
@sjd78  
> In the transform code only 1 stat prop needs special handling, and that is to parse a JSON string to normal JS. Generic processing of the stats is reasonable to me.

Generally I agree. Parsing values based on their type (string vs number) is OK.
I was referring more to the entire procedure which has checks like: 
```
if (stat.name === 'elapsed.time') {
...
if (stat.kind !== 'gauge') continue
...
 if (stat.name === 'disks.usage' && !!datum) {
...
/^(memory|cpu|network|disks)\.(.*)?$/.exec(stat.name)
```
 
> I wouldn't spend time adding unit tests in this PR. We'd want more unit tests on api/transform work ultimately, so we can create a new issue to track time and progress adding those.

I would start small:) we have a bug scenario so let's create a failing test case and fix the code so that it will start passing.

bamsalem:
let's sum up:
I summed up what most of us seemed to agree on:
It will be better to set datum to array if there is value or undefined in case of empty array and to access the single element we should use new entry, `firstDatum` that will hold the first element in the array or `undefined` in case of empty array.
I took the changes @sjd78  suggested [here](https://github.com/oVirt/ovirt-web-ui/pull/1358#discussion_r567121574) and I adjusted them to hold array of elements.
Regarding the unit test IMO it won't help to find changes in run time, so I think it unnecessary here.

src/ovirtapi/types.js
```diff
@@ -24,7 +24,8 @@ export type ApiVmStatisticType = {
 }
 
 export type StatisticValueType = {
-  datum: number | Array<number>,
+  firstDatum: number | void,
+  datum: Array<number>,
```
sjd78, line 28: `number` isn't the only thing the datum can be ... they can be number or string or objects even.  So probably better to be at least a `number | string | Object`.

sjd78 (changes requested):
The `firstDatum` use looks good and the transform changes look good.  Just a question about and small tweak on the type structure and then it'll be good to go.

sjd78:
> @sjd78
> 
> > In the transform code only 1 stat prop needs special handling, and that is to parse a JSON string to normal JS. Generic processing of the stats is reasonable to me.
> 
> Generally I agree. Parsing values based on their type (string vs number) is OK.
> I was referring more to the entire procedure which has checks like:
> 
> ```
> if (stat.name === 'elapsed.time') {
> ...
> if (stat.kind !== 'gauge') continue
> ...
>  if (stat.name === 'disks.usage' && !!datum) {
> ...
> /^(memory|cpu|network|disks)\.(.*)?$/.exec(stat.name)
> ```
> 

Ah, I got ya now.  Oh well, it's just complicated enough now to keep you on your toes.

> > I wouldn't spend time adding unit tests in this PR. We'd want more unit tests on api/transform work ultimately, so we can create a new issue to track time and progress adding those.
> 
> I would start small:) we have a bug scenario so let's create a failing test case and fix the code so that it will start passing.

That's a reasonable approach. I forget about writing tests beforehand as a kind of i/o spec...

@bamsalem - Do you want to create a handful of simple test cases and put them in a file `transform.test.js`?  Something that would just show how empty data, and a few of the expected stats, should be transformed?

Some sample inputs:
```js
  [
    {
      "kind": "gauge",
      "type": "integer",
      "unit": "bytes",
      "values": {
        "value": [
          { "datum": 1073741824 }
        ]
      },
      "name": "memory.installed",
    },
  ]
```
```js
    [{
      "kind": "gauge",
      "type": "decimal",
      "unit": "percent",
      "values": {
        "value": [
          { "datum": 0.61 }
        ]
      },
      "name": "cpu.current.total",
      "description": "Total CPU used",
    }]
```
```js
  [
    {
      "kind": "gauge",
      "type": "string",
      "unit": "none",
      "values": {
        "value": [
          {
            "detail": "[{\"path\":\"/boot\",\"total\":\"952840192\",\"used\":\"100130816\",\"fs\":\"ext4\"},{\"path\":\"/run/media/sdickers/Fedora-SB-ostree-x86_64-33\",\"total\":\"2874408960\",\"used\":\"2874408960\",\"fs\":\"iso9660\"}]"
          }
        ]
      },
      "name": "disks.usage",
      "description": "Disk usage, in bytes, per filesystem as JSON (agent)",
    },
  ]
```
```js
  [
    {
      "kind": "counter",
      "type": "integer",
      "unit": "seconds",
      "values": {
        "value": [
          {
            "datum": 1670253
          }
        ]
      },
      "name": "elapsed.time",
      "description": "Elapsed VM runtime",
    }
  ]
```
```js
  [
    {
      "kind": "gauge",
      "type": "decimal",
      "unit": "percent",
      "values": {
        "value": [
          {
            "datum": 1
          },
          {
            "datum": 2
          },
          {
            "datum": 3
          },
        ]
      },
      "name": "memory.usage.history",
      "description": "List of memory usage history, sorted by date from newest to oldest, at intervals of 30 seconds",
    },

  ]
```
Running just the tests you're working on can be done like this `yarn test part/of/path` e.g. `yarn test transform`

src/ovirtapi/types.js
bamsalem: Do you have any example for string or object, from what I tested is always numbers, undefined is for empty array...

bamsalem:
> @bamsalem - Do you want to create a handful of simple test cases and put them in a file `transform.test.js`? Something that would just show how empty data, and a few of the expected stats, should be transformed?

@sjd78 if you think that it would help and make that piece of code more readable I'll do it...

src/ovirtapi/types.js
sgratch: @bamsalem just check the code on backend: https://github.com/oVirt/ovirt-engine/blob/9059df4d26f5d327c87495ee9e19cb1876ad92b7/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/utils/StatisticResourceUtils.java#L36-L74

BigDecimal, Long and of couse String should be supported as well.

sgratch (approved):
LGTM

> > @bamsalem - Do you want to create a handful of simple test cases and put them in a file `transform.test.js`? Something that would just show how empty data, and a few of the expected stats, should be transformed?
> 
> @sjd78 if you think that it would help and make that piece of code more readable I'll do it...

Adding tests is always welcomed for stability even though in this specific case it won't prevent bugs like this.
Ben, if it won't take you too much then please add it  but please don't spend too much time on that so that we'll be able to merge it soon.

src/ovirtapi/transform.test.js
```diff
@@ -0,0 +1,343 @@
...
+    },
+    },
+  },
+}
+describe(testData.title, () => {
+  const result = VmStatistics.toInternal({ statistics: testData.test })
+  test(testData.title, () => {
+    expect(result).toEqual(testData.expect)
+  })
+})
```
sgratch, line 343: There is no test for a use case of an empty statistics (value is undefined) as represented here https://gerrit.ovirt.org/112146

sgratch (dismissed):

src/ovirtapi/types.js
sjd78: > Do you have any example for string or object, from what I tested is always numbers, undefined is for empty array...

Currently, a datum will be a String only for the "disks.usage" statistic.  You need to have a running VM with a valid guest agent running to get that value.  A current install of Fedora or SilverBlue should do the trick.

See: [VmStatisticalQuery.java](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/VmStatisticalQuery.java) for the full list of VM stats currently generated.

src/ovirtapi/transform.test.js
```diff
@@ -0,0 +1,343 @@
+/* eslint-env jest */
+
+import { VmStatistics } from './transform'
+
+const testData = {
+  title: 'test vm statistics',
+  test: [
```
sjd78, line 7: Please break each test into its own test case so they can fail independently.

You can make the `describe` a loop.  Here is an example:
https://github.com/oVirt/ovirt-web-ui/blob/a151047be41d90d29c40fde5d7b532ffd5e46729/src/intl/messages.test.js#L53-L75

Basically the `describe.each()` call runs 1 test per iterator item.  Make your test data look more like:
```jsx
const TEST_DATA = [
  {
    title: 'test 1',
    test: [
      ...
    ],
    expect: [
    ]
  },
  {
    title: 'test 2',
    test: ...,
    expect: ...,
  },
  ...
]
```

src/ovirtapi/transform.test.js
sjd78: So just have a:
```js
  {
    title: 'undefined input',
    test: undefined,
    expect: { ... }
  }
```

and probably a few other base cases like that.

sjd78 (changes requested):
Base code changes LGTM.

Good first cut at the unit tests. Use `describe.each()` to have a test run at a finer gained level so when it fails, it is easier to spot the error.  Model after the intl messages testing and you should be good.

src/ovirtapi/transform.test.js
bamsalem: IMO It shouldn't be tested because it's tested before the call, how the `VmStatistics.toInternal` function is implemented it's not a test case because it's tested before call: https://github.com/oVirt/ovirt-web-ui/blob/40774f083130850f7712d5fa7ce6e4affa3f2c2f/src/ovirtapi/transform.js#L216-L218
and the flow types is not include `undefined` option.
WDYT?

src/ovirtapi/transform.test.js
bamsalem: Done.

src/ovirtapi/transform.test.js
bamsalem: IMO this shouldn't be tested since the flow types is not including undefined and in the one place we call the function we tests  this case: https://github.com/oVirt/ovirt-web-ui/blob/40774f083130850f7712d5fa7ce6e4affa3f2c2f/src/ovirtapi/transform.js#L216-L218.
If we still want to test it we should edit the function.

sjd78: I update the unit test - all the cases I could think of are now covered.

sjd78 (approved):
LGTM

Issue #1360, opened 2021-01-18
Create VM Wizard - Disk invalid size tooltip is not consistent


sgratch:
On the "Create Virtual Machine" wizard -> "Storage" step, when adding/editing a Disk and setting an invalid disk size value, the validation tooltip shape (e.g. "please select a value that is no less than one") is not consistent with https://github.com/oVirt/ovirt-web-ui/pull/1289:

![image](https://user-images.githubusercontent.com/18169498/104946957-9815ea80-59c3-11eb-9dfb-83ad8a80f88b.png)

PR #1361, opened 2021-01-20
Remove disk size inconsistent tooltip

bamsalem:
The tooltip discussed on #1360 is a side effect of HTML min max validations (the tooltip is part of the input that appears on error).

Like we talked in the meeting I removed the min max values and as well I implemented  the validation in the JS level.
As a result of implementing the validations in the JS level the field’s value can be no value out of range so the notifiers became unnecessary and were removed too.
 the tooltip:
image

P.S: It creates the same behavior as edit existing VM disk

Fixes: #1360 .

### DIFF
--- a/src/components/CreateVmWizard/steps/Storage.js
+++ b/src/components/CreateVmWizard/steps/Storage.js
@@ -18,7 +18,6 @@ import {
   Checkbox,
   DropdownKebab,
   EmptyState,
-  FormGroup,
   FormControl,
   MenuItem,
   Table,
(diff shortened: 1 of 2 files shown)


bamsalem:
ci add to whitelist

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -241,24 +240,18 @@ class Storage extends React.Component {
...
           const row = this.state.editing[rowData.id]
           const sizeGiB = row.size / (1024 ** 3)
-          const { invalidSizeValue } = row
 
           return <div className={style['disk-size-edit']}>
-            <FormGroup
-              validationState={invalidSizeValue && 'error'}
-              className={style['form-group-edit']}
-            >
+            <div>
```
hstastna, line 245: Could you please make it clear why you removed `FormGroup` component? Or why do we prefer `div` instead of `FormGroup`? Usually, `FormControl` is used inside of `FormGroup`. It seems to be working and being displayed okay as it is. I am just asking :)

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -241,24 +240,18 @@ class Storage extends React.Component {
...
                 step={1}
                 value={sizeGiB}
                 className={style['disk-size-form-control-edit']}
                 onChange={e => this.handleCellChange(rowData, 'size', e.target.value)}
               />
-            </FormGroup>
+            </div>
             <span className={style['disk-size-edit-label']}>GiB</span>
             <div>
               <InfoTooltip id={`${idPrefix}-${value}-size-edit-info-tooltip`} tooltip={msg.diskEditorSizeCreateInfoTooltip()} />
```
hstastna, line 257: Could you please fix this message? Use "extend", not extends". Thanks.
Check also: https://github.com/oVirt/ovirt-web-ui/commit/bb683aaf83804f029c8c7fae7398bf082d726a83#diff-2ad0bd2e35e6b177ef9487ee23d8cbeed8061f518530fd769652907c66f4860cL208

hstastna (commented):
@bamsalem Please, check my comments in the code. Otherwise, LGTM :)

src/components/CreateVmWizard/steps/Storage.js
bamsalem: `FormControl` itself is spreading over the whole cell and hides every other "sibling" component so it have to be wrapped by another component, It could be `FormGroup` but I thought `div` is better because it's simpler and enough for this job...

src/components/CreateVmWizard/steps/Storage.js
bamsalem: Sure :)

sgratch (approved):
LGTM

sjd78 (approved):
LGTM

sjd78:
@bamsalem - please rebase

PR #1372, opened 2021-02-25
Change locale and display email via Account Settings

rszwajko:
Note that this PR depends on patch [core: migrate users.options to user_profiles](https://gerrit.ovirt.org/c/ovirt-engine/+/113128) which creates a new REST API endpoint `/users/123/options`.

Feature summary:
1. allow user to change the locales and persist them on the server
2. display user's email address (read only) in the Account Settings
2. the locales follow the same convention as oVirt landing page (as requested in [this review](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359))
2. limitations (details provided in [this review](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359))
  ** page reload is required since we have no "hot reload" capability 
  ** locales are saved in the local storage as optimisation to prevent second reload after locale is fetched from the server
2. give higher priority to locales passed in URL. They will override the locales persisted on the server. Note that similar as in Web Admin they will not be persisted.
3. refactoring - pass locales form URL detected by intl to AppConfiguration (previously configuration parsed the locales independent from the intl code)

Implementation (key points):
1. change the structure of Redux `options` slice to clearly split local and remote options
2. fetch all remote options but pick only supported (recognized by this version). This allow us to avoid breaking data used by Web Admin (and other parties). Previously this required costly merging - see details in section [Backward and Forward compatibility](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-645990615)
2. remote options behave now very similar to SSH keys. Both `/sshpublickeys` and `/options` endpoints are views over the same data - `user_profiles` table.
3. `locale` property is added as standalone user profile property in `user_profiles` table to allow easy re-use in Web Admin (if we decide so)


ovirt-infra:

sgratch:
ci add to whitelist

rszwajko:
Added additional feature to this PR - ability to automatically save user locales.
This was needed to address the following scenario:
1. on user's first login our intl code has inferred some non-default locale i.e. "de". 
2. no locale is stored on the server(yet) so UI loads using the inferred locale
3. user enters Account Settings 
4. inferred locale is visible as chosen (in the language drop down). Unfortunately we display in the same way locale persisted on the server.

With the auto-save feature the locale will be persisted on the server without any user actions so after entering Account Settings the displayed language will be in sync with the value persisted on the server. 
Note that once the locale is persisted it can be only changed via Account Settings. URL-provided locale will not overwrite the value stored on the server (although it will overwrite settings on the client temporarily).

This partially addresses concerns raised in this [review](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359) 
> We can decide that the locale chosen via landing page is the default for all users and each user can change it via user settings. Once changed via user setting, it will change the default engine settings stored on local storage as well. So each time the user select a locale it will influence both landing page and user settings.

sgratch:
> Note that this PR depends on patch core: migrate users.options to user_profiles which creates a new REST API endpoint /users/123/options.

 @rszwajko we need to consider updating the minimum oVirt version be compatible with REST enpoint for 4.4.5. 
Since current version.major, version.minor is set to 4.4:
https://github.com/oVirt/ovirt-web-ui/blob/a151047be41d90d29c40fde5d7b532ffd5e46729/src/version.js.in#L6-L8 
then we need to consider reading and verifying the version.full_version field instead for z stream:
http://ovirt.github.io/ovirt-engine-api-model/master/#types/version

rszwajko:
> @rszwajko we need to consider updating the minimum oVirt version be compatible with REST enpoint for 4.4.5.
> Since current version.major, version.minor is set to 4.4:

I think we should be compatible with the minor version (4.4). However we might disable features not available in releases < 4.4.5 . Feature flag can be created based on the release number. It seems that this information is already available (`build` property) - see here: 
https://github.com/oVirt/ovirt-engine/blob/4a938c5155e18c1f9255908ae45be64d366dcc95/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/types/VersionMapper.java#L17

rszwajko:
@sgratch @sjd78 
Remote options are now disabled based on `version.build`

rszwajko:
Change summary:
1. Auto-exporting user locales has been re-implemented. In the new version the logic is in the saga layer not in `LocaleReloader`. This solves issues with errors logged on the server after reloading the page.
2. Tests in prod mode revealed that when user is accessing VM Portal via landing page then always a locale is passed in the URL. 
From functional point of view, it means that locale from Account Settings will never be used unless user enters the UI directly using link. For this reason it seems a better choice to ignore the locales from landing page if we have locales saved on the server.
3. enforce a common prefix to all properties used by VM Portal - this way the properties are much easier to manage at DB level

All changes were implemented as separate commits (to be squashed before merge). Force push was necessary to delete original commit that introduced auto-exporting.

rszwajko:
Change summary:
1. remove code that was hiding remote options for API version < 4.4.5 - the expected behaviour is to fail
2. added delay before setting `loadingFinished` flag - it's a workaround for page reload being too fast in newest versions.

sgratch:
@rszwajko Few verification comments, didn't dig into the code yet:
1. please add the info regarding the email addition to PR title as well (e.g.change user locale and display email via account...)
2. local storage usage issue: 
if a user is logged in when db vmPortal.locale property doesn't exist or exists and set to language "A" but local storage locale is stored with language set to "B" then the  storage local is the one displayed on settings model. This seems wrong since after the user is logged in, the properties should been read from the server or taken from landing page if it's the 1st login.
3. local storage usage issue (seems related to 2): 
When a user is logged in for the first time but local storage consists of a locale from other user previously logged in, for some reason the local is set to be as the previous user instead of the one selected on landing page.
4. I'm not sure about the delay added before setting loadingFinished flag  to avoid cases in which validation of SSH key fails while locale is saved correctly. This seems ok now but once other properties will be added to settings and might fail as well then it will be much more complicated to read the list of failures on that given time before reloading.
Why not deciding that saving settings is done as a commit block only when all in dialog is valid? This way - either an error message is displayed with a list of failures or all is succeeded which is an easy/short message to read. it's not that a bad user experience since we use that concept for all other dialogs/cards.
5. Regarding landing page - user settings local integration:
This changes the current behavior for non first logins. Consider adding a reset/CB button for user settings locale so that at least user can change it's locale based on landing page for next logins.

rszwajko:
>  1. please add the info regarding the email addition to PR title as well 
>     (e.g.change user locale and display email via account...)
> 

Done.

>     2. local storage usage issue:
>        if a user is logged in when db vmPortal.locale property doesn't exist or exists and set to language "A" 
>        but local storage locale is stored with language set to "B" then the  storage local is the one displayed
>        on settings model. This seems wrong since after the user is logged in, the properties should been read
>        from the server or taken from landing page if it's the 1st login.
> 

Currently (in this patch) the locale are discovered as follows:
1. local storage  then
2. URL then
3. browser then
4. defaults

The base scenarios are:
1.  `vmPortal.locale` exists  on the server but no value in local storage exists
    1. at first locale is discovered - since there is no value local storage one of the remaining sources is used i.e. URL
    2. after `vmPortal.locale` is fetched  a value in local storage is created
    3. a page reload is triggered
    4. value from local storage (which just was created) is used. 
2.  `vmPortal.locale` exists on the server, local storage also exists but has different value
    1. at first value from local storage is used
    2. then the flow is the same for case 1) - after fetching value from the server the value in local storage is overwritten
3.  no `vmPortal.locale` on the server, no value in local storage
    1. at first locale is discovered based on other sources then local storage
    2. user can change the locale via Account Settings - the value will be persisted on the server and in the local storage
    3. a page reload is triggered
    4. value from local storage (which just was created) is used. 
4.  no `vmPortal.locale` on the server but a value in local storage exists (i.e. left from different login)
    1. at first value from local storage is used (this source has the highest priority)
    2. similar as in case 3) user can change it via Account Settings

All of above scenarios worked for me(at some point in the lifecycle of this PR).

>     3. local storage usage issue (seems related to 2):
>        When a user is logged in for the first time but local storage consists of a locale from other user 
>        previously logged in, for some reason the local is set to be as the previous user instead of 
>        the one selected on landing page.
> 

Yes, this is case 4) described above. The whole advantage of using local storage is that it works across different sessions. However we do not store user ID there so it works for all accounts.  As long as the same person logs in using different accounts then this is beneficial because each account inherits the same locale (during first log in). If that heuristic is incorrect then user needs to manually change the locale via Account Settings. Locale from landing page (the one passed via URL/cookie) has lower priority then local storage so it's ignored. 

>     4. I'm not sure about the delay added before setting loadingFinished flag  to avoid cases 
>         in which validation of SSH key fails while locale is saved correctly. This seems ok now 
>         but once other properties will be added to settings  and might fail as well then it will be 
>         much more complicated to read the list of failures on that given time before reloading.
>        Why not deciding that saving settings is done as a commit block only when all in dialog is valid? 
>        This way - either an error message is displayed with a list of failures or all is succeeded which
>        is an easy/short message to read. it's not that a bad user experience since we use that concept
>        for all other dialogs/cards.
> 

Delay is always a workaround. The most proper solution would be to remove the root cause of the problem - remove the need of page reload which violates the single-page app pattern. However this requires more investment into our intl framework to implement "hot-reload" feature. This was already discussed in [previous review](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359)

Commit block approach is not possible because:
1. endpoints are different `/sshpublickeys` and `/options`
2. each property is different entity and is validated separetely

>     5. Regarding landing page - user settings local integration:
>        This changes the current behavior for non first logins. Consider adding a reset/CB button for
>        user settings locale so that at least user can change it's locale based on landing page for next logins.

Do you mean the following flow?
1. reset locale to remove existing locale "B" saved on the server
2. logout from VM Portal
3. go to landing page
4. chose locale "A"
5. login to VM Portal (locale "A" is saved on the server)

Why not just chose the  locale "A" via Account Settings?

Regarding the priority of locales passed from landing page:  there are some problems to solve if we want to keep current behaviour for non-first login:
First, value visible on Account Setting screens should be in sync with the value stored on the server (see [this scenario](https://github.com/oVirt/ovirt-web-ui/pull/1372#issuecomment-788054909)). If value from URL overwrites that value on the client then we should also save it on the server. Basically this means extending auto-export feature to work each time locale in the URL is provided. This was discussed before - let me quote it here:
>>  Of course we could persist the chosen language but this creates some problems:
>>  1. error handling in case it failed
>
> if it failed then en_US is chosen
>>  2. unwanted language change (i.e. clicking on the link from some email)
>
> clicking on the link will change the default the same as changing via landing page.
>>  3. double reload (another place where "hot reload" is required)
>
> not sure it will be supported by us.

Second, this should work both ways - change via Account Settings should influence the locale preselected on the landing page. Otherwise user will constantly reset/overwrite his custom settings with the defaults from landing page.
If user is already logged in then we could retrieve the user profile property `vmPortal.locale` and use it in [LocaleFilter](https://github.com/oVirt/ovirt-engine/blob/12a5b774087a730ca33340b19f0adff0f9dbd2f5/backend/manager/modules/utils/src/main/java/org/ovirt/engine/core/utils/servlet/LocaleFilter.java#L95).
However if user is not logged in then the only way is to save it in the cookie. Unfortunately the `locale` cookie is marked as `httpOnly` and therefore not accessible to the JS. We would need to change it or use REST API.
Of course cookie won't work for clean/different browser - user would then need to be careful and re-select the right locales on the landing page.

sgratch:
@rszwajko 

2. 
> local storage usage issue:
> The base scenarios are:....

scenarios 1-3  look ok to me. 

>  use case 4: no vmPortal.locale on the server but a value in local storage exists (i.e. left from different login)
>     at first value from local storage is used (this source has the highest priority)
>     similar as in case 3) user can change it via Account Settings

This seems wrong to me since from my perspective if few users are using the same browser, each one of them should set his own locale for first login based on landing page or defaults and not "inherit" other users locale as their default. 
The local storage locale is hidden and doesn't mean anything since it might be an old/non relevant value.  

========================================================
3.
> Yes, this is case 4) described above. The whole advantage of using local storage is that it works across different sessions. However we do not store user ID there so it works for all accounts. As long as the same person logs in using different accounts then this is beneficial because each account inherits the same locale (during first log in). If that heuristic is incorrect then user needs to manually change the locale via Account Settings. Locale from landing page (the one passed via URL/cookie) has lower priority then local storage so it's ignored.
> 
> Note however that if we would implement hot-reloading of locales then we can drop local storage - switching locales will be then very cheap and we can wait until the data is fetched from the server.

Check my answer for 2. 

==================================================================

4.

> Delay is always a workaround. The most proper solution would be to remove the root cause of the problem - remove the need of page reload which violates the single-page app pattern. However this requires more investment into our intl framework to implement "hot-reload" feature. This was already discussed in previous review
> 
> Commit block approach is not possible because:
> 
>     endpoints are different /sshpublickeys and /options
>     each property is different entity and is validated separately

By "blocking approach" I mean literally and not actually since we can update separately and still display one message only after verifying all settings. If succeeded that we can go with the update, otherwise it failed for all.
Again, think of the scenario when there are 3-4 settings options, all failed and only the locale update was succeeded - the user should quickly read the message with all details and then figure out what was changed and what not. I can't see the benefit of this.
We can separate it to different tabs while locale will be on a dedicated one and updated separately.  

Bottom line, we can leave it as is when only SSH, locale and refresh interval can be updated but once there will be more then it's not acceptable to display user messages for few secs only :( The user might not even remember what was the original failed update value...

==================================================================
5.

> Regarding landing page - user settings local integration:

> Do you mean the following flow?
> 
>     reset locale to remove existing locale "B" saved on the server
>     logout from VM Portal
>     go to landing page
>     chose locale "A"
>     login to VM Portal (locale "A" is saved on the server)
> 
> Why not just chose the locale "A" via Account Settings?

The user doesn't need to logout in order to fo back to landing page and we will anyway need to implement "reset settings" :) but specifically here I meant something like enable/disable persisted locale, as we have for grids on webadmin. 
I just think to enable users working as today and avoid saving locale on both server and local storage. So if they used to work as today and want to avoid saving their locale, we can let them do that since we know that the integration is not completely supported. If they want to enable that then it will be as logged in for the first time. 
It is just a suggestion.

rszwajko:
2.
> This seems wrong to me since from my perspective if few users are using the same browser, each one of them should set his own locale for first login based on landing page or defaults and not "inherit" other users locale as their default.
> The local storage locale is hidden and doesn't mean anything since it might be an old/non relevant value.

I think it works the same as `locale` cookie set by landing page or any other local browser state: saved passwords, search history, etc

4.
> By "blocking approach" I mean literally and not actually since we can update separately and still display one message only after verifying all settings. If succeeded that we can go with the update, otherwise it failed for all.

If I understand correctly you would like to perform additional client side validation before sending data to the server.
This is of course possible but public SSH key is a complex data type and I'd rather avoid parsing it in the UI.

> Again, think of the scenario when there are 3-4 settings options, all failed and only the locale update was succeeded - the user should quickly read the message with all details and then figure out what was changed and what not. 
>

It already works this way - all error messages are displayed in one notification:
1. if there was only one property involved then it's just the message
2. if there were more properties and only some failed (partial success) then those failed are listed in the message

> Bottom line, we can leave it as is when only SSH, locale and refresh interval can be updated but once there will be more then it's not acceptable to display user messages for few secs only :( The user might not even remember what was the original failed update value...
> 

Every notification is displayed for few seconds and then dismissed. If the delay-hack works correctly then user experience should be good. Consider following example:
1. user updates SSH key and language
2. SSH key update failed but language was changed successfully
3. error message appears via component specific notifications
4. shortly after the notification a native browser window appears asking if user wants to leave the page and lose the data
5. at this point user may read the error message and make his decision:
    1.  cancel reload and continue editing i.e. correct SSH key and submit again OR
    2. confirm reload in which case the invalid key will be lost 


 5
> specifically here I meant something like enable/disable persisted locale, as we have for grids on webadmin.
> I just think to enable users working as today and avoid saving locale on both server and local storage. So if they used to work
> as today and want to avoid saving their locale, we can let them do that since we know that the integration is not completely 
> supported. If they want to enable that then it will be as logged in for the first time.
> It is just a suggestion.

Opt-out sounds good to me but this is a good candidate for a follow-up PR.

sgratch:
@rszwajko 
As a reply to https://github.com/oVirt/ovirt-web-ui/pull/1372#issuecomment-791576420:
**issues 2+3:**

> I think it works the same as locale cookie set by landing page or any other local browser state: saved passwords, search history, etc

I agree but here we added a new use case which seems problematic:
A new user tries to login  for the 1st time and set the user local on landing page to X and then the default set to him is Y based on local storage which is not necessarily used by the same user. This is annoying since we don't want to force users to use user settings if they don't have to. And it's different from today's behaviour.
I can't see the disadvantages of deciding that if no vmPortal.locale set on the server then we don't care if locale exists on local storage or not:
-at first locale is discovered based on other sources then local storage - landing page or defaults
-then user can change the locale via Account Settings

**issue 4:**

> If I understand correctly you would like to perform additional client side validation before sending data to the server.
> This is of course possible but public SSH key is a complex data type and I'd rather avoid parsing it in the UI.

yes, I agree.

> 
> It already works this way - all error messages are displayed in one notification:
> 
>     if there was only one property involved then it's just the message
>     if there were more properties and only some failed (partial success) then those failed are listed in the message

Let's take an example of 4 properties updated together by the user via account settings dialog:
- 2 properties failed to be updated.
- 2 succeeded, among them the locale property. 

The actual behaviour:
1. The user will see a notification popup message with a list of failed properties which he will need to read quickly for understanding what's the problem before it will disappear after locale reload. No event with the errors details is added to the events drawer for history tracing.
2. The account settings dialog after reload is not somehow marked/highlighted with failed to update properties, so the user needs to remember what was updated and what was not yet.

This is acceptable and easy to manage when there are only 3 properties on the dialog, but once there will be more then user experience seems bad to me.

One option for improving will be to differ the locale settings from all others by putting that on a dedicated separate tab. So that locale can't be updated with other properties. 
Another option can be to add a web-ui event with all failure details to the notification drawer.
This fix can be postponed for a separate PR, since now we have only 3 properties to set.

> Every notification is displayed for few seconds and then dismissed. If the delay-hack works correctly then user experience should be good. Consider following example:
> 
>     user updates SSH key and language
>     SSH key update failed but language was changed successfully
>     error message appears via component specific notifications
>     shortly after the notification a native browser window appears asking if user wants to leave the page and lose the data
>     at this point user may read the error message and make his decision:
>         cancel reload and continue editing i.e. correct SSH key and submit again OR
>         confirm reload in which case the invalid key will be lost

If user tries to cancel reload and update the ssh value then the error sometimes disappear:
image
 
And once the dialog is long then blocking the whole dialog for reading the error is not so convenient, but it can be a solution for now as long as it works ok.

**Issue 5:**

> Opt-out sounds good to me but this is a good candidate for a follow-up PR

I agree.


**new issue 6:**
Please note that the following exception reproduced to me on Chrome:
image

sjd78 (approved):
Some ordering of locale logic and user data fetches at a different point during login time could help, but it is not a blocker.

The changes look ok.

sgratch:
After reviewing and running a basic verification then this LGTM, except all mentioned issues found on code review that can wait for next release since they are not blockers. Therefore we can release 4.4.5 with current state and consider fixing those for the next release.

The current issues found on code review that we should consider handling on a separate follow-up PRs are:
- enhancements suggested by Scott: Some ordering of locale logic and user data fetches at a different point during login time could help, but it is not a blocker.
- error handling (few issues raised, check "issue 4" of previous comment https://github.com/oVirt/ovirt-web-ui/pull/1372#issuecomment-793934804)
- Since "landing page" locales and "user settings" for web-ui locales integration is partially supported, consider enable/disable persisted locale as mentioned on "issue 5" of previous comment https://github.com/oVirt/ovirt-web-ui/pull/1372#issuecomment-793934804

isaranova:
LGTM

PR #1374, opened 2021-03-07
Use translatable messages instead hard coded messages

bamsalem:
This fix should solve the issue described in [1801433](https://bugzilla.redhat.com/1801433).

The cause of this issue is that some of the messages were hard coded without using `messages.js` file, as a result some of the messages/words weren't translate.

In this PR I extracted the hard coded strings to the `messages.js` file so they can be translated.

I tried to find all the hard coded strings, if you find something I missed I'll add it.

Fixes: https://bugzilla.redhat.com/1801433 .


sgratch (changes requested):
@bamsalem 
Changes look good.
Missing translations still found on dashboard:
1. The Notification drawer title. I see it is translated so it's something with the code itself:
image
2. The NavigationConfirmationModal Yes/No button labels:
image

bamsalem:
@sgratch
Thanks, I fixed that :)
image
image

sgratch (approved):
LGTM

PR #1375, opened 2021-03-08
Using translated messages in "About" dialog

bamsalem:
In this PR I changed the static json file, instead using the json as messages holder I used it as keys holder.
Now we can use the same mechanism of config json file that point to the right branding but instead of static strings we can show translated messages.

the changes we should do on rhev json file are:

```
{
  "BRAND_NAME": "aboutDialogRhvBRAND_NAME",
  "LEGAL_INFO_LINK_TEXT": "aboutDialogRhvLEGAL_INFO_LINK_TEXT",
  "LEGAL_INFO_LINK_URL": <LEGAL_INFO_LINK_URL>,// if needed
  "DOCUMENTATION_LINK": <DOCUMENTATION_URL>,
  "ISSUES_TRACKER_URL": <ISSUES_URL>,
  "aboutDialogRhvISSUES_TRACKER_TEXT": "aboutDialogOvirtISSUES_TRACKER_TEXT"
}
```

Fixes: https://bugzilla.redhat.com/1542428


sjd78:
@bamsalem , @sgratch - Does anything actually need to be translatable that isn't already?

Running my dev engine with the dummy locale + RHV branding:
```
> ENGINE_URL=https://chopper.local.dickerson-web.com/ovirt-engine \
    BROWSER=none \
    KEEP_ALIVE=20 \
    BRANDING=~/projects/rhevm-branding-rhev/web-ui-branding \
    yarn start
```

Gives this About dialog:
VM Portal - Google Chrome_022

The following are already translatable:
  - app name ... `msg.vmPortal()`
  - version string ... `msg.aboutDialogVersion({ version })`
  - api version string ... `msg.aboutDialogApiVersion({ brandName, version })`
  - the please report issue string ... `msg.aboutDialogReportIssuesLink({ link })`

What cannot be translated are the app's fixed string, and I don't see a need to have them be translatable:
  - product name / brand name (we don't translate product names)
  - legal info (we don't translate this anymore)
  - the product name of the issue tracker (we don't translate product names)

In short, I don't think this PR is needed unless product names need to be translated (which I do not think is necessary).

Looking at Zanata, there are missing translations for a few keys on the about dialog.  Here is what it currently looks like for Spanish:
ovirt-web-ui (1 6) to Spanish - Zanata Web Translation — Mozilla Firefox_023

The BZ will be solved once the next rounds of translations are done and pulled into source.

sjd78 (commented):
I don't think these changes are necessary to solve the BZ.

sgratch:
@bamsalem @sjd78 
> I don't think these changes are necessary to solve the BZ.
I agree. I verified that on both u/s and d/s and they both look ok.
Therefore, I'm closing this PR and re-target the bug to 4.4.6 so that it can be verified after next translation cycle.

PR #1376, opened 2021-03-11
A temp fix for reading new format of config val MaxNumOfVmCpus

sgratch:
Since "MaxNumOfVmCpus" config value was changed to a Map format on backend in
patch https://gerrit.ovirt.org/113576, then we need to fix that on web-ui as well by parsing the config val for supporting that per version.

As a temp & quick solution, this value in just hard-coded in case of engine >=4.4.5 in this PR
This should be replaced by the actual parsing in a followed patch.


hstastna:
This PR does (temporarily) fix the issues in Create VM Wizard regarding the number of Virtual CPUs. And now it is possible to create a new VM. 👍🏽 ✨

PR #1382, opened 2021-03-16
Show VM's IP address if the vm is up and guest agent is running on the vm

bamsalem:
According to this [BZ comment](https://bugzilla.redhat.com/show_bug.cgi?id=1441741#c4) the guest_info data has been removed from the nics data and is available in `reporteddevices` entity.

So I added the `reporteddevices` to the `VM_FETCH_ADDITIONAL_DEEP` array so the data will be fetched and showed in the vmDetails page, due to the REST response changes I changed the `Nic.toInternal` function as well to map the data correctly.

Before:
image

 After:
image

Fixes: [BZ#1924610](https://bugzilla.redhat.com/1924610)

### DIFF
--- a/src/components/VmDetails/cards/NicsCard/NicListItem.js
+++ b/src/components/VmDetails/cards/NicsCard/NicListItem.js
@@ -20,7 +20,7 @@ import EllipsisValue from '_/components/EllipsisValue'
  * If _edit_ then render the appropriate action buttons linked to provided
  * handler functions.
  */
-const NicListItem = ({ idPrefix, nic, vmStatus, vnicProfileList, isEditing, onEdit, onDelete }) => {
+const NicListItem = ({ idPrefix, nic, vmStatus, vnicProfileList, isEditing, onEdit, onDelete, showNicIPs }) => {
   const { msg } = useContext(MsgContext)
   const canEdit = !!onEdit
   const canDelete = !!onDelete
@@ -45,7 +45,7 @@ const NicListItem = ({ idPrefix, nic, vmStatus, vnicProfileList, isEditing, onEd
       <Grid>
         <Row>
           <Col cols={4} wrapExpand className={style['ip4-container']} id={`${idPrefix}-ipv4`}>
-            { nic.ipv4.length > 0 &&
+            { showNicIPs && nic.ipv4.length > 0 &&
               nic.ipv4.map((ip4, index) =>
                 <EllipsisValue
                   tooltip={ip4}
@@ -57,7 +57,7 @@ const NicListItem = ({ idPrefix, nic, vmStatus, vnicProfileList, isEditing, onEd
             }
           </Col>
           <Col cols={8} wrapExpand className={style['ip6-container']} id={`${idPrefix}-ipv6`}>
-            { nic.ipv6.length > 0 &&
+            { showNicIPs && nic.ipv6.length > 0 &&
               nic.ipv6.map((ip6, index) =>
                 <EllipsisValue
                   tooltip={ip6}
@@ -143,6 +143,7 @@ NicListItem.propTypes = {
   vmStatus: PropTypes.string.isRequired,
   vnicProfileList: PropTypes.object.isRequired,
   isEditing: PropTypes.bool.isRequired,
+  showNicIPs: PropTypes.bool.isRequired,
 
   onEdit: PropTypes.func,
   onDelete: PropTypes.func,
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -715,10 +715,11 @@ const Cluster = {
 //
 const Nic = {
   toInternal ({ nic }: { nic: ApiNicType }): NicType {
+    const { mac: { address: nicMacAddress = '' } = {} } = nic
     const ips =
       nic.reported_devices && nic.reported_devices.reported_device
         ? nic.reported_devices.reported_device
-          .filter(device => !!device.ips && !!device.ips.ip)
+          .filter(device => !!device.ips && !!device.ips.ip && device.mac && device.mac.address === nicMacAddress)
           .map(device => device.ips.ip)
           .reduce((ips, ipArray) => [...ipArray, ...ips], [])
         : []
--- a/src/ovirtapi/types.js
+++ b/src/ovirtapi/types.js
@@ -114,6 +114,26 @@ export type ApiClusterType = Object
 export type ClusterType = Object
 
 export type ApiNicType = Object
+
+type IpType = {
+  address: string,
+  version: 'v4' | 'v6'
+}
+
+export type ReportedDevicesType = {
+  description: string,
+  href: string,
+  id: string,
+  ips?: { ip: Array<IpType> },
+  mac: { address: string },
+  name: string,
+  type: string,
+  vm: {
+    href: string,
+    id: string
+  }
+}
+
 export type NicType = {
   id: string,
   name: string,
--- a/src/sagas/index.js
+++ b/src/sagas/index.js
@@ -108,7 +108,7 @@ const VM_FETCH_ADDITIONAL_DEEP = [
   'cdroms',
   'disk_attachments.disk',
   'graphics_consoles',
-  'nics',
+  'nics.reporteddevices',
   'permissions',
   'sessions',
   'snapshots',
--- a/src/components/EllipsisValue/index.js
+++ b/src/components/EllipsisValue/index.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import { xor } from '_/propTypeExtras'
 import style from './style.css'
 import classnames from 'classnames'
+import Tooltip from '_/components/tooltips/Tooltip'
 
 class EllipsisValue extends React.Component {
   constructor (props) {
@@ -46,14 +47,18 @@ class EllipsisValue extends React.Component {
     const { className, id, children, tooltip } = this.props
 
     if (children) {
-      return <span
+      const ellipsisValue = (<span
         className={classnames(style['field-value'], className)}
         id={id}
-        title={this.state.isOverflow ? tooltip : ''}
         ref={this.ref}
       >
         {children}
-      </span>
+      </span>)
+
+      return this.state.isOverflow ? (<Tooltip tooltip={tooltip} id={`ellipsis_tooltip${id}`} >
+        {ellipsisValue}
+      </Tooltip>)
+        : (ellipsisValue)
     }
     return null
   }
--- a/src/components/VmDetails/cards/DetailsCard/index.js
+++ b/src/components/VmDetails/cards/DetailsCard/index.js
@@ -624,6 +624,7 @@ class DetailsCard extends React.Component {
 
     const idPrefix = 'vmdetail-details'
 
+    const showIPs = vm.get('status') === 'up'
     const isPoolVm = vm.getIn(['pool', 'id']) !== undefined
     const canEditDetails = vm.get('canUserEditVm') && !isPoolVm
     let pool = null
@@ -637,8 +638,13 @@ class DetailsCard extends React.Component {
     const hostName = hosts && hosts.getIn([vm.get('hostId'), 'name'])
 
     // IP Addresses
-    const ip4Addresses = !vm.has('nics') ? [] : vm.get('nics').reduce((ipSet, nic) => [...ipSet, ...nic.get('ipv4')], [])
-    const ip6Addresses = !vm.has('nics') ? [] : vm.get('nics').reduce((ipSet, nic) => [...ipSet, ...nic.get('ipv6')], [])
+    const ip4Addresses = [...(new Set(!vm.has('nics') || !showIPs
+      ? []
+      : vm.get('nics').reduce((ipSet, nic) => [...ipSet, ...nic.get('ipv4')], [])))]
+    const ip6Addresses = [...(new Set(
+      !vm.has('nics') || !showIPs
+        ? []
+        : vm.get('nics').reduce((ipSet, nic) => [...ipSet, ...nic.get('ipv6')], [])))]
 
     // FQDN
     const fqdn = vm.get('fqdn')
--- a/src/components/VmDetails/cards/NicsCard/index.js
+++ b/src/components/VmDetails/cards/NicsCard/index.js
@@ -173,6 +173,7 @@ class NicsCard extends React.Component {
               <Row key={nic.id} id={`${idPrefix}-${nic.name}`}>
                 <Col style={{ display: 'block' }}>
                   <NicListItem
+                    showNicIPs={showNicIPs}
                     idPrefix={`${idPrefix}-${nic.name}`}
                     nic={nic}
                     vmStatus={vmStatus}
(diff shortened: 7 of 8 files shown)


src/ovirtapi/transform.js
```diff
@@ -202,8 +204,15 @@ const VM = {
...
+      const reportedDevicesArray =
+        (vm.reported_devices && vm.reported_devices.reported_device)
+          ? vm.reported_devices.reported_device.filter(device => device.type === 'network')
+          : []
+
+      const reportedDeviceMap = arrayToMap(reportedDevicesArray, device => device.mac.address)
+
       parsedVm.nics = vm.nics.nic.map(
-        nic => Nic.toInternal({ nic })
+        nic => Nic.toInternal({ nic, reportedDevice: reportedDeviceMap[nic.mac.address] })
```
rszwajko, line 215: We assume here that `nic.mac` can be matched with `reportedDevice.mac` however this seems not to be true all the time (tested on brq by adding `?follow=reporteddevices` to `OvirtApi.getVmNic()`) i.e. for query [1] the device returned by backend has has the same MAC (not matching any NIC) for all reported cards. 
Note also that `mac` element is optional (see [ReportedDeviceMapper](https://github.com/oVirt/ovirt-engine/blob/07969f6cd0878d39a42d4feab0871e86a22e501b/backend/manager/modules/restapi/types/src/main/java/org/ovirt/engine/api/restapi/types/ReportedDeviceMapper.java#L27)) so we need to filter out invalid entities.

We could avoid those problems by adding `?follow=nics.reporteddevices` - this way the backend will do that for us. 

[1] .../api/vms/8f8bb2b8-34c1-4e5a-b4ec-c5301bfc14ac/nics?follow=reporteddevices

src/ovirtapi/types.js
```diff
@@ -112,6 +112,8 @@ export type ApiClusterType = Object
 export type ClusterType = Object
 
 export type ApiNicType = Object
+export type ReportedDevicesType = Object
```
rszwajko, line 115: please provide the type definition - exact types are preferred (`{| ... |}`)

rszwajko (changes requested):

hstastna (commented):
The changes look good to me and the IP address is displayed as expected in the VM Details and NICs cards. But Radek's comment https://github.com/oVirt/ovirt-web-ui/pull/1382/files#r595343562 worth checking :)

src/ovirtapi/transform.js
bamsalem: OK, in the code it can be optional but from what I know each network interface card should have it's own unique MAC address so I'm no sure if it's "over kill" in the code or there is any possible scenario that the mac will be undefined?

> We could avoid those problems by adding ?follow=nics.reporteddevices - this way the backend will do that for us.

In that case it's return all the ips for each nic and than we gather the wrong data, and it's duplicated in the view:
image

The reason  I used the MAC as unique is because I didn't found any relation between the nic and the reported data except MAC address.
but we are definitely can add the `?follow=reporteddevices` to the fetchVmNics call

bamsalem: In the case you raised in your comment:
> i.e. for query [1] the device returned by backend has has the same MAC (not matching any NIC) for all reported cards.

The filtering by MAC will filter it, and that's what expected, because the vm is down and it should not have an ip.

rszwajko: > we gather the wrong data, and it's duplicated in the view
>

Backend returns duplicated data?

> relation between the nic and the reported data except MAC address
> 
The  relation between devices and nics is based on device UUID - this is basically what  is returned from `.../nics/123?follow=reporteddevices`.

bamsalem: > Backend returns duplicated data?

Yes,
first nic:
image
second nic:
image

You can take a look on my screenshots, or try it on this vm on brq-dev: 95a41852-2898-4b51-8131-6e4315b65668
It's filtered in the backend

rszwajko: >> Backend returns duplicated data?
>
> Yes,
>

That's strange.
1. Might be a bug. Then we should report it and let the backend fix it.
2. Might be stale data i.e. old card still visible (but the new one is already in the list) or just delay/synchronization issues when reading the config from underlying hardware.

In both cases I would remove the MAC matching algorithm and treat the backend as the ultimate source of truth. If duplicated addresses look ugly then we can remove duplicates in the React component.(view layer). But let's treat thme just as strings (duplicated strings in a list of strings) - no custom logic.
The proper solution would be probably to display both nick name and nick address (then duplicated addresses would make sense).

src/sagas/index.js
```diff
@@ -113,6 +113,7 @@ const VM_FETCH_ADDITIONAL_DEEP = [
   'sessions',
   'snapshots',
   'statistics',
+  'reporteddevices',
```
rszwajko, line 116: please use `nics.reporteddevices`

src/ovirtapi/transform.js
```diff
@@ -707,13 +715,10 @@ const Cluster = {
 //
 //
 const Nic = {
-  toInternal ({ nic }: { nic: ApiNicType }): NicType {
+  toInternal ({ nic, reportedDevice = {} }: { nic: ApiNicType, reportedDevice?: ReportedDevicesType }): NicType {
```
rszwajko, line 718: Reported devices are part of nic structure and should be parsed inside `toInternal`.

src/ovirtapi/transform.js
```diff
@@ -202,8 +203,15 @@ const VM = {
     }
 
     if (vm.nics && vm.nics.nic) {
+      const reportedDevicesArray =
+        (vm.reported_devices && vm.reported_devices.reported_device)
```
rszwajko, line 207: With `follow?nics.reporteddevices` reported devices are part of the nic structure. There is no need to handle them externally.

rszwajko (dismissed):

src/ovirtapi/transform.js
```diff
@@ -708,10 +708,11 @@ const Cluster = {
 //
 const Nic = {
   toInternal ({ nic }: { nic: ApiNicType }): NicType {
+    const { mac: { address: nicMacAddress = '' } = {} } = nic
     const ips =
       nic.reported_devices && nic.reported_devices.reported_device
         ? nic.reported_devices.reported_device
-          .filter(device => !!device.ips && !!device.ips.ip)
+          .filter(device => !!device.ips && !!device.ips.ip && device.mac && device.mac.address === nicMacAddress)
```
rszwajko, line 715: @sjd78  @sgratch 
Please comment on Mac filtering. [My opinion](https://github.com/oVirt/ovirt-web-ui/pull/1382#discussion_r607664825) is that we should try to "correct" backend - just present what we receive.

rszwajko (commented):
Code looks OK in this version - except MAC filtering.

src/sagas/index.js
```diff
@@ -113,6 +113,7 @@ const VM_FETCH_ADDITIONAL_DEEP = [
   'sessions',
   'snapshots',
   'statistics',
+  'nics.reporteddevices',
```
sjd78, line 116: No need to specify the nics twice.

A single `'nics.reported_devices'` is preferred.

src/ovirtapi/transform.js
sjd78: I don't see any need to evaluate the reported_device's mac address unless multiple devices are returned. I tested a few running VMs and only ever had 1 returned.  The backend probably matches things up for us.

I wonder how a multiport passthrough nic would be reported.  That would be quite a rare occurrence.  Probably 1 nic for each port so it wouldn't be a problem anyway.

sjd78 (dismissed):
Only 1 minor issue, everything else looks good.

src/ovirtapi/transform.js
rszwajko: @sjd78
Then we should report a bug and in the meantime just remove duplicates in the view (which does make sense since we show all the addresses together).

sjd78: > @sjd78
> Then we should report a bug and in the meantime just remove duplicates in the view (which does make sense since we show all the addresses together).

So filter it in the view instead of in the transform?  I'd rather have the transform just make the data nice and neat for the view.  A BZ on the rest side can be opened, but probably wouldn't be handled in time for us to use so we'd still need to filter anyway.

rszwajko: @sjd78 @bamsalem 
>  I'd rather have the transform just make the data nice and neat for the view. 

As far as I understand we use the same data to display NICs details. IMHO NICs may have the same IPs and MACs in certain scenarios by design (migration, failover) or because of the delay(hardware device vs virtual device) or simply due to user error (incorrect configuration). The widget displaying the IP address summary should be prepared to handle that.

> A BZ on the rest side can be opened, but probably wouldn't be handled in time for us to use so we'd still need to filter anyway.

+1

hstastna (approved):
LGTM 👍🏽

src/ovirtapi/transform.js
sjd78: > @sjd78 @bamsalem
> 
> > I'd rather have the transform just make the data nice and neat for the view.
> 
> As far as I understand we use the same data to display NICs details. IMHO NICs may have the same IPs and MACs in certain scenarios by design (migration, failover) or because of the delay(hardware device vs virtual device) or simply due to user error (incorrect configuration). The widget displaying the IP address summary should be prepared to handle that.
> 

:facepalm: 

Ok so it will all depend on how the nic given to the VM (whose mac address is fixed in the VM definition) is tied to the guest agent's MAC and IP address reporting.  To make it even more crazy, just think of a VM setup where (1) the nic's mac address is cloned so the reported mac won't match the nic's mac, or (2) if the system is setup to have multiple macvlan adpaters on top of the VM's nic, or (3) virtual nics bleeding through via the reported devices but don't really have a parent device (e.g. a tap/tun device).  If there is a macvlan setup, each new macvlan derived nic has it's own mac address.  There can basically be as many as you want, but they're all "children" of the base nic....

SO.....

Makes sense to do a few things:
 1. Don't filter out the "extra" devices `reported_devices` in the transform
 2. Just show all of the info on the Nics Card for each Nic (so no change)
 3. On the Details Card, only show the unique set of IP4/IP6 addresses (so add IPs to a set instead of a list)
 4. Open a BZ so the infra/guest agent guys can make sure the reporting is being done properly (until that is "fixed" the Nics card will show extra info per nic)

@rszwajko - I think number 3 is what you were getting at originally.

> > A BZ on the rest side can be opened, but probably wouldn't be handled in time for us to use so we'd still need to filter anyway.
> 
> +1

sjd78 (changes requested):
See comment on the mac filtering bit

src/ovirtapi/transform.js
sgratch: I think current behaviour with mac filtering is ok since vmGuestAgentInterface.getMacAddress should fit nic.getMacAddress based on this backend code:
https://github.com/oVirt/ovirt-engine/blob/99bfdd6d1f3bf27c930260aa08921c6da99142e8/backend/manager/modules/restapi/jaxrs/src/main/java/org/ovirt/engine/api/restapi/resource/BackendNicHelper.java#L41

In addition, the resulted list of ips fit the list shown on webadmin.
There is a bug on rest so we can open BZ for the rest query but the filtering seems ok anyway to me.

sgratch (changes requested):
Code LGTM except two issues:
- The VM IP address field should be set only when VM is running (see relevant IP Address tooltip:
"This field is only available when the VM is running").
So when VM is not running, the N/A should be displayed.

- it seems that the filed width is not enough for default resolution, The Ips are cut off. Can we widen it a bit?

hstastna:
>     * it seems that the filed width is not enough for default resolution, The Ips are cut off. Can we widen it a bit?

I experienced that, too, but I think that displaying the IPs also in the NICs card should be enough.

sgratch:
> > it seems that the filed width is not enough for default resolution, The Ips are cut off. Can we widen it a bit?
> 
> I experienced that, too, but I think that displaying the IPs also in the NICs card should be enough.

All IPv6 are cut off when displayed with default browser resolution and there is no way to view the whole ip. If overlaying the ip would show the whole string or if there was a way to manually widen this column then fine. 
But currently it seems wrong since there is no point on showing a value that is always cut off.

Anyway, this can be handled separately on a different PR.

bamsalem:
> * it seems that the filed width is not enough for default resolution, The Ips are cut off. Can we widen it a bit?

Probably it can be widen but it might create inconsistency and other UI issues...

there is a mechanism for fields overflow:
image
This mechanism used the `title` html attribute, I fixed it and now it's using the `Tooltip` component.
@sgratch @hstastna ,
WDYT on this solution?

sgratch:
> > * it seems that the filed width is not enough for default resolution, The Ips are cut off. Can we widen it a bit?
> 
> Probably it can be widen but it might create inconsistency and other UI issues...
> 
> there is a mechanism for fields overflow:
> image
> This mechanism used the `title` html attribute, I fixed it and now it's using the `Tooltip` component.
> @sgratch @hstastna ,
> WDYT on this solution?

Looks good to me.

src/ovirtapi/transform.js
sjd78: As long as all we want to do it show any IP addresses reported by the guest agent explicitly with the Nic's mac address, sure.  Doing the `Set` work I suggested in `DetailsCard` will remove any duplicate IPs found.

AFAICT, the REST call should then make sure only the proper guest agent reported devices are attached to the actual nic.  The `BackendNicHelper` looks like it should do that, but the real calls don't seem to be working that way.

sjd78 (changes requested):
Add Set filtering for the IP address displays

src/components/VmDetails/cards/NicsCard/NicListItem.js
```diff
@@ -44,7 +44,7 @@ const NicListItem = ({ idPrefix, nic, vmStatus, vnicProfileList, isEditing, onEd
       <Grid>
         <Row>
           <Col cols={4} wrapExpand className={style['ip4-container']} id={`${idPrefix}-ipv4`}>
-            { nic.ipv4.length > 0 &&
+            { vmStatus === 'up' && nic.ipv4.length > 0 &&
```
sgratch, line 47: This is already handled here: https://github.com/oVirt/ovirt-web-ui/blob/4e018d11bb74d118315804b36949c501e04b7ef5/src/components/VmDetails/cards/NicsCard/index.js#L101

src/components/VmDetails/cards/NicsCard/NicListItem.js
```diff
@@ -56,7 +56,7 @@ const NicListItem = ({ idPrefix, nic, vmStatus, vnicProfileList, isEditing, onEd
             }
           </Col>
           <Col cols={8} wrapExpand className={style['ip6-container']} id={`${idPrefix}-ipv6`}>
-            { nic.ipv6.length > 0 &&
+            { vmStatus === 'up' && nic.ipv6.length > 0 &&
```
sgratch, line 59: same here

sgratch (commented):
@bamsalem Please check inline comment.
Other than that it looks good.

sgratch (approved):
LGTM

PR #1383, opened 2021-03-17
Read user name, password and domain from env variable

rszwajko:
When launching in dev mode, allow passing user, password and domain
as shell variables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ build
 npm-debug.log
 .idea
 .vscode/
+.env*
 
 # automake.sh / automake / autoconfig generated files
 Makefile
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -59,6 +59,15 @@ to accomplish this. To start the server:
 When asked, provide a valid user name (in the form of `user@profile`) and password so
 the application can start in the context of a logged in user.
 
+#### .env.* files
+
+You can group environment variables using `.env.envname` files. For example, for environment `foo`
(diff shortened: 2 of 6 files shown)


ovirt-infra:

rszwajko:
Added grouping env variables using dotenv library (in particular using [this advice](https://github.com/motdotla/dotenv#what-happens-to-environment-variables-that-were-already-set))

sjd78:
ci add to whitelist

sjd78 (changes requested):
LGTM (oops, I didn't hit approve)

sjd78 (approved):
LGTM

sjd78 (changes requested):
Ok - so `dotenv` should be added as a "devDependency" with its current version.  Right now it is only working because `dotenv` is a transitive dependency (at 5.0.1).  We'll probably need to update ovirt-engine-nodejs-modules after that.

bamsalem (approved):
Except @sjd78 review, LGTM
I like this change :)
Just an idea, we don't must to do it, but it will be nice if we can add `.env.example` file.

rszwajko:
@bamsalem 
> Just an idea, we don't must to do it, but it will be nice if we can add `.env.example` file.
>
IMHO documentation is enough. `.env` files are explicitly excluded from git to prevent accidental commit. If we provide an example file then someone might modify it and commit by mistake.

sjd78 (approved):
LGTM

PR #1384, opened 2021-03-17
Use semver to verify API level

rszwajko:
REST API sends version numbers as strings. However the required version
numbers are defined as numbers. As result, the strict comparison(===)
always fails.

Example: 
```
> compareVersion({major: '4',minor:'3'}, {major: 4, minor:4})
> true
```

### DIFF
--- a/ovirt-web-ui.spec.in
+++ b/ovirt-web-ui.spec.in
@@ -17,7 +17,7 @@ Source0:        https://github.com/oVirt/ovirt-web-ui/archive/%{source_basename}
 BuildArch: noarch
 
 # nodejs-modules embeds yarn and requires nodejs
-BuildRequires: ovirt-engine-nodejs-modules >= 2.0.41-1
+BuildRequires: ovirt-engine-nodejs-modules >= 2.0.42-1
 
 %description
 This package provides the VM Portal for %{product}.
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
     "react-test-renderer": "16.8.6",
     "redux": "4.0.0",
     "redux-saga": "0.16.0",
+    "semver": "^7.3.4",
     "victory": "32.3.1"
   },
   "resolutions": {
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -388,7 +388,7 @@ export const messages: { [messageId: string]: MessageType } = {
   hotPlugConfirmTitle: 'Apply Changes',
   hoursShort: 'h',
   htmlPleaseReferToDocumentationForMoreInformation: 'Please refer to <a href="{documentationUrl}" target="_blank">documentation</a> for more information.',
-  htmlUnsupportedOvirtVersionFoundButVersionAtLeastRequired: '<strong>Unsupported {version} {productName} version,</strong> found but version at least {requiredVersion} is required.',
+  htmlUnsupportedOvirtVersionFoundButVersionAtLeastRequired: '<strong>Unsupported {version} {productName} version</strong> found, but at least version {requiredVersion} is required.',
   icon: 'Icon',
   ifVmIsRunningClickToAccessItsGraphicsConsole: 'If the virtual machine is running, click the protocol name to access its Graphical Console.',
   inPreview: 'In Preview',
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -28,6 +28,7 @@ import type {
   ApiUserType, UserType,
   UserOptionType,
   RemoteUserOptionsType,
+  VersionType,
 } from './types'
 
 import { isWindows } from '_/helpers'
@@ -1024,6 +1025,16 @@ const User = {
   toApi: undefined,
 }
 
+const Version = {
+  toInternal ({ major = 0, minor = 0, build = 0 }: any): VersionType {
+    return {
+      major: Number(major),
+      minor: Number(minor),
+      build: Number(build),
+    }
+  },
+}
+
 //
 // Export each transforms individually so they can be consumed individually
 //
@@ -1055,4 +1066,5 @@ export {
   User,
   RemoteUserOptions,
   RemoteUserOption,
+  Version,
 }
--- a/src/ovirtapi/types.js
+++ b/src/ovirtapi/types.js
@@ -249,3 +249,9 @@ export type EventType = {
   description: string,
   severity: string
 }
+
+export type VersionType = {|
+  major: number,
+  minor: number,
+  build: number
+|}
--- a/src/sagas/login.js
+++ b/src/sagas/login.js
@@ -2,7 +2,7 @@ import { all, call, put, takeEvery, select } from 'redux-saga/effects'
 import pick from 'lodash/pick'
 
 import Product from '_/version'
-import Api from '_/ovirtapi'
+import Api, { Transforms } from '_/ovirtapi'
 import AppConfiguration from '_/config'
 import OptionsManager from '_/optionsManager'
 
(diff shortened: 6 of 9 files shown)


ovirt-infra:

sjd78:
Seems reasonable, but what about just switching to `semver` comparison? https://www.npmjs.com/package/semver

Importing just the specific function will let webpack leaf shake out things not used and minimize the new include size

rszwajko:
@sjd78 
To my surprise `full_version` is not always a valid version according to semver. Even when `loose` option is used the following code fails:
```
export function compareVersion ({ full_version: actualFullVersion }, required) {
  console.log(`compareVersion(), actual=${actualFullVersion}, required=${JSON.stringify(required)}`)
  // semver specification requires at least MAJOR.MINOR.PATCH format (additional labels are allowed)
  // sameverGte throws TypeError if invalid string is passed
  const actualValidated = semverValid(actualFullVersion, { loose: true })
  return !!actualValidated && semverGte(actualFullVersion, `${required.major}.${required.minor}.0`, { loose: true })
}
```
Unfortunately a very common version like `4.4.6_master` is invalid so we cannot ignore that.
A possible workaround would be to build the actual version number the same way as we build required version but then the semver lib doesn't bring much value.

mwperina:
If you need, you can fetch version ```product-info``` element from /ovirt-engine/api:

```
<product_info>
<name>Red Hat Virtualization Manager</name>
<version>
<build>1</build>
<full_version>4.4.1.10-0.1.el8ev</full_version>
<major>4</major>
<minor>4</minor>
<revision>0</revision>
</version>
</product_info>
```

and you can create require version manually using just ```major```, ```minor``` and ```revision``` fields

sjd78:
> @sjd78
> To my surprise `full_version` is not always a valid version according to semver. Even when `loose` option is used the following code fails:
> 
> ```
> export function compareVersion ({ full_version: actualFullVersion }, required) {
>   console.log(`compareVersion(), actual=${actualFullVersion}, required=${JSON.stringify(required)}`)
>   // semver specification requires at least MAJOR.MINOR.PATCH format (additional labels are allowed)
>   // sameverGte throws TypeError if invalid string is passed
>   const actualValidated = semverValid(actualFullVersion, { loose: true })
>   return !!actualValidated && semverGte(actualFullVersion, `${required.major}.${required.minor}.0`, { loose: true })
> }
> ```
> 
> Unfortunately a very common version like `4.4.6_master` is invalid so we cannot ignore that.
> A possible workaround would be to build the actual version number the same way as we build required version but then the semver lib doesn't bring much value.

As @mwperina mentioned, we could build the version string clean ourselves since we already pull that full data struct.

But, how would this look (in snippet form and barely tested - the extra import could be paired down to only the stuff needed so the library doesn't take much extra space in the packaged bundle)?

**src/version.js.in**
```js
const Product = {
  ...
  ovirtApiVersionRequired: '>= 4.4.0',
}
```

**src/sagas/login.js** (I guess this is actually unchanged)
```js
  const actual = oVirtMeta['product_info']['version']
  const required = Product.ovirtApiVersionRequired
  const passed = compareVersion(actual, required)
```

**src/utils.js**
```js
import semver from 'semver'

export function compareVersion (actual, required) {
  console.log(`compareVersion(), actual=${actual.full_version}, required=${required}`)
  return semver.satisfies(semver.coerce(actual.full_version), required)
}
```

Then the required string would look like a normal `Requires:` string in the rpm spec file, AND can look like a semver string from package.json dependencies blocks.  If we need to compareVersion in code elsewhere (I took all of those out previously) in future, it's easiser to read "> 4.4.6" than "{ major: 4, minor: 4, build: 6 }"...

sjd78 (dismissed):
See https://github.com/oVirt/ovirt-web-ui/pull/1384#issuecomment-811621360

rszwajko:
@sjd78 
Thanks for pointing to `semver.coerce()` - it does a very good job and is actually able to transform (invalid) `4.4.6_master` into `4.4.6`. However building the strings manually is still safer. 
Please check the newest code. The patch fixes the problem with invalid version comparison(main reason for this PR) and adds semver library which open doors for future improvements (suggested above).

sjd78:
ci test

sjd78:
Looks good.  Running CI again with the nodejs-modules update patch in place.

Questions:
 - Does it make sense to add the ability to do build number number checks at this point?
 - Does it make sense to convert the api require version in file `versions.js.in` to use a semver version instead of the major/minor struct that is there now?

rszwajko:
>   Does it make sense to add the ability to do build number number checks at this point?
>

Definitely yes but we don't have a functionality that would use it (yet). Adding a code that is not used is ...hard to test:) I would add it with first feature that requires it. The obvious candidate is max supported `cluster level` which changes with build number.

>   Does it make sense to convert the api require version in file `versions.js.in` to use a semver version instead of the major/minor struct that is there now?
>

If it's only `versions.js.in`  then OK. We can parse it anyway.
As for replacing version structure in the code - I'm not sure (yet):
1. parsing the full version string in the code to extract fields adds noise and  using full version might not always be possible. If we would have more cases that require individual access then I would leave the structure.
2. ovirt is not semver-compliant which requires extra care.  `coerce` function does a good job but it returns `null` when string is not good enough. The safest way would be to create semver string on our own (as in this PR)

sjd78:
> > Does it make sense to add the ability to do build number number checks at this point?
> 
> Definitely yes but we don't have a functionality that would use it (yet). Adding a code that is not used is ...hard to test:) I would add it with first feature that requires it. The obvious candidate is max supported `cluster level` which changes with build number.
> 

Yeah that one is cluster version level not API level, so that's more "interesting".

The other part about activating different code paths ... some sagas used to all the API endpoints differently depending on which version of the API was available.  Something like that could be done to keep supporting the user ssh key direct vs part of user options.

Either way, I'm good with adding the functionality once it is needed.

> > Does it make sense to convert the api require version in file `versions.js.in` to use a semver version instead of the major/minor struct that is there now?
> 
> If it's only `versions.js.in` then OK. We can parse it anyway.
> As for replacing version structure in the code - I'm not sure (yet):
> 
>   1. parsing the full version string in the code to extract fields adds noise and  using full version might not always be possible. If we would have more cases that require individual access then I would leave the structure.

The way you construct the API version is ok.  It's more about how we specify the version required and which semver call is used to check.  Using a string in `versions.js.in` of `>= 4.4.4` and using a base compare from the built API version to the spec string is what I was referring to.

> 
>   2. ovirt is not semver-compliant which requires extra care.  `coerce` function does a good job but it returns `null` when string is not good enough. The safest way would be to create semver string on our own (as in this PR)

I'm assuming you mean the "full_version" field isn't a proper semver version string.  Constructing the API version string the way you did is not a problem at all.


**In summary**, leaving things as-is sounds good to me since nothing right now needs a build number check.  Adding that if/when it is needed is the way to go.

sjd78 (approved):
LGTM

sgratch (dismissed):
Worth handling https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/OvirtApiCheckFailed.js as well (adding build numbers).

And by the way worth also rephrasing the
`msg. htmlUnsupportedOvirtVersionFoundButVersionAtLeastRequired` to something like:
"Unsupported x.y.z <product> version found, but at least version x1.y1.z1 is required."

rszwajko:
> Worth handling https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/OvirtApiCheckFailed.js as well (adding build numbers).
> 

We are not adding support for `build` number. `versions.js.in` still contains only major/minor part. Current usage of `build` part is implementation detail (`semver` format requires it).
As discussed [above](https://github.com/oVirt/ovirt-web-ui/pull/1384#issuecomment-817901456)  build number will be added with first functionality that will require it. Right now we have no such functionality.

sgratch:
> > Worth handling https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/OvirtApiCheckFailed.js as well (adding build numbers).
> 
> We are not adding support for `build` number. `versions.js.in` still contains only major/minor part. Current usage of `build` part is implementation detail (`semver` format requires it).
> As discussed [above](https://github.com/oVirt/ovirt-web-ui/pull/1384#issuecomment-817901456) build number will be added with first functionality that will require it. Right now we have no such functionality.

I am aware of that, but since we already read the actual build number then I thought it can be nice to display that as part of the error. We are the ones managing the required version build number anyway so as long as it is hard-coded to zero then we are fine. Anyway, it's not that important.

But we still need to somehow rephrase msg.htmlUnsupportedOvirtVersionFoundButVersionAtLeastRequired:
"Unsupported {version} {productName} version**,**found but **version at least** {requiredVersion} is required.' =>
"Unsupported {version} {productName} version found**,** but **at least version** {requiredVersion}  is required."

We can open a separate issue for this rephrasing.

sgratch:
@rszwajko 

> Message has been changed as requested.

please fix the comma as well:

> "Unsupported {version} {productName} version` ,` found but version at least {requiredVersion} is required.' =>
> "Unsupported {version} {productName} version found` , `but at least version {requiredVersion} is required."

sgratch (approved):
LGTM

PR #1386, opened 2021-03-21
Allow users to select/deselect save memory for running vm snapshots

bamsalem:
Fixes: [BZ#1919195](https://bugzilla.redhat.com/1919195).

Before this Change every snapshot created from web-ui for running vm were includes the memory automatically without option to change/deselect the memory saving option (like users can change in webAdmin).

In this Change I added the option to select/deselect memory saving only  for **running VMs**, and now users can decide if they want to save memory or not.

Screenshots for running VMs:

Before:
Running vm:
Running
Shut Down vm:
image

After:
Running vm:
Running
Shut Down vm (no changes):
Shut Down


src/ovirtapi/transform.js
```diff
@@ -486,6 +486,7 @@ const Snapshot = {
   toApi ({ snapshot }: { snapshot: SnapshotType }): ApiSnapshotType {
     return {
       description: snapshot.description,
+      persist_memorystate: snapshot.persistMemoryState,
```
rszwajko, line 489:
```
Cannot return object literal because boolean [1] is incompatible with string [2] in property persist_memorystate.

     src/ovirtapi/transform.js
     486|   toApi ({ snapshot }: { snapshot: SnapshotType }): ApiSnapshotType {
     487|     return {
     488|       description: snapshot.description,
     489|       persist_memorystate: snapshot.persistMemoryState,
     490|     }
     491|   },
     492| }

     src/ovirtapi/types.js
 [2]  50|   persist_memorystate?: string
        :
 [1]  59|   persistMemoryState: boolean,
```

rszwajko (commented):
Looks OK - there is one type issue to investigate (see below)

src/ovirtapi/transform.js
hstastna: This error makes sense to me, especially when I look at https://github.com/oVirt/ovirt-web-ui/blob/master/src/ovirtapi/transform.js#L481

hstastna (dismissed):
LGTM, just fix the error Radek has found :) Thank you, Ben!

hstastna:
One more thing: @bamsalem  is there any change regarding the "Shut Down vm"? I don't think so. Also from the description of the PR, it seems that the changes are related only to the running VM. So the presence of the screenshots for "Shut Down vm" seem to me confusing.

src/ovirtapi/transform.js
bamsalem: Great input :)
Thanks.

bamsalem:
@hstastna 
> One more thing: @bamsalem is there any change regarding the "Shut Down vm"? I don't think so. Also from the description of the PR, it seems that the changes are related only to the running VM. So the presence of the screenshots for "Shut Down vm" seem to me confusing, I don't see any change there.

No there is no change regarding the  "Shut Down vm", maybe I had to mention it explicitly.
I mentioned it in the first comment.

src/ovirtapi/transform.js
```diff
@@ -486,6 +486,7 @@ const Snapshot = {
   toApi ({ snapshot }: { snapshot: SnapshotType }): ApiSnapshotType {
     return {
       description: snapshot.description,
+      persist_memorystate: snapshot.persistMemoryState ? 'true' : 'false',
```
hstastna, line 489: Is there any reason for not using `toString()` here? Can `snapshot.persistMemoryState` be other than boolean?

hstastna (commented):
LGTM, just check some of my notes. Thanks! ;)

rszwajko (approved):

sgratch (changes requested):
LGTM except a suggestion to fix the dialog msg: 
"All attached disks will be included in the snapshot. Content of memory **will** be included only if VM is running." -> 
"All attached disks will be included in the snapshot. Content of memory **can** be included only if VM is running."

src/ovirtapi/transform.js
```diff
@@ -486,6 +486,7 @@ const Snapshot = {
   toApi ({ snapshot }: { snapshot: SnapshotType }): ApiSnapshotType {
     return {
       description: snapshot.description,
+      persist_memorystate: snapshot.persistMemoryState.toString(),
```
sjd78, line 489:
```suggestion
      persist_memorystate: convertBool(snapshot.persistMemoryState),
```

Also, the `persist_memorystate` type in `ApiSnapshotType` should be a boolean not a string as per add POST: http://ovirt.github.io/ovirt-engine-api-model/master/#services/snapshots/methods/add and the Snapshot struct: http://ovirt.github.io/ovirt-engine-api-model/master/#types/snapshot

sjd78 (changes requested):
The message change suggested by @sgratch makes sense since the option is now explicitly selected.

src/ovirtapi/transform.js
bamsalem: @sjd78 according to this [link](http://ovirt.github.io/ovirt-engine-api-model/master/#types/snapshot):
it should be a [boolean](http://ovirt.github.io/ovirt-engine-api-model/master/#types/boolean): so it might be "false" OR "true".
https://github.com/oVirt/ovirt-web-ui/blob/2886959acd3588132d5fc276a9d4df043de19644/src/ovirtapi/types.js#L50
I can change this type to: `persist_memorystate: "true" | "false"`.
But if I understood you correctly it shouldn't be changed more than explicitly mention the possible values, if I didn't can you please explain what did you meant?
Regarding the suggested change it converts it the opposite direction (string to bool).

src/ovirtapi/types.js
```diff
@@ -47,7 +48,7 @@ export type ApiSnapshotType = {
   snapshot_type?: string,
   date?: number,
   snapshot_status?: string,
-  persist_memorystate?: string
+  persist_memorystate: BoolStringType
```
sgratch, line 51: I don't think this is needed. Please check how the NicType.plugged is handled as an example.
BTW, the values for api Boolean type can be either all combinations of upper/lower cases for true/false and also 0/1 as mentioned: http://ovirt.github.io/ovirt-engine-api-model/master/#types/boolean

sgratch (changes requested):

src/ovirtapi/types.js
bamsalem: > Represents the false and true concepts used in mathematical logic.
>
>The valid values are the strings false and true.
>
>Case is ignored by the engine, so for example False and FALSE also valid values. However the server will always return lower case values.

The right values for the api are "true" OR "false" strings, any case combination.

> For backwards compatibility with older versions of the engine, the values 0 and 1 are also accepted. The value 0 has the same meaning than false, and 1 has the same meaning than true. Try to avoid using these values, as support for them may be removed in the future.

`NicType.plugged` is an internal types, in web ui we saving boolean as true/false values not strings, but for the API we should convert this value to string value (`"true"` or `"false"` with any combination of lower/upper case).
I don't know why it works now but it probably work because it relies on backward compatibility, it sends true false as values.

The right way to do it is to convert the values of true/false boolean type to strings, and that what I did here, boolean is a primitive type and the server requires string.

sgratch: > `NicType.plugged` is an internal types, in web ui we saving boolean as true/false values not strings, but for the API we should convert this value to string value (`"true"` or `"false"` with any combination of lower/upper case).
> I don't know why it works now but it probably work because it relies on backward compatibility, it sends true false as values.
> 
> The right way to do it is to convert the values of true/false boolean type to strings, and that what I did here, boolean is a primitive type and the server requires string.

It seems to me that we should handle all API Boolean attributes the same way. 
The `Nic.plugged` attribute http://ovirt.github.io/ovirt-engine-api-model/master/#types/nic is just another example for that and it should behave the same as for snapshot.persistMemoryState:
https://github.com/oVirt/ovirt-web-ui/blob/7f65e3ef47106d5028a52e8b2a8c9eeef795aa7a/src/ovirtapi/transform.js#L723
https://github.com/oVirt/ovirt-web-ui/blob/7f65e3ef47106d5028a52e8b2a8c9eeef795aa7a/src/ovirtapi/transform.js#L740

Based on this, the `persist_memorystate` should be declared as Boolean and not String

sjd78: > I don't know why it works now but it probably work because it relies on backward compatibility, it sends true false as values.
> 

Because JSON and the JSON parser should handle it all just fine.  Keep in mind the rest api also supports XML encoding of all of the data, and that handles boolean values differently.

> The right way to do it is to convert the values of true/false boolean type to strings, and that what I did here, boolean is a primitive type and the server requires string.

Based on what evidence?  Does the call fail if the POST body has `persist_memorystate: true` instead of `persist_memorystate: "true"`?

src/ovirtapi/transform.js
```diff
@@ -478,14 +482,15 @@ const Snapshot = {
...
-      persistMemoryState: snapshot.persist_memorystate === 'true',
+      persistMemoryState: convertBool(snapshot.persist_memorystate),
       isActive: snapshot.snapshot_type === 'active',
     }
   },
 
   toApi ({ snapshot }: { snapshot: SnapshotType }): ApiSnapshotType {
     return {
       description: snapshot.description,
+      persist_memorystate: convertToBoolString(snapshot.persistMemoryState),
```
sjd78, line 493: I don't think this is necessary unless the call is actually failing if a raw boolean value is sent.

Is the call failing was a raw boolean?

sjd78 (changes requested):
Questions about the boolean handling

src/ovirtapi/transform.js
bamsalem: AFAIK raw value converted to 1/0 values, according the documentation we should avoid using those value since it should be removed in the future: 

> For backwards compatibility with older versions of the engine, the values 0 and 1 are also accepted. The value 0 has the same meaning than false, and 1 has the same meaning than true. Try to avoid using these values, as support for them may be removed in the future.

For now you are right but the call may fail in the future.
So, if we are doing it let's do it right.

src/ovirtapi/transform.js
```diff
@@ -51,6 +51,10 @@ function convertBool (val: ?string): boolean {
   return val ? val.toLowerCase() === 'true' : false
 }
 
+function convertToBoolString (val: ?boolean): BoolStringType {
```
rszwajko, line 54: For converting truthy values it's better to use `Boolean(val).toString()` . No need for custom function here.

sgratch: So if we decide to go this way, we need to change that for all other API Boolean attributes handled on web-ui (which are declared now as Boolean instead of Strings). 
Otherwise the code is not consistent. Check: https://github.com/oVirt/ovirt-web-ui/pull/1386/#discussion_r611568797.
I can add a separate issue for that.

sgratch:
@bamsalem please handle the conflict

bamsalem:
> @bamsalem please handle the conflict

Done.

src/ovirtapi/transform.js
sjd78: > AFAIK raw value converted to 1/0 values, according the documentation we should avoid using those value since it should be removed in the future:

If we send the snapshot request payload like this:
```json
{
  "description": "Test1",
  "persist_memorystate": false
}
```
it is valid JSON and the flag is a boolean.  I tested it and everything works just fine.  The JSON unmarshalling layer and the rest api model mappings work just fine.  Things like this just need to be tested because documentation isn't always accurate.

> 
> > For backwards compatibility with older versions of the engine, the values 0 and 1 are also accepted. The value 0 has the same meaning than false, and 1 has the same meaning than true. Try to avoid using these values, as support for them may be removed in the future.
> 
> For now you are right but the call may fail in the future.

I doubt very much that anyone would introduce a breaking change like this.

> So, if we are doing it let's do it right.

The issue is how we handle boolean values generally in POST/PUT/DELETE requests to the rest api.  Sharon's comment in the review thread is valid...

src/ovirtapi/transform.js
sjd78: Sending JSON booleans works just fine for anything value that is a "Boolean".  But...the engine will send back the field as a string that we transform back to a boolean.

For this field specifically, here is the JAXB generated Java code for the rest api model object:
```java
    @XmlElement(name = "persist_memorystate", type = String.class)
    @XmlJavaTypeAdapter(Adapter1 .class)
    @XmlSchemaType(name = "boolean")
    protected Boolean persistMemorystate;
```

The `Adapter1` class, also generated by JAXB, handles mapping Strings to Boolean and Boolean back to String:
```java
public class Adapter1
    extends XmlAdapter<String, Boolean>
{
    public Boolean unmarshal(String value) {
        return (org.ovirt.engine.api.utils.BooleanParser.parseBoolean(value));
    }

    public String marshal(Boolean value) {
        if (value == null) {
            return null;
        }
        return (javax.xml.bind.DatatypeConverter.printBoolean((boolean)(boolean)value));
    }
}
```

And the `BooleanParser` looks like this:
```java
public class BooleanParser {
    public static boolean parseBoolean(String value) {
        if ("true".equalsIgnoreCase(value) || "1".equals(value)) {
            return true;
        }
        if ("false".equalsIgnoreCase(value) || "0".equals(value)) {
            return false;
        }
        throw new InvalidValueException(
            "Value \"" + value + "\" isn't a valid boolean, it should be \"true\" , \"false\", \"0\" or \"1\""
        );
    }
}
```

The mismatch is how we define the value's type on our Api flow type.  Since it'll be received as "true" or "false" from the engine, it is technically a string.  Probably the best course of action is to make sure that all boolean types on the Api side should be (1) a generic string, or (2) a `string | boolean`, or (3) a `ApiBooleanType = "true" | "false"`.  Following that, all `toApi` functions would then use a function like `toApiBoolean(value)` so all boolean are handled the same way.

```js
function toApiBoolean (value: any): ApiBooleanType {
  return Boolean(value).toString()
}
```

(sorry for being long winded)

sgratch: > The mismatch is how we define the value's type on our Api flow type. Since it'll be received as "true" or "false" from the engine, it is technically a string. Probably the best course of action is to make sure that all boolean types on the Api side should be (1) a generic string, or (2) a string | boolean, or (3) a ApiBooleanType = "true" | "false". 

We can treat the value's type on our API flow as a Boolean since we know it should work as expected or as a ApiBooleanType = "true" | "false" just to avoid confusion for next time.

> Following that, all toApi functions would then use a function like toApiBoolean(value) so all boolean are handled the same way.
> 
> function toApiBoolean (value: any): ApiBooleanType {
>   return Boolean(value).toString()
> }

If we'll choose to define as ApiBooleanType then we should fix all place in a followup PR.

src/ovirtapi/transform.js
```diff
@@ -478,14 +478,15 @@ const Snapshot = {
...
-      persistMemoryState: snapshot.persist_memorystate === 'true',
+      persistMemoryState: convertBool(snapshot.persist_memorystate),
       isActive: snapshot.snapshot_type === 'active',
     }
   },
 
   toApi ({ snapshot }: { snapshot: SnapshotType }): ApiSnapshotType {
     return {
       description: snapshot.description,
+      persist_memorystate: Boolean(snapshot.persistMemoryState).toString(),
```
sjd78, line 489: Go ahead and convert over to a `type ApiBooleanType = "true" | "false"` and use a function like:

```js
function toApiBoolean (value: any): ApiBooleanType {
  return Boolean(value).toString()
}
```

to handle the transform.  A followup PR, like @sgratch suggested, to have the API boolean type applied across all of the types and transforms.

sjd78 (dismissed):
Just a request to tweak the boolean handling on the api.

src/ovirtapi/transform.js
bamsalem: I decided to implement it differently because I got a flow error:
image
please let me know WDYT :)

src/ovirtapi/transform.js
hstastna: Hmm, and what about the sentence _The valid values are the strings `false` and `true`._
in http://ovirt.github.io/ovirt-engine-api-model/master/#types/boolean ?
@sjd78 I am little bit confused 😄

hstastna (approved):
LGTM. And the error mentioned above already fixed :)

sjd78 (approved):
LGTM.

@bamsalem, @sgratch - Don't forget that the rest of the API types and `toApi` functions need to be reviewed and convert any `boolean` types to our new `ApiBooleanType`.  Probably best to open a new issue first.

sgratch:
> LGTM.
> 
> @bamsalem, @sgratch - Don't forget that the rest of the API types and `toApi` functions need to be reviewed and convert any `boolean` types to our new `ApiBooleanType`. Probably best to open a new issue first.

Done: #1406

sgratch (approved):
LGTM

Issue #1385, opened 2021-03-21
web-ui crashes on schedule update


bamsalem:
While I tested another change on vmDetails page of vm: `0f3880fc-fdb8-43c0-b352-a5eaba757412` on brq-dev
web ui crashes it seems like it's not related to the change I tested:
![image](https://user-images.githubusercontent.com/64131213/111900422-2f86d080-8a3b-11eb-914e-4ce1ee02a92e.png)

I couldn't reproduce that but it's worth checking.

hstastna:
I checked that VM in web-ui, also some other ones, played a little bit with snapshots, details, disks, NICs and wasn't able to reproduce the problem, on web-ui master branch or also web-ui on brq-dev. I suggest to close this issue. We can reopen it later anytime.

sgratch:
@bamsalem It seems that the  api/disks/c6a9cb2b-9e6c-4130-9f3c-6e9bedc4ae76/permissions?follow=role.permits returns 500 internal server error and that might cause the crash on permissionsToInternal since permissions.map was undefined. Not sure if worth fixing since it's an extreme case.

bamsalem:
OK,I know that is a server error but I raised it here so we can decide if it's worth fixing or create any checks before we calling permissions.map.
If you think that's an extreme case so let's close that issue :)
Better safe than sorry ;)

sgratch:
@bamsalem we can easily verify that permissions map is defined so let's do it. I'll assign this to you.

PR #1399, opened 2021-03-31
Add check/validation if permission response is an array, to prevent web ui from crashing

bamsalem:
Hi guys,
it's a small fix that should prevent web ui from crashing in case of backend 500 error.
It's an extreme edge case but it should be handled.
Fixes: #1385 .

### DIFF
--- a/src/sagas/utils.js
+++ b/src/sagas/utils.js
@@ -162,7 +162,10 @@ export function* delayInMsSteps (count = 20, msMultiplier = 2000) {
 
 export function* fetchPermits ({ entityType, id }) {
   const permissions = yield callExternalAction(`get${entityType}Permissions`, Api[`get${entityType}Permissions`], { payload: { id } })
-  return getUserPermits(Api.permissionsToInternal({ permissions: permissions.permission }))
+  if (permissions && Array.isArray(permissions.permission)) {
+    return getUserPermits(Api.permissionsToInternal({ permissions: permissions.permission }))
+  }
+  return []
 }
 
 export const PermissionsType = {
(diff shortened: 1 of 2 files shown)


src/sagas/utils.js
```diff
@@ -162,7 +162,10 @@ export function* delayInMsSteps (count = 20, msMultiplier = 2000) {
 
 export function* fetchPermits ({ entityType, id }) {
   const permissions = yield callExternalAction(`get${entityType}Permissions`, Api[`get${entityType}Permissions`], { payload: { id } })
-  return getUserPermits(Api.permissionsToInternal({ permissions: permissions.permission }))
+  if (permissions && Array.isArray(permissions.permission)) {
+    return getUserPermits(Api.permissionsToInternal({ permissions: permissions.permission }))
```
rszwajko, line 166: `Api.permissionsToInternal` is called in 2 places - the other place already had the check so the problem is not that rare. This suggest that it would be better to have the fix in one common place. 
How about simply `toInternal ({ permissions = [] }` ?

bamsalem: @rszwajko It's a great idea so I added the default value as you suggested.
IMO we don't want to "work" with the response without check its content so I kept the if statement.
Thanks :)

rszwajko: I would remove the checks now and rely only on default value for error handling but it depends what flow we want here:
current error handling (no action in case of error) does not clear existing permissions which might be a desired side effect..
We can as well have both the explicit check(main path) and default value (as safety net)

rszwajko (approved):

hstastna (approved):
I would remove the checks, too, and would be fine with `toInternal ({ permissions = [] }` but it's up to you. Otherwise LGTM.

sgratch (approved):
LGTM, it looks ok to me to leave the checks as is.

Issue #1389, opened 2021-03-24
user settings - confirmation dialog for user settings is missing


sgratch:
Confirmation dialog for user settings is missing, e.g.: "are you sure you want to make changes to the following settings: language, UI refresh interval."

For more details, check https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359 (issue 14)

sgratch:
Please note that this was already implemented in the original PR but removed, so need to just reuse the existing code. Check with @rszwajko for more details.

Issue #1387, opened 2021-03-22
user settings - Change refresh interval via Account Settings


sgratch:
Add an option to change the global refresh interval used to refresh data in the UI.

PR #1401, opened 2021-04-04
Adding confirmation modal on saving user settings changes

bamsalem:
This PR prompting confirmation before saving changes in user settings.
Before this PR the settings were automatically saved, now before saving the changes will require confirmation.
I added the changed settings to modal's message so the user can see which fields were changed.
Screenshots:
image

image

Fixes: #1389.

### DIFF
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -120,7 +120,13 @@ const Settings = ({ draftValues, onSave, lastTransactionId, onCancel,
         />
       }
     </div>
-    <SettingsToolbar onSave={handleSave} onCancel={onCancel} enableSave={!!pendingChanges.length} />
+    <SettingsToolbar
+      onSave={handleSave}
+      onCancel={onCancel}
+      enableSave={!!pendingChanges.length}
+      changes={pendingChanges}
+      translatedLabels={translatedLabels}
+    />
     {children}
   </React.Fragment>
 }
--- a/src/components/Settings/SettingsToolbar.js
+++ b/src/components/Settings/SettingsToolbar.js
@@ -5,9 +5,13 @@ import { Toolbar } from 'patternfly-react'
 import { msg } from '_/intl'
 
 import style from './style.css'
+import ConfirmationModal from '_/components/VmActions/ConfirmationModal'
 
-const SettingsToolbar = ({ onSave, onCancel, enableSave }) => {
+const SettingsToolbar = ({ onSave, onCancel, enableSave, translatedLabels, changes = [] }) => {
   const [container] = useState(document.createElement('div'))
+  const [showConfirmModal, setShowConfirmModal] = useState(false)
+  const idPrefix = 'settings_toolbar'
+
   useEffect(() => {
     const root = document.getElementById('settings-toolbar')
     if (root) {
@@ -16,8 +20,38 @@ const SettingsToolbar = ({ onSave, onCancel, enableSave }) => {
     return () => root && root.removeChild(container)
   })
 
+  const onConfirm = () => {
+    onClose()
+    onSave()
+  }
+
+  const onClose = () => {
+    setShowConfirmModal(false)
+  }
+
+  const buildConfirmationModalSubContent = () => (
+    <ul className={style['changes-list']}>{
+      changes.map(name => {
+        const value = translatedLabels[name] || name
+        return (<li key={`${idPrefix}_li_${value}`}>{value}</li>)
+      })
+    }
+    </ul>
+  )
+
   return ReactDOM.createPortal(
     <Toolbar className={style['toolbar']}>
+      <ConfirmationModal
+        show={showConfirmModal}
+        title={msg.saveChanges()}
+        body={msg.saveSettingsChangesConfirmation()}
+        subContent={buildConfirmationModalSubContent()}
+        onClose={onClose}
+        confirm={{
+          title: msg.yes(),
+          onClick: onConfirm,
+        }}
+      />
       <Toolbar.RightContent>
         <button
           onClick={e => {
@@ -32,7 +66,7 @@ const SettingsToolbar = ({ onSave, onCancel, enableSave }) => {
           disabled={!enableSave}
           onClick={e => {
             e.preventDefault()
-            onSave()
+            setShowConfirmModal(true)
           }}
           className='btn btn-primary'
         >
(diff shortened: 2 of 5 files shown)


hstastna (commented):
The code changes LGTM. However, I suggest to change the look to this one:
confirmation_modal

Two things:
- I am missing the `?` in the original sentence, in the confirmation modal. The question without the question mark does not feel so good to me

- the string with changed settings was not so clear: same font as the question, items in one line separated by commas - this can be problematic especially when having much more settings in the future

So I tried to put the string with names of the changed settings to the `subContent` prop (instead of `body`) of the `ConfirmationModal` component and added some css (`white-space: pre-wrap;`) to the appropriate element with the subcontent of the component, to achieve new line for each changed setting, together with this change:
```javascript
const mapChangeNamesToString = () => changes.map(name => translatedLabels[name] || name).join('\n')
```
WDYT?

rszwajko:
Note that we have actually some prototypes.

1. Confirmation with list of VMs impacted - in the original PR #1088  the confirmation dialog was limited to (multi) VM settings. Please check the last screenshot in this comment: https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-543721868
2. Confirmation is case of conflicts - https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-645990615

Looking at those solution I think that bullet pointed list looks  the best

hstastna:
> Note that we have actually some prototypes.
> 
>     1. Confirmation with list of VMs impacted - in the original PR #1088  the confirmation dialog was limited to (multi) VM settings. Please check the last screenshot in this comment: [#1088 (comment)](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-543721868)
> 
>     2. Confirmation is case of conflicts - [#1088 (comment)](https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-645990615)
> 
> 
> Looking at those solution I think that bullet pointed list looks the best

Thanks Radek, those prototypes look good to me, too. Just... let's not overcomplicate it. I think this does not have to be the best ;)

rszwajko (approved):

bamsalem:
I made few changes:
image
@hstastna please take a look :)

hstastna (changes requested):
In my previous review, I suggested some very simple way to achieve the modal info more readable. Now there are more code changes and I'm not sure if it worth regarding the modal look. I still prefer the simpler version. But that's just my preference.

_Settings changes will be made to the following Settings:_ and _Press 'Yes' to confirm these changes._ are unnecessary sentences with no extra information for the user. The user had to read more stuff before they click on some button. I suggest to remove them from the modal. The rest LGTM :)

rszwajko:
@hstastna 
>  I suggest to remove them from the modal. 
>
OK for me.  Using the list alone is good  enough.

sgratch (changes requested):
Generally it looks good, few suggestions:
- Any reason for removing the bullets from the bulleted list? 

image

- Please reorder the list based on dialog's appearance (language, ssh...)
- This is more related to locale reload issue: 
  if the locale is changed then due to reloading, Notifications fields are not persisted and therefore disabled so it seems strange to notify that they are going to be changed. 
@rszwajko  is this fixed with #1387?

rszwajko:
> Please reorder the list based on dialog's appearance (language, ssh...)
>

This is not really possible in the current implementation - there is no logical link between the order of properties in the store/state and the order of fields in the UI.  However the problem should be mitigated when we split the screen into subscreens.
 
> This is more related to locale reload issue:
>  if the locale isep changed then due to reloading, Notifications fields are not persisted and therefore disabled so it seems strange to notify that they are going to be changed.
>  @rszwajko  is this fixed with #1387?

Yep.  As for now we don't know which fields are persisted on the server and which are not as framework treats them in a generic way.

bamsalem:
> * Any reason for removing the bullets from the bulleted list?

No, It's just few changes that weren't staged, I pushed them now :)
 
> * Please reorder the list based on dialog's appearance (language, ssh...)

Before: 
image

After:
image

rszwajko (changes requested):
Can we handle ordering as separate PR? It's not connected with this issue.

Regarding proposed implementation:
1. Maintaining an explicit ordering makes refactorings/enhancements harder. Please infer the order from the existing structures.
2. Global ordering is problematic as more settings are added - it's better to keep the order relative to the section (and then infer order at section level).
3. Aggregate properties into objects instead passing them as separate arrays (and tranlsating back and forth). The main case here are `translatedLabels`, `displayOrder` and `changes`.

bamsalem:
> Can we handle ordering as separate PR? It's not connected with this issue.

I've removed the order for now.

> Regarding proposed implementation:
> 
> 1. Maintaining an explicit ordering makes refactorings/enhancements harder. Please infer the order from the existing structures.
> 2. Global ordering is problematic as more settings are added - it's better to keep the order relative to the section (and then infer order at section level).
> 3. Aggregate properties into objects instead passing them as separate arrays (and tranlsating back and forth). The main case here are `translatedLabels`, `displayOrder` and `changes`.

I have these changes in my machine let's decide if it should be part of this PR or separate PR.

rszwajko (approved):

hstastna (changes requested):
The confirmation modal itself LGTM. But I am getting the error with the changes in this PR (not present on master):
err

sgratch (approved):

bamsalem:
> The confirmation modal itself LGTM. But I am getting the error with the changes in this PR (not present on master):
> err

Fixed by me in #1404 .

sgratch:
> The confirmation modal itself LGTM. But I am getting the error with the changes in this PR (not present on master):
> err

@hstastna This error seems due to merged issue https://github.com/oVirt/ovirt-web-ui/issues/1403 (not rebased on top of this PR)

hstastna:
> > The confirmation modal itself LGTM. But I am getting the error with the changes in this PR (not present on master):
> > err
> 
> @hstastna This error seems due to merged issue #1403 (not rebased on top of this PR)

I had the freshest code when testing this PR today, and despite the PR #1404 was merged yesterday, I was still getting the error with the changes in this PR. So I had to mention it. But now after merging this PR, the error I've mentioned above is not reproducible anymore. Strange 😄

isaranova:
LGTM

Issue #1388, opened 2021-03-24
user settings - breadcrumbs should be fixed to locate "Account settings" not under "Virtual Machines"


sgratch:
In breadcrumb the "Account settings" is under "Virtual Machines" (i.e. Virtual Machines->Account Settings). That looks a little bit confusing since it's not related only to VMs.
This was raised here: https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359

PR #1405, opened 2021-04-12
Create general breadcrumbs root to resolve the confusion about "Virtual Machines" Breadcrumbs root

bamsalem:
The old breadcrumbs root was "Virtual machines" and it was a little bit confusing since the settings not related only to virtual machines.
In this PR I changed the old root to home icon instead of "Virtual Machines" and I added a special case for `vmListPage` that adds the "Virtual Machines" to the breadcrumbs path:
image
image
image

Fixes: #1388 .

### DIFF
--- a/src/components/Breadcrumb/index.js
+++ b/src/components/Breadcrumb/index.js
@@ -4,8 +4,11 @@ import { Link } from 'react-router-dom'
 import { connect } from 'react-redux'
 import { msg } from '_/intl'
 
+const NONE_VM_ROUTES = [ '/settings' ]
+
 const buildPath = (vms, branches) => {
   const res = []
+  const isVmPath = !branches.find(branch => NONE_VM_ROUTES.includes(branch.match.path))
 
   for (const branch of branches) {
     if (typeof branch.route.title === 'function') {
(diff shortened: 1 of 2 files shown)


sgratch (commented):
Before drilling into the code, a general comment regarding VM details hirarchy:
I think that it should be located under "Virtual machines" as originaly designed i.e.
Home icon->Virtual Machines->__a

> image

bamsalem:
> Before drilling into the code, a general comment regarding VM details hirarchy:
> I think that it should be located under "Virtual machines" as originaly designed i.e.
> Home icon->Virtual Machines->__a

Changed :)

image

sgratch (approved):
LGTM

hstastna (approved):
LGTM, too :)

src/components/Breadcrumb/index.js
```diff
@@ -3,11 +3,24 @@ import PropTypes from 'prop-types'
...
 
 const buildPath = (vms, branches) => {
-  const res = []
+  const res = [{
+    title: (<HomeIcon style={{ fontSize: '1rem' }} />),
+    url: '/',
+  }]
 
   for (const branch of branches) {
+    const { match: { path = '' } = {} } = branch
```
rszwajko, line 15: The purpose of `routes.js` is to have a readable overview of  all routes. Dynamic path replacement with hardcoded paths goes in  the opposite direction.

I would expect following structure in `routes.js`
1. root (home) with path/URL `/`. It has 2 sub-branches:VM Machines and Account Settings. Clicking on the root should redirect to VM Machines.
2.  VM Machines with path `/vms` (note the change `vm` -> `vms`).  Single VM would use path `/vms/{id}`.
3. Account Settings with path `/settings` (no change here)

rszwajko (dismissed):
I like the look but I think that all routes should be defined in `routes.js` in one structure.

src/components/Breadcrumb/index.js
rszwajko: @sjd78 
Please share you opinion. Should we model this fix via branch structure(`routes.js`) or is the fix in the visual layer enough?

sgratch: @bamsalem I agree with Radek that we should put all the branches logic in routes.js.
Is it a complicated fix to do?

bamsalem: Not at all, you can review the changes :)

sgratch (approved):
LGTM

hstastna (approved):
LGTM, too :)

isaranova:
LGTM

Issue #1366, opened 2021-02-09
Create VM Wizard- create disks storage domain issues


sgratch:
This is a follow up issue for an already merged PR #1323 regarding storage domains management during a creation of a VM with disks (original commect: https://github.com/oVirt/ovirt-web-ui/pull/1323#issuecomment-775954746):

**1.** When creating a VM based on a template while the template includes more than 1 disk and is optimized to "server" or a clone is required due to SD permissions then the following error still occurs:
"CREATE_VM failed [Cannot add VM. One or more provided storage domains are either not in active status or of an illegal type for the requested operation.]"

This was already mentioned above on issue 3 of https://github.com/oVirt/ovirt-web-ui/pull/1323#issuecomment-728239827 and claimed to be fixed:
>  3.       When trying to create a VM based on a "server" template with two disks OR when trying to create a VM based on a template with no permissions for the template's two disks SD, then for both cases the VM creation failed on backend with the following error (I guess the issue is due to the two disks that needed to be cloned at once, while image is locked or something like that):
> 
> Fixed! There was an error in the way multiple disk changes were being merged to the new VM request. Creation of VMs from multi-disk templates cloning the disk to a new storage domain works just fine now.

**2.** when there are no permissions for all SDs , both previous and next buttons are disabled:
![image](https://user-images.githubusercontent.com/18169498/107372053-aac89e80-6aed-11eb-96e2-c970527a0494.png)

**3.** when there are no permissions for a template disks SD and the user can choose a different SD, the previous button is disabled till all SD are selected. This is wrong:
![image](https://user-images.githubusercontent.com/18169498/107372326-f5e2b180-6aed-11eb-9d57-6768e2a142c4.png)

**4.** when the user tries to create a disk not based on  a template and there are no permissions for current SDs, the SD field is marked as red and buttons are disabled, but no message is displayed:
![image](https://user-images.githubusercontent.com/18169498/107373293-0e070080-6aef-11eb-84ea-1d24f252d166.png)

sjd78:
> This is a follow up issue for an already merged PR #1323 regarding storage domains management during a creation of a VM with disks (original commect: [#1323 (comment)](https://github.com/oVirt/ovirt-web-ui/pull/1323#issuecomment-775954746)):
> 
> **1.** When creating a VM based on a template while the template includes more than 1 disk and is optimized to "server" or a clone is required due to SD permissions then the following error still occurs:
[...]

PR #1407, opened 2021-04-20
Update create VM step navigation and disk create checks

sjd78:
Fixes: #1366

Creating a VM from a template with disks defined, and if the disks need to be cloned to a new storage domain, navigation buttons needed to be enabled differently.  Now, once on the storage domain wizard step:

  - The forward button is enabled ONLY when all of the disks are correct and no edit is active.

  - The back button is always enabled, including when:

    - an edit is active

    - no storage domain is available for the template disks

    - disk storage domains must be selected, but have not yet been selected (i.e. edits are active)

When no storage domains are available, and no disks have been defined by the selected template, the `EmptyState` place holder will now show a "cannot create a disk" message.  This message replaces the "Create Disk" button.

---
Active edit:
VM Portal - Google Chrome_025

No SD available:
VM Portal - Google Chrome_026

SD must be selected but has not yet been:
VM Portal - Google Chrome_028

SD has been selected, everything OK, forward enabled:
VM Portal - Google Chrome_029

Everything OK by default, no active edits, forward enabled:
VM Portal - Google Chrome_027

### DIFF
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -49,20 +49,20 @@ const DEFAULT_STATE = {
   showCloseWizardDialog: false,
   stepNavigation: {
     basic: {
+      valid: false,
       preventEnter: false,
-      preventExit: true,
     },
     network: {
+      valid: true,
       preventEnter: false,
-      preventExit: false,
     },
     storage: {
+      valid: false,
       preventEnter: false,
-      preventExit: false,
     },
     review: {
+      valid: true,
       preventEnter: false,
-      preventExit: false,
     },
   },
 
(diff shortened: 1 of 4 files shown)


sjd78:
**NOTE** - A user needs to have `DiskCreator` and `DiskProfileUser` permissions on a storage domain to be able to both create the disk (or clone a disk from another) and to attach the newly created disk to the VM.

sgratch (changes requested):
It looks now that the "Create VM button" is always disabled:
image

rszwajko:
@sjd78 
Please check also the following scenario:

1. edit a NIC
2. trigger validation error (i.e. by duplicate name)
3. navigate back to Basic Settings
4. go back to Network Settings

Result: next button is disabled but the edited row is not visible and cannot be corrected.

sjd78:
> @sjd78
> Please check also the following scenario:
> 
>     1. edit a NIC
> 
>     2. trigger validation error (i.e. by duplicate name)
> 
>     3. navigate back to Basic Settings
> 
>     4. go back to Network Settings
> 
> 
> Result: next button is disabled but the edited row is not visible and cannot be corrected.

Thanks for finding this one!  I now have the Networking step validate itself on entry in the same way the Storage step does.  Much better!

sgratch (approved):
LGTM

Issue #1411, opened 2021-04-22
User permit calculations are including extra permissions


sjd78:
In some cases, if a user is assigned read permissions to an entity, that user can see all of the permissions assigned to that entity.  The `entityPermissionsToUserPermits` function does not handle this situation properly.

VM Portal version number: 1.6.9

oVirt version number: 4.4.6

Browser and version: N/A

Steps to Reproduce:
1. Create 2 users, userA and userB
2. Using webadmin / Admin Portal, assign permissions to a Cluster as follows:
  - userA: UserRole
  - userA: VmCreator
  - userB: UserRole
3. Open web-ui / VM Portal as userB

Actual results: The **Create VM** button is displayed, the wizard can get to the review step but the creation is blocked in the REST layer for lack for correct permissions

Expected results: No **Create VM** button is displayed

Looks like since userB has UserRole, when the permissions for the cluster is fetched, the permissions for userA are fetched even with the `Filter: true` header active.  The code for `entityPermissionsToUserPermits` already has checks that run for an admin user to ensure a fetched permission is active for the user since admin users don't normally fetch filtered.

sjd78:
In my dev-engine for **Cluster_DC-1_A**:
![oVirt Open Virtualization Manager - Google Chrome_030](https://user-images.githubusercontent.com/3985964/115761207-193c9f00-a370-11eb-9809-baead43627e3.png)

userB id: 
```sh
> curl -k -u"userB@internal:user" -H"Accept: application/json" \
  -H"Filter:true" \
  "https://engine-dev/ovirt-engine/api/users/7124321c-4441-4581-902e-2418b7d6e630" \
  jq '{user_name:.user_name, id:.id}'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1508    0  1508    0     0   8870      0 --:--:-- --:--:-- --:--:--  8870
{
  "user_name": "userB@internal-authz",
  "id": "7124321c-4441-4581-902e-2418b7d6e630"
}
```

Cluster permissions with `Filter:true`: 
```sh
> curl -k -u"userB@internal:user" -H"Accept: application/json" \
  -H"Filter:true" \
  "https://dev-engine/ovirt-engine/api/clusters/39a7e3f3-cd0f-4bde-84af-d99170da7b79?follow=permissions.role" \
  | jq '{name:.name, permissions:.permissions.permission}'  
[...]

PR #1412, opened 2021-04-22
Update `entityPermissionsToUserPermits()` to correctly filter permissions

sjd78:
Fixes: #1411

The assumption that fetching permissions with `Filter: true` will
only return permissions that **apply** to requesting user was wrong.
All permissions that can be **read by** the requesting user are
returned.  Additional filtering needs to be done to make sure the
permissions that can be seen apply to the user (or a group the user
belongs to).

Now `entityPermissionsToUserPermits()` will verify that a permission
applies to the user before adding the permission's permits to the
set of permits for an entity.


rszwajko (approved):

sgratch (approved):
LGTM

Issue #1393, opened 2021-03-24
user settings - consider enable/disable option for persisted locales


sgratch:
Since "landing page" locales and "user settings" for web-ui locales integration is partially supported, consider enable/disable persisted locale as mentioned on "issue 5" of previous comment https://github.com/oVirt/ovirt-web-ui/pull/1372#issuecomment-791576420

rszwajko:
**Background**

We have 3 storage options right now:
1.    server-side - persisted as part of user profile i.e. refresh interval
2.    client side - persisted in local storage i.e. locale or legacy console settings
3.    Redux store (in-memory) - i.e. notifications

Those options are not mutually exclusive - all options are in Redux store and locale is both server and client side(currently).

**Basic use case**

Some user might depend on passing the locales via URL (from landing page) and might not want to have this option persisted on the server. Currently this is the only item we want to support.

**Generic use case**

In the future we can expect more options that  need opt-out feature. IMHO the first candidate are (global) console settings.
Adding multiple "opt-out" check boxes or switches would make the UI unreadable so my idea was to group such options under a dedicated section.
There are few naming and layout options (to be decided which fits the best):
1. section: simply "Storage" or "Persistence" or similar
2. storage modes :
    1. on/off (via switch  or checkbox) - "on" means that option is saved on the server
    2.  server-side/local/in-memory (via dropdown) - the most flexibility

**Simple mockup**

![storage_options](https://user-images.githubusercontent.com/64194103/116229728-2504d880-a757-11eb-9fba-9c6ebcae936b.png)

rszwajko:
Original User Settings design is available here -> https://marvelapp.com/prototype/94h0g56/screen/59286965

PR #1414, opened 2021-04-26
Labels for Account Settings

rszwajko:
Labels for following sections:

1. Console options - planned layout is section per console type: VNC, Spice, RDP, Serial
2. Resetting to default Account Settings
3. Server persistence opt-out feature - choosing the settings storage type (local or server-side)


ovirt-infra:

bamsalem (approved):
Since it's only strings and no actual code to test, without test it locally, LGTM

sgratch:
ci add to whitelist

src/intl/messages.js
```diff
@@ -408,6 +410,8 @@ export const messages: { [messageId: string]: MessageType } = {
     message: 'Loading ...',
     description: 'Pending request to the server indicator label.',
   },
+  localStorage: 'Local',
+  localStorageTooltip: 'Settings are persisted by the browser locally.',
```
sgratch, line 414: What is this?
Is it referring to https://github.com/oVirt/ovirt-web-ui/issues/1393? I don't think/sure we need to enable this for all settings options. We discussed that only in the context of the locales. 
WDYT?

src/intl/messages.js
```diff
@@ -516,13 +520,16 @@ export const messages: { [messageId: string]: MessageType } = {
   pendingChanges: 'Pending Changes',
   pleaseEnterValidHostName: "Please enter a valid host name. Only lower-case and upper-case letters, numbers, and '_','-','.' are allowed.",
   pleaseEnterValidVmName: "Please enter a valid virtual machine name. Only lower-case and upper-case letters, numbers, and '_','-','.' are allowed.",
+  preferredConsole: 'Preferred Console',
+  preferredConsoleTooltip: 'The chosen console will be the first console displayed on a VM card',
```
sgratch, line 524: "The chosen console **type** will..."

src/intl/messages.js
```diff
@@ -516,13 +520,16 @@ export const messages: { [messageId: string]: MessageType } = {
   pendingChanges: 'Pending Changes',
   pleaseEnterValidHostName: "Please enter a valid host name. Only lower-case and upper-case letters, numbers, and '_','-','.' are allowed.",
   pleaseEnterValidVmName: "Please enter a valid virtual machine name. Only lower-case and upper-case letters, numbers, and '_','-','.' are allowed.",
+  preferredConsole: 'Preferred Console',
```
sgratch, line 523: I would change it to "console type"...

src/intl/messages.js
```diff
@@ -566,10 +584,15 @@ export const messages: { [messageId: string]: MessageType } = {
   selectTheBootableDeviceTooltip: 'Select the bootable devices in the order you would like them to boot.',
   sendShortcutKey: 'Send Key',
   sendCtrlAltDel: 'Ctrl+Alt+Del',
+  serialConsole: 'Serial Console',
+  serialConsoleOptions: 'Serial Console Options',
```
sgratch, line 588: Serial console settings is not needed on web-ui

src/intl/messages.js
```diff
@@ -566,10 +584,15 @@ export const messages: { [messageId: string]: MessageType } = {
...
+  serverStorage: 'Server-side',
+  serverStorageTooltip: 'Settings are persisted on the server.',
```
sgratch, line 590: This should be considered in the context of locale settings only

src/intl/messages.js
```diff
@@ -566,10 +584,15 @@ export const messages: { [messageId: string]: MessageType } = {
...
   sessionExpired: {
     message: 'Your session is about to timeout due to inactivity.',
     description: 'Primary message for SessionTimeout modal component',
   },
+  settingsStorage: 'Settings storage',
```
sgratch, line 595: Why we need this?

src/intl/messages.js
```diff
@@ -582,12 +605,16 @@ export const messages: { [messageId: string]: MessageType } = {
...
+  smartcard: 'Smartcard',
   smartcardEnabled: 'Smartcard enabled:',
+  smartcardTooltip: 'Smartcard is a physical card that can be used to authenticate to VMs connected to a SPICE console.',
   snapshot: 'Snapshots',
   snapshotRestore: 'Restore Snapshot',
   snapshotDelete: 'Delete Snapshot',
   snapshotInfo: 'All attached disks will be included in the snapshot. Content of memory can be included only if VM is running.',
   snapshotsTooltip: 'VM snapshots.',
+  spiceOptions: 'SPICE Options',
+  spiceProxy: 'SPICE Proxy',
```
sgratch, line 617: This is not required/planned as part of console options

src/intl/messages.js
```diff
@@ -699,8 +726,10 @@ export const messages: { [messageId: string]: MessageType } = {
   utilizationNoDataAvailableMessage: 'Utilization data is only available when the VM is running.',
   utilizationNoHistoricData: 'No historic data available',
   utilizationNoNetStats: 'Network utilization is not currently available for this VM.',
+  usbAutoshare: 'USB Auto-Share',
```
sgratch, line 729: This is not required/planned as part of console options

src/intl/messages.js
```diff
@@ -699,8 +726,10 @@ export const messages: { [messageId: string]: MessageType } = {
...
   useBrowserBelow: 'Please use one of the browsers below.',
   useCtrlAltEnd: 'Use Ctrl+Alt+End',
+  useLocalDrives: 'Use Local Drives',
```
sgratch, line 732: This is not required/planned as part of console options

src/intl/messages.js
```diff
@@ -537,12 +544,23 @@ export const messages: { [messageId: string]: MessageType } = {
...
   remove: 'Remove',
   removeVm: 'Remove VM',
   removeVmQustion: 'Are you sure you want to remove the VM? All of the data attached to the VM will be removed as well.',
+  reset: {
+    message: 'Reset',
+    description: 'Label for the Reset button',
+  },
+  resetSettings: 'Reset settings',
+  resetSettingsQuestion: 'Are you sure you want to reset your account settings?',
+  resetSettingsWarning: 'Resetting your account settings will clear all of the individual settings that were applied to it. Your account settings will be returned to the default settings. Please confirm that you want to reset your account settings before making these changes.',
```
sgratch, line 563: I would add:
"default settings (**installation defaults**)"
since those defaults are the hardcoded ones...

sgratch (changes requested):
Few comments:

- "Automatic console connection" "automatically connect to the console..." - is missing
- account settings tabs (general, console, notifications...) are missing but we're still not sure about those so let's leave it out for now. The only Ttabs we can be sure of are maybe "general" and "console" so maybe adding only those (?)
- Please consider handling https://bugzilla.redhat.com/show_bug.cgi?id=1950323 on the way..

rszwajko:
>   "Automatic console connection" "automatically connect to the console..." - is missing
>

Yes. We have already `connectAutomatically` label which seems to be used before. Auto-connect is also a VM specific settings (not part of the Account Settings).
 
>   account settings tabs (general, console, notifications...) are missing but we're still not sure about those so let's leave it out for now. The only Ttabs we can be sure of are maybe "general" and "console" so maybe adding only those (?)
> 

We have already labels for: General, Console, Notifications and Advanced options

>   Please consider handling https://bugzilla.redhat.com/show_bug.cgi?id=1950323 on the way..
>

This will require gerrit patch as it's Web Admin specific.

src/intl/messages.js
rszwajko: > Is it referring to #1393?

Yep. It will be (for now) only for language but my plan is to create  Advanced Options section and put it there.
My reasons:
1. adding this switch/checkbox next to language dropdown looks awkward
2. we are adding more options that require opt-out i.e. console options and notifications. Adding a switch/checkbox for every property would look bad.

src/intl/messages.js
rszwajko: Providing a correct naming gets tricky here... just "console type" could confuse users as this is not a console that will be used for **every** VM. It will be the first choice of course but console needs to be supported by the VM.

src/intl/messages.js
rszwajko: AFAIK SSH key is for serial console. The original label was a bit more descriptive - now it's reflected in the tooltip only. Originally it was planned in Console section.

src/intl/messages.js
rszwajko: Just wanted to cover all cases here. We have actually 3 options:
1. server-side - persisted as part of user profile i.e. refresh interval
2. client side - persisted in local storage i.e. locale
3. in-memory - not persisted at all - i.e. notifications

src/intl/messages.js
rszwajko: I plan to use it as section under Advanced Options to allow opt-out (currently only language).

src/intl/messages.js
rszwajko: 1. SPICE Options is a label for a section that groups all SPICE options (there are few of them)
2. SPICE Proxy is just in case here - it's present in the Web Admin UI

src/intl/messages.js
rszwajko: Added just in case to be compliant with Web Admin console options.

src/intl/messages.js
rszwajko: Added just in case to be compliant with Web Admin console options.

src/intl/messages.js
rszwajko: > default settings (installation defaults)
>

IMHO `installation defaults` sounds more complex then simple `default settings`. I would stick to the original design here.

rszwajko:
1. Added `(Default)` label  - to be used in Select Box to mark the default options.
2. corrected `console -> console type`

src/intl/messages.js
sgratch: Right, but "Preferred Console" is also tricky from the same reason. That's why we might need a tooltip for explaining it.

src/intl/messages.js
sgratch: I'm not sure that opt-out for console options and notification are really needed. I can't think of a reason for this but ok.

src/intl/messages.js
sgratch: I prefer the current design (which fits the current implementation).
It can be in the console section.

src/intl/messages.js
sgratch: Let's have a separate discussion regarding this. It think that it might confuse users and not really needed to have those options. It might also complex our solution. 
For most settings - option 1 should be enabled unless we didn't implement this yet (notifications) or unless we have gaps that might break existing functionality like the locale.
option 2 is an internal solution since we also persisted the locale on server as well.

src/intl/messages.js
sgratch: I prefer a different label, more user friendly, like "persist account settings". This will require a tooltip as well....

src/intl/messages.js
sgratch: Not all options on webadmin are relevant for web-ui.
It was already decided about those that are relevant as part of Laura's design, based on older user portal version and customer requirements.
I prefer not to add settings that we are not planning to implement :)

src/intl/messages.js
sgratch: Same here

src/intl/messages.js
sgratch: Same here

src/intl/messages.js
sgratch: > > "Automatic console connection" "automatically connect to the console..." - is missing
> 
> Yes. We have already `connectAutomatically` label which seems to be used before. Auto-connect is also a VM specific settings (not part of the Account Settings).
> 

let's change to the labels suggested by Laura.


> > account settings tabs (general, console, notifications...) are missing but we're still not sure about those so let's leave it out for now. The only Ttabs we can be sure of are maybe "general" and "console" so maybe adding only those (?)
> 
> We have already labels for: General, Console, Notifications and Advanced options
> 

+1

> > Please consider handling https://bugzilla.redhat.com/show_bug.cgi?id=1950323 on the way..
> 
> This will require gerrit patch as it's Web Admin specific.

Right, my mistake.

sgratch:
@rszwajko 
We might also add a tooltip (as appearing under each settings tab in origibal design) of:
image

rszwajko:
ci please test

rszwajko:
ci test please

src/intl/messages.js
rszwajko: technically we support it (but it's disabled):
https://github.com/oVirt/ovirt-web-ui/blob/53f33a8dc918b23e0f63498c4049fc70f9c794ab/src/sagas/console/rdpBuilder.js#L155

src/intl/messages.js
rszwajko: We might end up with leaving the current design but I would like to have some plan B here. Can we include them for now and decide later?

src/intl/messages.js
rszwajko: How about Advanced Options -> Storage -> Persist language +  checkbox or  switch?

sgratch: I don't like the "storage" label name. It's confusing with the storage settings for a VM.

sgratch (approved):
Overall it LGTM. If needed then we'll fix that later on.
- "Automatic console connection" - is missing. I prefer this suggested by Laura over the old connectAutomatically label
- "Enter serial console ssh key" -  is missing. I prefer this suggested by Laura over the old "ssh key" label, regardless to the section is resides in
- Few of the added labels are not surely needed.

PR #1415, opened 2021-04-26
Support default option in SelectBox

rszwajko:
Mark default options on Account Settings screen for:
1. language
default_language
2. refresh interval
default_referesh_interval

### DIFF
--- a/src/components/SelectBox.js
+++ b/src/components/SelectBox.js
@@ -16,6 +16,14 @@ function getItems ({ sort, items, locale }) {
   return sort ? sortedBy(items, 'value', locale) : items
 }
 
+const MarkAsDefault = withMsg(({ msg, value, isDefault }) => {
+  if (!isDefault) {
+    return value
+  }
+  return (<React.Fragment>{value}{NOBREAK_SPACE}<i>{msg.defaultOption()}</i></React.Fragment>
+  )
+})
+
 /*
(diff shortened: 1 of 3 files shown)


src/components/SelectBox.js
```diff
@@ -60,10 +60,13 @@ class SelectBox extends React.Component {
   }
 
   render () {
-    const { id, disabled } = this.props
+    const { id, disabled, msg } = this.props
 
     const selectedItem = this.state.items.find(item => item.id === this.state.selected)
     const validationClass = this.getValidationClass()
+    const markAsDefault = () => {
```
sjd78, line 67: I don't think this needs to be a function....

You should be able to do:
```jsx
const MarkAsDefault = <i>{msg.defaultOption()}</i>
```

and then later just do:
```jsx
  {selectedItem && selecteItem.default && <MarkAsDefault />}
```

sjd78 (changes requested):
Just a suggestion on the code style in `SelectBox`

hstastna (approved):
LGTM :)

rszwajko:
@sjd78 
Changes:
1. as suggested moved more responsibility to `MarkAsDefault` component
2. renamed the flag to`isDefault` to prevent name clash with `default` as reserved word

sjd78 (approved):
LGTM

PR #1420, opened 2021-04-28
Retrieve current graphic consoles

rszwajko:
This PR depends on patch: https://gerrit.ovirt.org/c/ovirt-engine/+/114302 and fixes bug: https://bugzilla.redhat.com/1940766
1. use detail=current_graphics_consoles to retrieve the list of
   available consoles (for active VMs only)
2. remove AutoSelect
3. use ClientModeConsoleDefault to set console priorities

Screenshot (VNC console has the highest priority):
console_current


src/sagas/index.js
```diff
@@ -191,33 +190,21 @@ function* fetchConsoles ({ payload: { vm } }) {
...
-    const consoles = Transforms.VmConsoles.toInternal({ consoles: apiConsoles })
-    // select the console to use based on availability and app/user configs
-    const selectedConsole =
-      consoles.length === 0 ? EMPTY_CONSOLES_LIST
-        : consoles.length === 1 ? consoles[0]
-          : consoles.find(c => c.protocol === defaultConsoleProtocol) || NO_DEFAULT_CONSOLE
-
-    yield put(setVmConsoles({ vmId, consolesList: consoles, selectedConsole }))
-  }
+  // for headless VM (no graphics_consoles defined) - API response is `{}`
```
sjd78, line 193: Is this saga, `fetchConsoles`, necessary anymore?

src/components/VmActions/index.js
```diff
@@ -127,63 +127,59 @@ class VmActions extends React.Component {
...
     } = this.props
     const isPoolVm = !!vm.getIn(['pool', 'id'], false)
     const isPool = !!pool && !isPoolVm
     const status = vm.get('status')
     const onStart = (isPool ? onStartPool : onStartVm)
 
-    const vncConsole = vm.get('consoles').find(c => c.get('protocol') === 'vnc')
+    const vncConsole = vm.get('consoles').find(c => c.get('protocol') === VNC)
     const hasRdp = isWindows(vm.getIn(['os', 'type']))
+    const defaultConsole = vm.get('defaultConsole')
```
sjd78, line 139: Is the `defaultConsole` being stored with the VM this way because you intend to make it user configurable in future?

Otherwise, it seems unnecessary to keep the system `defaultConsole` stored in the VM when it can just be accessed in `config`.

sjd78 (commented):
Looks ok, just a few questions before approving

src/sagas/index.js
rszwajko: we can remove it

sjd78 (commented):
These changes LGTM.  Hopefully the restapi patch can be progressed soon.

Question - and I didn't test this scenario - what happens if this change runs on an engine without the patch?  Is the console functionality lost completely?  If yes, we should probably put a more specific backend version check at login to ensure things are going to work.

rszwajko:
> Question - and I didn't test this scenario - what happens if this change runs on an engine without the patch? Is the console functionality lost completely?
>

Yes, there will be no consoles on the dashboard. We could use `follow=graphics_consoles&current` to address this scenario - if the engine supports the new detail then it will ignore `follow`. Old engine will ignore the detail and return `follow`. The `current` flag makes it 2x faster then `follow` alone.

rszwajko:
@sjd78 
The backward compatibility option (`follow=graphics_consoles&current`) depends on features in patch https://gerrit.ovirt.org/c/ovirt-engine/+/114302
Until the patch is merged I would wait with implementing them in the UI.

rszwajko:
Changes:
1. use additional `current` flag when fetching VM details to retrieve
   graphic consoles info from vm_dynamic. This is ~2x faster
   then standard `follow=graphics_consoles`.
2. for backward compatibility use `follow=graphics_consoles&current`
   when (shallow) fetching multiple VMs. If the new detail is supported
   by the server then those params will be ignored.

sjd78 (approved):
The backward compatibility works.  I verified against a 4.4.5 engine.

Code LGTM

sgratch (changes requested):
@rszwajko 
The changes look good and working well except the following:
- For headless Win based VMs, the RDP option is dismissed for both dashboard and VM details views.
- The `ClientModeVncDefault`  config parameter is ignored for setting the default console, i.e. in case `ClientModeVncDefault` is set to `NoVnc` and `ClientModeConsoleDefault` is set to `vnc` or the VM supports only vnc, then the 'VNC console browser' option should be set as a default on dashboard (this is how it behaves on webadmin as well).
- For both new engine and backward compatibility for old engines:
It seems that using the vm_dynamic is faster, but we still need to be sure that using `?follow=graphics_consoles&current` or the new `detail=current_graphics_consoles`  on dashboard won't re-generate bug https://bugzilla.redhat.com/show_bug.cgi?id=1795457 as was when using `?follow=graphics_consoles`. This bug was the whole motivation for the original fix. When checking the code, it seems ok but I can't be sure since we don't know where the sql query:  `select * from getdisksvmguid` is actually generated by the `follow=graphics_consoles` call. 
I suggest asking the performance QE team to verify before marking the BZ as modified. Unless we can verify that in the code.

rszwajko:
> For headless Win based VMs, the RDP option is dismissed for both dashboard and VM details views.

It's not a regression. This fix was introduced in https://github.com/oVirt/ovirt-web-ui/commit/148f73159aa72d65f4e5743ecb5d780404efb663

> The ClientModeVncDefault config parameter is ignored for setting the default console, i.e. in case ClientModeVncDefault is set to NoVnc and ClientModeConsoleDefault is set to vnc or the VM supports only vnc, then the 'VNC console browser' option should be set as a default on dashboard (this is how it behaves on webadmin as well).

Fixed by introducing a clear separation between protocols and UI consoles available for the end user. The source of the confusion was that VNC protocol maps into 2 UI consoles.

>  I suggest asking the performance QE team to verify before marking the BZ as modified. Unless we can verify that in the code.

OK. The approach seems safe to me as it is used by Web Admin (with comparable number of VMs). However we could do more research on this.

src/components/VmActions/index.js
```diff
@@ -136,12 +137,13 @@ class VmActions extends React.Component {
...
     const vncConsole = vm.get('consoles').find(c => c.get('protocol') === VNC)
     const hasRdp = isWindows(vm.getIn(['os', 'type']))
-    const defaultConsole = vm.get('defaultConsole')
+    const defaultUiConsole = config.get('defaultUiConsole')
     let consoles = []
 
     consoles = vm.get('consoles').map((c) => ({
       priority: 0,
       protocol: c.get('protocol'),
+      uiConsole: toUiConsole(NATIVE, c.get('protocol')),
```
sgratch, line 146: This doesn't seem incorrect to me. I think that in case Spice is not supported and vnc is supported then the default ui conosle type should be based on ClientModeVncDefault as well, i.e in case noVnc is supported then this should be the default.

src/constants/console.js
```diff
@@ -13,7 +13,15 @@ export const CONSOLE_IN_USE = 'IN_USE'
 export const CONSOLE_LOGON = 'LOGON'
 
 export const RDP_ID = 'rdp'
+// console protocols
 export const VNC = 'vnc'
-export const NO_VNC = 'noVNC'
 export const SPICE = 'spice'
 export const RDP = 'rdp'
+
+// VNC modes sent from the backend (ClientConsoleMode)
```
sgratch, line 21: ClientConsoleMode -> ClientModeVncDefault ?

sgratch (dismissed):
> It's not a regression. This fix was introduced in 148f731

Based on webadmin and from current design aspect, we don't really support an option to disable RDP since RDP is always enabled for win vms, even headless ones. 
We did miss that for older versions and added this functionality back on https://github.com/oVirt/ovirt-web-ui/pull/1342. 
So if we want to be compatible with webadmin, we need to add this.  
Nevertheless since it was not supported till 1.6.8, it's ok for me to leave it as is until someone will complain. Not that important.

src/constants/console.js
rszwajko: I was referring to `VncConsoleModel.ClientConsoleMode` enum but referring to config property is better.

src/components/VmActions/index.js
rszwajko: The current code provides the requested functionality.  

Note that we always support both VNC flavors if VNC protocol is present:
1.  `NoVnc` is added when `vncConsole` flag is t`true`. 
2.  The `Native` VNC is added by directly mapping the list of available protocols (that's why `toUiConsole()` is used with first parameter set to `NATIVE`).


This logic was inherited from previous versions.

rszwajko:
> > It's not a regression. This fix was introduced in [148f731](https://github.com/oVirt/ovirt-web-ui/commit/148f73159aa72d65f4e5743ecb5d780404efb663)
> 
> Based on webadmin and from current design aspect, we don't really support an option to disable RDP since RDP is always enabled for win vms, even headless ones.
> We did miss that for older versions and added this functionality back on #1342.
> So if we want to be compatible with webadmin, we need to add this.
> Nevertheless since it was not supported till 1.6.8, it's ok for me to leave it as is until someone will complain. Not that important.

We can fix it under separate issue.

src/components/VmActions/index.js
sgratch: @rszwajko the issue raised here is regarding the default console displayed for the VM card in case that the defaultConsole is not supported by that VM.

For reproducing:
set `ClientModeConsoleDefault`=`spice`, `ClientModeVncDefault`=`NoVnc` and  set the VM to support only `VNC`.

Expected result: the default option should be `BrowserVnc`.
Actual result:  the default option is `VNC`.
This seems like not supported by the current code but it was supported by the previous version and by webadmin.

> > > It's not a regression. This fix was introduced in [148f731](https://github.com/oVirt/ovirt-web-ui/commit/148f73159aa72d65f4e5743ecb5d780404efb663)
> > 
> > 
> > Based on webadmin and from current design aspect, we don't really support an option to disable RDP since RDP is always enabled for win vms, even headless ones.
> > We did miss that for older versions and added this functionality back on #1342.
> > So if we want to be compatible with webadmin, we need to add this.
> > Nevertheless since it was not supported till 1.6.8, it's ok for me to leave it as is until someone will complain. Not that important.
> 
> We can fix it under separate issue.

Sure, please open a separate issue for this

rszwajko: > @rszwajko the issue raised here is regarding the default console displayed for the VM card in case that the defaultConsole is not supported by that VM.
> 
> For reproducing:
> set `ClientModeConsoleDefault`=`spice`, `ClientModeVncDefault`=`NoVnc` and set the VM to support only `VNC`.
> Expected result: the default option should be BrowserVnc.
> Actual result: the default option is VNC.
> This seems like not supported by the current code but it was supported by the previous version and by webadmin.
>

OK. I understand the use case now. However I'm not sure if it applies to the current design. Previous version and Web Admin provide a single button with black-box style logic (no indication what console will be chosen by the algorithm).
In the new design it's transparent to the user:
1. all available consoles are visible (the default console goes first)
2. the default global console send from the server (server's default console) is displayed in Account Settings (see below - feature available soon). Due to that there is only one such default (we don't have secondary default  for cases like `ClientModeConsoleDefault=spice`, `ClientModeVncDefault=NoVnc` )
3. user can override(completely) the server's default by setting a preferred console in Account Settings (available soon). Once again it's only one value (a limitation of SelectBox UI).
4. user can override the global console by setting per VM value (available in the next phase)

The goal is to have the logic here simple and transparent that's why only 2 levels are used:
1. global - either server's default console or console set by the user in Account Settings
2. per VM (next phase)

If the default is not supported by the given VM (for any reason) then consoles are displayed in default order (as before). Of course we could provide fallback but things will get complex i.e. imagine the case where:
1. ClientModeConsoleDefault`=`spice`, `ClientModeVncDefault`=`NoVnc`
2. preferred console is RDP
3. VM is Linux with VNC only

I'm not convinced if the added value justifies the complexity:) 

image

sgratch:
@rszwajko 
I agree with all you wrote above regarding the design, **except** the specific scenario of which the user specifically sets the Vnc mode to noVnc (`ClientModeVncDefault`=`noVnc`) and only VNC console type is supported for that VM.
This case is currently working for webadmin and I think that it should work the same for web-ui if we supports a default console now.
 
> Previous version and Web Admin provide a single button with black-box style logic (no indication what console will be chosen by the algorithm).

Not accurate. On webadmin you can currently choose the actual console type per VM ,but the default console (console button) is based on two things: the supported consoles and the global config ClientModeVncDefault value. So in case the user sets `ClientModeVncDefault`=`noVnc` and if vnc is the only client supported for a VM, this will be the auto default console opened while clicking on console button via webadmin.

> Due to that there is only one such default (we don't have secondary default for cases like ClientModeConsoleDefault=spice, ClientModeVncDefault=NoVnc )

We don't need to support a secondary default. But in a very specific case in which:
1. the VM is configured to run with only VNC console type
2. ClientModeVncDefault=noVnc
3. for the new console options settings: the global default is not RDP for Win VMs or VNC)
 
=> then the default console which appears as the first on list should be noVnc and not VNC. 

> Of course we could provide fallback but things will get complex i.e. imagine the case where:
> 
>     ClientModeConsoleDefault=spice, ClientModeVncDefault=NoVnc
>     preferred console is RDP
>     VM is Linux with VNC only
> 
in that case, the default should be NoVnc and not VNC. 
The same goes If VM is Linux with Spice only then Spice should be the default even if ClientModeConsoleDefault=vnc, ClientModeVncDefault=NoVnc and preferred console is RDP.

rszwajko:
@sgratch 
> It shouldn't be too complicated to fix for that specific scenario.
>

Of course we can address this specific scenario. Please check the newest commit.

sgratch (commented):
The code now LGTM.
But there are performance test results that still needed to be examined before merging. They are discussed via mails.

sgratch:
@rszwajko please rebase

sgratch:
The performance test results look ok so we can merge the PR.

PR #1422, opened 2021-04-30
Switch formatDateFromNow() to use moment.humanize()

rszwajko:
image


ovirt-infra:

sjd78:
ci test

src/helpers.js
```diff
@@ -250,23 +251,10 @@ export function getFormatedDateTime (timestamp) {
...
-    suffix = suffixes[currentIndex]
-    elapsed = Math.floor(elapsed / divitions[currentIndex])
-    currentIndex++
-  } while (divitions[currentIndex] <= elapsed && currentIndex < suffixes.length)
-
-  return msg.timeAgo({ time: `${elapsed}${suffix}` })
+export function formatDateFromNow (date, msg, now = new Date()) {
+  // 'now' parameter allows reliable testing
+  // uses standard moment.relativeTimeThreshold values
+  return moment.duration(moment(date).diff(moment(now))).humanize(true)
```
sjd78, line 257: Why not just use `formatUptimeDuration` in `src/utils/format.js`?  Then the uptime string and the age of snapshot would all follow the same format on the VM Details page.

rszwajko: My assumption was that we want to keep single number "humanized" format.

sjd78: Historically, there was a bunch of debate over the format of the text for how long ago type of things.  The debate around the uptime one was annoyingly long.

So at this point, as long as `humanize()` has localized text provided by moment, then we're good.  If not, we'll need to drop back to the `moment-duration-format` and our own localization strings.

sjd78: @rszwajko , @sgratch , @hstastna - As long as the humanized string are translated, I don't mind using it instead of a more uptime string format.

src/helpers.js
```diff
@@ -250,23 +251,10 @@ export function getFormatedDateTime (timestamp) {
...
-
-  let currentIndex = 0
-  do {
-    suffix = suffixes[currentIndex]
-    elapsed = Math.floor(elapsed / divitions[currentIndex])
-    currentIndex++
-  } while (divitions[currentIndex] <= elapsed && currentIndex < suffixes.length)
-
-  return msg.timeAgo({ time: `${elapsed}${suffix}` })
+export function formatDateFromNow (date, msg, now = new Date()) {
```
sjd78, line 254: A few nit-picks:

  - I suggest moving this function to `src/utils/format.js` to be in the same place as `formatUptimeDuration`.

  - Maybe a slightly different name?  `formatHowLongAgo`

  - `msg` isn't needed since `humanize(true)` handles that for us

  - `s/date/pastDate/` maybe?

sjd78 (changes requested):

rszwajko:
@sjd78 
1. implemented suggested changes
2. (successfully) tested that 'moment' supports all languages required by ovirt

sjd78 (approved):
LGTM

sgratch:
@rszwajko for some reason both pt_BR and zh_CN locales are not translated by 'moment' for me for snapshots duration. Please approve that both are working as expected for snapshots before merging

rszwajko:
> @rszwajko for some reason both pt_BR and zh_CN locales are not translated by 'moment' for me for snapshots duration. Please approve that both are working as expected for snapshots before merging

@sgratch  @sjd78 
This is regression after #1433. It seems that adding content to file `src/intl/translated-time-durations.json` disabled some translations.

sjd78:
> > @rszwajko for some reason both pt_BR and zh_CN locales are not translated by 'moment' for me for snapshots duration. Please approve that both are working as expected for snapshots before merging
> 
> @sgratch @sjd78
> This is regression after #1433. It seems that adding content to file `src/intl/translated-time-durations.json` disabled some translations.

Really?  This code is what extends the moment locale to add our translations for moment-duration-format which does not ship with any translations:
https://github.com/oVirt/ovirt-web-ui/blob/04f3a7e10ba4fa93946a84b38f61a574850a8eb9/src/intl/initialize.js#L82-L137

Does humanize() use the same keys?  I thought that `moment-duration-format` used keys unique to the plugin.

moment locale customization: https://momentjs.com/docs/#/customization/

sgratch:
@rszwajko @sjd78 

> Really? This code is what extends the moment locale to add our translations for moment-duration-format which does not ship with any translations

It seems that for fixing this we just need to rename translated-time-durations.json locale entries as follows (it's case sensitive):
pt-BR":  => pt-br":
"zh-CN": => "zh-cn":

rszwajko:
> It seems that for fixing this we just need to rename translated-time-durations.json locale entries as follows (it's case sensitive):
> pt-BR": => pt-br":
> "zh-CN": => "zh-cn":

Actually the upper-case version seems correct to me. Let's limit the scope of the changes to moment initialization.
Please check PR #1436

sgratch (approved):
LGTM

isaranova:
LGTM

Issue #1406, opened 2021-04-19
convert all API Boolean attributes used by "toApi" functions to our new ApiBooleanType


sgratch:
As triggered and mentioned by https://github.com/oVirt/ovirt-web-ui/pull/1386#discussion_r614455764, we should go over all "toApi" functions and and convert API Boolean attributes handled on web-ui to the new ApiBooleanType type (as done for persist_memorystate in PR # 1386).

PR #1423, opened 2021-05-02
Change boolean into "true"/"false" strings in toApi functions to adjust to API interface.

bamsalem:
> As triggered and mentioned by #1386 (comment), we should go over all "toApi" functions and and convert API Boolean attributes handled on web-ui to the new ApiBooleanType type (as done for persist_memorystate in PR # 1386).

In this PR I changed all the boolean types (that I found) in `toApi` functions to adjust to the API interface.
Fixes: #1406 .

### DIFF
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -281,7 +281,7 @@ const VM = {
       bios: vm.hasOwnProperty('bootMenuEnabled')
         ? {
           boot_menu: {
-            enabled: vm.bootMenuEnabled,
+            enabled: toApiBoolean(vm.bootMenuEnabled),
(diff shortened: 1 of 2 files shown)


sjd78 (changes requested):
You should also adjust the related types.

sjd78 (changes requested):
I only saw one other potentials:
  - vm.bootMenuEnabled

The type work looks good.  Going forward, if the type is just an Object, then creating it from the api docs is great but not strictly necessary.  For examples, I wouldn't worry about dealing with the `ApiVmType`...

sjd78 (approved):
LGTM

Issue #1390, opened 2021-03-24
user settings - reset option back to default/installation settings is missing


sgratch:
Original User Settings design is available here ->https://docs.google.com/document/d/1apA6E4zT21JSJgeO2V3Q_LhK9xq4SOwNeD-gIsG5pNY/edit?pli=1#heading=h.mkd1w7uwr9hs

For more details, check https://github.com/oVirt/ovirt-web-ui/pull/1088#issuecomment-734439359 (issue 12)

rszwajko:
**Original design**

https://marvelapp.com/prototype/94h0g56/screen/59286966

IMHO we can skip the message next to the button ("Default settings applied") and just disable the button if there is nothing to reset (defaults are used).
Adding the label explaining the state might be required  when/if the settings hierarchy would get complex but right now it seem not necessary.

PR #1424, opened 2021-05-05
Add option to reset global user settings to default/installation settings

bamsalem:
This PR Resolves: #1390 .
In this PR I added the option to reset the Global settings to the default/installation settings.

The default values (for now) are locale, showNotifications and refreshInterval, in the future we should support also console options.
I added new button on the toolbar left side - "Reset Settings":
image.
The button is disabled if the `curruntValues` are equals to the default.
I used the save mechanism so apply the changes and to prevent from the save button activation I added a flag `isReset` that helps us to identify if the current changes are reset or normal saving.

image

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -12,6 +12,7 @@ const AppConfiguration = {
   pageLimit: 20,
   schedulerFixedDelayInSeconds: 60,
   notificationSnoozeDurationInMinutes: 10,
+  showNotificationsDefault: true,
 
   consoleClientResourcesURL: 'https://www.ovirt.org/documentation/admin-guide/virt/console-client-resources/',
   cockpitPort: '9090',
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -69,6 +69,7 @@ export const messages: { [messageId: string]: MessageType } = {
   cd: 'CD',
   cdCanOnlyChangeWhenVmRunning: 'CD can only be changed when the VM is running.',
   cdromBoot: 'CD-ROM',
+  changesResetSuccessfully: 'Settings have been successfully reset',
   changesSavedSuccesfully: {
     message: 'Changes to settings saved succesfully!',
     description: 'Message displayed when all user settings have been saved successfully',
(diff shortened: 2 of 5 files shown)


src/config.js
```diff
@@ -12,6 +12,8 @@ const AppConfiguration = {
   pageLimit: 20,
   schedulerFixedDelayInSeconds: 60,
   notificationSnoozeDurationInMinutes: 10,
+  defaultLanguage: 'en',
```
rszwajko, line 15: default language is defined in `_/intl` -  we can use it directly (no need to create prop here) i.e check https://github.com/oVirt/ovirt-web-ui/pull/1415

src/intl/messages.js
```diff
@@ -69,6 +69,7 @@ export const messages: { [messageId: string]: MessageType } = {
   cd: 'CD',
   cdCanOnlyChangeWhenVmRunning: 'CD can only be changed when the VM is running.',
   cdromBoot: 'CD-ROM',
+  ChangeSuccessfullyReset: 'Settings have been successfully reset',
```
rszwajko, line 72: typo - not camel case

rszwajko (changes requested):
1. I would recommend enabling reset only if there is no  pending user changes - otherwise user input will get lost. Perhaps we could then use this info instead of reset flag.
2. confirmation message is missing - we have translated messages for it.

sjd78 (approved):
LGTM

sgratch (commented):
@bamsalem any reason for avoid resetting the notificationSnoozeDuration back to its init default  value of AppConfiguration.notificationSnoozeDurationInMinutes =10 ?
I think that this option should also be marked as default (a follow up for https://github.com/oVirt/ovirt-web-ui/pull/1415). 
Other than that, looks good.

bamsalem:
> @bamsalem any reason for avoid resetting the notificationSnoozeDuration back to its init default value of AppConfiguration.notificationSnoozeDurationInMinutes =10 ?
> I think that this option should also be marked as default (a follow up for #1415).
> Other than that, looks good.

Done.

sgratch:
\

> > @bamsalem any reason for avoid resetting the notificationSnoozeDuration back to its init default value of AppConfiguration.notificationSnoozeDurationInMinutes =10 ?

+1

> > I think that this option should also be marked as default (a follow up for #1415).

Still need to handle this but we can open a separate issue for that.

> > Other than that, looks good.
> 
> Done.

sgratch (approved):

isaranova:
LGTM

PR #1428, opened 2021-05-13
Add persistLanguage setting to web-ui account settings

bamsalem:
This PR adds new setting to web ui account settings.
This settings enable users to decide if they want to persist the language in the backend and then no matter which device they use the web-ui it always use the same language.
If they decide not to persist language The web ui will use the query string locale parameter (that sent from the login page via url).
If there is no value from the qs the system will use the persisted value, and if there is no persisted value it will stay the same as before this PR.
Resolves: #1393 .
image


rszwajko (changes requested):

bamsalem:
@sgratch 
> IMHO you don't need any changes here. The locale discovered by our intl scripts will be used as Redux store default. If `localeFromUrl` was detected then it will be used automatically (because locale from local storage should be cleared after disabling persistence)

Should we remove the locale property from the backend and local storage or just ignore it in case of disabling the `persistLocale` property?
In our last conversation we talked about ignoring the backend property...

src/components/UserSettings/GlobalSettings.js
```diff
@@ -5,7 +5,7 @@ import { connect } from 'react-redux'
 import { push } from 'connected-react-router'
 import { saveGlobalOptions } from '_/actions'
 import { FormControl, Switch } from 'patternfly-react'
-import { withMsg, localeWithFullName, DEFAULT_LOCALE } from '_/intl'
+import { withMsg, localeWithFullName, locale, DEFAULT_LOCALE } from '_/intl'
```
rszwajko, line 8: `locale` should not be accessed directly (similar to `msg` object) - both are provided by `withMsg` HOC (check other usages in the code)

src/components/UserSettings/GlobalSettings.js
```diff
@@ -145,7 +145,8 @@ class GlobalSettings extends Component {
 
   buildSections (onChange, translatedLabels) {
     const { draftValues } = this.state
-    const { config, msg } = this.props
+    const { config, msg, currentValues: { persistLocale, language } } = this.props
```
rszwajko, line 148: IMHO `persistLocale` should be read (only) from `draftValues` to allow updating both language and persist  in one save.

src/components/UserSettings/GlobalSettings.js
```diff
@@ -189,6 +191,26 @@ class GlobalSettings extends Component {
...
+            title: msg.persistLanguage(),
+            tooltip: msg.persistLanguageTooltip(),
+            body: (<Switch
+              id={`${idPrefix}-persist-language`}
+              bsSize='normal'
+              title='normal'
+              value={draftValues.persistLocale}
+              onChange={(e, persist) => {
+                onChange('persistLocale')(persist)
+                if (persist === false) {
```
rszwajko, line 207: This seems unnecessary to me. Please check how `notificationSnoozeDuration` is disabled  based on `showNotifications` flag.

src/components/UserSettings/GlobalSettings.js
```diff
@@ -298,10 +321,11 @@ export default connect(
     },
     currentValues: {
       sshKey: options.getIn(['ssh', 'key']),
-      language: options.getIn(['remoteOptions', 'locale', 'content']),
+      language: options.getIn(['remoteOptions', 'locale', 'content']) || locale,
```
rszwajko, line 324: the default value should be provided via the Redux store

src/intl/LocaleReloader.js
```diff
@@ -3,27 +3,35 @@ import PropTypes from 'prop-types'
...
+   * 1. locale provided by URL (<server>/?locale=en_US) and cookie is used if the persistLanguage value is equals to false.
    * 2. no data from the server (yet) - UI will try to gues user locale. When user settings will be fetched
    *    from the server and incorrect locale was guessed then locale will get hot-reloaded.
    * 3. no property on server exists but UI was launched using non-default locale(i.e. from URL) - save that locale on the server.
    */
   useEffect(() => {
-    if (localeFromStore !== locale) {
-      console.warn(`reload due to locale change: ${locale} -> ${localeFromStore}`)
-      reloadMsg(localeFromStore)
+    if (persistLocale) {
```
rszwajko, line 23: If MsgContext is correctly initialized then this should not be necessary. The correct locales should flow down from GlobalErrorBoundary.

src/reducers/options.js
```diff
@@ -18,7 +16,11 @@ const defaultOptions: UserOptionsType = {
   remoteOptions: {
     locale: {
       id: undefined,
-      content: locale,
+      content: undefined,
```
rszwajko, line 19: the store should always contain meaningful data

src/sagas/options.js
```diff
@@ -88,7 +87,7 @@ function* fetchUserOptions (action: Object): any {
 
   yield put(A.loadUserOptions(remoteOptions))
 
-  if (!remoteOptions.locale) {
+  if (!remoteOptions.locale && remoteOptions.persistLocale && remoteOptions.persistLocale.content) {
```
rszwajko, line 90: we should skip import only if  `remoteOptions.persistLocale === false`

src/sagas/options.js
```diff
@@ -147,10 +146,11 @@ function* saveRemoteOption ([ name, value ]: any): any | ResultType {
     change: name })
 }
 
-function* saveGlobalOptions ({ payload: { sshKey, showNotifications, notificationSnoozeDuration, language, refreshInterval }, meta: { transactionId } }: SaveGlobalOptionsActionType): Generator<any, any, any> {
-  const { ssh, locale, refresh } = yield all({
+function* saveGlobalOptions ({ payload: { sshKey, showNotifications, notificationSnoozeDuration, language, refreshInterval, persistLocale }, meta: { transactionId } }: SaveGlobalOptionsActionType): Generator<any, any, any> {
+  const { ssh, locale, refresh, persistLoc } = yield all({
     ssh: call(saveSSHKey, ...Object.entries({ sshKey })),
-    locale: call(saveRemoteOption, ...Object.entries({ locale: language })),
+    locale: (persistLocale || persistLocale === undefined) ? call(saveRemoteOption, ...Object.entries({ locale: language })) : {},
```
rszwajko, line 152: OK. It seems we need to check also the current value in the store (via `select()`) if persist flag was not changed.

src/sagas/options.js
```diff
@@ -188,6 +199,22 @@ function* saveGlobalOptions ({ payload: { sshKey, showNotifications, notificatio
...
+  if (optionId) {
+    const userId = yield (select(({ config }) => config.getIn(['user', 'id'])))
+    yield call(
+      callExternalAction,
+      'persistUserOption',
+      Api.deleteUserOption,
+      A.deleteUserOption({ optionId, userId }),
+      true,
+    )
+    yield call(fetchUserOptions, A.fetchUserOptions({ userId }))
```
rszwajko, line 215: why we need to fetch all options here?

src/sagas/options.js
```diff
@@ -162,7 +162,18 @@ function* saveGlobalOptions ({ payload: { sshKey, showNotifications, notificatio
...
+
+  if (!persistLoc.error && persistLoc.change && !persistLoc.sameAsCurrent) {
+    const { name, value } = persistLoc.data
+    yield put(A.setOption({ key: [ 'remoteOptions', name ], value }))
+
+    if (value && value.content === false) {
+      yield call(deleteRemoteOption, { optionName: 'locale' })
+    } else if (value && value.content === true) {
+      const userId = yield (select(({ config }) => config.getIn(['user', 'id'])))
+      yield call(fetchUserOptions, A.fetchUserOptions({ userId }))
```
rszwajko, line 175: why we cannot just call `yield call(exportInferredLocale)` ?

src/sagas/options.js
```diff
@@ -188,6 +199,22 @@ function* saveGlobalOptions ({ payload: { sshKey, showNotifications, notificatio
...
 }
+function* deleteRemoteOption ({ optionName, optionId }: Object): any {
+  if (!optionId && optionName) {
+    optionId = yield (select(state => state.options.getIn(['remoteOptions', optionName, 'id'])))
+  }
+  if (optionId) {
+    const userId = yield (select(({ config }) => config.getIn(['user', 'id'])))
+    yield call(
+      callExternalAction,
+      'persistUserOption',
```
rszwajko, line 210: `methodName` should be deleteUserOption

rszwajko (dismissed):

src/components/UserSettings/GlobalSettings.js
bamsalem: It's disable the `notificationSnoozeDuration` but it will also change the value in the backend, so IMO it's different case...

src/intl/LocaleReloader.js
bamsalem: there is an edge case when this component initialized before the redux store and we got undefined: as value:
image

bamsalem:
> @bamsalem
> Few issues/comments:
> 
> Reset option issues (now that it's merged):
> 
> * Does persist have a default value and should be reset as part of reset settings?
> * The Language menu option is still supporting reset option even if persist is disabled - and once clicking on reset, remote locale is saved/persisted. We need to ignore locale in case of persist=false or always set locale to default once setting persist=false.
> 

IMHO Persist Language should have a default value and it should be true, so we should not ignore the language on reset.

> Others:
> 
> * When enabling the persist language before the save, the language menu is still disabled.

Fixed.
> * when disabling the persist locale and saving it, the save and reset settings buttons are both enabled.

I cant' reproduce it van you please review it after the latest changes and let me know if it's still reproduced?
> * When login for the first time after restart with persist = false, the language menu displays current selected language (why needed?) and language is saved/persisted.

After adding default value for persistLocale it should store the data because it should persist the locale and now I changed this code so it should not happen: https://github.com/bamsalem/ovirt-web-ui/blob/persist-language-task/src/sagas/options.js#L149-L155
> * I missed that part which we decided to avoid  saveLocaleToLocalStorage for both modes (persisted=true/fasle). I thought we remove local and remote storage only in case of persist=false.

we decided to avoid using local storage because it should be used only for error to fetch the locales from the backend,
in any other case it will replace the value from localStorage with the backend value, and can make confusion.

sgratch:
@bamsalem 
- Please fix the main comment based on up to date design

- After playing a bit with persist=false and true, an extra "do not disturb" field appears. For reproducing, enable the persist locale and change the language and then apply (save) both:
image

 
> > Reset option issues (now that it's merged):
> > 
> > * Does persist have a default value and should be reset as part of reset settings?
> > * The Language menu option is still supporting reset option even if persist is disabled - and once clicking on reset, remote locale is saved/persisted. We need to ignore locale in case of persist=false or always set locale to default once setting persist=false.
> 
> IMHO Persist Language should have a default value and it should be true, so we should not ignore the language on reset.

Now when clicking on reset while persist=false, the actual used locale is changed to the one set via the url parameter and not the default English:
image

 
> > Others:
> > 
> > * When enabling the persist language before the save, the language menu is still disabled.
> 
> Fixed.

+1, looks ok now

> 
> > * when disabling the persist locale and saving it, the save and reset settings buttons are both enabled.
> 
> I cant' reproduce it van you please review it after the latest changes and let me know if it's still reproduced?
> 

It is still reproduced (mainly after reset), trying to record such a scenario.


> > * When login for the first time after restart with persist = false, the language menu displays current selected language (why needed?) and language is saved/persisted.
> 
> After adding default value for persistLocale it should store the data because it should persist the locale and now I changed this code so it should not happen: https://github.com/bamsalem/ovirt-web-ui/blob/persist-language-task/src/sagas/options.js#L149-L155
> 

+1, looks ok now

> > * I missed that part which we decided to avoid  saveLocaleToLocalStorage for both modes (persisted=true/fasle). I thought we remove local and remote storage only in case of persist=false.
> 
> we decided to avoid using local storage because it should be used only for error to fetch the locales from the backend,
> in any other case it will replace the value from localStorage with the backend value, and can make confusion.

But how we'll handle errors on fetching backend data in current case without the local storage?
Anyway, in case of a browser running 4.4.6 and then upgrading to 4.4.7, the local storage locale is never deleted.

rszwajko:
Changes:
1. keep the language enabled when persistence is disabled. Reasons:
   1. there is no link between server side persistence and ability to 
     change language in general. Merging both functionalities requires
     extra effort to explain this to the user i.e. property should be 
     named similar to "Persist and enable language change".
   2. leaving language active solves some corner cases i.e. when user
     changed the language and toggled persistence flag at the same time
2. explicitly delete "options" property from local storage to clean-up
  unused data
3.  export inferred locale asynchronously to prevent overlapping
      with current saveGlobalOptions saga

sgratch (dismissed):
@rszwajko 
Few comments and (didn't review all yet):
> keep the language enabled when  persistence is disabled

In that case we need to indicate that changing the language is somehow not persistent otherwise it's confusing: when persist=false and language is changed via settings, the message appears is: "Changes to settings saved succesfully!". Worth changing to something that indicates that it's not persisted.


In addition I encountered 3 wrong scenarios:
1. When setting persist=false and changing the language at one save, the previous locale property is removed and persistLocale is set to false but the new locale property for the updated new language is created - which leads to strange behvious
2. When locale is set to English default and persist=false and then reset settings is done, the English locale new property is not created as it should, only persistLocale is updated to true.
3. When login with persist=false and locale used by url and then set "reset options", only the vmPortal.persistLocale is set to true while locale property=English  is not created as part of reset settings. This leads to strange behaviors.

rszwajko:
> > keep the language enabled when  persistence is disabled
> 
> In that case we need to indicate that changing the language is somehow not persistent otherwise it's confusing
>

It's the same case as with notifications. The locale is saved in the Redux store but not persisted on the server. 

> In addition I encountered 3 wrong scenarios:
> 

It seems that all 3 were caused by the same issue -  locale being set to quickly (before `persistLocale`).

sgratch:
> > > keep the language enabled when  persistence is disabled
> > 
> > 
> > In that case we need to indicate that changing the language is somehow not persistent otherwise it's confusing
> 
> It's the same case as with notifications. The locale is saved in the Redux store but not persisted on the server.

Yes and no. Technically it is the same. But from user perspective, changing the notification is always non persistent. But when changing the language, it depends on the language-persist settings which is on a different section so FMPOV it is confusing. If we always enable changing the locale, we somehow need to indicate if it's going to be persistent or not without the need to check the advanced section every time. 

> 
> > In addition I encountered 3 wrong scenarios:
> 
> It seems that all 3 were caused by the same issue - locale being set to quickly (before `persistLocale`).

Issue 1 is solved but 2-3 are still reproduced. We need to decide if this is the correct design: after reset settings is applied, should we treat it as after 1st login (url locale might be enforced) or should we store the default English locale and avoid setting the locale read from url?


- A new issue found:
Reading the following locales from url fails: ?locale=pt_BR, ?locale=zh_CN

rszwajko:
> > It's the same case as with notifications. The locale is saved in the Redux store but not persisted on the server.
> 
> Yes and no. Technically it is the same. But from user perspective, changing the notification is always non persistent. But when changing the language, it depends on the language-persist settings which is on a different section so FMPOV it is confusing. If we always enable changing the locale, we somehow need to indicate if it's going to be persistent or not without the need to check the advanced section every time.

What solution do you suggest?


> -A new issue found:
> Reading the following locales from url fails: ?locale=pt_BR, ?locale=zh_CN

This is not a regression (checked on commit https://github.com/oVirt/ovirt-web-ui/commit/c4b08d262f71283f4c1e1a5c3400ad5658eb635d)

rszwajko:
@sgratch 
>> 1. When locale is set to English default and persist=false and then reset settings is done, the English locale new property is not created as it should, only persistLocale is updated to true.
>> 2. When login with persist=false and locale used by url and then set "reset options", only the vmPortal.persistLocale is set to true while locale property=English is not created as part of reset settings. This leads to strange behaviors.
>>
> Issue 1 is solved but 2-3 are still reproduced.

Hopefully fixed (this time). 
I was able to reproduce scenario 1 in a very specific use case only:
1. start with locale stored on the server side
2. disable persistence (this used to delete the option on the server but did not clear the ID in Redux)
3. during the same browser session reset options - this triggers regular save so existing ID was used
4. saving `locale` property failed due to invalid ID

After scenario 1 was fixed I was no longer able to reproduce scenario 2.

>  We need to decide if this is the correct design: after reset settings is applied, should we treat it as after 1st login (url locale might be enforced) or should we store the default English locale and avoid setting the locale read from url?
>

IMHO default state should be the same for all scenarios so English locale should be used.

sgratch:
@rszwajko 

> > Yes and no. Technically it is the same. But from user perspective, changing the notification is always non persistent. But when changing the language, it depends on the language-persist settings which is on a different section so FMPOV it is confusing. If we always enable changing the locale, we somehow need to indicate if it's going to be persistent or not without the need to check the advanced section every time.
> 
> What solution do you suggest?

We can either add a tooltip in case the persist=false for notifying the user or maybe just add a successful operation message instead of "Changes to settings saved successfully!" for language only, e,g "Changes to language setting is not saved in the server"

> 
> > -A new issue found:
> > Reading the following locales from url fails: ?locale=pt_BR, ?locale=zh_CN
> 
> This is not a regression (checked on commit [c4b08d2](https://github.com/oVirt/ovirt-web-ui/commit/c4b08d262f71283f4c1e1a5c3400ad5658eb635d))

Correct, let's open a separate issue for this.

>     issue 1 is solved but 2-3 are still reproduced.
> 
> Hopefully fixed (this time).

LGTM

> IMHO default state should be the same for all scenarios so English locale should be used.

+1

rszwajko:
@sgratch 
> We can either add a tooltip in case the persist=false for notifying the user
>

Done. Please check the screenshot below.

image

src/intl/messages.js
```diff
@@ -512,12 +512,20 @@ export const messages: { [messageId: string]: MessageType } = {
   operatingSystemInstalledOnVm: 'Operating system installed on the virtual machine.',
   optimizedFor: 'Optimized For',
   optionalUserDescriptionOfVm: 'Optional user description of the virtual machine.',
+  optionIsNotSavedOnTheServer: {
+    message: 'This option is not saved on the server. {persistenceReEnableHowTo}',
+    description: 'Inform the user that server-side persistence is not enabled for this option. If it can be re-enabled then provide translated instructions (parameter persistenceReEnableHowTo). Otherwise leave it empty.',
```
sgratch, line 517: @rszwajko Why did you split the message into 2 parts `optionIsNotSavedOnTheServer` and  `persistenceReEnableHowTo` and what do you mean by "If it can be re-enabled then provide translated instructions". Is there a use case in which it can't be re-enabled?

rszwajko: I would like to re-use the `optionIsNotSavedOnTheServer` message for all future options and some of them might be local only (i.e. like notifications today).

sgratch (approved):
LGTM

isaranova:
LGTM

PR #1435, opened 2021-05-26
Avoid using translated strings in  "key" property

rszwajko:
React uses "key" property to identify items in collections and reports
error if the assigned value is not unique.
After translation some fields may share the same name which violates
this requirement.

Steps to recreate:
1. change language to Japanese
2. go to Account Settings -> Notifications
3. note that both `Do not disturb` and `Do not disturb for` share the same translation
4. check console for errors similar to:
```
Encountered two children with the same key, `%s`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.%s 妨害しない 
    in Section (created by SettingsBase)
    in div (created by SettingsBase)
```
5. change notification settings and save - confirmation dialog will be displayed
6. note that listed changed properties are the same 
7. check console for errors similar to:
```
Encountered two children with the same key, `%s`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.%s settings_toolbar_li_妨害しない 
    in ul (created by SettingsToolbar)
    in div (created by ConfirmationModal)
```

### DIFF
--- a/src/components/Settings/SettingsBase.js
+++ b/src/components/Settings/SettingsBase.js
@@ -44,8 +44,8 @@ const Section = ({ name, section }) => (
       { section.tooltip && <InfoTooltip id={`${name}-info-tooltip`} tooltip={section.tooltip} /> }
     </h3>
     { section.fields.map((field) => (
-      <FormGroup key={field.title} className={style['settings-field']}>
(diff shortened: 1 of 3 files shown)


sgratch (approved):
LGTM

PR #1436, opened 2021-05-26
Use lower case locale for updating moment

rszwajko:
Internally moment.js seems to use lower case identifiers for locales.
Standard locale format (i.e. pt-BR)  works for initialization but
not for updating built-in translations.

Check #1422  for details how to re-create.


sgratch (approved):
LGTM

Issue #1438, opened 2021-06-01
Group permissions don't work in ovirt-web-ui-1.6.9-1


opersh:
Description of problem: group permissions don't work in ovirt-web-ui-1.6.9-1

VM Portal version number: 1.6.9-1

oVirt version number: 4.4.6

Browser and version: N/A

Steps to Reproduce:
1. Using webadmin / Admin Portal create UserA and GroupA, add UserA to GroupA
2. Granted permission "VmCreate" to the group GroupA
3. Add GroupA to cluster or host permissions 
4. Open web-ui / VM Portal as UserA

Actual results:  No **Create VM** button is displayed

Expected results:  The **Create VM** button is displayed

If I granted permission "VmCreate" to the UserA and add UserA to cluster or host permissions, then **Create VM** button is displayed.

If I install ovirt-web-ui-1.6.8-1, then group permissions work and **Create VM** button is displayed.

sjd78:
Ok, so this is frustrating.  The `/ovirt-engine/api/users/<user_id>/groups` returns the "DomainUserGroups" which is technically different than regular engine "Groups".  The "DomainUserGroups" don't necessarily exist in the ovirt-engine.  So as a mechanism to lookup a user's ovirt groups, this endpoint is wrong.

For example. on my dev engine, I use the `ovirt-aaa-jdbc-tool` tool to:
  - Create 2 groups, "GroupFoxtrot" and "GroupEmpty"
  - Create user "UserFoxtrot"
  - Add user "UserFoxtrot" to "GroupFoxtrot"

Via the api, I queried a bunch of things (links and actions removed for space)

The user:
```sh
$ curl -k -u"UserFoxtrot@internal:user" -H"Accept: application/json" -H"Filter:true" "https://engine-dev.ovirt:8443/ovirt-engine/api/users/f986fee3-0254-468a-9ce2-e46b1c80c1c8" | jq '.'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1526    0  1526    0     0  10380      0 --:--:-- --:--:-- --:--:-- 10310
{
  "department": "",
  "domain_entry_id": "37646238383365632D646132332D343937392D386238632D666634633766643335303235",
  "email": "",
  "last_name": "User",
  "namespace": "*",
  "principal": "UserFoxtrot",
  "user_name": "UserFoxtrot@internal-authz",
  "user_options": {
[...]

PR #1441, opened 2021-06-04
Cross reference user domain groups with ovirt groups

sjd78:
At login, we want to know the uuids of the ovirt added domain user groups for the logged in user.  Group membership drives permission look ups.  Since the `/api/users/<userid>/groups` returns the domain user groups, additional work is required to figure out which of those groups are ovirt groups.

Now at login, the domain groups are cross referenced with the ovirt know user groups (i.e. domain groups that have been added
to ovirt in the Administration > Users > Group panels).  This result of the cross reference is a list of the ovirt uuids of the user groups the user is a member of.  With the ovirt uuids the permission calculations used to enable/disable functions of the app will work correctly.

Resolves: #1438

The group id problem surfaced as a result of the fix for issue #1411, PR #1412.

### DIFF
--- a/src/sagas/base-data.js
+++ b/src/sagas/base-data.js
@@ -1,5 +1,5 @@
 import Api, { Transforms } from '_/ovirtapi'
-import { put, select, takeLatest } from 'redux-saga/effects'
+import { put, select, takeLatest, all, call } from 'redux-saga/effects'
 import {
   canUserUseTemplate,
   canUserUseCluster,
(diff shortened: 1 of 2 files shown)


sgratch (commented):
The code fix LGTM.

There is an issue that I'm not sure if related to this fixor not,  in which for user's ovirt groups > 1 then only permission roles from the first group are handled.

src/sagas/base-data.js
```diff
@@ -108,16 +108,61 @@ export function* fetchAllVnicProfiles (action) {
   }
 }
 
+/**
+ * Fetch the user's ovirt known groups.  To do so, we need to fetch and cross reference the
+ * users's domain groups with the ovirt known groups.  Users can have more domain groups
+ * than ovirt groups.  The cross reference will turn domain groups membership into ovirt
+ * group uuids.  Permission checks require ovirt group uuids, not the domain entity ids.
+ */
```
rszwajko, line 116: > Users can have more domainGroups then ovirt groups
>

The opposite situation happened during my tests:
1. `/api/groups` returns 3 entries i.e. `Everyone`, `ovirt-administrator`
2. `/api/users/{id}/groups` returns empty array

Is this behavior correct?

src/sagas/base-data.js
```diff
@@ -108,16 +108,61 @@ export function* fetchAllVnicProfiles (action) {
...
-    groupsInternal.push(EVERYONE_GROUP_ID)
+  const {
+    domainGroups,
+    ovirtGroups,
+  } = yield all({
+    domainGroups: call(function* (userId) {
+      const domainGroups = []
+
+      const dgs = yield callExternalAction('userDomainGroups', Api.userDomainGroups, { payload: { userId } })
+      if (dgs && dgs.group) {
```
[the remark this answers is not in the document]
sjd78: Makes sense.

Although - it is possible to over use destructuring with defaults....

src/sagas/base-data.js
sjd78: Well, the `/api/groups` will return all of the groups that the user has access to see.  So yeah, if your user is an admin user, they can see all of the groups but don't need to actually belong to any.  I'll update the comment a bit.

sjd78:
> There is an issue that I'm not sure if related to this fixor not, in which for user's ovirt groups > 1 then only permission roles from the first group are handled.

So the bug is when a user belongs to >1 group and the entity has different permissions for each group?  The function that turns permissions into permits will iterate through all of the permissions returned by the entity fetch and map the matched permissions to the role to the permits.

Is there an issue or BZ opened on it yet?

rszwajko (approved):

hstastna (approved):
LGTM

sgratch:
> > There is an issue that I'm not sure if related to this fixor not, in which for user's ovirt groups > 1 then only permission roles from the first group are handled.
> 
> So the bug is when a user belongs to >1 group and the entity has different permissions for each group? The function that turns permissions into permits will iterate through all of the permissions returned by the entity fetch and map the matched permissions to the role to the permits.
> 
> Is there an issue or BZ opened on it yet?

After re-verifying, the problem is not related to user's groups permissions while belonging to more than one group.
The problem is that if a user is added to another oVirt group, the user will not be marked as a member of this group till next login (since fetchUserGroups is called only in login time).
Since this is part of the design and not a regression then it's ok and not considered an issue to handle.

sgratch (approved):
LGTM

isaranova:
LGTM

PR #1433, opened 2021-05-17
Translation update, 17-May-2021

sjd78:
Updated from https://zanata.phx.ovirt.org/iteration/view/ovirt-web-ui/1.7

Resolves: #1432

----
```console
> yarn intl:report                                                                                                                                   (zanata_May17|●2…3⚑13)
yarn run v1.22.5
$ babel-node scripts/intl/report.js --presets env,flow
English key count: 548

Untranslated keys per locale:
╔════════╤════════════════╤═════════════════════════════════════════╗
║ locale │ % untranslated │ keys                                    ║
╟────────┼────────────────┼─────────────────────────────────────────╢
║ cs     │ 72.3%          │ >20% of keys are untranslated [396/548] ║
╟────────┼────────────────┼─────────────────────────────────────────╢
║ it     │ 84.7%          │ >20% of keys are untranslated [464/548] ║
╚════════╧════════════════╧═════════════════════════════════════════╝


Multiple value count: 26
╔════════════════════════╤═════════════════════════════════════════╤═════════════════════════════════════════════╤═════════════════╗
║ Text                   │ Key                                     │ Locales With Key                            │ Locales w/o Key ║
╟────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────╢
║ Create Virtual Machine │ addNewVm                                │ 7: de, es, fr, ja, ko, pt-BR, zh-CN         │ 2: cs, it       ║
║                        │ createVmWizardButtonCreate              │ 7: de, es, fr, ja, ko, pt-BR, zh-CN         │ 2: cs, it       ║
╟────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────╢
║ Cancel                 │ cancel                                  │ 9: cs, de, es, fr, it, ja, ko, pt-BR, zh-CN │ 0:              ║
║                        │ createVmWizardButtonCancel              │ 9: cs, de, es, fr, it, ja, ko, pt-BR, zh-CN │ 0:              ║
╟────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────╢
║ Details                │ cardTitleDetails                        │ 7: de, es, fr, ja, ko, pt-BR, zh-CN         │ 2: cs, it       ║
║                        │ details                                 │ 7: de, es, fr, ja, ko, pt-BR, zh-CN         │ 2: cs, it       ║
[...]

Issue #1432, opened 2021-05-17
Update translations for ovirt 4.4.7 (ovirt-web-ui 1.7.0)


sjd78:
Update the translations for the project based on the 1.7 version in Zanata.

This correlates to the version of ovirt-web-ui that will be included in ovirt-4.4.7.

sjd78:
All related issues for the 17-May-2021 translation updates:
  - ovirt-web-ui: https://github.com/oVirt/ovirt-web-ui/issues/1432
  - ovirt-engine-ui-extensions: https://bugzilla.redhat.com/1961331
  - ovirt-engine: https://bugzilla.redhat.com/show_bug.cgi?id=1961396

PR #1443, opened 2021-06-08
Translation update, Japanese VM status string updates

sjd78:
Resolves: #1444 

This is a follow up of Issue #1432 and PR #1433


sgratch (approved):
LGTM

PR #208, opened 2017-05-31
Improve design of "Connect automatically"

mareklibra:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/207
Depends on: https://github.com/oVirt/ovirt-ui-components/pull/43

PR #1446, opened 2021-06-09
Console options

rszwajko:
**NOTE** that this PR depends on #1420  **AND**  #1456

### Add Console section to Account Settings

Added options (all options are persisted on the server):
1. preferredConsole - the console that will be the first console
   available to the user (if supported by the VM)
2. VNC options:
   a) full screen mode
   b) secure attention (mapping Ctrl+Alt+Del to Ctrl+Alt+End)
3. SPICE option
   a) full screen mode
   b) secure attention
   c) smartcard

Moved options:
1. SSH key moved from General section to Console - the key is only
  relevant for Serial Console access.

### Preserve the order and group options by section

Extracted to #1456 
    
### Apply custom Console settings
    
1. enhance vv files using the values configured via Account Settings
2. let preferredConsole override default console sent from the server

### DIFF
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -66,7 +66,7 @@ export function* downloadVmConsole (action) {
     if (isSpice || !isNoVNC) {
       const legacyOptions = getLegacyOptions({ vmId })
       const options = isSpice
-        ? yield getSpiceConsoleOptions({ legacyOptions, usbAutoshare, usbFilter })
+        ? yield getSpiceConsoleOptions({ legacyOptions, usbAutoshare, usbFilter, vmId })
         : yield getVncOptions({ legacyOptions })
 
       data = adjustVVFile({ data, options })
@@ -86,6 +86,12 @@ export function* downloadVmConsole (action) {
   }
 }
 
+/**
+ * Legacy options were saved in browser's local storage.
+ * The UI for setting theose options was removed in previous versions.
+ * However there is still a (small) chance that the data is still there.
+ * Note that legacy options are per VM and therefore should overwrite global defaults.
+ */
 function* getLegacyOptions ({ vmId }) {
   const options = yield select(state => state.options.getIn(['options', 'consoleOptions', vmId]))
   if (options) {
@@ -95,18 +101,22 @@ function* getLegacyOptions ({ vmId }) {
   yield getConsoleOptions(getConsoleOptionsAction({ vmId }))
 }
 
-function* getSpiceConsoleOptions ({ legacyOptions, usbAutoshare, usbFilter }) {
-  const newOptions = yield select(({ options }) => ({
+function* getSpiceConsoleOptions ({ legacyOptions, usbAutoshare, usbFilter, vmId }) {
+  const smartcardEnabledOnVm = yield select(({ vms }) => vms.getIn(['vms', vmId, 'display', 'smartcardEnabled']))
+  const newOptions = yield select(({ options, vms }) => ({
     fullscreen: options.getIn(['remoteOptions', 'fullScreenSpice', 'content']),
     ctrlAltDelToEnd: options.getIn(['remoteOptions', 'ctrlAltEndSpice', 'content']),
     smartcardEnabled: options.getIn(['remoteOptions', 'smartcardSpice', 'content']),
   }))
-  return {
+
+  const merged = {
     ...newOptions,
     ...legacyOptions,
     usbFilter,
     usbAutoshare,
   }
+  merged.smartcardEnabled = smartcardEnabledOnVm && merged.smartcardEnabled
+  return merged
 }
 
 function* getVncOptions ({ legacyOptions }) {
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -62,14 +62,14 @@ export function* downloadVmConsole (action) {
     /**
      *Download console if type is spice or novnc is running already
      */
-    if (data.indexOf('type=spice') > -1 || !isNoVNC) {
-      let options = yield select(state => state.options.getIn(['options', 'consoleOptions', vmId]))
-      if (!options) {
-        console.log('downloadVmConsole() console options not yet present, trying to load from local storage')
-        options = yield getConsoleOptions(getConsoleOptionsAction({ vmId }))
-      }
+    const isSpice = data.indexOf('type=spice') > -1
+    if (isSpice || !isNoVNC) {
+      const legacyOptions = getLegacyOptions({ vmId })
+      const options = isSpice
+        ? yield getSpiceConsoleOptions({ legacyOptions, usbAutoshare, usbFilter })
+        : yield getVncOptions({ legacyOptions })
 
-      data = adjustVVFile({ data, options, usbAutoshare, usbFilter })
+      data = adjustVVFile({ data, options })
       fileDownload({ data, fileName: `console.vv`, mimeType: 'application/x-virt-viewer' })
       yield put(setConsoleStatus({ vmId, status: DOWNLOAD_CONSOLE }))
     } else {
@@ -83,6 +83,40 @@ export function* downloadVmConsole (action) {
     if (openInPage || isNoVNC) {
       yield put(push('/vm/' + vmId + '/console/' + consoleId))
     }
+  }
+}
+
+function* getLegacyOptions ({ vmId }) {
+  const options = yield select(state => state.options.getIn(['options', 'consoleOptions', vmId]))
+  if (options) {
+    return (options.toJS && options.toJS()) || options
+  }
+  console.log('downloadVmConsole() console options not yet present, trying to load from local storage')
+  yield getConsoleOptions(getConsoleOptionsAction({ vmId }))
+}
+
+function* getSpiceConsoleOptions ({ legacyOptions, usbAutoshare, usbFilter }) {
+  const newOptions = yield select(({ options }) => ({
+    fullscreen: options.getIn(['remoteOptions', 'fullScreenSpice', 'content']),
+    ctrlAltDelToEnd: options.getIn(['remoteOptions', 'ctrlAltEndSpice', 'content']),
+    smartcardEnabled: options.getIn(['remoteOptions', 'smartcardSpice', 'content']),
+  }))
+  return {
+    ...newOptions,
+    ...legacyOptions,
+    usbFilter,
+    usbAutoshare,
+  }
+}
+
+function* getVncOptions ({ legacyOptions }) {
+  const newOptions = yield select(({ options }) => ({
+    fullscreen: options.getIn(['remoteOptions', 'fullScreenVnc', 'content']),
+    ctrlAltDelToEnd: options.getIn(['remoteOptions', 'ctrlAltEndVnc', 'content']),
+  }))
+  return {
+    ...newOptions,
+    ...legacyOptions,
   }
 }
 
--- a/src/sagas/console/vvFileUtils.js
+++ b/src/sagas/console/vvFileUtils.js
@@ -1,15 +1,12 @@
 
-export function adjustVVFile ({ data, options, usbAutoshare, usbFilter }) {
-  // __options__ can either be a plain JS object or ImmutableJS Map
-  console.log('adjustVVFile options:', options)
-
-  if (options && ((options.get && options.get('fullscreen')) || options.fullscreen)) {
+export function adjustVVFile ({ data = '', options: { fullscreen, ctrlAltDelToEnd, smartcardEnabled, usbAutoshare, usbFilter } = {} } = {}) {
+  if (fullscreen) {
     data = data.replace(/^fullscreen=0/mg, 'fullscreen=1')
   }
 
   const pattern = /^secure-attention=.*$/mg
   let text = 'secure-attention=ctrl+alt+del'
-  if (options && ((options.get && options.get('ctrlAltDelToEnd')) || options.ctrlAltDelToEnd)) {
+  if (ctrlAltDelToEnd) {
     text = 'secure-attention=ctrl+alt+end'
   }
   if (data.match(pattern)) {
@@ -20,20 +17,14 @@ export function adjustVVFile ({ data, options, usbAutoshare, usbFilter }) {
     data = data.replace(/^\[virt-viewer\]$/mg, `[virt-viewer]\n${text}`) // ending \n is already there
   }
 
-  const isSpice = data.indexOf('type=spice') > -1
-
-  if (usbFilter && isSpice) {
+  if (usbFilter) {
     data = data.replace(/^\[virt-viewer\]$/mg, `[virt-viewer]\nusb-filter=${usbFilter}`)
     data = data.replace(/^usb-filter=null\n/mg, '') // remove an extra 'usb-filter=null' line if present
   }
 
-  if (options && isSpice) {
-    const smartcardEnabled = options.get ? options.get('smartcardEnabled') : options.smartcardEnabled
-    data = data.replace(/^enable-smartcard=[01]$/mg, `enable-smartcard=${smartcardEnabled ? 1 : 0}`)
-  }
+  data = data.replace(/^enable-smartcard=[01]$/mg, `enable-smartcard=${smartcardEnabled ? 1 : 0}`)
 
-  // make USB Auto-Share to be enabled/disabled in VM Portal according to the SpiceUsbAutoShare config value
-  data = data.replace(/^enable-usb-autoshare=.*$/mg, `enable-usb-autoshare=${usbAutoshare.toString() === 'true' ? 1 : 0}`)
+  data = data.replace(/^enable-usb-autoshare=.*$/mg, `enable-usb-autoshare=${usbAutoshare ? 1 : 0}`)
 
   console.log('adjustVVFile data after adjustment:', data)
   return data
--- a/src/actions/options.js
+++ b/src/actions/options.js
@@ -72,7 +72,22 @@ export function loadUserOptions (userOptions: RemoteUserOptionsType): LoadUserOp
   }
 }
 
-export function saveGlobalOptions ({ values: { sshKey, language, showNotifications, notificationSnoozeDuration, refreshInterval, persistLocale } = {} }: Object, { transactionId }: Object): SaveGlobalOptionsActionType {
+export function saveGlobalOptions ({
+  values: {
+    sshKey,
+    language,
+    persistLocale,
+    showNotifications,
+    notificationSnoozeDuration,
+    refreshInterval,
+    preferredConsole,
+    fullScreenVnc,
+    ctrlAltEndVnc,
+    fullScreenSpice,
+    ctrlAltEndSpice,
+    smartcardSpice,
+  } = {},
+}: Object, { transactionId }: Object): SaveGlobalOptionsActionType {
   return {
     type: C.SAVE_GLOBAL_OPTIONS,
     payload: {
@@ -82,6 +97,12 @@ export function saveGlobalOptions ({ values: { sshKey, language, showNotificatio
       showNotifications,
       notificationSnoozeDuration,
       refreshInterval,
+      preferredConsole,
+      fullScreenVnc,
+      ctrlAltEndVnc,
+      fullScreenSpice,
+      ctrlAltEndSpice,
+      smartcardSpice,
     },
     meta: {
       transactionId,
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -17,7 +17,13 @@ export type SaveGlobalOptionsActionType = {
     persistLocale?: boolean,
     showNotifications?: boolean,
     notificationSnoozeDuration?: number,
-    sshKey?: string
+    sshKey?: string,
+    preferredConsole?: string,
+    fullScreenVnc?: boolean,
+    ctrlAltEndVnc?: boolean,
+    fullScreenSpice?: boolean,
+    ctrlAltEndSpice?: boolean,
+    smartcardSpice?: boolean
   |},
   meta: {|
     transactionId: string
--- a/src/components/Settings/SettingsBase.js
+++ b/src/components/Settings/SettingsBase.js
@@ -62,13 +62,16 @@ Section.propTypes = {
 }
 
 const SettingsBase = ({ name, section }) => {
+  const sections = section.sections ? Object.entries(section.sections) : [[name, section]]
   return (
     <div className={style['search-content-box']}>
-      <Card key={name} className={style['main-content']}>
-        <div className={style['main-content-container']}>
-          <Section name={name} section={section} />
-        </div>
-      </Card>
+      { sections.map(([name, section]) =>
+        <Card key={name} className={style['main-content']}>
+          <div className={style['main-content-container']}>
+            <Section name={name} section={section} />
+          </div>
+        </Card>
+      )}
     </div>
   )
 }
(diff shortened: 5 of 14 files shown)


rszwajko:
### New screen
screen

### Grouping options by sections - confirmation dialog will all possible options changed
confirmationAllOptions
### Grouping options by sections - error notification
notification

### Preferred console - available consoles
preferredConsole

sgratch (changes requested):
Generally it looks ok but there are few design issues to discuss and few things that didn't work for me:
- If possible, It is better to extract the 'grouping options by sections code' fix to a different PR since it's not really related to the console settings fix.

- dialog layout comments:
     - Why separating into VNC and Spice sections is needed within the console dialog? I think that we can have one unified section while 'full screen' and  'Ctrl+Alt+End' can be relevant for VNC, NoVnc and Spice altogether and smartcard field that will be relevant only for Spice (we can add this info into as a tooltip if you think it's needed). This will make the dialog smaller and simpler. I know that on console options it is separated but not sure it's really needed to separate. unifying them will make the layout much better.
     - The 'console' title is relevant for this whole section tab and not specifically for the "preferred console" (including the tooltip of "this settings featured...." which repeats for all other section titles).
    - I suggest to move the ssh key/serial console option to the top or bottom of the dialog since it is a different console separate from all other Graphical Console options. I would also suggest to rename "Preferred Console" label to "preferred Graphical Console" if we decide to leave all other sections titles of "spice options" and "VNC options"

- It seems that the 'Full screen mode' is not polled from backend successfully as part of refresh, I started to receive an error of "Failed to save changes to settings" while it seems that web-ui tries ro remove properties with uuid that doesn't exist anymore.

- All consoles can't be opened for a specific user with the following error: Failed to retrieve VM console details Conflict. Not sure it it's related to this fix anyway.

- full screen mode should be handled for noVnc as well.

- for smartcard: it's better to read current smartcard enabled conf from VM since it's part of backend VM configuration regarless to user settings:  http://ovirt.github.io/ovirt-engine-api-model/master/#types/display/attributes/smartcard_enabled. This might be used by curomers already and we don't want to override their settings. I suggest that if the smartcard is enabled for a VM then enable it regardless to global settings.

- the Ctrl+Alt+End mapping is not working for me. It is always enabled for both spice and vnc.

- If default console is RDP then why the default console for non win VMs is Spice? It should be changed to VNC since VNC is the initial default for graphic consoles.

- The 'Automatic console connection' setting is not included on this PR yet as mentioned by original design https://marvelapp.com/prototype/94h0g56/screen/59286975 and required here https://bugzilla.redhat.com/show_bug.cgi?id=1781241. 

- as a follow up fix: need to support all new console settings as part of 'reset settings'

rszwajko:
> * f possible, It is better to extract the 'grouping options by sections code' fix to a different PR since it's not really related to the console settings fix.
>

In the proposed layout some options have the same labels i.e. Full screen mode. To make the confirmation dialog readable we need to also include the section label.

> 
> * dialog layout comments:
>   
>   * Why separating into VNC and Spice sections is needed within the console dialog? I think that we can have one unified section while 'full screen' and  'Ctrl+Alt+End' can be relevant for VNC, NoVnc and Spice altogether and smartcard field that will be relevant only for Spice (we can add this info into as a tooltip if you think it's needed). This will make the dialog smaller and simpler. I know that on console options it is separated but not sure it's really needed to separate. unifying them will make the layout much better.

Advantages of fine grained options:
1. self-documenting - we have a clear link which option is supported by which console type. One-big-bag design requires providing that information in the tooltip which is not user friendly and gets complex with additional properties (and there are many of them that can be added in the future) 
2. option meaning is stable in time - the general rule for configuration option is that you can add them but once added they should keep their original meaning.  With one-option-multiple-consoles approach the flag stored in the DB could change meaning with time i.e. Full screen could start working for RDP or stop working for SPICE without user consent. This might cause all kinds of problems i.e. during upgrade/downgrade or version mismatch between VM Portal and engine. With fine-grained options we can remove or add a console type without any risk of regression.
3. future-proof - looking at the list of supported (but hidden) options i.e. for [RDP](https://github.com/oVirt/ovirt-web-ui/blob/master/src/sagas/console/rdpBuilder.js) we can expect that with time the list of options will grow

>   * The 'console' title is relevant for this whole section tab and not specifically for the "preferred console" (including the tooltip of "this settings featured...." which repeats for all other section titles).
>

I agree. In the current design it seemed the best way. What would you suggest instead?

>   * I suggest to move the ssh key/serial console option to the top or bottom of the dialog since it is a different console separate from all other Graphical Console options. 
>

I have no user data to support my claim however serial console seems quite advanced way of accessing a VM. This suggests that it's rather rarely used and can be moved to the bottom.

> I would also suggest to rename "Preferred Console" label to "preferred Graphical Console" if we decide to leave all other sections titles of "spice options" and "VNC options"

OK. However this would mean loosing all the translations. Perhaps we can postpone it and implement with other similar changes before next translation cycle?

rszwajko:
>    * The 'Automatic console connection' setting is not included on this PR yet as mentioned by original design https://marvelapp.com/prototype/94h0g56/screen/59286975 and required here https://bugzilla.redhat.com/show_bug.cgi?id=1781241.
>

This option is per VM only and logically belongs to VM settings (and not to global Account Settings)
1. the original UI was part of VM Details - see here https://github.com/oVirt/ovirt-web-ui/pull/208
2. the VM ID was saved in the local storage: https://github.com/oVirt/ovirt-web-ui/blob/853a86b6c300dda1500a74d8e6cac638e692a4be/src/optionsManager.js#L24

rszwajko:
> * It seems that the 'Full screen mode' is not polled from backend successfully as part of refresh, I started to receive an error of "Failed to save changes to settings" while it seems that web-ui tries ro remove properties with uuid that doesn't exist anymore.

> * All consoles can't be opened for a specific user with the following error: Failed to retrieve VM console details Conflict. Not sure it it's related to this fix anyway.

> * the Ctrl+Alt+End mapping is not working for me. It is always enabled for both spice and vnc.

Please describe(or record) the steps needed to recreate the behaviour. Note that the console configuration is logged in the browser's JS console so you can check the content of *.vv file (see below).

```
[debug]  adjustVVFile data after adjustment: [virt-viewer]
type=vnc
....
fullscreen=1
...
secure-attention=ctrl+alt+end
....
```

sgratch:
@rszwajko please rebase

rszwajko:
> * full screen mode should be handled for noVnc as well.
>

Done.

> * for smartcard: it's better to read current smartcard enabled conf from VM since it's part of backend VM configuration regarless to user settings: http://ovirt.github.io/ovirt-engine-api-model/master/#types/display/attributes/smartcard_enabled. This might be used by curomers already and we don't want to override their settings. I suggest that if the smartcard is enabled for a VM then enable it regardless to global settings.
>

Smartcard flag in VM configuration seems to define a capability. If smartcard is supported by the VM then user might enble or disable it. 

> * if default console is RDP then why the default console for non win VMs is Spice? It should be changed to VNC since VNC is the initial default for graphic consoles. 

fixed as part of #1420 

> * as a follow up fix: need to support all new console settings as part of 'reset settings'
>

Done.

sgratch (dismissed):
> @rszwajko 
> 
> - If possible, It is better to extract the 'grouping options by sections code' fix to a different PR since it's not really related to the console settings fix.

> In the proposed layout some options have the same labels i.e. Full screen mode. To make the confirmation dialog readable we need to also include the section label.

Sure, my original comment was just regarding extracting the code to a separate PR since not all grouping is relevant for the console section so we do mix two issues within one PR. 

> 
> - dialog layout comments:
>      - Why separating into VNC and Spice sections is needed within the console dialog? I think that we can have one unified section while 'full screen' and  'Ctrl+Alt+End' can be relevant for VNC, NoVnc and Spice altogether and smartcard field that will be relevant only for Spice (we can add this info into as a tooltip if you think it's needed). This will make the dialog smaller and simpler. I know that on console options it is separated but not sure it's really needed to separate. unifying them will make the layout much better.

> Advantages of fine grained options.....

OK, I'm convinced :) I still don't think it's necessary but we can leave it as is.  

> - The 'console' title is relevant for this whole section tab and not specifically for the "preferred console" (including the tooltip 
> of "this settings featured...." which repeats for all other section titles).

>  - I suggest to move the ssh key/serial console option to the top or bottom of the dialog since it is a different console separate from all other Graphical Console options. I would also suggest to rename "Preferred Console" label to "preferred Graphical Console" if we decide to leave all other sections titles of "spice options" and "VNC options"

I suggest to have something similar to the following layout:
image

> rename "Preferred Console" label to "preferred Graphical Console"
>   OK. However this would mean loosing all the translations. Perhaps we can postpone it and implement with other similar changes before next translation cycle?

Sure

> 
> - It seems that the 'Full screen mode' is not polled from backend successfully as part of refresh, I started to receive an error of "Failed to save changes to settings" while it seems that web-ui tries ro remove properties with uuid that doesn't exist anymore.

This is not reproduced to me now and I don't remember the exact scenario so let's ignore for now.

> 
> - All consoles can't be opened for a specific user with the following error: Failed to retrieve VM console details Conflict. Not sure it it's related to this fix anyway.
> 

I had this issue reproduced few times for a specific user but it isn't related to this PR so once reproduced we can open a separate issue. 
> 
> - full screen mode should be handled for noVnc as well.
> Done

It seems that the full screen is enabled but layout for the novnc console is not working properly (it's not displayed on full screen). Check how it works while clicking the 'full screen' button from the noVnc console dialog. I remember there were few customizations for this to work: https://github.com/oVirt/ovirt-web-ui/pull/1092

> 
> - for smartcard: it's better to read current smartcard enabled conf from VM since it's part of backend VM configuration regarless to user settings:  http://ovirt.github.io/ovirt-engine-api-model/master/#types/display/attributes/smartcard_enabled. This might be used by curomers already and we don't want to override their settings. I suggest that if the smartcard is enabled for a VM then enable it regardless to global settings.

> Smartcard flag in VM configuration seems to define a capability. If smartcard is supported by the VM then user might enble or disable it.

The smartcard flag of the VM configuration is actually enabled/disabled the smartcard and not just define a capability. 
Since we don't want to change the settings for existing customers and since smartcard is not enabled by default then what about deciding that smartcard will be enabled for a VM only if enabled by the VM configuration OR enabled on a global level by user settings?  

> 
> - the Ctrl+Alt+End mapping is not working for me. It is always enabled for both spice and vnc.

It seems working correctly now.

> 
> - If default console is RDP then why the default console for non win VMs is Spice? It should be changed to VNC since VNC is the initial default for graphic consoles.

> fixed as part of  #1420

+1


> 
> - The 'Automatic console connection' setting is not included on this PR yet as mentioned by original design https://marvelapp.com/prototype/94h0g56/screen/59286975 and required here https://bugzilla.redhat.com/show_bug.cgi?id=1781241. 
> This option is per VM only and logically belongs to VM settings (and not to global Account Settings)

You are right that for user portal it was supported per VM on VM level. So we can have two options for handling this for user settings:
1. Enable it on VM level only: add a new switch field for enabling this as part of VM details card and persist this settings as a vmportal property while the value will be the latest 'automatic console connection' VM id. No global settings will be existed but only one property is needed for this settings. Similar to the local storage solution
2. Enable this on a global level by deciding that it is enabled only if there is 1 running VM assigned to this user. If there is more than one then we can disable this. 


> 
> - as a follow up fix: need to support all new console settings as part of 'reset settings'   
> 
> Done

+1

rszwajko:
> I suggest to have something similar to the following layout:
> image
>

This is possible but it's a big change in the design and in the current framework (sections with widgets but without fields).

> > * full screen mode should be handled for noVnc as well.
> >   Done
> 
> It seems that the full screen is enabled but layout for the novnc console is not working properly (it's not displayed on full screen). Check how it works while clicking the 'full screen' button from the noVnc console dialog. I remember there were few customizations for this to work: #1092
> 

Differently then native VNC we never supported auto-fullscreen for noVNC. This has been added in this PR (as requested in review above). However I had no env that I could use to fully test this feature (proxy needs to be configured). I was only able to test that the browser console starts correctly in full screen and after connection error displays error page.
If this feature will cause problems I would rather remove it from this PR and finish it under separate issue.

> > * for smartcard: it's better to read current smartcard enabled conf from VM since it's part of backend VM configuration regarless to user settings:  http://ovirt.github.io/ovirt-engine-api-model/master/#types/display/attributes/smartcard_enabled. This might be used by curomers already and we don't want to override their settings. I suggest that if the smartcard is enabled for a VM then enable it regardless to global settings.
> 
> > Smartcard flag in VM configuration seems to define a capability. If smartcard is supported by the VM then user might enble or disable it.
> 
> The smartcard flag of the VM configuration is actually enabled/disabled the smartcard and not just define a capability.
>
Yes. This is my understanding too.

> Since we don't want to change the settings for existing customers and since smartcard is not enabled by default then what about deciding that smartcard will be enabled for a VM only if enabled by the VM configuration OR enabled on a global level by user settings?
> 

According to my research "smartcard" option is not used by VM Portal users since the time it was removed from the UI (except of those that still have the right flag in the browser's local storage).
Looking at (last?) [screen that used this functionality](https://github.com/oVirt/ovirt-web-ui/blob/56b134fd0462fcd380c057a209e703d0a817ade8/src/components/ConsoleOptions/index.js#L80) (version 1.4)  we can see that enabled smartcard capability allowed the users to set the option (on or off). Missing capability resulted in disabling the UI widget and forced "no smartcard" mode.

>   2. Enable this on a global level by deciding that it is enabled only if there is 1 running VM assigned to this user. If there is more than one then we can disable this.
>

This is entirely different functionality - instead of connecting to one specified  VM (that is running) we will connect to ANY machine (that's  the ONLY running machine). This is quite a  specific use case. I think we need a separate issue for that.

sgratch:
@rszwajko 

> > I suggest to have something similar to the following layout:
> > image
> 
> This is possible but it's a big change in the design and in the current framework (sections with widgets but without fields).

The current section divisions is pretty problematic , a section for console, for serial console and then for spice and vnc. So where is the section for RDP and noVnc? Why the serial in in the middle etc...
it should be re-ordered and it seems important. How much time you estimate for changing it?
We can take a quicker approach and fixthe simple parts by moving the ssh to the buttom, move the console title to a dedicated section with it's tooltip (remove this tooltip from other sections) and create a new section for preferred console. So the alignment won't be perfect but at least the devision will be more correct. WDYT?

> 
> > > * full screen mode should be handled for noVnc as well.
> > >   Done
> > 
> > 
> > It seems that the full screen is enabled but layout for the novnc console is not working properly (it's not displayed on full screen). Check how it works while clicking the 'full screen' button from the noVnc console dialog. I remember there were few customizations for this to work: #1092
> 
> Differently then native VNC we never supported auto-fullscreen for noVNC. This has been added in this PR (as requested in review above). However I had no env that I could use to fully test this feature (proxy needs to be configured). I was only able to test that the browser console starts correctly in full screen and after connection error displays error page.
> If this feature will cause problems I would rather remove it from this PR and finish it under separate issue.

We already support fullscreen mode for noVnc and it is working, we just need to enable it correctly as part of user settings:
image
We can open a follow up issue for enabling the noVnc full screen for user settings if it's problematic to fix now.

> 
> > > * for smartcard: it's better to read current smartcard enabled conf from VM since it's part of backend VM configuration regarless to user settings:  http://ovirt.github.io/ovirt-engine-api-model/master/#types/display/attributes/smartcard_enabled. This might be used by curomers already and we don't want to override their settings. I suggest that if the smartcard is enabled for a VM then enable it regardless to global settings.
> > 
> > 
> > > Smartcard flag in VM configuration seems to define a capability. If smartcard is supported by the VM then user might enble or disable it.
> > 
> > 
> > The smartcard flag of the VM configuration is actually enabled/disabled the smartcard and not just define a capability.
> 
> Yes. This is my understanding too.
> 
> > Since we don't want to change the settings for existing customers and since smartcard is not enabled by default then what about deciding that smartcard will be enabled for a VM only if enabled by the VM configuration OR enabled on a global level by user settings?
> 
> According to my research "smartcard" option is not used by VM Portal users since the time it was removed from the UI (except of those that still have the right flag in the browser's local storage).
> Looking at (last?) [screen that used this functionality](https://github.com/oVirt/ovirt-web-ui/blob/56b134fd0462fcd380c057a209e703d0a817ade8/src/components/ConsoleOptions/index.js#L80) (version 1.4) we can see that enabled smartcard capability allowed the users to set the option (on or off). Missing capability resulted in disabling the UI widget and forced "no smartcard" mode.

We are not based on older functionality since it's totally different. We are currently working on user settings for a global level and not sure that we'll ever support a VM level mode. Since we assume that there are still customers that using smartcards for their win VMs on webadmin then we need to somehow support it on web-ui on a global level. 
We can also use that approach of enabling/disabling for all VMs so that:
if user settings smartcard is disabled on a global level - it is disabled for all vms regardless to VM's configuration. 
And if it is enabled on global level, only VMs that are configured to use it via their configuraiton will be actually using smartcard. 
So the actual decision is done by the admin and the user can toggle it. This sounds ok to me.

> 
> > 1. Enable this on a global level by deciding that it is enabled only if there is 1 running VM assigned to this user. If there is more than one then we can disable this.
> 
> This is entirely different functionality - instead of connecting to one specified VM (that is running) we will connect to ANY machine (that's the ONLY running machine). This is quite a specific use case. I think we need a separate issue for that.

It is. what about the other sugested solution?

sgratch:
@rszwajko 

> Sure, my original comment was just regarding extracting the code to a separate PR since not all grouping is relevant for the console section so we do mix two issues within one PR.

You can extract the non console grouping code to a separate PR.

rszwajko:
### Modified layout

1. `Console` - the first section has no fields (this is allowed in the current design) and as suggested in the review is kind of header for this screen
2. the second section has no title(technically empty string as title) and one field `Preferred console`. The layout suggested in the review is possible (with some hacks) but then the `Console` section looses it's "header" function as  the second section is more visible (see the second screenshot).

**Chosen layout**
image

**Suggested layout**
image

### Confirmation for preferred console

image

rszwajko:
> We already support fullscreen mode for noVnc and it is working, we just need to enable it correctly as part of user settings
> ...
> We can open a follow up issue for enabling the noVnc full screen for user settings if it's problematic to fix now.
>
Full screen mode for noVNC was/is not connected to options saved in local storage. In this sense it was/is not supported to use custom settings for this functionality. Implementing this functionality requires some more time (and testing) therefor I've removed this commit. We can implement in follow up issue.

> On web-ui since it is a global settings then it can behave differently :
> once smartcard is enabled for the VM via VM's configuration and the global settings 'smartcard' option is enabled - smartcard > will be enabled for that VM.
> If the global settings 'smartcard' option is disabled, it will be disabled regardless to VM's configuration.
>

Done. Note that to achieve this functionality default value for smartard was changed to **enabled** (this way it's the VM config that decides).

> We can 'move' the local storage property that stores vm-id to be remotely persistent on a global level, as suggested in previous comment. 
> 
Yes. Logically it was global - only one VM was allowed per user.


> Whatever approach we will use, we still need to manage a global disable/enable option for auto connect so that the user won't need to go VM by VM. The older ui solution of enabling/disabling by going to each VM's details card is ugly.
>
It was not necessary to go one-by-one. Selecting the option on one VM was overriding the previous setting. 
Since this part of the design requires more discussion it cannot be added in this PR - a follow-up is necessary.

rszwajko:
Force pushed to re-base on top of #1456 .
Additionally the defaults for `smartcardSpice` have been moved to App configuration.

rszwajko:
Force-pushed after #1456  has been merged. No changes in the code.

src/sagas/console/index.js
```diff
@@ -86,6 +86,50 @@ export function* downloadVmConsole (action) {
   }
 }
 
+/**
+ * Legacy options were saved in browser's local storage.
+ * The UI for setting theose options was removed in previous versions.
+ * However there is still a (small) chance that the data is still there.
+ * Note that legacy options are per VM and therefore should overwrite global defaults.
+ */
+function* getLegacyOptions ({ vmId }) {
```
sjd78, line 95: Man that is a VERY small chance.  The switches for console config hasn't existed since before the VM Detail's screen was redesigned at least 2 years ago.

src/sagas/console/vvFileUtils.js
```diff
@@ -1,15 +1,12 @@
 
-export function adjustVVFile ({ data, options, usbAutoshare, usbFilter }) {
-  // __options__ can either be a plain JS object or ImmutableJS Map
-  console.log('adjustVVFile options:', options)
-
-  if (options && ((options.get && options.get('fullscreen')) || options.fullscreen)) {
+export function adjustVVFile ({ data, options: { fullscreen, ctrlAltDelToEnd, smartcardEnabled, usbAutoshare, usbFilter } }) {
```
sjd78, line 2: Minor housekeeping: No default value for the `options:` destructure?

sjd78 (commented):
One minor comment.  Otherwise, everything else looks good and works well.

src/sagas/console/index.js
rszwajko: Yes, I agree. However removing the legacy mechanism was not in scope of this PR. Beside that ....the code for reading the values is still in our code base and most likely working:)

src/sagas/console/vvFileUtils.js
```diff
@@ -20,20 +17,14 @@ export function adjustVVFile ({ data, options, usbAutoshare, usbFilter }) {
...
 
-  if (options && isSpice) {
-    const smartcardEnabled = options.get ? options.get('smartcardEnabled') : options.smartcardEnabled
-    data = data.replace(/^enable-smartcard=[01]$/mg, `enable-smartcard=${smartcardEnabled ? 1 : 0}`)
-  }
+  data = data.replace(/^enable-smartcard=[01]$/mg, `enable-smartcard=${smartcardEnabled ? 1 : 0}`)
 
-  // make USB Auto-Share to be enabled/disabled in VM Portal according to the SpiceUsbAutoShare config value
-  data = data.replace(/^enable-usb-autoshare=.*$/mg, `enable-usb-autoshare=${usbAutoshare.toString() === 'true' ? 1 : 0}`)
+  data = data.replace(/^enable-usb-autoshare=.*$/mg, `enable-usb-autoshare=${usbAutoshare ? 1 : 0}`)
```
sgratch, line 27: `usbAutoshare ? 1 : 0`
This isn't correct since it's a string and not a boolean

sgratch (changes requested):
@rszwajko 

The code LGTM except a minor fix.

> Modified layout

> the second section has no title(technically empty string as title) and one field Preferred console. The layout suggested in the review is possible (with some hacks) but then the Console section looses it's "header" function as the second section is more visible (see the second screenshot).

I agree that it was preferable to display within the confirmation/error dialogs the list of fields with the 'Console' title per each console changed field instead of those internal titles (e.g. 'console' instead of 'vnc options'), since for navigating to the right section for updating, the 'Console' is the section's visible name. But I think it's ok like it is now as well.

> full screen mode should be handled for noVnc as well.

Please open a follow up issue for this.

> smartcard' option

LGTM

> The 'Automatic console connection' setting

Please open a follow up issue for this.

src/sagas/console/vvFileUtils.js
sjd78: Yeah it looks like it is passed through from the `fetchServerConfiguredValues()` saga untransformed.  Which means it'll be `"true"` or `"false"`.

So maybe worth it to have a `convertBool()` call in either the `fetchServerConfiguredValues()` saga call or in the `SET_USB_AUTOSHARE` config reducer?

rszwajko: Note that also default value has been changed - previously it was `1` which suggested that the supported values are [`0`,`1`]

rszwajko:
Force pushed changes:
1. error handling and tests for `adjustVVFile()` method
2. transforming `SpiceUsbAutoShare` into boolean upon arrival

sjd78 (approved):
LGTM

sgratch (approved):
LGTM

sgratch:
The code looks good and follow up issues were opened.

Issue #1442, opened 2021-06-07
Review docker container build (locally, via make, via CI on std-ci or travisci)


sjd78:
We need to review the usefulness and validity of the container build of ovirt-web-ui.

There are a few items to consider:
 1. Do we still need the container build?
 2. Should the container be running the webpack dev server or use another build/deploy mechanism that does not require compiling every time it runs?
 3. Is docker hub the best place to continue publishing the container?  Maybe quay.io or something else RH provides?
 4. Should the CI build scripts login to the target container registry to avoid rate limiting etc?
 5. ~~Should travisci still be used to build the container or can that be ported to std-ci?~~ travis-ci.org is shutdown as of June 15, so the container build will need to be ported to another ci build environment (travis-ci.com or std-ci).
 6. Should we migrate from docker to podman/buildah?

Currently, the build via travisci is frequently failing.  For example see https://travis-ci.org/github/oVirt/ovirt-web-ui/builds/773732433#L351.  Additionally, travis-ci.org is due to shutdown in favor of travis-ci.com.  We may need to migrate anyway.

sjd78:
@michalskrivanek , @sgratch , @rszwajko , @hstastna - Thoughts and comments?

michalskrivanek:
IMHO it's not widely used, if ever. So it's certainly something we can consider to ditch. Otherwise we would have to fix and improve 2-6...probably not worth it

sjd78:
Good article about dockerizing react app https://mherman.org/blog/dockerizing-a-react-app/

It isn't 100% applicable (missing the user login from ENV vars, missing the api proxy), but would be a good base.

sjd78:
**Note**: travis-ci.org is shutdown as of June 15.

PR #1447, opened 2021-06-14
Disable the docker container builds on travisci

sjd78:
Due to (1) consistent docker container build problems on travisci,
and (2) container issues to be considered in issue #1442, disable
the docker container build.  Any updates to the docker hub image
will need to be manual until #1442 is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,30 @@
 language: node_js
 
 services:
-  - docker
+  # - docker
 
 script:
   - yarn run build
-  - docker build --tag web-ui .
-  - docker run -d --rm -it -e ENGINE_URL=https://localhost/ovirt-engine -p 3000:3000 --name web-ui web-ui
-  - >
-    for i in {1..7}; do
-      docker logs web-ui
-      docker logs web-ui 2>&1  | grep -Pzl 'error'
-      if [ $? -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


sjd78:
@sgratch , @rszwajko , @hstastna - I'd rather see travisci pass with just running the build by itself than keep on failing (most of the time at least) with the docker build included.

hstastna (approved):
I agree with disabling the docker container build for now.

rszwajko:
I have no opinion on that.

sjd78:
Note: travis-ci.org is no longer running builds, so this PR is a bit of a moot point.  If we want automated container builds, it'll need to be done in std-ci or migrate to travis-ci.com.

sgratch:
ci test

sgratch:
> I'd rather see travisci pass with just running the build by itself than keep on failing (most of the time at least) with the docker build included.

@sjd78 note that now after travis-ci.org is no longer running builds, the travisci is totally dismissed on all PRs except here where it's always pass due to running the build only.
I'm actually not sure what is preferable for now. Let's wait after 4.4.7 is released, check with the community if docker container is ever used and decide.

sjd78:
Closing this PR because travisci.org is not running builds anymore so the change is not required.  Leave #1442 open to track future container deploy work.

Issue #1434, opened 2021-05-20
user settings - support marking a default option for the  SelectBox of the notificationSnoozeDurationInMinutes


sgratch:
This is a follow up for the fix https://github.com/oVirt/ovirt-web-ui/pull/1415.
As mentioned on https://github.com/oVirt/ovirt-web-ui/pull/1424#pullrequestreview-663432907, worth marking the "Do not disturb for" time duration default of AppConfiguration.notificationSnoozeDurationInMinutes as default as well.

PR #1449, opened 2021-06-15
Mark a default option for the Notifications' SelectBox in Account Settings

hstastna:
Issue: https://github.com/oVirt/ovirt-web-ui/issues/1434

This PR supports marking a default option for the SelectBox of the `notificationSnoozeDurationInMinutes`, in Account Settings, which is 10 minutes.

**Before:**
default_before2
default_before

**After:**
default_after2
default_after


sgratch (approved):
LGTM

Issue #1439, opened 2021-06-02
user settings - display user settings confirmation dialog fields in order based on dialog appearace 


sgratch:
This is a follow up issue for https://github.com/oVirt/ovirt-web-ui/pull/1427#issuecomment-851608502.
It's better to set the order of fields within the user settings confirmation dialog to fit user settings dialog fields/tabs order appearance. e.g.:
Language
SSH key
UI refresh
Do not disturb
Do not disturb for
Advanced settings

Current order looks like this:
![image](https://user-images.githubusercontent.com/18169498/120471879-ec879880-c3ad-11eb-9be8-d1a13e77ea39.png)

PR #1456, opened 2021-06-18
Preserve the order and group options by section

rszwajko:
Extracted from PR #1446 
Resolves: #1439
Screenshots: [see this comment](https://github.com/oVirt/ovirt-web-ui/pull/1446#issuecomment-857799268)

Background:
Confirmation dialog and error notification need to display the list
of changed options. However a flat list is no longer suitable:
1. some options have the same translated name
2. option list has grown significantly
3. options are split into sub-screens (not all edited options are
   visible to the user)

Changes:
1. display a list of changed sections
2. within each changed section display a list of changed options
3. preserve the same order of sections and options as in the UI

### DIFF
--- a/src/components/Settings/ChangesList.js
+++ b/src/components/Settings/ChangesList.js
@@ -1,0 +1,43 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import style from './style.css'
+
+const ChangesList = ({ changes = [], translatedLabels = [] }) => {
+  // start with translated labels to preserve the order of sections/fields
+  const filteredSectionTree = translatedLabels
+    .filter(({ fieldName }) => changes.find(changeName => changeName === fieldName))
+    // group fields by section (keep the order of sections)
+    // output { "some section": ["field A", "field B", ...], ...}
+    .reduce((acc, { fieldTitle, sectionTitle }) => {
+      if (acc[sectionTitle]) {
+        acc[sectionTitle].push(fieldTitle)
+      } else {
+        acc[sectionTitle] = [fieldTitle]
+      }
+      return acc
+    }, {})
+
+  return (
+    <ul className={style['section-list']}>
+      {
+        Object.entries(filteredSectionTree)
+          .map(([sectionTitle, fields]) => (
+            <li key={sectionTitle}>
+              {sectionTitle}
+              <ul className={style['field-list']}>
+                { fields.map(fieldTitle => <li key={fieldTitle}>{fieldTitle}</li>)}
+              </ul>
+            </li>
+          ))
+      }
+    </ul>
+  )
+}
+
+ChangesList.propTypes = {
+  changes: PropTypes.array,
+  translatedLabels: PropTypes.array.isRequired,
+}
+
+export default ChangesList
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -8,6 +8,7 @@ import CounterAlert from '_/components/CounterAlert'
 import { generateUnique } from '_/helpers'
 import { MsgContext } from '_/intl'
 import style from './style.css'
+import ChangesList from './ChangesList'
 
 const changedInTheMeantime = ({ currentValues = {}, baseValues = {}, draftValues = {}, sentValues = {} }) => {
   return Object.keys(currentValues).filter(name =>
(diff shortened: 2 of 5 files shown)


src/components/Settings/ChangesList.js
```diff
@@ -0,0 +1,42 @@
...
+    // group fields by section (keep the order of sections)
+    // output { "some section": ["field A", "field B", ...], ...}
+    .reduce((acc, { fieldTitle, sectionTitle }) => {
+      if (acc[sectionTitle]) {
+        acc[sectionTitle].push(fieldTitle)
+      } else {
+        acc[sectionTitle] = [fieldTitle]
+      }
+      return acc
+    }, {})
```
sjd78, line 20: Nit-pick: Indent one more level

Side-note: gotta get eslint working again!

src/components/Settings/ChangesList.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import style from './style.css'
+
+export const ChangesList = ({ changes = [], translatedLabels = [] }) => {
```
sjd78, line 6: I'd rather see the only component in this file be the default export and drop the {} on the import

```suggestion
export default const ChangesList = ({ changes = [], translatedLabels = [] }) => {
```

src/components/Settings/Settings.js
```diff
@@ -8,6 +8,7 @@ import CounterAlert from '_/components/CounterAlert'
 import { generateUnique } from '_/helpers'
 import { MsgContext } from '_/intl'
 import style from './style.css'
+import { ChangesList } from './ChangesList'
```
sjd78, line 11:
```suggestion
import ChangesList from './ChangesList'
```

src/components/Settings/SettingsToolbar.js
```diff
@@ -6,13 +6,13 @@ import { MsgContext } from '_/intl'
 
 import style from './style.css'
 import ConfirmationModal from '_/components/VmActions/ConfirmationModal'
+import { ChangesList } from './ChangesList'
```
sjd78, line 9:
```suggestion
import ChangesList from './ChangesList'
```

src/components/UserSettings/GlobalSettings.js
```diff
@@ -148,147 +148,147 @@ class GlobalSettings extends Component {
...
-          {
+          ((name) => ({
             title: msg.username(),
-            name: 'username',
-            body: <span>{config.userName}</span>,
-          },
-          {
+            name,
+            body: <span>{config[name]}</span>,
+          }))('userName'),
```
sjd78, line 163: IDK - this style makes it a bit more difficult to read at the expense of have fewer specific values.  Feels a bit over complicated.  Any strong opinions?

sjd78 (commented):
Approach is reasonable, more a bunch of nit-pick/style questions.

sgratch:
ci test

src/components/UserSettings/GlobalSettings.js
rszwajko: I agree, that for this field it's overkill (forced by the framework). The change was required for editable fields. There we need to use the same identifier in 4 places: 
1. name(used for React `key` property and for confirmation/error dialog), 
2. id, 
3. `draft[name]` (reading value from the state) and 
4. `onChange(name)` (storing value in the state).

Ensuring that the same string is used in all those places has become a problem.
We could move/hide this in the `Settings*` layer however every widget consumes the input differently which adds extra complexity.

sjd78: So I'm meh about the whole thing.  I'm ok with it if you're ok with it.

sgratch (approved):
LGTM

sjd78 (approved):
LGTM

isaranova:
Screenshot_2021-07-15 VM Portal
Shouldn't the consoles options be under a first line Console instead of double dot indent?

rszwajko:
@isaranova  @sgratch 
> Shouldn't the consoles options be under a first line Console instead of double dot indent?
>

Yep. That was the original plan (as visible on original [screenshots](https://github.com/oVirt/ovirt-web-ui/pull/1446#issuecomment-857799268)). This has been changed as a followup to @sgratch [comments](https://github.com/oVirt/ovirt-web-ui/pull/1446#pullrequestreview-684303525).
The idea was to provide some kind of header/title for the entire section. However this is not possible in the current design which is flat section-based layout.

isaranova:
Well, it does not look very consistent with the other indents that start with the tile title, but since it has been agreed on I guess it works as expected.

LGTM

PR #1457, opened 2021-06-20
Update/fix dependencies of handlebars, bootstrap-select and jquery

sgratch:
Update handlebars to 4.7.7
Fix dependencies of bootstrap-select to use only 1.13.6
Fix dependencies of jquery to use only 3.5.1


sgratch:
ci test

sgratch:
I ran some basic regression tests and it LGTM.

PR #1467, opened 2021-06-25
refactor sagas - refactor login managed base-data, user, group, and role fetches

sjd78:
Do some work on sagas.  The work includes:

1. Rearrange the user, group and role related sagas

2. Remove the get/fetch actions for login managed base-data fetches that do not refreshed after login.  Relevant entities touched:
   - clusters
   - hosts
   - iso files
   - operating systems
   - storage domains
   - templates
   - vnic profiles
 
3. _moved to #1468:_ ~~Refactor the disk related sagas to fetch the permissions and to calculate the permits on the front end.  This follows existing permission -> permit -> canUser* calcualtions done in the rest of the app.  Any remaining permission/permit only fetching support have been removed to avoid and future confusion.~~

4. _moved to #1469:_ ~~Refactored the snapshot disks and nics fetching.~~


hstastna (approved):
I think moving `fetchCurrentUser` function to _base-data.js_ wasn't necessary, but it is okay, too. LGTM.

rszwajko (approved):

PR #1468, opened 2021-06-26
refactor sagas - improve disk permit/permissions calculations

sjd78:
Refactor the disk related sagas to fetch the permissions and to calculate the permits on the front end. This follows existing permission -> permit -> canUser* calculations done in the rest of the app. Any remaining permission/permit only fetching support have been removed to avoid and future confusion.

Reduce the number of rest calls required to permit disks that are fetched:
  - on single vm fetch (disk_attachments.disk.permissions)
  - on update a vm's disks (disk.permissions)
  - on update a single disk attachment (disk.permsissions)

Remove unused direct fetch permit/permission functionality. Now, all permissions are fetched along with entities that need. The permission to permit mapping is all done on in the app.  `canUse*` constants are calculated during the fetch after the transforms.

### DIFF
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -101,9 +101,7 @@ function getPoolColor (id: string): string {
 //
 const VM = {
   toInternal ({ vm }: { vm: ApiVmType }): VmType {
-    const permissions = vm.permissions && vm.permissions.permission
-      ? Permissions.toInternal({ permissions: vm.permissions.permission })
-      : []
+    const permissions = Permissions.toInternal({ permissions: vm?.permissions?.permission })
 
     const parsedVm: Object = {
       name: vm.name,
@@ -407,9 +405,7 @@ const Template = {
       baseTemplateId: template.version && template.version.base_template ? template.version.base_template.id : undefined,
     }
 
-    const permissions = template.permissions && template.permissions.permission
-      ? Permissions.toInternal({ permissions: template.permissions.permission })
-      : []
+    const permissions = Permissions.toInternal({ permissions: template?.permissions?.permission })
 
     return cleanUndefined({
       id: template.id,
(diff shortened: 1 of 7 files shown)


sjd78:
For example, before this PR, a VM with 2 disks will generate at least 3 requests to fully fetch the VM's details.  The `permissions?rollow=role.permits` calls happen once for each disk.  The `disks` and `nics` calls are for snapshots and are addressed in #1469:
VM Portal - Google Chrome_038

After the PR:
VM Portal - Google Chrome_039

src/sagas/disks.js
```diff
@@ -1,21 +1,54 @@
...
+  return internalDisk
+}
+
+function* fetchVmDisks ({ vmId }) {
+  const diskattachments = yield callExternalAction('diskattachments', Api.diskattachments, { payload: { vmId } })
+
+  const internalDisks = []
+
+  if (diskattachments && diskattachments.disk_attachment) {
+    for (const attachment of diskattachments.disk_attachment) {
```
rszwajko, line 44: can we use destructuring (or `.?` operator) to get the attachments? 
```
const { disk_attachment: attachments = []} = yield ...
// or
const attachements = diskattachments?.disk_attachment ?? [] 
```
can we use `yield all([...]) here ? i.e.
```
const internalDisks = yield all(attachments.map(a => call(transformAndPermitDiskAttachment, a))
```

src/ovirtapi/transform.js
```diff
@@ -537,8 +537,11 @@ const Snapshot = {
 // VM -> DiskAttachments.DiskAttachment[] -> Disk
 const DiskAttachment = {
   toInternal ({ attachment, disk }: { attachment?: ApiDiskAttachmentType, disk: ApiDiskType }): DiskType {
-    // TODO Add nested permissions support when BZ 1639784 will be done
-    return cleanUndefined({
+    const permissions = disk.permissions && disk.permissions.permission
```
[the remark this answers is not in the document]
sjd78: There are 7 places that follow this pattern in `transform.js` already.  They can all be switched!

src/sagas/disks.js
sjd78: > can we use destructuring (or `.?` operator) to get the attachments?
> 
> ```
> const { disk_attachment: attachments = []} = yield ...
> // or
> const attachements = diskattachments?.disk_attachment ?? [] 
> ```

I don't feel the need to always destructure or pull out bits of objects.  The if condition using `?.` does make sense.

> 
> can we use `yield all([...]) here ? i.e.
> 
> ```
> const internalDisks = yield all(attachments.map(a => call(transformAndPermitDiskAttachment, a))
> ```

We could but it wouldn't matter since there are no async actions in `transformAndPermitDiskAttachment()`.  I'd really only use `yield all(...)` if there are async/fetch bits.

rszwajko (approved):

PR #1469, opened 2021-06-26
refactor sagas - optimize snapshot disks and nics fetching

sjd78:
Refactored the snapshot disks and nics fetching to optimize request patterns:

  - only fetch for snapshots that are not 'active' since we don't display the 'active' snapshot anyway

  - fetch all disks and nics for each non active snapshot in parallel to minimize the overall wall clock wait time

Note: The saga helper `parallelFetchAndPopulateSnapshotDisksAndNics()` directly updates the provided snapshots array objects with the fetched data.  The calling saga needs to put the updates to the store as appropriate.

### DIFF
--- a/src/sagas/index.js
+++ b/src/sagas/index.js
@@ -231,16 +231,8 @@ export function* fetchSingleVm (action) {
       internalVm.cdrom = yield fetchVmCdRom({ vmId: internalVm.id, current: true })
     }
 
-    // NOTE: Snapshot Disks and Nics are not currently (Sept-2018) available via
-    //       additional/follow param on the VM/snapshot fetch.  We need to fetch them
-    //       directly.
-    for (const snapshot of internalVm.snapshots) {
-      const follows = yield all([
-        call(fetchVmSnapshotDisks, { vmId: internalVm.id, snapshotId: snapshot.id }),
-        call(fetchVmSnapshotNics, { vmId: internalVm.id, snapshotId: snapshot.id }),
-      ])
-      snapshot.disks = follows[0]
-      snapshot.nics = follows[1]
+    if (!shallowFetch) {
+      yield parallelFetchAndPopulateSnapshotDisksAndNics(internalVm.id, internalVm.snapshots)
     }
 
     yield put(updateVms({ vms: [internalVm], copySubResources: shallowFetch }))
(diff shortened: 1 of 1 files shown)


sjd78:
## Example 1 - A VM with no snapshots

Before the PR, a VM with no snapshots is still requesting disks and nics for the 'active' snapshot even though there are none:
VM Portal - Google Chrome_042

After the PR, the same VM skips pulling details for the active snapshot.  We save 2 extra requests:
VM Portal - Google Chrome_041

## Example 2 - A VM with a bunch of snapshots

Before the PR, for this VM, a total of 13 requests are executed to get all of the VM's details.  The 2 requests for disks (`permissions?follow=role.permits` requests) are addressed in #1468. Also notice the waterfall pattern for the requests:
VM Portal - Google Chrome_043

src/sagas/index.js
```diff
@@ -520,23 +510,41 @@ export function* fetchVmSnapshots ({ vmId }) {
...
+  }
+
+  // NOTE 1: Snapshot Disks and Nics are not currently (Sept-2018) available via
+  //         additional/follow param on the VM/snapshot fetch.  We need to fetch them
+  //         directly.
+  // NOTE 2: Looks like the `?follows=snapshots` returns the active and regular
+  //         snapshots.  The active snapshot does not contain links to disks and nics,
+  //         but the regular ones do.  That is probably the reason why the VM call
+  //         `?follows=snapshots.disks` fails.
+  yield all(
```
rszwajko, line 531: 1. `map()` usually has no side effects - here we use it only for side effects (fetching and mutating)
2. `all()` could fail on any request (not sure if it will leave the snapshots partially updated/corrupted or not)

How about sth like (pseudo-code)):
```
const [id2nics, id2disks] = yield all([
   call( all, Object.fromEntries(nonActive.map({id}=> [id, call(fetchVmSnapshotNics,....)]))),
   call( all, Object.fromEntries(nonActive.map({id}=> [id, call(fetchVmSnapshotDisks,....)]))),
])

if(id2nics) {
   snapshots.forEach(s => s.nics = id2nics[s.id] ?? [])
} else {
  //  trigger user notification about nics
}
```
This should give us all-or-nothing updates and a way to notify user about failed operation

sjd78: >   1. `map()` usually has no side effects - here we use it only for side effects (fetching and mutating)
>

`map()` always turns one set of things into another set of things - in this case, one snapshot to one disks fetch plus one nics fetch.  Mapping over things to get a set of actions to call for each thing is a pretty standard form in sagas. 

I'll make a few changes to remove the `map(entity to callback)` to see if I like how it looks better.

> 
>   2. `all()` could fail on any request (not sure if it will leave the snapshots partially updated/corrupted or not)
> 

This is no different than previous.  All that is changing is fetching them all at the same time instead of iterating though each snapshot.  If the call failed before, the VM's snapshot data would be no better or worse than before this PR.

> 
> How about sth like (pseudo-code)):
> 
> ```
> const [id2nics, id2disks] = yield all([
>    call( all, Object.fromEntries(nonActive.map({id}=> [id, call(fetchVmSnapshotNics,....)]))),
>    call( all, Object.fromEntries(nonActive.map({id}=> [id, call(fetchVmSnapshotDisks,....)]))),
> ])
> ```

Oddly, this would not result in the same make-all-fetches-at-once behavior as I was going for.  Only a flat array or flat object would do that.  Annoying for sure.

> ```
> if(id2nics) {
>    snapshots.forEach(s => s.nics = id2nics[s.id] ?? [])
> } else {
>   //  trigger user notification about nics
> }
> ```
> 
> This should give us all-or-nothing updates and a way to notify user about failed operation

It would take more than that ... probably a try/catch/finally clause to really do that right.  Plus, that is a functional change that introduces a new notification.  I don't think we need to do that kind of change now.  I'll add a TODO note about maybe detecting that, and a few other things that are fetched differently from the full VM.

Ideally, the rest endpoint would actually work with the follows call and remove the need to do these calls all together.

rszwajko (approved):

hstastna (approved):
LGTM :)

src/sagas/index.js
```diff
@@ -461,28 +453,56 @@ export function* fetchVmSnapshots ({ vmId }) {
...
-    //       directly.
-    for (const snapshot of snapshotsInternal) {
-      const follows = yield all([
-        call(fetchVmSnapshotDisks, { vmId, snapshotId: snapshot.id }),
-        call(fetchVmSnapshotNics, { vmId, snapshotId: snapshot.id }),
-      ])
-      snapshot.disks = follows[0]
-      snapshot.nics = follows[1]
-    }
+    snapshotsInternal = snapshots.snapshot.map(snapshot => Api.snapshotToInternal({ snapshot }))
```
sgratch, line 456: @sjd78 
Whenever trying to delete,create a snapshot, the following exception occurs:
image

It seems that Api.snapshotToInternal => Transforms.Snapshot.toInternal.
Please fix on a follow up PR.

Issue #1459, opened 2021-06-22
user settings - support the  'automatic console connection' option


sgratch:
This is a follow up for #1446 and a discussion of the implementation started there: https://github.com/oVirt/ovirt-web-ui/pull/1446#pullrequestreview-684303525 and appears also here https://github.com/oVirt/ovirt-web-ui/pull/1446#issuecomment-859490923 etc.

 The 'Automatic console connection' setting was part of the original design plan https://marvelapp.com/prototype/94h0g56/screen/59286975 and is required here https://bugzilla.redhat.com/show_bug.cgi?id=1781241.

PR #1470, opened 2021-06-28
[WIP] Add autoconnect to Account Settings

rszwajko:
### Connect automatically (overview and tooltip)

image

### Typeahead filtering

image

### Selected VM

image

### Confirmation dialog

image

### No VMs available

image

### Unknown VM (not fetched yet or not existing)

image

image


rszwajko:
@sgratch 
Open topics:
1. fetch missing VMs (not loaded yet on the Dashboard)
2. refactor logic around `ConsoleConfirmationModal` to allow opening console without UI component

src/sagas/console/index.js
```diff
@@ -175,11 +166,20 @@ export function* openConsoleModal (action) {
...
-  yield getConsoleOptions(getConsoleOptionsAction({ vmId: action.payload.vmId }))
+  yield getConsoleOptions(Actions.getConsoleOptions({ vmId: action.payload.vmId }))
+}
+
+export function* autoConnect (vmId) {
+  console.warn('inside autoconnect', vmId)
+  const vm = yield call(fetchSingleVm, Actions.getSingleVm({ vmId, shallowFetch: true }))
+  console.warn('autoconnect fetch', vm)
+  if (vm && vm.id && vm.status !== 'down') {
+    yield downloadVmConsole(Actions.downloadConsole({ vmId, hasGuestAgent: vm.ssoGuestAgent }))
```
sjd78, line 183: `downloadVmConsole()` is a misleading name these days.  Maybe rename that function to `downloadOrOpenVmConsole()` since it will acutally push a use to the novnc console if things are configured that way...

src/components/UserSettings/VmSelect.js
```diff
@@ -0,0 +1,71 @@
+import React, { useState } from 'react'
+import { Select, SelectOption, SelectVariant } from '@patternfly/react-core'
+import { connect } from 'react-redux'
+import { withMsg } from '_/intl'
+import PropTypes from 'prop-types'
+
+const VmSelect = ({ vms, onChange, selected, msg }) => {
+  const [open, setOpen] = useState(false)
+
+  const options = vms.valueSeq().toJS()
```
sjd78, line 10: So looking at the PF4-react docs, you give the select the FULL set of possible values and let the component itself filter the list based on the input text.  That's a bit weird.

It would be nice to have a slimmed down VM fetch that only does "fetch all of the user's VMs, but only the names and ids".  Then it would be independent of the full VM details list.  And if it can be lazy fetched after the user clicks or starts typing a name, that would be nice.  The fetch could also add a search by name and require a user to type at least 1 letter before the fetch actually happens.

sjd78 (changes requested):
Only a few comments so far. I don't see anything problematic with the approach.

I vote +1 to ignore the potential to have a auto connect VM id stored in local storage from some number of years ago.  Make the users pick the auto connect VM anew from this new config option.

Finally, it would be nice to have a visual indicator on the VM Card list for the VM that is flagged to auto connect.  Something like a different color background on the card, or a different top border line accent color, or an icon in the corner. 

Icons :https://www.patternfly.org/v4/guidelines/icons
Potential icons: `DesktopIcon`. `ExternalLinkAltIcon`, `EyeIcon`

src/sagas/options.js
```diff
@@ -86,12 +87,29 @@ function* fetchUserOptions (action: Object): any {
...
+  if (isLogin && !locale && persistLocale) {
     // locale is not saved on the server and locale persistence is enabled
     yield put({ type: C.EXPORT_LOCALE })
   }
+
+  if (isLogin) {
+    // TODO
+    // read & remove autoconnect from local storage
+    // export to the server if value exist
+  }
```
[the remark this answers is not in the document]
rszwajko: @sjd78  @sgratch 
OK. Should we extend that to other local storage based flags that stopped being supported together with `autoconnect` (smartcard, fullscreen, etc) ?

src/components/UserSettings/VmSelect.js
rszwajko: > So looking at the PF4-react docs, you give the select the FULL set of possible values and let the component itself filter the list based on the input text.

Yes. This gives the best user experience when filtering the list - server side query will never keep up with typing speed.

> It would be nice to have a slimmed down VM fetch that only does "fetch all of the user's VMs, but only the names and ids". 

Oh yes:)

> And if it can be lazy fetched after the user clicks or starts typing a name, that would be nice. The fetch could also add a search by name and require a user to type at least 1 letter before the fetch actually happens.

IMHO the query is too slow for that. 

I see 2 solutions:
1. use [view more](https://www.patternfly.org/v4/components/select#view-more-with-checkboxesl) feature to let the user load next chunk of data (similar to infinite scroller) . The feature is beta and not available in our version of react-core (but PF4 is releasing fast).
2. lazy load all VMs when user enters console tab. This is not very elegant but assuming that at least 20 VMs are loaded then how many more (on avarage) require fetching? 20? 40?

src/sagas/options.js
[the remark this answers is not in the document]
sjd78: I vote +1,000

src/components/UserSettings/VmSelect.js
sjd78: > I see 2 solutions:
> 
>   1. use [view more](https://www.patternfly.org/v4/components/select#view-more-with-checkboxes) feature to let the user load next chunk of data (similar to infinite scroller) . The feature is beta and not available in our version of react-core (but PF4 is releasing fast).
> 

That would be nice for sure, but we always need to be careful about pf4-react breaking changes.  They've happened all to frequently in the past.

>
>   2. lazy load all VMs when user enters console tab. This is not very elegant but assuming that at least 20 VMs are loaded then how many more (on avarage) require fetching? 20? 40?
>

That would be a nice thing to know, but ovirt/RHV doesn't collect statistics from installs so we can't really know for sure.  The load more option with displaying what has already been fetched (which if they hard refreshed on options would be 0 VMs) sounds like the best way to go.

src/actions/types.js
```diff
@@ -12,6 +12,7 @@ export type LoadUserOptionsActionType = {
 export type SaveGlobalOptionsActionType = {
   type: C.SAVE_GLOBAL_OPTIONS,
   payload: {|
+    autoconnect?: string,
```
sgratch, line 15: autoconnect => autoConnect

sgratch (changes requested):
@rszwajko 
Please see inline comments below.

> Open topics:
> 
>     1. fetch missing VMs (not loaded yet on the Dashboard)

IMPOV we don't have to handle this. I agree with Scott's comment:
"The load more option with displaying what has already been fetched (which if they hard refreshed on options would be 0 VMs) sounds like the best way to go."


>     2. refactor logic around ConsoleConfirmationModal to allow opening console without UI component
> 
IMPOV it's not that important to handle as well for now. If the console is already used or sso is not supported then the user will need to interactively click the confirmation dialog.
If it is notified by an event (don't remember) then we can also dismiss the conformation in that case.

Few more issues to consider

3. If opening the console fails after logged in (VM is invalid, VM is headless, console is invalid etc) need to consider if to inform the user and how.

4. Do we treat Pool's VMs as regular VMs in this context? I think we should.

src/components/UserSettings/VmSelect.js
rszwajko: > Why not marking the the selection as invalid or clear the field?

In some case it should be possible but generally missing details does not mean that the VM is invalid. VM Portal is just not able to retrieve the details at the given moment.

rszwajko:
> > 
> > 2. refactor logic around ConsoleConfirmationModal to allow opening console without UI component
> > 
> 
> 
> IMPOV it's not that important to handle as well for now. If the console is already used or sso is not supported then the user will need to interactively click the confirmation dialog.

PR #1482 is needed for that. Confirmation dialog is currently tied to VmAction buttons.

> 
>     1. If opening the console fails after logged in (VM is invalid, VM is headless, console is invalid etc) need to consider if to inform the user and how.

PR #1482 should handle that.

> Do we treat Pool's VMs as regular VMs in this context? I think we should.

Not sure if it was supported before but we can give a try.

src/ovirtapi/types.js
```diff
@@ -232,6 +232,7 @@ export type GlobalUserSettingsType = {|
 |}
 
 export type RemoteUserOptionsType = {|
+  autoconnect?: UserOptionType<string>,
```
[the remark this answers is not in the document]
rszwajko: @sjd78  can you help:) ?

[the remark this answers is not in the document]
sjd78: `autoconnect` makes sense and is self-consistent through all of the code

src/actions/types.js
sjd78: I think in this context `autoconnect` as-is works ok.  No need to change.

sjd78 (approved):
Code LGTM, and It works pretty well too.

Would be nice if the toast message for "no console since the VM isn't running" was a bit more descriptive.  Suggestion: "Cannot open console for {vmName}, VM is not running."  It is not a blocking problem.

The toast message for discarding an invalid autoconnect VM is good.

src/ovirtapi/types.js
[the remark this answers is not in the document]
sgratch: it's not consistent with other options convention if name consists of multiple words but ok.

sgratch (changes requested):
@rszwajko 
Generally looks great. Few issues:

1. Currently the console is opened automatically after every browser refresh since we can't distinguish between real user login to hard refresh action. The original feature is meant for automatically connecting to a single running virtual machine after logged in only. It might be very annoying that a console is opened per each browser opening/refresh. I suggest checking with @Artur Socha regarding the options to identify the login state. This is a basic requirement for this feature so let's check the options to solve and then decide if to fix now or later.

2. Error notification - in case of a console error that triggers a toast notification , the error is not persisted anywhere. Since opening the console is an auto operation and if the user missed this error, there is no way for him to track the reason then It worth either displaying the last error as a text info within the auto console settings or as part of events drawer (vanished after refresh but better than nothing for now). 

3. When autoconnect is set for a VM and the VM is removed, the next autoconnect fails on `deleteUserOption` with the following error:
image
and vmPortal.autoconnect property is not removed from DB.

4. This is a minor side issue that will be solved once issue 1 above will be handled:
   When noVnc console dialog is the current opened dialog, the auto connected console after hard refresh is not always opened  properly since after hard refresh, the original refresh is activated in addition to opening the auto console and those two are running    simultaneously and might cause problems.
  For example use case for reproducing: 
  a. set console settings -> preferred console type to noVnc.
  b. set console settings -> Connect automatically to be opened for VM `a`
  c. open noVnc console for VM `a` e.g. vm/a/console/BrowserVnc and check it's opened properly.
  d. hard refresh the browser

   Expected result: noVnc console for VM a should be opened successfully
    Actual result: noVnc console is failed to open with "Disconnected from Console" error since duplicated post requests for the proxyticket and ticket are sent to websocket proxy server.

   BTW, strange behavior exists even when current opened console is noVnc for VM `a` and auto connect is set for noVnc of VM `b`. On that case the order is not consistent since both consoles are opened and sometimes console of VM `a` is the one that is opened last.

rszwajko:
> 1. Currently the console is opened automatically after every browser refresh since we can't distinguish between real user login to hard refresh action. The original feature is meant for automatically connecting to a single running virtual machine after logged in only. It might be very annoying that a console is opened per each browser opening/refresh. I suggest checking with @arso regarding the options to identify the login state. This is a basic requirement for this feature so let's check the options to solve and then decide if to fix now or later.
>

It seems not possible with current SSO but lets check it with @arso 
 
> 2. Error notification - in case of a console error that triggers a toast notification , the error is not persisted anywhere. Since opening the console is an auto operation and if the user missed this error, there is no way for him to track the reason then It worth either displaying the last error as a text info within the auto console settings or as part of events drawer (vanished after refresh but better than nothing for now).
> 

Actually I was convinced that the notification drawer **will** contain the error messages. This is a bug (or missing feature) in the notification framework. Can we fix it as a follow-up issue?

> 3. When autoconnect is set for a VM and the VM is removed, the next autoconnect fails on `deleteUserOption`
>

Fixed by skipping undefined options (when looking for the option to be deleted)

> 4. This is a minor side issue that will be solved once issue 1 above will be handled:
>        When noVnc console dialog is the current opened dialog, the auto connected console after hard refresh is not always opened  properly 

Fixed by disallowing autoconnect in this scenario.

sgratch:
@rszwajko 

> > 1. Currently the console is opened automatically after every browser refresh since we can't distinguish between real user login to hard refresh action. The original feature is meant for automatically connecting to a single running virtual machine after logged in only. It might be very annoying that a console is opened per each browser opening/refresh. I suggest checking with @arso regarding the options to identify the login state. This is a basic requirement for this feature so let's check the options to solve and then decide if to fix now or later.
> 
> It seems not possible with current SSO but lets check it with @arso

Let's wait a day or two for Artur to respond.

> 
> > 1. Error notification - in case of a console error that triggers a toast notification , the error is not persisted anywhere. Since opening the console is an auto operation and if the user missed this error, there is no way for him to track the reason then It worth either displaying the last error as a text info within the auto console settings or as part of events drawer (vanished after refresh but better than nothing for now).
> 
> Actually I was convinced that the notification drawer **will** contain the error messages. This is a bug (or missing feature) in the notification framework. Can we fix it as a follow-up issue?

Note that other  "internal" event messages like for example `actionFeedbackShutdownVm` does appear on notification drawer while console event messages aren't. It seems like a problem with our usage of triggering autoconnect for APP_CONFIGURED rather than a general issue since if changing APP_CONFIGURED action to any other for line:

`takeEvery(C.APP_CONFIGURED, autoconnect)`

then the event does appear on drawer. 

> 
> > 1. When autoconnect is set for a VM and the VM is removed, the next autoconnect fails on `deleteUserOption`
> 
> Fixed by skipping undefined options (when looking for the option to be deleted)

+1

> 
> > 1. This is a minor side issue that will be solved once issue 1 above will be handled:
> >    When noVnc console dialog is the current opened dialog, the auto connected console after hard refresh is not always opened  properly
> 
> Fixed by disallowing autoconnect in this scenario.

+1

rszwajko:
> > > 1. Error notification - in case of a console error that triggers a toast notification , the error is not persisted anywhere. Since opening the console is an auto operation and if the user missed this error, there is no way for him to track the reason then It worth either displaying the last error as a text info within the auto console settings or as part of events drawer (vanished after refresh but better than nothing for now).
> > 
> > 
> > Actually I was convinced that the notification drawer **will** contain the error messages. This is a bug (or missing feature) in the notification framework. Can we fix it as a follow-up issue?
> 
> Note that other "internal" event messages like for example `actionFeedbackShutdownVm` does appear on notification drawer while autoconnect console event messages aren't. 

Fixed. The issue was caused by deleting all events (local and server) when fetching events from the server.

arso:
> @rszwajko
> 
> > > 1. Currently the console is opened automatically after every browser refresh since we can't distinguish between real user login to hard refresh action. The original feature is meant for automatically connecting to a single running virtual machine after logged in only. It might be very annoying that a console is opened per each browser opening/refresh. I suggest checking with @arso regarding the options to identify the login state. This is a basic requirement for this feature so let's check the options to solve and then decide if to fix now or later.
> > 
> > 
> > It seems not possible with current SSO but lets check it with @arso
> 
> Let's wait a day or two for Artur to respond.

@rszwajko @sgratch I have only had a quick look so far but I already have an theory to verify. There is an oauth token info endpoint that potentially could be used here to distinguish between 'login' and 'refresh' actions. My assumption is based on the information that is carried within its response that tells whether current session (token) is 'active'. However, I need to confirm it (most likely that will happen tomorrow).

sgratch:
> > > > 1. Error notification - in case of a console error that triggers a toast notification , the error is not persisted anywhere. Since opening the console is an auto operation and if the user missed this error, there is no way for him to track the reason then It worth either displaying the last error as a text info within the auto console settings or as part of events drawer (vanished after refresh but better than nothing for now).
> > > 
> > > 
> > > Actually I was convinced that the notification drawer **will** contain the error messages. This is a bug (or missing feature) in the notification framework. Can we fix it as a follow-up issue?
> > 
> > 
> > Note that other "internal" event messages like for example `actionFeedbackShutdownVm` does appear on notification drawer while autoconnect console event messages aren't.
> 
> Fixed. The issue was caused by deleting all events (local and server) when fetching events from the server.

LGTM. There are other issues with events handling (old events suddenly appear, clear events failed with insufficient permission0 bu those are not related to this PR and will be handled separately.

sgratch:
> @rszwajko @sgratch I have only had a quick look so far but I already have an theory to verify. There is an oauth token info endpoint that potentially could be used here to distinguish between 'login' and 'refresh' actions. My assumption is based on the information that is carried within its response that tells whether current session (token) is 'active'. However, I need to confirm it (most likely that will happen tomorrow).

@arso 
Can you please take a look on this?

rszwajko:
> > @rszwajko @sgratch I have only had a quick look so far but I already have an theory to verify. There is an oauth token info endpoint that potentially could be used here to distinguish between 'login' and 'refresh' actions. My assumption is based on the information that is carried within its response that tells whether current session (token) is 'active'. However, I need to confirm it (most likely that will happen tomorrow).
> 
> @arso
> Can you please take a look on this?

As checked with @arso we don't have the requested functionality in the current SSO implementation. However we can base our logic on session age i.e. for assume first login if `sessionAge < 5sec`. The required data can be easily provided by server side as part of `userInfo` object - see [this patch](https://gerrit.ovirt.org/c/ovirt-engine/+/116709) for details.
Note that since we don't have that capability currently in VM Portal this is an enhancement (different PR?).

sgratch:
> As checked with @arso we don't have the requested functionality in the current SSO implementation. However we can base our logic on session age i.e. for assume first login if `sessionAge < 5sec`. The required data can be easily provided by server side as part of `userInfo` object - see [this patch](https://gerrit.ovirt.org/c/ovirt-engine/+/116709) for details.
> Note that since we don't have that capability currently in VM Portal this is an enhancement (different PR?).

This looks as a quick and good solution and can definitely be implemented in a follow up PR. 
Not sure how it behaves for invalid/non active sessions which are younger than 5 secs old. It seems that it will behave correctly since checked as part of backend patch and even if won't, it's still better than what we currently have.

sjd78:
> > As checked with @arso we don't have the requested functionality in the current SSO implementation. However we can base our logic on session age i.e. for assume first login if `sessionAge < 5sec`. The required data can be easily provided by server side as part of `userInfo` object - see [this patch](https://gerrit.ovirt.org/c/ovirt-engine/+/116709) for details.
> > Note that since we don't have that capability currently in VM Portal this is an enhancement (different PR?).
> 
> This looks as a quick and good solution and can definitely be implemented in a follow up PR. Not sure how it behaves for invalid/non active sessions which are younger than 5 secs old. It seems that it will behave correctly since checked as part of backend patch and even if won't, it's still better than what we currently have.

Should we set this PR as blocked until the session enhancements are ready?

Once any SSO changes are available, changes will need to be made in 2 places to be used.  One is in the dev server startup script and the other is in the index.jsp generated for the prod/war build.  Both will need to be tested out.

Before those are done, would it make any sense just to use a "session" aged cookie or local storage to track the "has the autoconnect been done for this session" check?  It wouldn't be "perfect" but it would be better than autoconnect on every refresh (which is what would happen in the first implementation of autoconnect in VM Portal).

rszwajko:
> Should we set this PR as blocked until the session enhancements are ready?
> 

IMHO it's not a blocker. I see it as enhancement. Looking at the last working version I don't see any code that would differentiate between first login and subsequent logins. Actually, if users were happy with first (legacy) implementation then they may even consider this feature a regression. 

Details: https://github.com/oVirt/ovirt-web-ui/blob/56b134fd0462fcd380c057a209e703d0a817ade8/src/saga/login.js#L84

> Once any SSO changes are available, changes will need to be made in 2 places to be used. One is in the dev server startup script and the other is in the index.jsp generated for the prod/war build. Both will need to be tested out.
> 

Ack

> Before those are done, would it make any sense just to use a "session" aged cookie or local storage to track the "has the autoconnect been done for this session" check? 

This would be only noticed by dev team - both this PR and the proposed enhancement are planned for 4.5.

> It wouldn't be "perfect" but it would be better than autoconnect on every refresh (which is what would happen in the first implementation of autoconnect in VM Portal).

@sgratch @sjd78 
Do we want to be consistent with first (legacy) implementation here? If yes then we should probably double check how it behaves and if needed provide compatibility mode (i.e. disabling the new feature).

sjd78:
> @sgratch @sjd78 Do we want to be consistent with first (legacy) implementation here? If yes then we should probably double check how it behaves and if needed provide compatibility mode (i.e. disabling the new feature).

That feature has been gone long enough that I think we're free to be unencumbered by what was there before.

sgratch:
@rszwajko @sjd78 @hstastna 
I think we can merge this PR now since it addressed all relevant feature requirements except the SSO enhancement (for identifying first page load after login versa subsequent page reloads).

This feature is targeted to 4.5 so we still have time for the SSO enhancement on web-ui side and no temporary solutions of cookies/local storage is needed.

Nevertheless, I do think it's important to fix that SSO issue regardless to older (legacy) implementation, since it's very annoying to face a console dialog every refresh/reload and since original feature deals with login only. Therefore I will leave [issue 1459](https://github.com/oVirt/ovirt-web-ui/issues/1459) open till this will be implemented on a follow up PR. 
@rszwajko please remember to handle this.

Issue #1448, opened 2021-06-14
Fail to read query string locales pt_BR, zh_CN from URL (landing page locales)


sgratch:
Reading the following locales from web-ui url fails: locale=pt_BR, locale=zh_CN.

For reproducing:
1. open engine landing page and select one of the following languages: Portuguese or Chinese.
2. make sure that this is the user's first login or that user settings->Advanced options->persisted language is disabled

Expected result: web-ui should load the chosen language by reading it from the url e.g.` ...ovirt-engine/web-ui/?locale=pt_BR`
Actual result: for those 2 languages, reading from URL failed

Seems like a format issue since the language code is expected to include a dash instead of underscore (e.g. pt-BR instead of pt_BR).

PR #1471, opened 2021-06-28
Load Portuguese or Chinese languages by reading locales from the url

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1448

Load the chosen language successfully in web-ui after reading query string locale 'pt_BR' or 'zh_CN' from the URL (landing page locales).

**Before:**
 before

**After:**
after

### DIFF
--- a/src/intl/initialize.js
+++ b/src/intl/initialize.js
@@ -64,11 +64,16 @@ export function coerceToSupportedLocale (locale: ?string): ?string {
     return 'en'
   }
 
-  if (BASE_LOCALE_SET.has(locale)) {
-    return locale
+  let commonLocale = locale.replace('_', '-')
+  const localeArray = commonLocale.split(/[-]/)
+  if (localeArray.length === 2) {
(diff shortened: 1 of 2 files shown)


sjd78 (changes requested):
Create a `initialize.test.js` file like `index.test.js` in that same folder and add some unit tests to make sure the function is doing what it is supposed to do.  That'll make it easier to verify your solution will work with other languages.

Tests should cover stuff like:
  - null to null
  - en_US, en_us, en-US, en-us, en to en
  - pt to null
  - pt-PT to null
  - pt_BR, pt_br, pt-BR, pt-br to pt-BR
  - nv to null
  - maybe some variations on the zn-CH locale...

sgratch (approved):
LGTM (just add unit tests as mentioned by Scott)

sjd78 (approved):
LGTM

sgratch (approved):

isaranova:
LGTM

PR #1475, opened 2021-07-07
refactor - remove unused sagas utils

sjd78:
Remove a few utility saga functions that are no longer in use.  We have better alternatives than what they provide.


rszwajko (approved):

hstastna (approved):
LGTM :)

PR #1476, opened 2021-07-07
Don't display manual refresh or setting actions in the header/navbar until the app is ready

sjd78:
Just like how the main `PageRouter` is not displayed until the app is marked as ready to run, do not show the web-ui app header/navbar actions that require the app to be ready until the app is ready.  The manual refresh action and the global settings action should not be available until the app is ready to run.

Note: The app is marked as ready to run as the last part of the login saga.

### DIFF
--- a/src/components/VmsPageHeader/index.js
+++ b/src/components/VmsPageHeader/index.js
@@ -18,7 +18,7 @@ import { Tooltip } from '../tooltips'
 /**
  * Main application header on top of the page
  */
-const VmsPageHeader = ({ onRefresh }) => {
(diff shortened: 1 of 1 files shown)


sjd78:
Example with network throttling to be able to notice the delay:
ezgif com-gif-maker (3)

The refresh and settings icons do not display until the loading indicator goes away and the VM list is rendered.

rszwajko (approved):

hstastna (approved):
LGTM, works as expected :)

sgratch (approved):
LGTM

Issue #1477, opened 2021-07-08
Edit an exising disk via VM details always resulted with an inactive disk


sgratch:
Go to VM details of an existing VM with a disk and edit the disk (by either change the bootable flag or by changing disk size) and click apply.
Actual result: the disk state always becomes inactive.
Expected result: the disk state stays active.

This is a regression for 1.7.0 version that doesn't reproduce for 1.6.9.

The root cause seems to be here: https://github.com/oVirt/ovirt-web-ui/blob/e5aa3f393b97d8f6a9ec6fcc0173855f901e3add/src/ovirtapi/transform.js#L585 caused by this fix https://github.com/oVirt/ovirt-web-ui/pull/1423/files#diff-83343b99424fb2a62ce7723a77810ebe7d4f3ccf8fd53ad4cd92c244902dc80a

If disk.active is undefined then toApiBoolean(disk.active) returns false.

We need to go over all places in code replaced with the toApiBoolean and check that we return the correct default value in case undefined

PR #1478, opened 2021-07-10
Boolean transform for API - keep undefined as undefined

sjd78:
Fixes: #1477

For boolean values to be sent to the REST API, the following values are expected:
  - `undefined` stays `undefined` ... no explicit value = no key in the JSON REST request = no change to the value
  - defined and truthy becomes the string 'true'
  - defined and falsey becomes the string 'false'

This change adds `undefined` handling.  `undefined` needs to stay as `undefined` not get converted to an explicit 'false'.

### DIFF
--- a/src/ovirtapi/transform.js
+++ b/src/ovirtapi/transform.js
@@ -57,8 +57,24 @@ function convertBool (val: ?string): boolean {
   return val ? val.toLowerCase() === 'true' : false
 }
 
-function toApiBoolean (value: any): ApiBooleanType {
-  return value ? 'true' : 'false'
+/**
+ * Convert the given value to an oVirt REST API boolean.  The output can be one
+ * of three states:
+ *
+ *   - `undefined`: If the value is `undefined` the output is `undefined`.  An `undefined`
+ *     value does not get sent in a request, and the backend will fill in the value from
+ *     the entity's existing value (or template value if it is a new entity).
+ *
+ *   - `"true"`: If the value is defined and truthy `true`, send the string "true" to
+ *     match the API's expected `true` value.
+ *
+ *   - `"false"`: If the value is defined and truthy `false`, send the string "false" to
+ *     match the API's expected `false` value.
+ *
+ * @returns undefined | "true" | "false"
+ */
+function toApiBoolean (value: any): ApiBooleanType | void {
+  return value === undefined ? undefined : value ? 'true' : 'false'
 }
 
 function convertInt (val: ?(number | string), defaultValue: number = Number.NaN): number {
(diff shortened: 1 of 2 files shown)


rszwajko (changes requested):
1. `undefined` **is** falsy in JS  -  redefinition of this contract might lead to bugs
2. changed signature triggers flow errors

IMHO we should just pass the correct value instead of  relying on (non-standard) behaviour of our REST API (partial update should be PATCH not PUT)

sjd78:
>   1. `undefined` **is** falsy in JS  -  redefinition of this contract might lead to bugs
> 

I can't recall a place where an `undefined` for a boolean value was intended to be `false`.  Values over the api need at least 2 options... `undefined` to not send as part of the request, or the actual value to send.  Just so happens that booleans have custom handlers on the rest side that forces to the strings `'true'` and `'false'`.  :shrug: 

>   2. changed signature triggers flow errors
> 
> 
> IMHO we should just pass the correct value instead of relying on (non-standard) behaviour of our REST API (partial update should be PATCH not PUT)

Does our REST API support PATCH over PUT for partial updates?  If it doesn't, it is a moot point.  If it does, then great, let's figure that one out in future.  I'm not looking for perfect here, just working as expected.

rszwajko:
> I can't recall a place where an `undefined` for a boolean value was intended to be `false`

[convertBool](https://github.com/oVirt/ovirt-web-ui/blob/8d9c8aec9653000acc5b0c135876af4788b50781/src/ovirtapi/transform.js#L56) works this way when parsing the response **from** the server

sjd78:
> > I can't recall a place where an `undefined` for a boolean value was intended to be `false`
> 
> [convertBool](https://github.com/oVirt/ovirt-web-ui/blob/8d9c8aec9653000acc5b0c135876af4788b50781/src/ovirtapi/transform.js#L56) works this way when parsing the response **from** the server

True, but we don't have a problem with data coming from the server, just going out to the server.  The `undefined` fix changes the behavior back to what it used to do anyway.  The rest of the app really only sets values on a POST that we want to be changed.  They're all partial updates in some sense. Plus, the model-api has basically everything as being optional, so it still work.

I'll fix the type errors -- void is perfectly acceptable for the relevant attributes according to the schema definitions on the rest entity objects.

sjd78:
@rszwajko, @hstastna, @sgratch 

I fixed the type errors by making the attributes optional.  This matches the api schema `maxOccurs="1" minOccurs="0"` defined for those attributes.

I didn't change `toApiBoolean()` itself, but I did:
  - Add a better comment to the function describing its behavior
  - Verify that the other places it is used will not be broken by this change

BTW, did you know you can fetch the RSDL and XSD for the api directly from the api?
  - https://engine.fqdn/ovirt-engine/api?rsdl
  - https://engine.fqdn/ovirt-engine/api?schema

I don't care about POST vs PUT at the moment.

hstastna (approved):
Interesting change, LGTM 👍🏽

sgratch (approved):
LGTM

@rszwajko 

> The Flow errors seem fixed. IMHO convertBool and toApiBoolean should be symmetric (following falsy nature of undefined in JS) but I leave it others to decide.

I generally agree but here we just fix it to work as before. Is there any option that the data coming from the server can be set as undefined? Based on api schema it can be either "false" or "true" so there shouldn't be such a problem. 

We can open a follow up PR to fix that if you think it's important.

isaranova:
LGTM

PR #1486, opened 2021-07-15
refactor `callExternalAction()` to log the API method with `Function.name`

sjd78:
Given:
  1. The `Function.name` attribute on every function will contain the function's name in the running source.

  2. The `TerserPlugin` webpack plugin allows us to keep class and function names instead of minimizing/mangling the names.

  3. The webpack production build configures `TerserPlugin` to keep class and function names intact.

The `callExternalAction()` helper saga can now use `Function.name` when logging the API call instead of relying on the callee passing in the method name as a string.  This saves having to maintain `methodName` and `method` everywhere `callExternalAction()` is used.


sgratch (approved):
LGTM

PR #1487, opened 2021-07-15
Remove `NewDiskDialog` actions/reducers

sjd78:
Remove some old `NewDiskDialog` actions, constants, and reducers that are no longer used.  The VM details `DisksCard` component, create VM `Storage` components, and the disk create sagas don't need them anymore.


sgratch (approved):
LGTM

Issue #1431, opened 2021-05-17
Calculation of CPU topology is incorrect for some vCPU values


sjd78:
There are some instances when changing the vCPU value of a VM (either in Create VM Basic Settings or in VM Details, Details card edit) that the topology is set incorrectly.

For example, for the vCPU value of **383**, the topology comes out as 1 socket, 1 core, 1 thread:
![VM Portal - Google Chrome_033](https://user-images.githubusercontent.com/3985964/118524795-8a236c80-b70c-11eb-96cb-6d96bd15fef2.png)

----

This issue was recently raised in another PR:

> This fix looks good to me except 2 small issues:
> 
>  1. Not related directly to this fix so we can open a separate issue for this:
>        the VCPU Topology on create vm basic settings step doesn't always correlated with Total Virtual CPUs. This still occurs, 

That is an interesting problem where the function that generates the topology falls down for some vCPU counts.  I'll open a new issue on that.

_Originally posted by @sjd78 in https://github.com/oVirt/ovirt-web-ui/issues/1381#issuecomment-840840827_

PR #1489, opened 2021-07-20
Display error message for total VCPU values which can't be factored

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1431

This PR allows to dispaly error message
- in the VM Details card and
- in the Create VM Wizard's Basic settings in Advanced options

for:
- total VCPU values which can't be factored,
- for some other buggy scenarios regarding incorrectly selected VCPU topology values from the drop downs, as there can be many combinations for the number of sockets, cores, threads, and we need to respect the maximum allowed values together with the product of the number of sockets, cores, threads which should be equal to the number of the total VCPUs.

TODO: (DONE)
- [x] disable _Save_ button for invalid value(s) - total VCPUs or regarding the VCPU topology, in the Create VM Wizard

**Before:**
VM Details card:
vmdetails_before1
The next picture shows an example of one of the buggy scenarios in the VM Details where after setting value `241` as the number of total VCPUs, there can be only two options in the drop down for the number of cores. If the user chooses number `1` in the drop down, the topology values become incorrect because of the limits of the sockets and cores don't allow to choose `241` in the appropriate drop downs (same scenario reproducible also in the Create VM Wizard): 
vcpu_issue
Create VM Wizard:
createvm_before1
createvm_before2

**After:**
after_factor
select_topology
afterr
wizard_after2


src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -543,7 +589,7 @@ class BasicSettings extends React.Component {
             />
           </FieldRow>
 
-          <FieldRow label={msg.cpus()} id={`${idPrefix}-cpus`} required>
+          <FieldRow label={msg.cpus()} id={`${idPrefix}-cpus`} required validationState={indicators.cpu || indicators.topology} errorMessage={!vCpuCountIsFactored ? msg.cpusBadTopology() : ''}>
```
sjd78, line 592: Break the attributes to one per line so it is easier to read.

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -679,33 +725,49 @@ class BasicSettings extends React.Component {
...
+              fieldCols={3}
+              label={msg.virtualSockets()}
+              id={`${idPrefix}-topology-sockets`}
+            >
               <SelectBox
                 id={`${idPrefix}-topology-sockets-edit`}
                 items={this.mapVCpuTopologyItems(vCpuTopologyDividers.sockets)}
                 selected={data.topology.sockets.toString()}
                 onChange={selectedId => { this.handleChange('topology', selectedId, { vcpu: 'sockets' }) }}
+                disabled={indicators.cpu}
```
sjd78, line 738:
```suggestion
                disabled={!!indicators.cpu}
```

This is causing an error on browser console:
> Warning: Failed prop type: Invalid prop `disabled` of type `string` supplied to `SelectBox`, expected `boolean`.

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -679,33 +725,49 @@ class BasicSettings extends React.Component {
...
+              fieldCols={3}
+              label={msg.coresPerSockets()}
+              id={`${idPrefix}-topology-cores`}
+            >
               <SelectBox
                 id={`${idPrefix}-topology-cores-edit`}
                 items={this.mapVCpuTopologyItems(vCpuTopologyDividers.cores)}
                 selected={data.topology.cores.toString()}
                 onChange={selectedId => { this.handleChange('topology', selectedId, { vcpu: 'cores' }) }}
+                disabled={indicators.cpu}
```
sjd78, line 752:
```suggestion
                disabled={!!indicators.cpu}
```

src/components/CreateVmWizard/steps/BasicSettings.js
```diff
@@ -679,33 +725,49 @@ class BasicSettings extends React.Component {
...
               tooltip={threadsTooltip}
+              validationState={indicators.topology}
+              errorMessage={!topologyValuesAreValid && vCpuCountIsFactored ? msg.cpusBadTopologySelection() : ''}
             >
               <SelectBox
                 id={`${idPrefix}-topology-threads-edit`}
                 items={this.mapVCpuTopologyItems(vCpuTopologyDividers.threads)}
                 selected={data.topology.threads.toString()}
                 onChange={selectedId => { this.handleChange('topology', selectedId, { vcpu: 'threads' }) }}
+                disabled={indicators.cpu}
```
sjd78, line 769:
```suggestion
                disabled={!!indicators.cpu}
```

sjd78 (changes requested):
Functionally looks good.  A few code style and a few prop type issues that need to be resolved.

sjd78 (approved):
LGTM

sgratch (commented):
Please check the following 2 issues:
**issue 1)** Regarding the validation of cpus topology:

select_topology

Can you please explain how do we get to this scenario in 1st place? 
On webadmin for example, once selecting total vcpus to be 241, the topology is auto selected to V, C, T of: 1, 241, 1 and can't be changed since there are no options for changing it. Why letting the user with options that we already know that are non valid? 

Another example is that for setting total cpus to be 340 for example, the user can select the following:
image

while on webadmin it is blocked since the options are more limited since cores num started from 17:
image



In addition, this PR doesn't cover all cases anyways since few are still verified and fail on backend:
image


Bottom line is that I don't think we need to cover all cases and it's ok to fail on backend. The important thing is to avoid creating a VM with one vcpu due to invalid architecture. I think we need to prefer backend failures as much as possible and just filter invalid options in advance where possible instead of current solution of displaying errors on ui for part of bad topology selection. The only important thing is to avoid creating a VM in that case. We just need to filter those options out within the selection box values.
The current solution causes strange behaviors. 


~~**issue 2:** I remember we raised this in the past and it's not solved by this PR and I'm still not sure if solved by pr 1491 as well (if it is solved then please ignore this):~~

~~when manually entering a total vcpu value (not necessarily the upper limit value), the cpu arch is sometimes auto set to invalid values.~~
~~For example, set the "total vcpus" to 33 and then remove one digit to leave only 3 value and the following error appears:~~

~~image~~
or:
~~image~~


~~As discussed, this is reproduced only for create VM and not for edit VM and seems related to the UI fields input manipulation.~~ <-- fixed by PR 1491 so not relevant anymore.

hstastna:
> Please check the following 2 issues:
> **issue 1)** Regarding the validation of cpus topology:
> 
> select_topology
> 
> Can you please explain how do we get to this scenario in 1st place?
> On webadmin for example, once selecting total vcpus to be 241, the topology is auto selected to V, C, T of: 1, 241, 1 and can't be changed since there are no options for changing it. Why letting the user with options that we already know that are non valid?

@sgratch, because this requires dealing with all the possible combinations of the numbers of sockets, cores, threads in advance. This PR is not about implementing this functionality. Not sure how this is implemented in webadmin but I consider it as not trivial. In this PR, a simple check is added and error message for invalid values of total VCPUs or VCPU topology.

> Another example is that for setting total cpus to be 340 for example, the user can select the following:
> image
> 
> while on webadmin it is blocked since the options are more limited since cores num started from 17:
> image

Yeah, that makes sense. It is because the user has chosen a value from some drop down (number of cores in this case) for which there isn't any valid combination with other possible values. They should pick another value from the drop down or to choose a different value as the number of total VCPUs. This is why the drop downs and the Total VCPUs field have the red border.
Originally, the error message said that the user should pick another value, as there are all the possible values present in the drop downs, but I was suggested to change the error message to how it is now.

> In addition, this PR doesn't cover all cases anyways since few are still verified and fail on backend:
> image

Yes. it is very probable that this PR doesn't cover all the buggy cases. But it covers more than https://github.com/oVirt/ovirt-web-ui/issues/1431. Anyway the example above feels like the validation of the maximum values does not work the best. 

> Bottom line is that I don't think we need to cover all cases and it's ok to fail on backend. The important thing is to avoid creating a VM with one vcpu due to invalid architecture. I think we need to prefer backend failures as much as possible and just filter invalid options in advance where possible instead of current solution of displaying errors on ui for part of bad topology selection. The only important thing is to avoid creating a VM in that case. We just need to filter those options out within the selection box values.
> The current solution causes strange behaviors.

Hmm, I haven't found anything strange. Except that the error messages are not 100% clear, IMHO.

> **issue 2:** I remember we raised this in the past and it's not solved by this PR and I'm still not sure if solved by pr 1491 as well (if it is solved then please ignore this):
> 
> when manually entering a total vcpu value (not necessarily the upper limit value), the cpu arch is sometimes auto set to invalid values.
> For example, set the "total vcpus" to 33 and then remove one digit to leave only 3 value and the following error appears:
> 
> image
> or:
> image
> 
> As discussed, this is reproduced only for create VM and not for edit VM and seems related to the UI fields input manipulation.

Yep, I was able to reproduce, but only sometimes. It appears after I press Enter a couple of times :D

src/intl/messages.js
```diff
@@ -120,6 +120,8 @@ export const messages: { [messageId: string]: MessageType } = {
   },
   coresPerSockets: 'Cores per Virtual Socket',
   cpus: 'Total Virtual CPUs',
+  cpusBadTopology: 'No valid topology exists for this vCPU count.',
```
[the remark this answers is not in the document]
hstastna: I don't feel it is more clear, but I can change that for sure ;)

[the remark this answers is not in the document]
sgratch: I don't think we need to discuss that with UX folks :)

My suggestion is just to make the message suggested a bit less technical since simple users doesn't necessarily know what `vCPU`means (doesn't appear as a field on dialogs) and which topology is not valid on the basic step dialog? CPU pinning topology? 

Here is the error as appears on webadmin:
"Incorrect number of Total Virtual CPUs. It is not possible to compose this number from the available Virtual Sockets and Cores per Virtual Sockets"

Since we don't have so much space for such an informative message, we shorten it a bit but note that there are no shortcuts and message is very informative.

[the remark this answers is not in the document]
sjd78: Sharon's suggestions make sense.

hstastna:
> > In addition, this PR doesn't cover all cases anyways since few are still verified and fail on backend:
> image

@sgratch this is weird because I am pretty sure those values are not above the maximum values, in this example. I was able to reproducew but it does not make sense. I suggest to create a separate issue for that. Not sure if it is the UI issue.

src/intl/messages.js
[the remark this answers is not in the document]
hstastna: Done ;)

sgratch:
> Yeah, that makes sense. It is because the user has chosen a value from some drop down (number of cores in this case) which is not valid according to other chosen values. The user should pick another value from some drop down or to choose a different value as the number of total VCPUs. This is why the drop downs and the Total VCPUs field have the red border.
> Originally, the error message said that the user should pick another value, as there are all the possible values present in the drop downs, but I was suggested to change the error message to how it is now.

The error for invalid total cpus is ok. 
The issue is regarding the error handling for the cpu architecture values. We needed to assume the user chooses sockets, cores and threads on that order up to down and filter out the invalid values in advance or just make the VM creation to fail on backend with the correct error for all cases.

The current scenario is that the user can choose all kinds of combinations while the values are dynamically changed when selecting different values in different order without any logic. If values are invalid then it failed on backend or it is turned to 1, 1, 1 or other invalid non chosen values with the error displayed but without letting the user remember what he tried to set. 
For example- the user set values to 1, 1, 384 but error appears for 16,1,8
Peek 2021-08-03 16-43

in addition, current web-ui blocks options like in case of total vcpus set to 9 and topology is set to S=1, C=3, T=3.
This is possible on webadmin:
image

 But is not possible on web-ui:
Peek 2021-08-03 17-17

So there are 2 issues here:
1. The current cpu topology options selection is not fully supported and working a bit strange. This is not part of this fix so we can open a separate issue for this.

2. Error handling for detected invalid CPU architecture:
If we decide to leave the current solution of displaying an error then please rephrase the error to:
"Selected CPU topology is invalid" (instead of vCPU)
And check if possible to display a fixed (empty) or the original invalid value in case of this error occurrence instead of switching to a different invalid values that the user didn't select.  


> Yep, I was able to reproduce, but only sometimes. It appears after I press Enter a couple of times :D

This is not reproduced to me anymore so seems fixed by PR 1491

> this is weird because I am pretty sure those values are not above the maximum values, in this example. I was able to reproduce, but it does not make sense. I suggest to create a separate issue for that. Not sure if it is the UI issue. You can see the maximum possible values of the number of sockets, cores, threads in the browser console in the picture below:

The combination of  340 (1, 170, 2)  is invalid on webadmin as well so it's totally ok that it failed on backend with our current solution.

hstastna:
> The error for invalid total cpus is ok.
> The issue is regarding the error handling for the cpu architecture values. We needed to assume the user chooses sockets, cores and threads on that order up to down and filter out the invalid values in advance...

I think the https://github.com/oVirt/ovirt-web-ui/issues/1431 is about weird topology values according to the total VCPUs value, not about changing the user experience. The user always could choose sockets, cores, threads randomly, not in any order, and the other values (other than actually changed value) were always changed according to the user's change and the number of total VCPUs.

> ... or just make the VM creation to fail on backend with the correct error for all cases.

I am not sure how I can achieve this here in the UI. First, the backend has to fail only when it should fail. Seems that for some cases this doesn't happen.

> The current scenario is that the user can choose all kinds of combinations while the values are dynamically changed when selecting different values in different order without any logic. If values are invalid then it failed on backend or it is turned to 1, 1, 1 or other invalid non chosen values with the error displayed but without letting the user remember what he tried to set.

Yes, it seems this is the original design. From my experience, it always worked like that. And when it turns to 1,1,1, it is because it is the only possible combination of values, according to their maximum. But unfortunately inconsistent with the total VCPUs value. And this was missing here: some check if the product of those values = number of total VCPUs.

> For example- the user set values to 1, 1, 384 but error appears for 16,1,8
Peek 2021-08-03 16-43

That's how the behavior is designed. After every change, the other values are changed, too, to respect maximum values and the number of total VCPUs. The number of sockets/threads change when changing the number of cores to 1. It seems it is not possible to set 1,1,whatever values, for 384 total VCPUs. We would have to have possibility to set 384 threads, but maximum is much less.

> in addition, current web-ui blocks options like in case of total vcpus set to 9 and topology is set to S=1, C=3, T=3.

Yes, this IS the problem, but different than we are dealing with in this PR. it always behaved like that. It follows from the design. The same for the VM Details card.

sgratch:
@hstastna 
> So there are 2 issues here:
> 
>     1. The current cpu topology options selection is not fully supported and working a bit strange. This is not part of this fix so we can open a separate issue for this.

Based on your reply, please open a separate new issue for this inconsistency and gaps with setting the cpu architecture (with the example for setting  S=1, C=3, T=3 for total vcpus=9 and maybe other examples raised here). 
Not sure if UX shoud help since it's a functional bug on the way we implement this but let's see.


> 
>     2. Error handling for detected invalid CPU architecture:
>     If we decide to leave the current solution of displaying an error then please rephrase the error to:
>     "Selected CPU topology is invalid" (instead of vCPU)
>     And check if possible to display a fixed (empty) or the original invalid value in case of this error occurrence instead of switching to a different invalid values that the user didn't select.
> 

Based on your reply - do you mean that we can't fix this use case?:

> values to 1, 1, 384 but error appears for 16,1,8

So the error appears but the CPU topology marked in red is switched to a suggested other topology which is also wrong but different than the original user's entering values?
If we can't display the original invalid topology set by the user then at least we can make sure that the cpu topology marked in red in case of a topology error is always empty or set to 0, 0, 0 or even 1, 1, 1.

hstastna:
>  And check if possible to display a fixed (empty) or the original invalid value in case of this error occurrence instead of switching to a different invalid values that the user didn't select

I have only a few questions on this: How this would work? How and when the remaining values would be chosen after the user makes some choice? If ever. When the validation of those values would happen?

It doesn't switch to a different values because of the error but because those values are the best possible values, according to the maximum allowed values (for the number of sockets, cores, threads).
It is possible to change everything, of course, but I think it is not a good idea. Not here and right now. We would have to change the behavior of that feature, to answer all of my questions above. I think such a change should be discussed with the UX folks and solved by a different issue/PR - it can take some time.

Actually it works the way that as soon as the user chooses some value in the drop down, the other values are chosen according to the user's selected value, with respect to the maximum allowed values and to the total VCPU value. Then the validation of all the chosen values happens because sometimes it is not possible to choose the values that would meet all the required conditions. Such a validation was missing there. If we wanted the validation before the two remaining values are chosen (+ the chosen values would be fixed), the error would appear immediately and very often. Only after the user chooses the third value corresponding to the maximum values and the total VCPUs, the error would disappear. I consider this quite annoying. Or we can have validation of those values later, only after the user clicks on Save button. But I consider this as a step back as we can check the values immediately and to display the corresponding errors. This would be also inconsistent with some other already present checks there, for example checking the maximum allowed value of the total VCPUs.

> So the error appears but the CPU topology marked in red is switched to a suggested other topology which is also wrong but different than the original user's entering values?

Again: user chooses some value. The rest of the values is set according to the chosen value. Then the validation, then the error.

> If we can't display the original invalid topology set by the user then at least we can make sure that the cpu topology marked in red in case of a topology error is always empty or set to 0, 0, 0 or even 1, 1, 1.

I am not sure if I understand. It depends on the user choice of the values, incl. the total VCPU. The Next or Save button is disabled in case of the invalid values.

sgratch:
> 
> @sgratch, I have only a few questions on this: How this would work? How and when the remaining values would be chosen after the user makes some choice? If ever. When the validation of those values would happen?

Let me summarize my concerns and suggestions:
As far as I checked, the backend is running all cpu topology validations for us  so in case of a wrong cpu topology, a backend error is diaplyed e.g. "Cannot add VM. Maximum number of threads per cpu exceeded" will be displayed. 

The main problem with our current topology calculations/validation logic s that it validates user input for vcpu=X and if not succeeded then:
1.  either switch to 1,1,1 (which leads to a VM created with vcpu=1 <> X) 
2. or switched to a suggested other topology which might also fail or include a result in which the product of the number of sockets, cores, threads is consistent with the number of the total VCPus (which leads to a VM created with vcpu != X).

So the backend is working correctly and the only problem that I can see is with this whole auto calculation for cpu topology.


As a quick solution to make it less confusing till refactoring might be done, we can either:
1. Cancel this whole auto calculation for cpu topology and let the VM created with the original user cpu topology selection. If wrong then it will fail on backend with the correct error.

2. Leave the code as is for now but at least avoid those jumping/invalid re-calculations by displaying a consistent values in case of a vcpu topology error. e.g.:
Peek 2021-08-05 14-47

This is easy to implement - if validateTopologyValues() returns false, display the error and just set the topology values to -1 for example

hstastna:
@sjd78 Any thoughts about this?
> Leave the code as is for now but at least avoid those jumping/invalid re-calculations by displaying a consistent empty values in case of a vcpu topology error. e.g.:

@sgratch I can't imagine avoiding those jumping/invalid re-calculations without canceling the whole auto calculation. But we can add another "jumping", by adding the condition you've already mentioned, after the validation. Those re-calculations happen before validation and it makes sense to me like that.

sgratch:
@hstastna 

> @sgratch I can't imagine avoiding those jumping/invalid re-calculations without canceling the whole auto calculation. But we can add another "jumping", by adding the condition you've already mentioned, after the validation. Those re-calculations happen before validation and it makes sense to me like that.

If it's not easy to reset those fields after the validation failure since it will break the behavior when choosing invalid values in a raw, then let's not invest more time on this now, merge this PR and open a follow up issue for the vCPU problems that were mentioned above.
Please open the issue before merging so that we won't forget.

src/intl/messages.js
```diff
@@ -120,6 +120,8 @@ export const messages: { [messageId: string]: MessageType } = {
   },
   coresPerSockets: 'Cores per Virtual Socket',
   cpus: 'Total Virtual CPUs',
+  cpusBadTopology: 'No valid CPU topology exists for this total Virtual CPUs count.',
+  cpusBadTopologySelection: 'Selected vCPU topology is invalid.',
```
sgratch, line 124: It was somehow hidden between all posts but as mentioned on https://github.com/oVirt/ovirt-web-ui/pull/1489#issuecomment-891911743

> 2. Error handling for detected invalid CPU architecture:
> If we decide to leave the current solution of displaying an error then please rephrase the error to:
> "Selected **CPU** topology is invalid" (instead of vCPU)

sgratch (commented):
LGTM, just suggest to rephrase the message.

A follow up issue was opened for open problems: #1503.

sgratch (approved):
LGTM

Issue #1473, opened 2021-07-07
Created VM has incorrect number of CPUs


isaranova:
Description of problem:
VM is created with incorrect number of CPUs that does not match the one in VM Wizard.

VM Portal version number:
ovirt-web-ui-1.7.1-0.20210706.gite5aa3f3.el8.noarch

oVirt version number:
ovirt-engine-4.4.8-0.0.master.20210706164306.git8e41812880d.el8.noarch

Browser and version:
firefox-82.0.2-1.fc31.x86_64

Steps:
1) Create VM with maximum number of CPUs (384 for my cluster)

![Screenshot_2021-07-07 VM Portal(3)](https://user-images.githubusercontent.com/39263962/124767783-67f1e200-df38-11eb-812b-d25f10915169.png)

3) Go to the VMs details and check the number of CPUs
The number of CPUs is only 1:

![Screenshot_2021-07-07 VM Portal(4)](https://user-images.githubusercontent.com/39263962/124767870-793aee80-df38-11eb-951e-7728fa6f4a26.png)

Actual results:
Wrong number of CPUs.

Expected results:
Correct number of CPUs or an error message.

Additional info:
On the previous version of webui, there was an error message, that such number of CPUs was not possible. However, the VM could still be created with that number, if user went back to the first step and clicked on the up arrow for the CPUs number (the number itself didn't change, stayed the maximum).

sjd78:
Very strange, if you type in 384 (the max vCPUs on a 4.4 compat version cluster), the topology fields are correct.  Therefore the CPU count will be saved properly.

If you then up arrow to try to go to 385, the change is prevented.  HOWEVER, the topology is broken.  The topology calc at the edge of the  range is not doing the right thing.

![ezgif com-gif-maker (2)](https://user-images.githubusercontent.com/3985964/124795045-a75f0500-df1d-11eb-81ee-759bd3537fef.gif)

sgratch:
This seems somehow related to https://github.com/oVirt/ovirt-web-ui/issues/1431
Assigning to Hilda for checking this as well.

PR #1491, opened 2021-07-27
Create VM with the correct number of the VCPUs, in CreateVmWizard

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1473

The issue was introduced together with implementing [advanced CPU topology ](https://github.com/oVirt/ovirt-web-ui/pull/1279/files#diff-ecc9f96259bb1bc0ee42cc7ff4de2121caf4ac2bfecc640b2cc798f976565489R328) feature in the _Basic Settings_ step of the _CreateVmWizard_.
 
After clicking on up arrow for the _Total Virtual CPUs_ field, `handleChange` was called and the next code was executed:
```javascript
      case 'cpus': {
        const { maxNumOfVmCpus } = this.grabCpuOptions()
        if (isNumberInRange(value, 0, maxNumOfVmCpus)) {
          changes.cpus = +value
        }
        changes.topology = this.getTopologySettings(changes.cpus)
        break
      }
```
If we reach the max value of the total VCPUs, this value does not change anymore when trying to increase it. But the change in the topology was always "detected", new values set. This was the issue. VCPU Topology values depend on the number of the total VCPUs so the topology values shoud change only with the change of the number of the total VCPUs. Inappropriate change of the topology values lead to the creation of the VM with incorrect number of the total VCPUs. 

**Before:**
before1
before2

**After:**
after1
after2

### DIFF
--- a/src/components/CreateVmWizard/steps/BasicSettings.js
+++ b/src/components/CreateVmWizard/steps/BasicSettings.js
@@ -316,9 +316,8 @@ class BasicSettings extends React.Component {
         const { maxNumOfVmCpus } = this.grabCpuOptions()
         if (isNumberInRange(value, 0, maxNumOfVmCpus)) {
           changes.cpus = +value
+          changes.topology = this.getTopologySettings(changes.cpus)
         }
-
-        changes.topology = this.getTopologySettings(changes.cpus)
         break
       }
 


sjd78 (approved):
LGTM

sgratch (approved):
LGTM

isaranova:
LGTM

PR #1493, opened 2021-07-29
add babel plugin fancy-console

sjd78:
Sometimes the logging enhancements done in `src/logger.js` breaks some error logging.  This happen in react error logging where the `console.error()` string substitution characters are used in react code.  To step around this problem, `src/logger.js` is paired back to just enable turning console logging on and off and babel is used to add fancy output to the various logging functions.  The babel plugin will enhance every `console.(log|info|warn|error)` encountered in the code it processes.  Dependency code won't be touched, or will be properly enhanced and retain original logging functionality.

The fancy logging is only applied in development (using the default enhanced console css features for a browser runtime) and test (using chalk strings for a nodejs runtime) builds.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,7 @@ EXTRA_DIST = \
 	ovirt-web-ui.spec.in \
 	README.md \
 	LICENSE \
+	babel-plugin \
 	branding \
 	src \
 	config \
@@ -38,6 +39,8 @@ EXTRA_DIST = \
 
 DISTCLEANFILES = $(PACKAGE)-$(VERSION).tar.gz \
 	aclocal.m4 \
+	config.log \
+	config.status \
 	configure \
 	install-sh \
 	missing \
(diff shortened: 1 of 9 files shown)


sjd78:
Same thing in ovirt-engine-ui-extensions in patch https://gerrit.ovirt.org/c/ovirt-engine-ui-extensions/+/115975.

sjd78:
**NOTE**: Babel output is cached, so for existing/developer environments, the cache may need to be cleared before the change from dynamic/logger.js logging enhancements to static/babel logging enhancements take effect:

```
  rm -rf node_modules/.cache
```

sjd78:
ci test

rszwajko (approved):
@sjd78 very interesting code!

sjd78:
I think CI may be failing because nodejs is version 10 (via ovirt-engine-nodejs-modules) and my local env on centos 8 is nodejs 12 and on fedora 33 is nodejs 14.  I'll see about bumping the nodejs version there so this works.

Running `yarn test` on f33 or centos8 works just fine.

Running with mock_runner fails just like CI.

PR #1495, opened 2021-08-04
when creating vm, send quota as empty object to rest-api instead of null

saifabusaleh:
When adding new vm from template from web-ui, there is no
option to select quota, so it takes always the template quota.

This caused situation that user can't add a VM if not
authorized for the template quota, without having the
option to change it.

This change changes web-ui to send empty quota object to rest-api, so this situation can be handled properly in rest-api and backend  

BZ url: https://bugzilla.redhat.com/1959186


sgratch (changes requested):
@saifabusaleh 
Two comments:

- Please squash those twp dup commits into one: 
"when creating vm, send quota as empty object to rest-api instead of null"
- in case the user is originally granted with the template's quota (either since he is an admin or since he granted that quota in 1st place) then the VM should be created with the template's quota instead of the 1st quota assigned to that user. 
This is most important for admins which use web-ui and usually have assigned with all quotas. The web-ui in that case should behave the same as webadmin. 
The fix can be done either on web-ui or rest api side.

ahadas:
>     * in case the user is originally granted with the template's quota (either since he is an admin or since he granted that quota in 1st place) then the VM should be created with the template's quota instead of the 1st quota assigned to that user.
>       This is most important for admins which use web-ui and usually have assigned with all quotas. The web-ui in that case should behave the same as webadmin.

That's not necessarily true
But anyway, we cannot really have everything without supporting quota properly in the vm portal
So if the admin wants that, the admin should create the VM from the webadmin..

ahadas:
@saifabusaleh please update the BZ url to https://bugzilla.redhat.com/show_bug.cgi?id=1991240 - I don't want us to spend time on over-specification so let's do our part and if this gets more complicated, the UX team can handle the VM portal side

saifabusaleh:
> @saifabusaleh please update the BZ url to https://bugzilla.redhat.com/show_bug.cgi?id=1991240 - I don't want us to spend time on over-specification so let's do our part and if this gets more complicated, the UX team can take over the VM portal side

Done.

sgratch:
@ahadas 
> That's not necessarily true - the template's quota says nothing about the quota of the vm
> But anyway, we cannot really have everything without supporting quota properly in the vm portal
> So if the admin wants that, the admin should create the VM from the webadmin..

As mentioned above. this regression is not relevant for admins only but for regular users as well, if they are assigned with more than one quota.
**Before this fix**: a user that is originally assigned with the template's quota can automatically create a VM via web-ui with that template's quota.
**After this fix:** a user that is originally assigned with the template's quota will always create a VM via web-ui with the first quota assigned to him in the list. He will never be bale to create a VM with other quotas assigned to him, even if they are assigned to the template.

So we fixed one problem but broke another one.
This can be handled on backend by verifying that the template's quota (if exists for non Blank) is assigned to the user before always choosing the first quota. 

If you decide not to fix that on backend then we need to check if it's ok to have this regression,  Maybe we can live with this new behavior unless someone will complain...

> 
> In the old user portal we used to have a view for admins which was not implemented in the vm portal
> So I don't think it's a fair assumption to think that admins really use the vm portal...

With new VM portal both admin/expert and non admin views were combined into one view that should support both type of users. We do know that admins use VM portal and we do allow this as part of the implementation, even though it's mainly used for regular users.
Anyway, the mentioned problem above is relevant for non admins as well.

ahadas:
> > That's not necessarily true - the template's quota says nothing about the quota of the vm
> > But anyway, we cannot really have everything without supporting quota properly in the vm portal
> > So if the admin wants that, the admin should create the VM from the webadmin..
> 
> As mentioned above. this regression is not relevant for admins only but for regular users as well, if they are assigned with more than one quota.
> **Before this fix**: a user that is originally assigned with the template's quota can automatically create a VM via web-ui with that template's quota.
> **After this fix:** a user that is originally assigned with the template's quota will always create a VM via web-ui with the first quota assigned to him in the list. He will never be bale to create a VM with other quotas assigned to him, even if they are assigned to the template.
> 
> So we fixed one problem but broke another one.

I agree that it introdcues a change in behaviour but I wouldn't consider this a regression since it's not a bug
There is no reason to believe that the quota on the template better fits the user - it might be that there's no more space or CPUs available on the template's quota and the first quota of the VM can accommodate the resources of the VM

> This can be handled on backend by verifying that the template's quota (if exists for non Blank) is assigned to the user before always choosing the first quota.

I would rather prefer to avoid this kind of logic - there is really no sense in preferring the template's quota over any other quota that the VM has access to
 
> If you decide not to fix that on backend then we need to check if it's ok to have this regression, Maybe we can live with this new behavior unless someone will complain...

+1
I think it would alright as users typically are not assigned with more than one quota and administrators have no reason to provision VMs from the VM portal (especially if they'd like to set the quota)

To put this effort in context - we never said that it's the best way to go.
The best way is to provide proper support for quota on the client side but as that requires significant amount of work and introduces extra complexity on a client that is designed to be simple, we look for a workaround - supporting quota in a reasonable way in a client that has no notion of quota

> > In the old user portal we used to have a view for admins which was not implemented in the vm portal
> > So I don't think it's a fair assumption to think that admins really use the vm portal...
> 
> With new VM portal both admin/expert and non admin views were combined into one view that should support both type of users. We do know that admins use VM portal and we do allow this as part of the implementation, even though it's mainly used for regular users.

I'd put it differently - instead of two views we now have a single view that doesn't have the advanced options admins were provided with using the user portal :)

sgratch:
> I agree that it introduces a change in behaviour (in case the "first user's quota" is different than the template's quota) but I wouldn't consider this a regression since it's not a bug
> There is no reason to believe that the quota on the template better fits the user - it might be that there's no more space or CPUs available on the template's quota and the first quota of the VM can accommodate the resources of the VM

+1

> I think it would be alright as users are typically not assigned with more than one quota + administrators have no reason to provision VMs from the VM portal (especially if they'd like to have control over the quota) + quota is not that commonly used
> To put this effort in context - we never said that it's the best way to go.
> The best way is to provide proper support for quota on the client side but as that requires significant amount of work and introduces extra complexity on a client that is designed to be simple, we look for a workaround - supporting quota in a reasonable way in a client that has no notion of quota

+1

> I'd put it differently - instead of two views we now have a single view that doesn't have the advanced options admins were provided with using the user portal :)

Which advanced options admins don't have except quota support :)?
The design and implementation should support both so it's not accurate...quota is an exception.

We'll merge this after https://gerrit.ovirt.org/#/c/ovirt-engine/+/116017/ is merged.

sgratch (approved):
LGTM
Should be merged only after https://gerrit.ovirt.org/#/c/ovirt-engine/+/116017/ is merged.

ahadas:
> LGTM
> Should be merged only after https://gerrit.ovirt.org/#/c/ovirt-engine/+/116017/ is merged.

merged

Issue #1440, opened 2021-06-02
user settings - need to inform the user that all sections/tabs actions (save/cancel/reset) are applied to all tabs


sgratch:
This is a follow up issue for https://github.com/oVirt/ovirt-web-ui/pull/1427#issuecomment-851608502
It is not clear for the user to understand that all user settings dialog actions (save/Cancel/Reset settings) are applied to all tabs in once and not only the current user settings tab.
Maybe it worth just mentioning it within a tooltip/confirmation dialog so that the user will be aware of that.

rszwajko:
Partially addressed in #1446 - confirmation dialog shows section titles.

PR #1498, opened 2021-08-10
User settings: Inform the user that cancel action is applied to all tabs

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1440

Allow displaying the additional info in the confirmation dialog for cancel action in the user settings section. Note that the save and reset actions already contain the information that the action is applied to all tabs/settings.

**Before:** (_Cancel_ and _Reset_ actions)
cancel_before
reset_done

**After:**
cancel_after
reset_after

---

**The following screen wasn't changed:**

_Save_ action:
save_done

### DIFF
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -46,6 +46,7 @@ export const messages: { [messageId: string]: MessageType } = {
     description: 'In sense of "human friendly name"',
   },
   allocatedVms: 'Allocated VMs',
+  allTabs: 'This action is applied to all tabs.',
   apiConnectionFailed: 'oVirt API connection failed',
   apiVersionCheckFailed: 'oVirt API version check failed',
   areYouSureYouWantToDeleteDisk: 'Are you sure you want to delete disk {diskName}?',
--- a/src/components/NavigationConfirmationModal/index.js
+++ b/src/components/NavigationConfirmationModal/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { Modal, Button, Alert } from 'patternfly-react'
 import { MsgContext } from '_/intl'
 
-const NavigationConfirmationModal = ({ show, onYes, onNo }) => {
+const NavigationConfirmationModal = ({ show, onYes, onNo, additionalNote }) => {
   const { msg } = useContext(MsgContext)
   const idPrefix = 'close-dialog-confim'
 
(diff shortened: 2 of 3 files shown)


rszwajko (approved):

sgratch (commented):
for cancel action - LGTM
For reset action - I'm not sure if the text "Resetting your account settings **will clear all of the individual settings** that were applied to it...." is clear enough for explaining that all tabs are reset since the user can currently see/manage only one tab and he might think that all settings within this tab only are cleared.
I would add "This action is applied to all tabs" text to this dialog as well but it's up to you if you think it's clear enough.

sjd78:
> for cancel action - LGTM
> For reset action - I'm not sure if the text "Resetting your account settings **will clear all of the individual settings** that were applied to it...." is clear enough for explaining that all tabs are reset since the user can currently see/manage only one tab and he might think that all settings within this tab only are cleared.
> I would add "This action is applied to all tabs" text to this dialog as well but it's up to you if you think it's clear enough.

The current reset message details text is: 

_"Resetting your account settings will clear all of the individual settings that were applied to it. Your account settings will be returned to the default settings. Please confirm that you want to reset your account settings before making these changes."_

It is a bit awkward but passable as the "nuke it all from space" action.

Possible re-write considering the header text _"Are you sure your want to reset your account settings?"_:

"The reset will apply the system defaults to ALL of your settings across all settings tabs.  Please confirm you want to fully reset your account settings."

sjd78 (approved):
Additional note work LGTM

hstastna:
@sgratch The message for the reset action is the original one. I wasn't changing it. I didn't add any extra text to it, as I consider it already so long and containing the info the user needs, IMO :)

hstastna:
Note that the suggested change of the reset message details text is done :)

src/intl/messages.js
```diff
@@ -574,7 +575,7 @@ export const messages: { [messageId: string]: MessageType } = {
   },
   resetSettings: 'Reset settings',
   resetSettingsQuestion: 'Are you sure you want to reset your account settings?',
-  resetSettingsWarning: 'Resetting your account settings will clear all of the individual settings that were applied to it. Your account settings will be returned to the default settings. Please confirm that you want to reset your account settings before making these changes.',
+  resetSettingsWarning: 'The reset will apply the system defaults to ALL of your settings across all settings tabs. Please confirm you want to fully reset your account settings.',
```
sgratch, line 578: I wouldn't touch this original message as rephrased by Laura back then since it will require re-translation for all languages which we currently try to avoid if not necessary.
If you think that leaving this long text and just adding a new clarification sentence `additionalNote` regarding applying to all settings tabs is too long and not essential then let's leave it as is, otherwise let's just add this `additionalNote`. I don't have a strong opinion on that

hstastna: Okay, I am moving the original message back.

sgratch (approved):
LGTM

isaranova:
LGTM

Issue #1505, opened 2021-08-20
I can't download console.rdp on VM Portal with Firefox


bishopd:
I am testing the lastest version of ovirt-engine 4.4.7.7-1 before updating.

oVirt VM Portal Version 1.7.0-1

when I log into the VM portal and want to download the Remote Destop file (console.rdp) from a Windows VM, the file I receive is console.vv instead of console.rdp. So I have to rename the file to console.rdp.
This problem is only with Firefox. Edge and google chrome don't have this problem.

Firefox version 91.0.1

![image](https://user-images.githubusercontent.com/12138756/130268724-3ad8d9ad-3b76-44ef-b2e5-ce074ee20e3d.png)

Thank you

sgratch:
@bishopd The format and content of the rdp downloaded file should be ok and valid, the only issue is with the rdp file name extension which is `.vv` instead of `.rdp`
Can you please approve that this is the case?

bishopd:
that's right.

PR #1507, opened 2021-08-25
Download Remote Destop file with the proper file extension .rdp

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1505

Allow downloading the VM console file _console.rdp_ instead of _onsole.vv_ from a Windows VM also for Firefox browser. The other browsers didn't have the problem with the incorrect file extension.

**Before:**
before

**After:**
after


sjd78 (approved):
LGTM

Issue #1458, opened 2021-06-22
user settings - console options full screen mode should be supported for noVnc as well


sgratch:
This is a follow up  for https://github.com/oVirt/ovirt-web-ui/pull/1446.

As part of user console options, if `"VNC"-"full screen mode"` option is enabled then only VNC consoles are opened on full screen.
We should enable opening noVnc on full screen mode for user settings as well.

This is already supported manually and works well:
![image](https://user-images.githubusercontent.com/18169498/122880248-7b9b2700-d342-11eb-8724-416f05abf858.png)

So just need to enable that for user settings.

hstastna:
@rszwajko @sgratch So does this issue depend on https://github.com/oVirt/ovirt-web-ui/pull/1482 ? Thanks! :)

rszwajko:
> So does this issue depend on #1482 ? 
>

IMHO yes. The refactoring changes the component significantly - it may help solving the problem but also may break the fix prepared for class-based component.

PR #1509, opened 2021-09-01
Support full screen mode for noVnc console in User settings

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1458

In this PR, the _Full screen mode_ Console option for noVnc is added to User settings . When this option is enabled, VNC Browser consoles are opened on full screen. It was possible to open the consoles on full screen manually, now it is possible to enable this option also in User settings.

**Before:**
vnc_before

**After:**
vnc_after


src/components/VmConsole/index.js
```diff
@@ -121,7 +122,7 @@ const VmConsole = ({
     if (isFullScreen) {
       onFullScreen()
     }
-  }, [isFullScreen])
+  }, [isFullScreen, fullScreenNoVnc])
```
rszwajko, line 125: it seems we can remove `fullScreenNoVnc` from dependency list at it is used only to set initial value to  `isFullScreen`

rszwajko (approved):
Looks good!

src/components/VmConsole/index.js
sjd78: Only time I can think of when the component would need to be sensitive to a change to `fullScreenNoVnc` would be if the options change while the console is displayed.  But even then, there would need to be more changes to pick up that change.

@hstastna - I'm ok with `VmConsole` just using the initial value of `fullScreenNoVnc` and not being sensitive to any changes.  It would be weird to have the full screen option get picked up from another tab (or rest call) and have the console toggle between modes all by itself!

sjd78 (commented):
Everything looks good, just need a followup on the `useEffect()` dependency list in `VmConsole`.

src/components/VmConsole/index.js
hstastna: Right, thanks, folks! Seems that we really don't need `fullScreenNoVnc` here, at least for now :D I was thinking what if this value changed while the console being displayed, but as Scott has mentioned, that would require some more changes ;)

sjd78 (approved):
LGTM

sgratch (commented):
@hstastna works great except one issue:
When opening noVnc from the VmConsole dialog itself, by choosing it via console-selector dropdown menu, the fullScreenNoVnc=true settings value is ignored and console is always opened with non full screen mode.

For reproducing:
1. Set fullScreenNoVnc=true via user settings dialog.
2. Open a noVnc console  which will be opened on full screen mode as required.
3. open the console-selector from within the VmConsole and choose Spice/Vnc/noVnc.
4. Note that for Spice/Vnc - console options full screen mode is considered while for noVnc it's not since it will always be opened on non full screen mode:
image

hstastna:
> @hstastna works great except one issue:
> When opening noVnc from the VmConsole dialog itself, by choosing it via console-selector dropdown menu, the fullScreenNoVnc=true settings value is ignored and console is always opened with non full screen mode.
> 
> For reproducing:
> 
>     1. Set fullScreenNoVnc=true via user settings dialog.
> 
>     2. Open a noVnc console  which will be opened on full screen mode as required.
> 
>     3. open the console-selector from within the VmConsole and choose Spice/Vnc/noVnc.
> 
>     4. Note that for Spice/Vnc - console options full screen mode is considered while for noVnc it's not since it will always be opened on non full screen mode:
>        image

Thank you, Sharon, for finding this! I suggest to fix this separately as this is not trivial to fix because of our current implementation.. WDYT? :)

isaranova:
LGTM

PR #1510, opened 2021-09-02
Create VM - default the first disk as bootable

sjd78:
When creating a new VM, if a new disk is the first disk being created,
default it to be bootable.  The next disk will allow resetting the
bootable flag as per normal.  New disks for templates with disks
will continue to work the same way as it does now.

Before when creating the first disk:
Screenshot 2021-09-02 at 16-05-08 VM Portal

After when creating the first disk:
Screenshot 2021-09-02 at 16-06-37 VM Portal


rszwajko (approved):

hstastna (approved):
Small but a useful change 😏  LGTM.

Issue #1504, opened 2021-08-18
VM details - Utilization sparkline charts tooltips include undefined values instead of a real numeric value


sgratch:
This is a regression reproduced already for version 1.7.0-1 (worked properly on 1.6.9-1)

For all  3 utilization sparkline charts within VM details dialog (CPU, memory and Networking), the charts tooltips are displayed with  "`undefined%"` text instead of real numeric values, e.g.:
![image](https://user-images.githubusercontent.com/18169498/129918736-44f430b9-1afd-40d2-b5a1-86d75f93c8a7.png)

Tooltips looked fine for version 1.6.9-1:
![image](https://user-images.githubusercontent.com/18169498/129918764-dc5342a2-3746-4f65-87a8-c437bd420468.png)

PR #1437, opened 2021-05-27
Upgrade PF4 packages and force lodash version

rszwajko:
Force lodash version ^4.17.21

Upgrade packages(together with dependencies):
1. @patternfly/react-charts to6.14.2
2. @patternfly/react-core to 4.97.2
3. @patternfly/react-tokens to 4.10.2

PR #1511, opened 2021-09-10
Display numeric values in Utilization sparkline charts

hstastna:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1504

Display the numeric values in the tooltips correctly, for all 3 Utilization sparkline charts within VM details dialog (_CPU, Memory_ and _Networking_).

After upgrade of the _@patternfly/react-charts_ package, `undefined` was displayed instead of the number. This was due to related changes in the package, `Chart` react component. The problem is related especially to `labels` prop of `ChartVoronoiContainer` used.

Note that without this commit, using the component and any version of the package higher than 5.0.0 (incl.) would lead to the original problem.

_Related sources:_
PF Area Charts: https://www.patternfly.org/v4/charts/area-chart/
package upgrade: https://github.com/oVirt/ovirt-web-ui/pull/1437

**Before:**
util_before

**After:**
util_after


sjd78 (approved):
LGTM

isaranova:
LGTM

sgratch:
@ isaranova 

> Just one thing I'm not sure about - why is there 0% instead of 100%? Shouldn't the graph show the blue value from the chart?

It depends if guest agent is installed or not.
If not installed then there is no statistics info to collect regarding the free guest OS memory  and therefore the historical data will be received from backend as 0's.
And since there is no info regarding free/cached/buffered memory as well then the donut chart memory usage  will always show 100% usage (0% free).

If guest agent installed then the actual sizes will be received for both real used/free memory and historical data.

For verification, please check that the same memory usage info per each such VM appears on webadmin vms grid under the 'Memory' trend column:

image

Based on my test, it seems the same info as appears on webadmin. Therefore, it doesn't seem like a bug to me.

> Also, why is it in percentage and not the correct value (GiB, MiB, etc.)?

The percentage units are displayed for all 3 Utilization sparkline charts as part of the design. The same units as appear on webadmin for this trend data.

isaranova:
Yes, you were right, having the guest agent installed on the VM. No bug then. LGTM

PR #1515, opened 2021-09-24
Refactor - Convert `activeRequests` reducer to ImmerJS

sjd78:
Convert the `activeRequests` reducer and related code from ImmutableJS to ImmerJS.

Avoid the use of `Map` and `Set` ES6 objects for holding state in redux.  They are not easily convertible/serializable to JSON and as a side-effect don't render properly in the redux devtool browser extension.


hstastna (approved):
LGTM :)

src/reducers/activeRequests.js
```diff
@@ -1,18 +1,25 @@
...
 
-const initialState = Set()
+const initialState: ActiveRequestsStateType = []
 
 const activeRequestsReducer = actionReducer(initialState, {
-  [ADD_ACTIVE_REQUEST] (state, { payload: requestId }) {
-    return state.add(requestId)
-  },
+  [ADD_ACTIVE_REQUEST]: produce((draft: ActiveRequestsStateType, { payload: requestId }) => {
+    const flatRequestId: string = JSON.stringify(requestId)
```
rszwajko, line 13: we do we need to stringify the ID?

sjd78: To make sure `draft.includes(flatRequestId)` and `draft.filter(request => request !== flatRequestId)` work no matter what the requestId type is.

rszwajko: I think we can skip stringify - previously we relied on the reference equality and we can continue doing so. However some comment explaining that would be helpful.
Check:
```
const a = { a: 'a' }
const set = Set().add(a)
const arr = [a]

arr.includes({ a: 'a' }) == set.has({ a: 'a' }) == false
arr.includes(a) == set.has(a) == true
```

sjd78: > I think we can skip stringify - previously we relied on the reference equality and we can continue doing so.

So I tried it and realized that relying on `Object.is()` or `===` object ref check will NOT work in this context.  I think it comes down to Immer doing its magic.  The object stored in the produced array is not the same as the array going in.

And of course, it is a pitfill: https://immerjs.github.io/immer/pitfalls#drafts-arent-referentially-equal

rszwajko: OK .But still using stringfy is confusing. Actually naming an object of this type `requestId` is confusing!
How about  destructuring: ` { payload: {method, url})` and later just `draft.find(r => r.url == url && r.method == method)` ?

PR #1517, opened 2021-09-24
Refactor - Convert `dataCenters` reducer to ImmerJS

sjd78:
Convert the `dataCenters` reducer and related code from ImmutableJS to ImmerJS.

Things covered
  - Reducer changed storage from an array to a hash of id to object
  - Adjust components to change from 'array find by id' to a straight
    keyed lookup

No sagas needed to be updated


src/components/VmDetails/cards/DetailsCard/index.js
```diff
@@ -692,7 +692,7 @@ class DetailsCard extends React.Component {
 
     // Data Center
     const dataCenterId = (clusters && clusters.getIn([clusterId, 'dataCenterId']))
-    const dataCenter = dataCenters && dataCenters.find(v => v.id === dataCenterId)
+    const dataCenter = dataCenters && dataCenters[dataCenterId]
```
rszwajko, line 695: consider `const dataCenterName = dataCenters?.[dataCenterId]?.name ?? msg.notAvailable()`

rszwajko (approved):

src/components/VmDetails/cards/DetailsCard/index.js
sjd78: Since the render handles an `undefined` value for `dataCenter`, doing this would be all that is needed:
```suggestion
    const dataCenter = dataCenters?.[dataCenterId]
```

but that's really just a simple style change at that point.  IDK if it makes sense to convert to optional chaining in these cases for everythign we touch going forward or to just leave them in place.

rszwajko (approved):

PR #1518, opened 2021-09-24
Refactor - Convert `roles` reducer to ImmerJS

sjd78:
Convert the `roles` reducer and related code from ImmutableJS to ImmerJS.

Things covered:
  - Convert from Set to Array in redux state to get good output with the redux devtool browser extension
  - Typing added to roles reducers and actions
  - Updated the util saga `entityPermissionsToUserPermits()` as needed

Note: Avoid the use of Map and Set ES6 objects for holding state in redux. They are not easily convertible/serializable to JSON and as a side-effect don't render properly in the redux devtool browser extension.


src/sagas/utils.js
```diff
@@ -168,26 +168,32 @@ export function* entityPermissionsToUserPermits (entity) {
...
     : []
 
-  const userGroups = yield select(state => state.config.get('userGroups'))
-  const userId = yield select(state => state.config.getIn(['user', 'id']))
-  const roles = yield select(state => state.roles)
+  const {
+    userGroups,
+    userId,
+    roles,
+  } = yield select(state => ({
```
rszwajko, line 175: consider: `yield select(({config, roles}) => ...`

src/sagas/utils.js
```diff
@@ -168,26 +168,32 @@ export function* entityPermissionsToUserPermits (entity) {
...
         console.info('Could not find role in redux state, roleId:', permission.roleId)
       } else {
-        permitNames.push(...role.get('permitNames', []))
+        permitNames.push(...role.permitNames)
       }
     }
   }
 
-  return new Set(permitNames)
+  return Array.from(new Set(permitNames))
```
rszwajko, line 196: `permitNames` seems already to be an array

rszwajko (approved):
Regarding this part of the commit message:
> Convert from Set to Array in redux state to get good output with the redux devtool browser extension

There are to additional reasons I've found:
1. in Immer they [need to be enabled](https://immerjs.github.io/immer/installation) explicitly by calling `enableMapSet()`
2. default `combineReducers()` [expects plain objects ](https://github.com/reduxjs/redux/issues/1499#issuecomment-315615439)

Generally it seems safer to use plain objects.

src/sagas/utils.js
sjd78: It is, but `Array.from(new Set(permitNames))` will remove all duplicates in the array.

sjd78:
> Regarding this part of the commit message:
> 
> > Convert from Set to Array in redux state to get good output with the redux devtool browser extension
> 
> There are to additional reasons I've found:
> 
>   1. in Immer they [need to be enabled](https://immerjs.github.io/immer/installation) explicitly by calling `enableMapSet()`

I did that step initially to try it out, so things with immer all worked properly.

> 
>   2. default `combineReducers()` [expects plain objects ](https://github.com/reduxjs/redux/issues/1499#issuecomment-315615439)
> 

Oh I didn't see that one.  I just didn't like not being able to see the data in redux!

> 
> Generally it seems safer to use plain objects.

For sure

hstastna (approved):
I am fine with these changes. LGTM 👍🏽

PR #1529, opened 2021-10-27
Autoconnect console only after login

rszwajko:
Fixes: #1459 

Connect automatically to the console at the first page load after
login but not on subsequent page refreshes.

First page load after login is recognized indirectly based
on the session age. Session older than ~~15~~ 60 sec indicates a page refresh.

### DIFF
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -268,22 +268,22 @@ function* autoconnect () {
 
   const {
     userId,
-    sessionAgeInSecAtPageLoad,
+    firstLogin,
     websocket,
     defaultVncMode,
     preferredConsole,
     autoconnectOption,
   } = yield select(({ config, options }) => ({
     userId: config.getIn(['user', 'id']),
-    sessionAgeInSecAtPageLoad: config.getIn(['user', 'sessionAgeInSecAtPageLoad'], 0),
+    firstLogin: config.getIn(['user', 'firstLogin'], true),
     websocket: config.get('websocket'),
     defaultVncMode: config.get('defaultVncMode'),
     preferredConsole: options.getIn(['remoteOptions', 'preferredConsole', 'content'], config.getIn(['defaultUiConsole'])),
     autoconnectOption: toJS(options.getIn(['remoteOptions', 'autoconnect'])),
   }))
 
-  if (sessionAgeInSecAtPageLoad > 15) {
-    console.warn(`Autoconnect aborted - based on the session age (${sessionAgeInSecAtPageLoad} sec) it's a page refresh not the first page load`)
+  if (!firstLogin) {
+    console.warn('Autoconnect aborted - page refresh detected')
     return
   }
 
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -268,17 +268,24 @@ function* autoconnect () {
 
   const {
     userId,
+    sessionAgeInSecAtPageLoad,
     websocket,
     defaultVncMode,
     preferredConsole,
     autoconnectOption,
   } = yield select(({ config, options }) => ({
     userId: config.getIn(['user', 'id']),
+    sessionAgeInSecAtPageLoad: config.getIn(['user', 'sessionAgeInSecAtPageLoad'], 0),
     websocket: config.get('websocket'),
     defaultVncMode: config.get('defaultVncMode'),
     preferredConsole: options.getIn(['remoteOptions', 'preferredConsole', 'content'], config.getIn(['defaultUiConsole'])),
     autoconnectOption: toJS(options.getIn(['remoteOptions', 'autoconnect'])),
   }))
+
+  if (sessionAgeInSecAtPageLoad > 15) {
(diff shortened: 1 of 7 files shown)


src/sagas/console/index.js
```diff
@@ -268,18 +268,25 @@ function* autoconnect () {
...
   } = yield select(({ config, options }) => ({
     userId: config.getIn(['user', 'id']),
+    sessionAgeInSecAtPageLoad: config.getIn(['user', 'sessionAgeInSecAtPageLoad'], 0),
     websocket: config.get('websocket'),
     defaultVncMode: config.get('defaultVncMode'),
     preferredConsole: options.getIn(['remoteOptions', 'preferredConsole', 'content'], config.getIn(['defaultUiConsole'])),
     autoconnectOption: toJS(options.getIn(['remoteOptions', 'autoconnect'])),
   }))
 
+  if (sessionAgeInSecAtPageLoad > 15) {
```
sjd78, line 285: I suggest making the 15 an item in the config file (`ovirt-web-ui.config` and referenced through `AppConfiguration` in `src/config.js`) to make it easier to change.

sjd78 (commented):
The solution is better than popping the autoconnect on every page reload!  Other than being able to push an attribute to the sso session itself from VM Portal, none of the potential solutions really close the various gaps / edge cases.

The first gap I can think of is if someone logs in on the welcome page with the login link in the top right corner and then they take longer than 15 seconds to click the link to open VM Portal...no autoconnect

The second gap is if they go to Admin portal first and then back out to VM Portal...no autoconnect (assuming the navigation and page loading takes >15 seconds).

rszwajko:
> Other than being able to push an attribute to the sso session itself from VM Portal, none of the potential solutions really close the various gaps / edge cases.
> 

I fully agree. However I see no traces of such functionality in the past. Actually looking at the code I don't see first-login working neither in previous Vm Portal [1] nor in Userportal [2].
It seems also that the functionality changed significantly over time. In previous Vm Portal the VM for autoconnect was chosen explicitly by the user (and stored in local storage).
In Userportal[3] there is no sign of chosing a VM for autoconnect - autoconnect works if there is only one reachable console (note that this matches the behaviour described in the docs).
```
public List<VmConsoles> getAutoConnectableConsoles() {
            // ...    
            for (UserPortalItemModel upItem : items) {
                if (!upItem.isPool() && upItem.getVmConsoles().canConnectToConsole()) {
                    autoConnectableConsoles.add(upItem.getVmConsoles());
                }
            }
        }
        return autoConnectableConsoles;
    }

    public boolean getCanConnectAutomatically() {
        return getAutoConnectableConsoles().size() == 1;
    }
```


> The first gap I can think of is if someone logs in on the welcome page with the login link in the top right corner and then they take longer than 15 seconds to click the link to open VM Portal...no autoconnect
> 
> The second gap is if they go to Admin portal first and then back out to VM Portal...no autoconnect (assuming the navigation and page loading takes >15 seconds).

IMHO those gaps are a blocker.


[1] https://github.com/oVirt/ovirt-web-ui/blob/56b134fd0462fcd380c057a209e703d0a817ade8/src/saga/login.js#L84
[2] https://github.com/oVirt/ovirt-engine/blob/f56fc7fb09e57a67236efe3e0e6aa6c588c9495e/frontend/webadmin/modules/userportal-gwtp/src/main/java/org/ovirt/engine/ui/userportal/utils/ConnectAutomaticallyManager.java#L102
[3] https://github.com/oVirt/ovirt-engine/blob/f56fc7fb09e57a67236efe3e0e6aa6c588c9495e/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/userportal/AbstractUserPortalListModel.java#L60

src/sagas/console/index.js
sjd78: Based on your followup comment to my review - I also suggest reorganizing the check out to be part of the login saga.  During login, have a function whose purpose is to determine if this is the user's first login.  On a page refresh, it'll run again and should turn off the flag.  Store the result in the config reduce somewhere and just check that flag in the console component.

This would do a few nice things.  First, the react component is just doing a straight lookup to redux, no further logic need apply.  Second, a function is established to do the calculation.

That `isFirstLogin()` function can be as nieve as we want to be able to cover whatever cases we want.  Maybe we stick with sso token age.  Maybe we do something with a cookie with an appropriate scope.  Maybe local storage also with a session type scope.  At the very least, we would have a single place to make that change going forward.

sjd78 (approved):
The updates look good.  I'm ok with pushing this forward notwithstanding the gaps identified.

Something to limit autoconnect is better than nothing!

sgratch (commented):
@rszwajko 
Please check minor comments below.

Regarding current gaps mentioned above, maybe increasing the default from 15 sec to a bit larger number (1 min ?) might probably help for the scenario of login via landing page and then opening vm portal after waiting...

Anyway, we can't really avoid those gaps scenarios from happening but I agree with Scott that it's not worth fixing them now. If users will complain then we can enhance this later on.

sgratch:
ci test

sgratch (approved):
LGTM

isaranova:
LGTM

PR #1532, opened 2021-11-09
copr: enable copr builds

sandrobonazzola:
Enable copr builds

Signed-off-by: Sandro Bonazzola <sbonazzo@redhat.com>

PR #1534, opened 2021-11-16
Enable copr build and refactor build to better support the copr build

sjd78:
## Enable copr build
The copr build should be setup using the [SCM `make srpm` method](https://docs.pagure.org/copr.copr/user_documentation.html#scm).  Anyone can setup a copr project, point the SCM at their github fork and have copr do automatic builds on your fork.

The base change is from PR #1532

## Refactor the build to better support the copr build.

After refactoring, the build now supports:
  - offline (ovirt-engine-nodejs-modules) builds (OFFLINE_BUILD=1)
  - online (yarn download) builds (OFFLINE_BUILD=0)
  - source only (for copr) release builds
  - source only (for copr) snapshot builds

spec file refactoring:
  - BuildRequires all necessary packages to actually run the rpm build.
  - `%prep` only unpacks sources
  - `%build` now does both `yarn install` and build
  - `SKIP_AUTOGEN=1` prevents double calling `autogen.sh` on the build's `yarn install`

**Note**: BuildRequires packages are installed by the copr chroots before the script is run.  This is peer functionality to stdci's `*.packages` files.

**Note**: By default, in copr all builds have access to the rpms already built in the same project.  An explicit repo url to access ovirt-engine-nodejs-modules is not needed as long as ovirt-web-ui and ovirt-engine-nodejs-modules are both in the same project.

build refactoring:
  - Removed yarn detection by `configure.ac` as it was blocking source only builds.  Inclusion via `BuildRequires` in the spec file is sufficient.
  - Updated PATH in Makefile
  - copr Makefile doesn't need to install nodejs-modules for srpm build


sjd78:
ci test

sjd78:
ci test

sjd78:
ci test

sjd78:
ci test

package.json
```diff
@@ -125,7 +125,7 @@
     "ua-parser-js": "^0.7.22"
   },
   "scripts": {
-    "preinstall": "test -e ./autogen.sh && ./autogen.sh --prefix=/usr --datarootdir=/share || :",
+    "preinstall": "scripts/yarn-preinstall.sh",
```
rszwajko, line 128: can we instead check for artifacts generated by `autogen.sh` ?

sjd78: TL;DR - No. There is deep dark magic going on here!

Basically, we need a balance about where and how to run autogen.sh in a number of different scenarios...
  - as a developer running `yarn install` after a repo clone or rebase
  - `yarn install` in ovirt-engine-nodejs-modules where only package.json and yarn.lock are available
  - under `make` either manually or by the build system

So, `autogen.sh` may not be available / not needed (nodejs-modules) or we always want to run it (developer) or we run it external to `yarn install` and can skip running it (build system).  It's all dark magic.

For the developer scenario, we only need to config files to be substituted in by autogen.  See:
https://github.com/oVirt/ovirt-web-ui/blob/4aee8f2646ed05296dda2bd164a6582edcaeb6b9/configure.ac#L40-L44

Moving it to a regular shell script makes it much easier to control when a `yarn install` will call `autogen.sh`.

rszwajko: OK. Please describe the reasons (mentioned above) in a comment in `yarn-preinstall.sh`

sjd78: I had to add the `test -e` back to the package.json script to support inclusion in nodejs-modules, but left the rest of the code as-is.  Added a big fat comment in yarn-preinstall.sh.

rszwajko (approved):
Looks good to me however I have a limited knowledge of how our build works. 
Some thoughts:
1. The comments in `yarn-preinstall.sh` were very helpful in understanding various build paths. It would be nice to describe also other parts of our build in a similar way(a follow-up PR?)
2. +1 for removing outdated project informoation!

PR #1536, opened 2021-11-19
Handle network errors during page refresh

rszwajko:
Steps to recreate:
1. enter main VM Portal page with VMs/pools
2. disable networking (i.e. using dev tools)
3. trigger manual refresh
Expected result: error notifications are displayed
Actual result: VM Portal crashes

Fixes:
1. error.status == 0  - if network is down the reported status code is
   zero which bypasses some of our checks. Fixed by returning
   the entire error object.
2. temporary network error causes VMs/pools to be removed from
   the dashboard. Fixed by removing only if the server reported 404.
3. use common filtering routine to process VMs and pools retrieved
   during page refresh

### DIFF
--- a/src/sagas/index.js
+++ b/src/sagas/index.js
@@ -203,32 +203,31 @@ export function* fetchSingleVm (action) {
 
   action.payload.additional = shallowFetch ? VM_FETCH_ADDITIONAL_SHALLOW : VM_FETCH_ADDITIONAL_DEEP
   const vm = yield callExternalAction(Api.getVm, action, true)
-  const error = vm.error ? vm.error.status ?? 'fetch-error' : null
-
-  let internalVm = null
-  if (vm?.id) {
-    internalVm = yield transformAndPermitVm(vm)
-
-    // If the VM is running, we want to display the current=true cdrom info. Due
-    // to an API restriction, current=true cdrom info cannot currently (Aug-2018)
-    // be accessed via the additional fetch list on the VM. Fetch it directly.
-    if (!shallowFetch && internalVm.status === 'up') {
-      internalVm.cdrom = yield fetchVmCdRom({ vmId: internalVm.id, current: true })
-    }
-
-    if (!shallowFetch) {
-      yield parallelFetchAndPopulateSnapshotDisksAndNics(internalVm.id, internalVm.snapshots)
-    }
+  if (vm?.error) {
+    return { vmId, error: vm.error }
   }
 
-  return { vmId, internalVm, error }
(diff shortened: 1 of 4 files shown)


src/sagas/index.js
```diff
@@ -258,21 +255,21 @@ function* fetchAndPutPools (action) {
...
   const { internalPool, error } = yield fetchSinglePool(action)
 
-  if (error) {
-    if (error === 404) {
-      yield put(updateVms({ removePoolIds: [action.payload.poolId] }))
-    }
-  } else {
+  if (error?.status === 404) {
+    yield put(updateVms({ removePoolIds: [action.payload.poolId] }))
+  } else if (!error && internalPool) {
```
sjd78, line 272: I prefer the old if structure to the new one ... `if (error) { do errors } else { do success }`.

src/sagas/index.js
```diff
@@ -203,35 +203,32 @@ export function* fetchSingleVm (action) {
...
   const { internalVm, error } = yield fetchSingleVm(action)
 
-  if (error) {
-    if (error === 404) {
-      yield put(updateVms({ removeVmIds: [action.payload.vmId] }))
-    }
-  } else {
+  if (error?.status === 404) {
+    yield put(updateVms({ removeVmIds: [action.payload.vmId] }))
+  } else if (!error && internalVm) {
```
sjd78, line 231: I prefer the old if structure to the new one ... `if (error) { do errors } else { do success }`.

sjd78 (changes requested):
Looks good in general.

sjd78 (approved):
LGTM

isaranova:
Unfortunately, I'm still experiencing this issue. The pool is still visible in the VM portal, but it does not happen always. Once a new pool is created in the Admin portal, it "replaces" the already removed pool in the VM portal.

Screenshot 2022-03-21 at 16-51-47 VM Portal
Screenshot 2022-03-21 at 16-50-25 VM Portal

rszwajko:
@isaranova 
> Unfortunately, I'm still experiencing this issue. The pool is still visible in the VM portal, but it does not happen always.
>

Problem can be re-created if there is a single pool in the system and that pool will be deleted.

sgratch:
> The problem will be tracked under #1573

@isaranova @rszwajko can we close this original issue as done since Radek opened a new one?

isaranova:
> > The problem will be tracked under #1573
> 
> @isaranova @rszwajko can we close this original issue as done since Radek opened a new one?

Yes

PR #1542, opened 2021-12-07
Support engine with HTTP in dev mode

rszwajko:
Fixes regression in `start.js` which prevents obtaining user info from engine if HTTP protocol is used.

### DIFF
--- a/scripts/start.js
+++ b/scripts/start.js
@@ -118,9 +118,11 @@ async function getUserInfo () {
 
   console.log('Connecting using provided credentials...')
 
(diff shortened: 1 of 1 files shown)


sjd78 (approved):
FWIW - every ovirt install has a `https` endpoint, even the source built dev installs.

Is there any particular reason why you need `http` other than the old `start.js` script supported it?

rszwajko:
> Is there any particular reason why you need http other than the old start.js script supported it?

Personal preference. From time to time I use plain HTTP when working with my dev engine.

PR #1546, opened 2021-12-10
Remove nullish fields from GET

rszwajko:
Enforce convention that unset fields are not present in the response.


sjd78 (commented):
Seems reasonable.  A few comments:

  - Probably should be on the other http method functions as well
  - I'd like to see some unit tests for `removeNullish()`, especially since it touches every request/response

sgratch:
Let's wait to see if this will be fixed by infra/backend or not and then decide how to proceed with this fix.

sgratch:
> Let's wait to see if this will be fixed by infra/backend or not and then decide how to proceed with this fix.

It should be fixed for 4.5 by Infra team: https://bugzilla.redhat.com/show_bug.cgi?id=2032917

PR #1547, opened 2021-12-13
Github Actions to implement CI / check-patch

sjd78:
Add `check.yaml` to replicate the std-ci **check-patch** stage
of CI.  This will run the FULL build from sources to srpm to rpm.
The full build includes linting, unit tests, dist build and rpm
packaging.

Two jobs are setup.  The first will run the full build in offline
mode using ovirt-engine-nodejs-modules.  The second will run an
online build directly downloading yarn dependencies.

Note: If the online build succeeds but the offline build fails, then
the **ovirt-engine-nodejs-modules** project probably needs to be
updated with a pre-seed for the PR being worked on.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,0 +1,92 @@
+#
+# CI for this project needs to do two things:
+#   1. Setup the environment for and run `yarn build`
+#   2. Build the distribution rpm for use in QE/OST/Integration testing
+#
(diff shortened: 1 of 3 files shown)


sjd78:
Replacing with #1548 since PRs from forks won't run the actions with current repo settings.  PRs from branches in the repo will run actions.

PR #1549, opened 2021-12-23
Migrate Create VM Wizard to PF4

rszwajko:
Depends on #1543 

Final part of the PF3 -> PF4 migration. The only remaining PF3 elements are:
1. `react-console` package - details in https://github.com/oVirt/ovirt-web-ui/pull/1525/commits/aa1e22787e6d0e7568686938f91ad7aa17f357b7
2. ~~`VmUserMessages` because PF4 replecement [Notification Drawer](https://www.patternfly.org/v4/components/notification-drawer) is still beta (missing features need to be re-implemented)~~  FIX IN #1584 
3. ~~PF3 styles - they are indirectly used across the product so a clean-up action is required~~  FIX IN #1584


rszwajko:
### Basic step

image

### NICs
image

image

image

image

### Disks

image

image

### Review step

image

### Finished 

image

image

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

sjd78:
### L&F / interaction issues (based on testing via #1594)

#### Basic Settings Step
- I can change the Memory before selection cluster/source/template or iso/OS but I can't change the total CPUs.  They should both be read only if other data needs to be selected first (e.g. to get the max counts per cluster level).

- The Cluster shows as in error if one isn't selected but only if a name is entered.  Not sure if "invalid" is the best choice, but it is reasonable.  Maybe just start the form with an error on the name and go from there?  No need to rehash the "direct the use through the form" debate we had a few years ago.

- The Cloud-init / sysprep section title seems crowded up against the enable checkbox.  Some top and bottom margins like in the PF3 version would look better.

- The `ExpandableSection` for the CPU topology looks good.

- Would be nice to had a slightly taller content area for the form to avoid scrolling, but it's manageable.

- As noted in the VM Details Details card, the Operation System dropdown needs to scroll.  It is painful to have to scroll the content area instead of the select box.  Same thing would probably happen with a lot of clusters or ISOs or Templates.  Seems like it would be a generic select box issue.

#### Networking Step
- It all looks and works great!

- Maybe the only thing would be static column widths, as the widths tend to jump around between view, edit, save.

#### Storage Step
- The table look and editing looks good.
 
- Picking templates with disks looks good.

- I did NOT test the various permutations of thin/preallocated against the optimized for type or base templates or creation that requires selecting a new target storage domain.

#### Review Step
- I miss the "Review and confirm setting" heading.

- The tooltip on the virtual CPUs doesn't seem to have the content lined up correctly.

- Create button looks good.

#### Create Step
- In progress looks good.

- Success looks good.

- Error looks good (easiest way is to create a VM with the same name as an existing VM).

- "View the VM" and Close are good and work as expected

src/components/CreateVmWizard/steps/Storage.js
```diff
@@ -483,286 +104,311 @@ class Storage extends React.Component {
...
+                            id: `${id}-save`,
+                            icon: (<CheckIcon/>),
+                            isDisabled: !isValid,
+                            onClick: () => onSave({ id, underConstruction }),
+                          },
+                          underConstruction && {
+                            ariaLabel: msg.cancel(),
+                            id: `${id}-cancel`,
+                            icon: (<TimesIcon/>),
+                            // TODO template cannot be edited again
```
sjd78, line 371: Is this still true?

sjd78 (commented):
All of the PF4 changes and the table refactoring look really really nice.

I was worried about loosing any of the disks type handling etc, but I realized that most of that is handled in the sagas and the Thin/Pre dropdown helps drive the underlying disk attributes.  All of the wizard code LGTM.

ovirt-infra:
All tests passed

rszwajko:
@sjd78 

> Basic Settings Step                                                     
>                                                                        
>I can change the Memory before selection cluster/source/template or iso/OS but I can't change the total CPUs. They should both be read only if other data needs to be selected first (e.g. to get the max counts per cluster level).
>                                                                        

This seems not a regression so I would rather fix it under a separate PR.
                                                                        
> The Cluster shows as in error if one isn't selected but only if a name is entered. Not sure if "invalid" is the best choice, but it is reasonable. Maybe just start the form with an error on the name and go from there? No need to rehash the "direct the use through the form" debate we had a few years ago.
>
                                                                        
Done. Validation state was not passed to `TextInput`. Note that focus styling covers (partially) error styling.
                                                                        
> The Cloud-init / sysprep section title seems crowded up against the enable checkbox. Some top and bottom margins like in the PF3 version would look better.
>
                                                                        
Done. Extra margin was added to sub-heading. Note that this looks best after PF3 styling was removed (until then there is double margin).
                                                                        
> Would be nice to had a slightly taller content area for the form to avoid scrolling, but it's manageable.
                                                                        
I agree. However the default Modal used by the Wizard is already `ModalVariant.large` which seems the largest available. Increasing the size even more requires more research. 
                                                                        
>  As noted in the VM Details Details card, the Operation System dropdown needs to scroll. It is painful to have to scroll the content area instead of the select box. Same thing would probably happen with a lot of clusters or ISOs or Templates. Seems like it would be a generic select box issue.
>                                                                      
Not a regression.
                                                                        
> Networking Step                                                         
>                                                                        
> Maybe the only thing would be static column widths, as the widths tend to jump around between view, edit, save.
>                                                                        
This might be tricky - I'd rather move it to another PR.
                                                                        
> Review Step                                                             
>                                                                      
>I miss the "Review and confirm setting" heading.                    
>                                                                       
Done

ovirt-infra:
All tests passed

Issue #1484, opened 2021-07-14
Handle SSO deprecation by removing the default attempt to use SSO whenever opening a console to a virtual machine


sgratch:
Currently whenever opening a console to a virtual machine and if guest_agent is installed, the user  will be logged in automatically. Otherwise an error appears.

Since the ovirt-guest-agent package was deprecated, SSO is deprecated for virtual machines as well and therefore need to stop using it by default for cluster levels 4.4 and above to avoid those errors.

Check https://bugzilla.redhat.com/show_bug.cgi?id=1869251 and https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/release_notes/deprecated_features_rhv for details

sgratch:
After discussing it offline with Virt/backend, few comments:

- The channel to ovirt-guest-agent was removed for RHEL 8+, Ubuntu 18+, and on ppc64: https://github.com/oVirt/ovirt-engine/blob/master/packaging/conf/osinfo-defaults.properties#L97 and also for Windows.

- Current web-ui implementation for choosing if to enable SSO logon is based only on reported `vm.sso.method` value. If it is set by backend to '`guest_agent`' and console is not used, then trying to logon using SSO. 
There is no checking if there's guest agent actually existed for the VM.

- I wouldn't add a backend cluster level configuration for this, because it's not really a cluster-level related feature - and if we drop it in, let's say CL 4.7, then people would just avoid upgrading to that CL

- Implementing a solution on backend side is more complicated since requires changing default configuration per VM and the user can still enable this, even if not supported. A better & easiest way is to block this on web-ui side for non supported VMs

Therefore, a suggested solution for implementing this on on web-ui side:
1. Dropping the 'annoying' console error popup for status `status: C.CONSOLE_LOGON`: https://github.com/oVirt/ovirt-web-ui/blob/b497e32eecf98ead2c12fdfc5691c53b12049dd5/src/sagas/console/index.js#L48.

i.e. never shows this popup anymore, for all console types and regardless to SSO status:
![image](https://user-images.githubusercontent.com/18169498/146933410-dca19539-5064-4197-bc14-deec4e541817.png)

Consider still creating the `cantOpenConsole` user event for still letting the user knows about failures.

2. Instead of only checking if `vm.sso.method`=='`guest-agent`', also check if guest-agent is installed for that vm (can be done via `applications` endpoint). So logic can be something like:
[...]

PR #1550, opened 2021-12-24
Ignore SSO failures when connecting to the console

rszwajko:
In case of unsuccessful SSO logon try to recover by connecting without SSO. The flow is the same as if the user manually ignored the error. This avoids disrupting the user with an error dialog that offers no real choice (beside ignoring the error).
The user will be still informed about unsuccessful logon via notification system (at warning level) so that he will be able to report the problem.

### DIFF
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -64,7 +64,7 @@ export const messages: { [messageId: string]: MessageType } = {
   bootSequence: 'Boot Sequence',
   cancel: 'Cancel',
   cannotUpdateCloudInitHostname: 'Host name cannot be synchronized with VM name because the VM name is not a valid host name.',
-  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues. Error message: {message}',
+  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Error message: {message}',
   cantOpenConsole: 'Can\'t open console: {message}',
   cardTitleDetails: 'Details',
   cd: 'CD',
--- a/src/intl/messages.js
+++ b/src/intl/messages.js
@@ -64,7 +64,7 @@ export const messages: { [messageId: string]: MessageType } = {
   bootSequence: 'Boot Sequence',
   cancel: 'Cancel',
   cannotUpdateCloudInitHostname: 'Host name cannot be synchronized with VM name because the VM name is not a valid host name.',
-  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues.',
+  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues. Error message: {message}',
   cantOpenConsole: 'Can\'t open console: {message}',
   cardTitleDetails: 'Details',
   cd: 'CD',
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -44,10 +44,7 @@ function* downloadOrOpenVmConsole ({
   if (hasGuestAgent && !skipSSO) {
     const result = yield callExternalAction(Api.vmLogon, { payload: { vmId } }, true)
     if (!result || result.status !== 'complete') {
-      const message = result?.error?.responseJSON?.fault?.detail ?? ''
-      yield put(Actions.addConsoleError({ vmId, vmName, consoleType, status: C.CONSOLE_LOGON, consoleId, logoutOtherUsers }))
-      yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantOpenConsole', params: { message } }, type: 'error' }))
-      return
+      yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantLogonToConsole' }, type: 'warning' }))
     }
   }
 
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -36,12 +36,12 @@ function* downloadOrOpenVmConsole ({
   consoleId,
   usbAutoshare,
   usbFilter,
-  hasGuestAgent,
+  attemptSsoOnOpenConsole,
   skipSSO,
   openInPage,
   logoutOtherUsers,
 }) {
-  if (hasGuestAgent && !skipSSO) {
+  if (attemptSsoOnOpenConsole && !skipSSO) {
     const result = yield callExternalAction(Api.vmLogon, { payload: { vmId } }, true)
     if (!result || result.status !== 'complete') {
       const message = result?.error?.responseJSON?.fault?.detail ?? ''
@@ -173,7 +173,7 @@ export function* openConsole ({
     usbAutoshare,
     usbFilter,
     vmName,
-    hasGuestAgent,
+    attemptSsoOnOpenConsole,
     consoleId,
     fqdn,
     domain,
@@ -185,7 +185,7 @@ export function* openConsole ({
     domain: config.get('domain'),
     username: config.getIn(['user', 'name']),
     vmName: vms.getIn(['vms', vmId, 'name']),
-    hasGuestAgent: vms.getIn(['vms', vmId, 'ssoGuestAgent']),
+    attemptSsoOnOpenConsole: vms.getIn(['vms', vmId, 'attemptSsoOnOpenConsole']),
     consoleId: idFromType({ consoleType, vm: vms.getIn(['vms', vmId]) }),
     fqdn: vms.getIn(['vms', vmId, 'fqdn']),
   }))
@@ -223,7 +223,7 @@ export function* openConsole ({
       vmId,
       usbAutoshare,
       usbFilter,
-      hasGuestAgent,
+      attemptSsoOnOpenConsole,
       consoleId,
       openInPage,
       skipSSO: skipSSO || doesVmSessionExistForUserId(sessionsInternal, userId),
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -45,10 +45,7 @@ function* downloadOrOpenVmConsole ({
     const result = yield callExternalAction(Api.vmLogon, { payload: { vmId } }, true)
     if (!result || result.status !== 'complete') {
       const message = result?.error?.responseJSON?.fault?.detail ?? ''
-      if (message) {
-        yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantOpenConsole', params: { message } }, type: 'warning' }))
-      }
-      yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantLogonToConsole' }, type: 'warning' }))
+      yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantLogonToConsole', params: { message } }, type: 'warning' }))
     }
   }
 
--- a/src/sagas/console/index.js
+++ b/src/sagas/console/index.js
@@ -44,6 +44,10 @@ function* downloadOrOpenVmConsole ({
   if (hasGuestAgent && !skipSSO) {
     const result = yield callExternalAction(Api.vmLogon, { payload: { vmId } }, true)
     if (!result || result.status !== 'complete') {
+      const message = result?.error?.responseJSON?.fault?.detail ?? ''
+      if (message) {
+        yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantOpenConsole', params: { message } }, type: 'warning' }))
+      }
       yield put(Actions.addUserMessage({ messageDescriptor: { id: 'cantLogonToConsole' }, type: 'warning' }))
     }
   }
--- a/src/components/VmActions/ConsoleNotificationsDialog.js
+++ b/src/components/VmActions/ConsoleNotificationsDialog.js
@@ -9,7 +9,6 @@ import * as Actions from '_/actions/console'
 
 import {
   CONSOLE_IN_USE,
-  CONSOLE_LOGON,
 } from '_/constants'
 
 import {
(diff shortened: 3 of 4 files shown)


rszwajko:
### Before
image

### After
image

sgratch (commented):
@rszwajko 
Removing this  error dialog that leaves no choice for the user seems ok to me, but keeping the SSO logon and this notification warning for all cases seems wrong.
Since SSO and guest agent are deprecated for specific host versions then I'm not sure if this is the correct way to go:

- For hosts that ovirt-guest-agent and therefore SSO are not supported anymore (deprecated) then there is no need to try running SSO logon by default and then displaying a general warning notification of  "SSO failed since guest-agent is not installed..please check if installed...contact if problem continues". The user has nothing to do with that info if this is not supported anymore and it's a misleading one.

- On the other hand, for hosts that still support running ovirt-guest-agent and ovirt-guest-agent is installed ((e.g. rhel <=7) then the SSO logon should run but might fail due to other reasons and therefore the real backend reason for logon failure (result.error) should be displayed as was up til now as part of the event/notification.

How about this suggested solution?: https://github.com/oVirt/ovirt-web-ui/issues/1484#issuecomment-998790111?
Maybe it's better to try running logon only if ovirt-guest-agent is installed and SSO is enabled. If not installed then we don't care if supported or not - we just avoid SSO logon (and can consider displaying an info event for notifying the user that this is not supported since guest-agent is not installed/deprecated). 

If ovirt-guest-agent is installed and SSO is enabled then we can run the same logon use case as today (except of removing this error dialog). In case of an error, a full notification warning with reason details should be displayed.

src/sagas/console/index.js
```diff
@@ -44,10 +44,7 @@ function* downloadOrOpenVmConsole ({
   if (hasGuestAgent && !skipSSO) {
```
sjd78, line 44: The workflow handled by this block needs to be updated to really solve the issue.

sjd78 (changes requested):
Swapping out the dialog for a toast is better UX, so I like how that part looks.

But...in the `openConsole()` saga, the lookup for `hasGuestAgent` needs to be different since it is no longer a fully valid check:
https://github.com/oVirt/ovirt-web-ui/blob/c97d2338c03e7c3d5433ed120f2d05d1cbd88ad8/src/sagas/console/index.js#L170-L190

And `ssoGuestAgent` is parsed from the VM here:
https://github.com/oVirt/ovirt-web-ui/blob/c97d2338c03e7c3d5433ed120f2d05d1cbd88ad8/src/ovirtapi/transform.js#L199

I'd suggest even renaming the `ssoGuestAgent` to something like `attemptSsoOnOpenConsole`.  That way more things can be added to that check and it can still makes semantic sense.

If data about the VM's containing entities (cluster, data center etc) is needed to figure out the value of the new `attemptSsoOnOpenConsole`, then it can be done in the fetch sagas, specifically `transformAndPermitVm()`.  By the time that transform saga runs, all of the hosts, clusters, etc should already be fetched and available via `yield select()`.

rszwajko:
@sgratch 

> Since SSO and guest agent are deprecated for specific host versions then I'm not sure if this is the correct way to go:
> 
> For hosts that ovirt-guest-agent and therefore SSO are not supported anymore (deprecated) then there is no need to try running SSO logon by default and then displaying a general warning notification of "SSO failed since guest-agent is not installed..please check if installed...contact if problem continues". 
> 

IMHO we should trust the configuration. It seems there is no other reliable way of determining if the OS installed on the VM supports guest agent. 

> The user has nothing to do with that info if this is not supported anymore and it's a misleading one.
>

User should always report such problem to the administrator. 

>  On the other hand, for hosts that still support running ovirt-guest-agent and ovirt-guest-agent is installed ((e.g. rhel <=7) then the SSO logon should run but might fail due to other reasons and therefore the real backend reason for logon failure (result.error) should be displayed as was up til now as part of the event/notification.
> 

+1

> How about this suggested solution?: #1484 (comment)?
>
> Maybe it's better to try running logon only if ovirt-guest-agent is installed and SSO is enabled. If not installed then we don't care if supported or not - we just avoid SSO logon (and can consider displaying an info event for notifying the user that this is not supported since guest-agent is not installed/deprecated).
> 
> If ovirt-guest-agent is installed and SSO is enabled then we can run the same logon use case as today (except of removing this error dialog). In case of an error, a full notification warning with reason details should be displayed.
> 

Regarding the second part of proposed solution:

> > Instead of only checking if vm.sso.method=='guest-agent', also check if guest-agent is installed for that vm (can be done via applications endpoint). So logic can be something like:
if the vm is set with ovirt-guest-agent reported + sshmethod=guest-agent and !skipSSO -> logon
otherwise -> don't logon
>> 

IMHO higher level of confidence is achieved by just trying SSO logon. Consider the possible outcomes of the extra check:
1. the call failed (i.e. due to network issue) - the only recovery action is to just try SSO logon
2. check failed (agent is reported as missing) but VM is marked as having a guest agent. Then we have following possibilities:
   1. configuration is correct and agent is really missing- the admin should investigate why guest agent got lost and re-install it.
   2. configuration is correct and agent is incorrectly reported as missing (i.e. not yet refreshed) - false error will be reported
   2. configuration is incorrect  - the admin should remove the flag manually.
3. check was successful (agent reported as installed) and VM is  configured to support guest agent - then we just perform SSO logon (which may fail due to various reasons including ...missing agent)

rszwajko:
@sjd78 

> the lookup for hasGuestAgent needs to be different since it is no longer a fully valid check

It still seems the best choice here - please check my response [above](https://github.com/oVirt/ovirt-web-ui/pull/1550#issuecomment-1020100629)

> if data about the VM's containing entities (cluster, data center etc) is needed to figure out the value of the new `attemptSsoOnOpenConsole`, then it can be done in the fetch sagas, specifically `transformAndPermitVm()`. By the time that transform saga runs, all of the hosts, clusters, etc should already be fetched and available via `yield select()`.

As per [this comment](https://github.com/oVirt/ovirt-web-ui/issues/1484#issuecomment-998790111) there is no cluster level setting for this and there are no plans to add it:
> I wouldn't add a backend cluster level configuration for this, because it's not really a cluster-level related feature - and if we drop it in, let's say CL 4.7, then people would just avoid upgrading to that CL

It seems that the only information that a given VM supports guest agent is just a flag on the VM itself. Note also (as per link above):
> the user can still enable this, even if not supported.

ovirt-infra:
Some tests failed

rszwajko:
Added notification with server-side SSO error - it's the same message as before but on `warning` level.

image

sjd78:
> @sgratch
> 
> > Since SSO and guest agent are deprecated for specific host versions then I'm not sure if this is the correct way to go:
> > For hosts that ovirt-guest-agent and therefore SSO are not supported anymore (deprecated) then there is no need to try running SSO logon by default and then displaying a general warning notification of "SSO failed since guest-agent is not installed..please check if installed...contact if problem continues".
> 
> IMHO we should trust the configuration. It seems there is no other reliable way of determining if the OS installed on the VM supports guest agent.

That makes perfect sense.  Trying to chase specific guest-agent versions or looking at reported installed software per OS or specific versions of specific packages is purely a loosing proposition.

----

> @sjd78
> 
> > the lookup for hasGuestAgent needs to be different since it is no longer a fully valid check
> 
> It still seems the best choice here - please check my response [above](https://github.com/oVirt/ovirt-web-ui/pull/1550#issuecomment-1020100629)

While the check itself makes sense to keep as is, I still thing renaming the data point in the transformed VM to `attemptSsoOnOpenConsole` makes sense.

> 
> > if data about the VM's containing entities (cluster, data center etc) is needed to figure out the value of the new `attemptSsoOnOpenConsole`, then it can be done in the fetch sagas, specifically `transformAndPermitVm()`. By the time that transform saga runs, all of the hosts, clusters, etc should already be fetched and available via `yield select()`.
> 
> As per [this comment](https://github.com/oVirt/ovirt-web-ui/issues/1484#issuecomment-998790111) there is no cluster level setting for this and there are no plans to add it:
> 
> > I wouldn't add a backend cluster level configuration for this, because it's not really a cluster-level related feature - and if we drop it in, let's say CL 4.7, then people would just avoid upgrading to that CL
> 
> It seems that the only information that a given VM supports guest agent is just a flag on the VM itself. Note also (as per link above):
> 
> > the user can still enable this, even if not supported.

All of that makes sense as well.

----

Net commentary: I understand the minimal nature of the change and with more context that does make sense.  Also, renaming the flag used by the code to something like `attemptSsoOnOpenConsole` still does make sense.  Maybe even a comment in the transform about it?

sjd78 (approved):
LGTM

ovirt-infra:
Some tests failed

sgratch:
@rszwajko @sjd78
 We can assume now that for most VMs, SSO is already deprecated and therefore the solution should be that either the  SSO is skipped for most cases or hidden as much as possible. The best solution is to handle that on backend but since it's more complicated then I suggested  to block that on VM portal side when there is no guest agent installed. This requires an extra query instead of the logon query that is currently running.

Anyway, If you prefer to attempt to logon to the VM every time a console is opened then it's ok as well. But at least hide the SSO attempt as much as possible or raise an informative but still minimal notification warning for the cases it's still supported and failed.

**Current situation**:  the following two notifications are generated almost every time a console is opened and SSO is deprecated:
1. "Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues."
2. Can't open console: [backed  error: Guest agent non responsive]

**My suggestion**:  display only one notification warning instead of two, include  the backend error message but skip the guideline to contact the admin if the problem continues.
Something like: "Single sign on can't run. Please check to see if the agent guest agent is running on your virtual machine [backed  error: Guest agent non responsive]"

sgratch (changes requested):
Please check my comment for rephrasing the SSO notifications

rszwajko:
@sgratch 
> We can assume now that for most VMs, SSO is already deprecated and therefore the solution should be that either the SSO is skipped for most cases or hidden as much as possible. 
>

1. AFAIK assumption applies to (some) new VMs using default settings
2. as for 4.5 release, SSO is still supported when creating a VM manually and can be enabled/disabled any time (see below)
image


> The best solution is to handle that on backend but since it's more complicated then I suggested to block that on VM portal side > when there is no guest agent installed. This requires an extra query instead of the logon query that is currently running.
> 
> Anyway, If you prefer to attempt to logon to the VM every time a console is opened then it's ok as well.
>

We do logon only if this is requested by the configuration: trusting the configuration seems the most reasonable approach.

> But at least hide the SSO attempt as much as possible or raise an informative but still minimal notification warning for the cases it's still supported and failed.
> 
> **Current situation**: the following two notifications are generated almost every time a console is opened and SSO is deprecated:
> 
>  1. "Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues."
> 
>  2. Can't open console: [backed  error: Guest agent non responsive]
>
> 
> **My suggestion**: display only one notification warning instead of two, include the backend error message but skip the guideline to contact the admin if the problem continues. Something like: "Single sign on can't run. Please check to see if the agent guest agent is running on your virtual machine [backed error: Guest agent non responsive]"
>


1. the (long) message is actually correct: user should contact the administrator in case of SSO errors as only the admin can investigate and fix them (including configuration error).
2.  re-phrasing the message will invalidate the translations - since this is a long and important message I would rather avoid it
3. we could display both messages but this makes the error (shorter message)  less visible.

ovirt-infra:
All tests passed

rszwajko:
The notification with the error message:
image

rszwajko:
> > 1. AFAIK assumption applies to (some) new VMs using default settings
> 
> The ovirt-guest-agent and SSO are not supported for VMs running RHEL 8 or later, and for Windows operating systems.. Customers usually upgrade to a newer oVirt verison to enjoy the new RHEL/Virt functionality and therefore need to upgrade their cluster versions to enjoy that. So they will eventually upgrade their clusters to 4.5 and above. Therefore most of their VMs will not support SSO anymore. Only old clusters that were not migrated yet, will still support that for a while until migrated.
>

Correct. This actually confirms my understanding: some new VMs won't support SSO. Old VMs and new VMs with "old" OS should work as before so we cannot "ignore" the flag.
 
> > 2. as for 4.5 release, SSO is still supported when creating a VM manually and can be enabled/disabled any time (see below)
> >    image
> 
> The correct solution was to disable this configuration option on webadmin for those unsupported VMs. I discussed that with Virt team and since it is more complicated to implement, we decided not to do that. So currently, the default of "use guest agent" is enabled even for VMs which we know that won't ever support that. It's a bug.  So the solution was to block that on web-ui only since it's much more easy to do.
>

Note that frontend also cannot distinguish unsupported VMs and ignoring a legal configuration flag would impact valid use cases.

> Therefore, raising a notification with a requirement to "Contact your administrator if the problem continues" is wrong since the problem will continue for those VMs :). We can't ask users to do that on things that we should identify as not supported anymore. It's only text message but it still looks strange to me.
>

As already discussed, incorrectly placed flag is a configuration error. User should report this to the admin and the admin should simply remove the option.

src/intl/messages.js
```diff
@@ -64,7 +64,7 @@ export const messages: { [messageId: string]: MessageType } = {
   bootSequence: 'Boot Sequence',
   cancel: 'Cancel',
   cannotUpdateCloudInitHostname: 'Host name cannot be synchronized with VM name because the VM name is not a valid host name.',
-  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues.',
+  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Contact your administrator if the problem continues. Error message: {message}',
```
sjd78, line 67:
```suggestion
  cantLogonToConsole: 'Single sign on failed. Please check to see if the guest agent is running on your virtual machine. Error message: {message}',
```

As per our discussions today, I think changing the error message to the suggested will be sufficient.

sjd78 (changes requested):

ovirt-infra:
All tests passed

sgratch (approved):
LGTM

isaranova:
LGTM

Issue #1558, opened 2022-01-27
vmdetails snapshots card: After deleting an existing snapshot succesfuly, the  'create snapshot' button is still disabled


sgratch:
The 'create snapshot' button enabled/disabled mechanism is wrong after deleting an existing snapshot for a VM

**Expected**:
When a snapshot operation is taking place via the vmdetails snapshots card, the 'create snapshot' button should be disabled for current VM till the deleting operation is completed and the lock is removed.

**Actual**:
The problem is that after deleting an existing snapshot via the snapshots card, the 'create snapshot' button stays disabled till hard refresh and because of that all other operations like delete/restore snapshots are also disabled.

rszwajko:
It seems related to the way we detect pending actions (check [this code](https://github.com/oVirt/ovirt-web-ui/blob/b36242b85b335ee2f9ea4143fc3941d521a37e99/src/components/VmDetails/cards/SnapshotsCard/index.js#L67)). More investigation is needed to verify if existing flow can be fixed.

Note that improving this area might also solve https://github.com/oVirt/ovirt-web-ui/issues/1559

PR #1579, opened 2022-03-30
Use ms as delay time unit when deleting snapshots

rszwajko:
Before, the number was interpreted as seconds which effectively
prevented checking if the snapshot was already deleted.

Resolves: #1558


ovirt-infra:
All tests passed

sjd78 (approved):
LGTM

isaranova:
Issue is still persistent for me:
Screenshot 2022-06-15 at 16-35-40 VM Portal

isaranova:
Retested on newer version. LGTM

PR #1585, opened 2022-04-21
Migrate to PF4 based react-console

rszwajko:
Depends on https://github.com/oVirt/ovirt-web-ui/pull/1564

 ### Re-implement console based on newest react-console
    
  Changes:
  1. drop portaled toolbars - react-console provides AccessConsoles
     component with built-in grid layout that includes toolbar
  2. drop support for native consoles on the console page:
     a) to enter the screen user needs to first chose BrowserVnc console.
        Only after disconnecting native consoles can be chosen
     b) react-console by default generates the *.vv files - more research
        is needed how to merge it with current flow (downloading)
     c) react-console uses one common 'Connect' button for all native
        consoles and internally picks the preferred connection type
     d) react-consoles prefers SPICE over VNC when both are available
  3. adapt VncConsole component from @patternfly/react-console
     to fix following issues:
     a) use 'binary' as default 'wsProtocols' to support Qemu < 5.0.0
     b) infinite re-connect loop in case of connection problem - problem
        caused by useEffect dependencies changing with every render and
        triggering 'Maximum update depth exceeded' warning
     c) add onConnect callback removed in newer version - required for
         handling focus
  
Reference-Url: https://github.com/novnc/noVNC/commit/c912230309806aacbae4295faf7ad6406da97617
Reference-Url: https://github.com/oVirt/ovirt-web-ui/pull/1525#issuecomment-976329142

###    Port VncConsole from TypeScript to plain JS
    
Source:      [@patternfly/react-console/src/components/VncConsole/VncConsole.tsx](https://github.com/patternfly/patternfly-react/blob/main/packages/react-console/src/components/VncConsole/VncConsole.tsx)
Commit:     https://github.com/patternfly/patternfly-react/commit/e0240737eb6abb091c3ac83c3de8fb68420d7578


ovirt-infra:
Some tests failed

ovirt-infra:
Some tests failed

ovirt-infra:
Some tests failed

rszwajko:
### Screenshots

image
image

ovirt-infra:
Some tests failed

ovirt-infra:
Some tests failed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

src/components/VmConsole/index.js
```diff
@@ -1,27 +1,32 @@
-import React, { useEffect, useState } from 'react'
+import React, {
+  useEffect,
+  useState,
+} from 'react'
 import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import { push } from 'connected-react-router'
-import $ from 'jquery'
```
sjd78, line 5: One less use of jquery!  Replace the `jquery.ajax` calls with standard `fetch` and we can drop it completely.

sjd78 (commented):
Code changes look clean.

sjd78:
### L&F Notes:

#### VM Details:
- On VM Details, if the VM is Off, the Console button doesn't do anything but it is not grayed out like the other disabled action buttons:
screenshot-localhost_3000-2022 06 13-11_42_16

- VM Details, Spice console via `console.vv` launches successfully
- VM Details, native VNC via `console.vv` launches successfully
- Browser console pages launches successfully

#### Browser Console:
- Functions work as expected:
  - disconnect,
  - reconnect,
  - send ctrl-alt-del,
  - fullscreen

- Directly opening console page for a VM works as expected

- Changing screen resolution on the VM works!

- Toolbar on the browser page has different padding than the VM Details toolbar.  Ideally they'd match.

sjd78 (approved):
Only 1 minor L&F issue (toolbar padding around buttons on browser console page), otherwise LGTM

src/components/VmConsole/index.js
rszwajko: This is a good follow-up PR ! `fetch` does not reject on error status so we could also clean up error handling in the saga layer -> https://github.com/oVirt/ovirt-web-ui/blob/bfd825b22b9b4acc14c3309f2832c1f8bd9e6e28/src/sagas/utils.js#L76

rszwajko:
@sjd78 
The dropdown uses the default [PF4 styling](https://www.patternfly.org/v4/components/dropdown#disabled). The font color for the disabled state is the same as for active state.As you correctly pointed this is different then styling for regular buttons. However I'm not sure if we should change it - it might be a feature :)

ovirt-infra:
All tests passed

sjd78 (approved):
LGTM

Issue #1572, opened 2022-03-16
Allow a user to consume ISO images without granting a `create disk` permission


sgratch:
Originally raised here: https://mail.google.com/chat/u/0/#chat/space/AAAA7lwAJb4/GVTtxRTSAZQ

The requirement is to allow a user to consume ISO image without granting a `create disk` permission on that storage domain. 

Trying to create a new role, that has:
```
- Disk - Disk Profile >> Attach Disk Profile
- Disk - Provisioning Operations >> Access Image Storage Domains
- Disk - Provisioning Operations >> Attach
```
but without having
`- Disk - Provisioning Operations >> Create ` permission.

A user who is granted that role can't reach ISO images by the VM Portal.

PR #1591, opened 2022-05-16
Allow selection of ISO images from storage domains tagged `canUserUseIsoImages`

sjd78:
Fixes: #1572

### Summary

A user that has the proper permits (`edit_vm_properties`) to edit a VM's
CD should be allowed to choose any ISO from any Storage Domain in which
they have an `attach_disk_profile` permit.

Add a flag to Storage Domains called `canUserUseIsoImages` to indicate
the current user can see and select and ISO held in that storage domain.
A user is not required to have `canUserUseDomain` on a SD to be able to
use the ISOs in the SD.

### Setup
1. As per issue #1572, create a new Role with the following permits:
  - 1563 / attach_disk_profile / Disk - Disk Profile >> Attach Disk Profile
  - 1106 / access_image_storage / Disk - Provisioning Operations >> Access Image Storage Domains
  - 1101 / attach_disk / Disk - Provisioning Operations >> Attach
  
2. For a Storage Domain with ISO images, assign a Permission to a User (or a Group) with the new role.

3. Permission the User (or Group) to be able to edit a VM

4. Open the VM Details page and edit the Details card.

5. All of the ISO images in the storage domain permissioned in step 2 should be visible.

Note: Depending on the user's permissions on the VM, they may only be able to change the CD while the VM is running.  In this case, the CD will revert to the base VM setting after a reboot.

### DIFF
--- a/src/components/utils/build-select-box-lists.js
+++ b/src/components/utils/build-select-box-lists.js
@@ -46,7 +46,7 @@ function createIsoList (storageDomains, dataCenterId = null) {
     .toList()
     .filter(storageDomain =>
       storageDomain.has('files') &&
-      storageDomain.get('canUserUseDomain') &&
+      storageDomain.get('canUserUseIsoImages') &&
       (dataCenterId === null
         ? true
         : storageDomain.getIn(['statusPerDataCenter', dataCenterId]) === 'active')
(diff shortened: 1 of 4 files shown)


ovirt-infra:
All tests passed

rszwajko (approved):
Code looks good!

PR #1599, opened 2022-06-09
Use finer RPM versions

tinez:
Using only the date part in RPM names is not fine enough - when a couple
of builds land in COPR with the same date, the one with the "biggest SHA"
wins, i.e.:

  1.8.2-0.20220607.gitf519b50.fc35 > 1.8.2-0.20220607.git26ac92c.fc35 > 1.8.2-0.20220607.git703b8f7.fc35

even though it might not actually be the latest build.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -73,7 +73,7 @@ srpm:	dist-gzip
 	@echo "srpm available at '$(TMPREPOS)'"
 	@echo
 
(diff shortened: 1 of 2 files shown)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

sgratch (approved):
LGTM

PR #1600, opened 2022-06-09
Table view for Virtual Machines screen

rszwajko:
~~Depends on https://github.com/oVirt/ovirt-web-ui/pull/1592~~ _The commit from #1592 this PR was dependent on has been cherry-picked to this PR._

This PR adds possibility to switch between card view and table view on the Virtual Machines screen. The card view remains the default but it's possible to change it (the change is persisted on the server as an user profile property).

Note that this PR is only about presentation layer: the underlying logic remains the same.
In particular:
1. both views present the same information to the user (no extra columns in table view)
2. data is lazy loaded using infinite scroller (no explicit paging) 
3. actions are per VM/pool (no multiple selection/bulk operations)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

src/components/VmsList/VmCardList.js
```diff
@@ -79,31 +96,77 @@ const VmCardList = ({ vms, alwaysShowPoolCard, fetchMoreVmsAndPools }) => {
...
+          {vmsAndPools.map(entity => (
+            <GalleryItem key={entity.get('id')}>{
+            entity.get('isVm')
+              ? <CardVm vm={entity} />
+              : <CardPool pool={entity} />}
+            </GalleryItem>
+          ))}
+        </Gallery>
+      )}
+      {tableView && (
```
sjd78, line 130: So why put the table view inside the `VmCardList` instead of closer to `VmsList`?  It would be more logical to have the table view vs card view render happen in `VmsList`.  That would just mean you need to double up on the infinite scroller or to have a TableListOrCardList wrapper...

ovirt-infra:
All tests passed

src/components/VmsList/VmCardList.js
rszwajko: @sjd78 indeed after moving the logic to `VmsList` the code looks much better! Note that after this refactoring the infinite scroller is now also a separate entity that could be replaced with another loader if needed.

rszwajko:
@sjd78 @sgratch 
Note that this PR depends on https://github.com/oVirt/ovirt-web-ui/pull/1592  0 - in particular it uses common filtering and sorting components.

sjd78 (approved):
LGTM

ovirt-infra:
All tests passed

sjd78:
> @sjd78 @sgratch Note that this PR depends on #1592 0 - in particular it uses common filtering and sorting components.

@sgratch, @rszwajko -- I cherry-picked the necessary commit from 1592 to this PR.  This PR is now standalone complete.

sgratch (commented):
@rszwajko 
Few comments/enhancements to consider:

1. IMHO both card/table view buttons on VMs dashboard dialog worth adding tooltips to easily differentiate between them:
image

2. Resetting the user's account settings via the account settings dialog reset this view option back to cards view. This is not an obvious expected behavior and even though it's not a destructive action, it is still worth adding a property/indication to the account settings dialog, as done for all other reset-able options.

3. When resizing the dialog by zooming in, the filter search text field is dismissed. This is reproduced regardless to this fix but now the chance is higher since less space is left. It's not obvious that the search text filed is hidden by the funnel button so maybe worth at least adding a tooltip:
    Default zoom:
    image

     Zoom in:
     image

4. Pool and Pool vms rows look exactly the same at the table view (same layout except the status new option) while on cards view there is a bit different layout (2 status fields instead of one) and it's more easy to distinguish between them. Maybe worth adding a tag to distinguish

5. The number of entities displayed on vms dashboard, as part of the filter results, includes a question mark which seems like a bug. E.g.: "`23 of ? Results`"
    
image

rszwajko:
@sgratch 
> Few comments/enhancements to consider:
> 
> 1. IMHO both card/table view buttons on VMs dashboard dialog worth adding tooltips to easily differentiate between them:
>

The icons look self-explanatory to me but I'll check how tooltips will behave in this scenario. Note that the [design guidlines](https://www.patternfly.org/v4/components/toggle-group/design-guidelines) and the examples don't use the tooltips.  
 
> 2. Resetting the user's account settings via the account settings dialog reset this view option back to cards view. This is not an obvious expected behavior and even though it's not a destructive action, it is still worth adding a property/indication to the account settings dialog, as done for all other reset-able options.
>

AFAIK we don't list the properties that will be removed during the reset operation.

> 3. When resizing the dialog by zooming in, the filter search text field is dismissed. This is reproduced regardless to this fix but now the chance is higher since less space is left. It's not obvious that the search text filed is hidden by the funnel button so maybe worth at least adding a tooltip:
>

According to PF4 [design guidelines](https://www.patternfly.org/v4/components/toolbar/design-guidelines/#the-toolbar-on-mobile) the primary use case here is the mobile browser where "on hover" support is poor.

>  4. Pool and Pool vms rows look exactly the same at the table view (same layout except the status new option) while on cards view there is a bit different layout (2 status fields instead of one) and it's more easy to distinguish between them. Maybe worth adding a tag to distinguish
> 

The data is the same in both views - only the layout is different. Of course we can add some extra tag but then it should be visible also on the card view. It would be also nice to filter based on such tag. I would rather create a follow-up PR to track this.

>     5. The number of entities displayed on vms dashboard, as part of the filter results, includes a question mark which seems like a bug. E.g.: "`23 of ? Results`"
>        

This is a feature :)
```
Visualise that the total number of VMs/Pools is unknown by displaying
a question mark as the total result count i.e. "10 of ? Results" 
```

ovirt-infra:
All tests passed

rszwajko:
@sgratch @sjd78 
Added 3 commits on top of already reviewed code:
1. Add tooltips to toolbar filter and card/table view toggle
2. Use 'many' to describe indeterminate number of results. This wording is recommended by PF4 for paging - see https://www.patternfly.org/v4/components/pagination/design-guidelines#indeterminate-pagination
3. Base filter selection on already selected filters - the existing code worked (somehow) but after refactoring it's deterministic

rszwajko:
## Screenshots

image

image

image

image

sgratch (commented):
All looks good,  agree with your comments above, except one comment which I'm not sure about:

> AFAIK we don't list the properties that will be removed during the reset operation.

It's true that the list of properties is not displayed during reset operation but it's strange that the user can change his current view from cards->table and then approach the "account settings" dialog and see that the "reset settings" option is enabled without knowing why. For all other reset-able options he can see the current chosen values as part of the account settings dialog itself.

A suggestion might be to add a read only view mode field as part of "advanced options" tab within the "account settings" dialog.

A followup issue was opened to handle this: https://github.com/oVirt/ovirt-web-ui/issues/1632

sgratch (approved):
This looks ok except a followup issue that will be handled separately for the reset options handling.

isaranova:
LGTM

PR #1604, opened 2022-06-16
Revert "Add Events view"

sgratch:
Reverts oVirt/ovirt-web-ui#1592

Due to few functional aspects of this solution, specifically concerning permissions and info that the user shouldn't been exposed to. Let's re-think about a better solution for this for next release. Since it's a new functionality, there is no problem with reverting for now


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

sjd78 (approved):
LGTM

VM event view is gona and I see no UI or console errors.

PR #1607, opened 2022-06-21
Recover after rejected fullscreen request

rszwajko:
Fullscreen request can be rejected due to various reasons. In such
case the isFullScreen flag needs to be cleared to synchronize the state
and the displayed UI.

Note that fullscreen request cannot be triggered automatically on page
load. Due to security reasons the API is restricted to user gestures.
This prevents opening the console automatically in full screen when
accessed directly via the link (i.e. by page refresh).


ovirt-infra:
All tests passed

sjd78 (approved):
LGTM

isaranova:
LGTM

ovirt-infra:
Some tests failed

Issue #1596, opened 2022-05-30
WIndows 11/2022 VMs for cluster level >=  4.6 -  failed to create/edit such VMs due to the TPM device


sgratch:
Since TPM is not set on backend automatically based on OS type and Cluster level version, the following 2 scenarios occur:

1. for `cluster level` >= 4.6: when trying to create a new VM with OS set to `Windows 11` or `2022`  or when editing an existing VM  and set OS to `Windows 11` or `2022`, the creation/edit failed in rest backend with the following error:
_`TPM device is required by the guest OS`_

   E.g.
![image](https://user-images.githubusercontent.com/18169498/171031478-45ba310b-062d-4bfa-b2fb-b4caf24b452d.png)
![image](https://user-images.githubusercontent.com/18169498/171032032-a206df43-6c3b-4a32-852b-c496082e8b72.png)

2) When trying to edit a VM with OS `Windows 11` or `2022` and  `cluster level` >= 4.6 by changing the OS to non supported TPM one, e.g. Linux, the following error appears:

   ![image](https://user-images.githubusercontent.com/18169498/171032575-f987a17a-913c-433f-853d-83de3a4c6988.png)

### Details:
On webadmin the TPM is enabled/disabled on frontend based on OS and cluster level: https://github.com/oVirt/ovirt-engine/blob/91bc0b8f9e4d0cbfa8799f860a008948f3e6ed0a/frontend/webadmin/modules/uicommonweb/src/main/java/org/ovirt/engine/ui/uicommonweb/models/vms/UnitVmModel.java#L4047

So for fixing we should either enable/disbale the TPM on backend  based on OS or the same logic as on webadmin frontend should be used by web-ui as well.

ahadas:
[...]

PR #1610, opened 2022-07-05
Set tpm_enabled flag according to OS requirements

rszwajko:
Fixes: #1596
Depends on: https://github.com/oVirt/ovirt-engine/pull/514

For both new and edited VMs:
1. set `tpm_enabled = true` if TPM support is required by the guest OS

For edited VMs:
1. set `tpm_enabled = false` if TPM is not supported by the guest OS.
2. require confirmation before disabling TPM as it may result in data
   loss.


ovirt-infra:
All tests passed

rszwajko:
### TPM flag on Summary step (Create VM Wizard)

image

### Confirmation message warning about potential data loss

image

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

sjd78 (approved):
- transform and transport changes LGTM
 - create wizard changes LGTM
 - VM details card OS change driven TPM change confirm modal changes LGTM

sgratch (changes requested):
@rszwajko 
1. The following exception occurs for the following use case:
    - edit an existing vm and change the OS from a required TPM OS to a non TPM supported OS (e.g. change OS from 2022 to Linux).
    - the confirmation dialog tpmNotSupportedByTheSystem appears.
    - click the ok button
    - the following exception appears:
       
       image
       image

2. When changing OS from a required TPM OS to a non TPM supported OS (e.g. change OS from 2022 to Linux) and then clicking on `cancel` button once the confirmation dialog `tpmNotSupportedByTheSystem` appears, from now on changing to any TPM supported OS (e.g. Window 2008) still causes the confirmation dialog `tpmNotSupportedByTheSystem` to appear. Only after exiting the vm details card edit mode, the confirmation dialog stopped appearing

3. Need to consider supporting cases in which vm.cluster level >= 4.6 and vm.TPM-required OS (win 11/2022) and vm.Chipset type <> UEFI. In those cases the Chipset type should be changed automatically without notifying the user to UEFI. same as supported on webadmin. This can be done on a follow up fix.

rszwajko:
@sgratch 
fixed the scenarios pointed above - please review again.

src/components/CreateVmWizard/helpers.js
```diff
@@ -163,12 +165,32 @@ const verifyOsIdToCluster = (selectedOsId, clusterId, { clusters, operatingSyste
   return selectedOs ? selectedOs.get('id') : '0'
 }
 
+const isTpmRequired = (operatingSystemId, operatingSystems) => {
+  return operatingSystems.getIn([operatingSystemId, 'tpmSupport']) === 'required'
+}
+
+const getTpmChange = (operatingSystemId, operatingSystems) => {
```
sgratch, line 172: Why declaring this method here and not in src/helpers.js.
It is not really used by the createVmWizard.

sgratch (commented):
@rszwajko 
Issues 1, 2 are fixed and looks good now.
For issue 3 - a followup fix will be created.

src/components/CreateVmWizard/helpers.js
rszwajko: no particular reason - we can change it

sgratch (approved):

sgratch:
OST failed 3 times in test_verify_add_hosts (didn't get to ui tests yet) so merging without waiting for OST to succeed.

PR #1613, opened 2022-07-13
regressions: cannot enter values into NumberInput

rszwajko:
The value passed to `onChange` callback is not the number but an event that encapsulates the value.


sjd78 (approved):
LGTM
event handlers follow the pattern in https://www.patternfly.org/v4/components/number-input

ovirt-infra:
Some tests failed

isaranova:
LGTM

PR #1615, opened 2022-07-14
Lock snapshot during restore operation

rszwajko:
This is follow-up to PR https://github.com/oVirt/ovirt-web-ui/pull/1594

The snapshot is locked on the backend  during the restore but the UI
needs to wait until the next refresh to receive this status change.
To provide better user experience we can lock the snapshot right
after restore is triggered. The next refresh will confirm or clear
the status (according to restore operation progress).


ovirt-infra:
All tests passed

PR #1616, opened 2022-07-19
build: add conditional build for offline/online builds

sjd78:
Instead of autoconf swapping in a 0 or 1 for a `%define` variable
in the `spec.in` file for `OFFLINE_BUILD`, add a conditional build.
See [1] for how they work.

With a conditional build, source rpms retain the ability to build with
different options depending on how `rpmbuild` is called.  This allows
an offline (using virt-engine-nodejs-modules) or online (using
`yarn install`) build from srpm to be selected at build time using
`--with` or `--without` build options.

Since at least copr builds chroots directly from a srpm, conditional
builds can be used to change build options.  With this change, an
"online build" can be enabled simply by configuring the target chroot
setting "without" to include "ovirt_use_nodejs_modules".

Notes:
  - An offline build with `ovirt-engine-nodejs-modules` is the default

  - An online build using standard `yarn install` can be achieved by
    using `--without ovirt_use_nodejs_modules` on the call to rpmbuild

  - Setting env var `OFFLINE_BUILD=0` when calling `make rpm` will
    add the `--without` option, and run an offline build.  This is
    the same behavior as previous.

[1] - https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html

**Note**: PR https://github.com/oVirt/ovirt-engine-ui-extensions/pull/58 adds the same functionality to ovirt-engine-ui-extensions with the same option name.  With both in place, online and offline builds can be configured for both packages with the same single option on a copr chroot config.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -64,6 +64,7 @@ TMPREPOS = tmp.repos
 RPMBUILD_ARGS :=
 RPMBUILD_ARGS += --define="_topdir `pwd`/$(TMPREPOS)"
 RPMBUILD_ARGS += $(if $(RELEASE_SUFFIX), --define="release_suffix $$RELEASE_SUFFIX")
+RPMBUILD_ARGS += $(if $(filter 0,$(OFFLINE_BUILD)), --without ovirt_use_nodejs_modules)
 
 srpm:	dist-gzip
 	rm -fr "$(TMPREPOS)"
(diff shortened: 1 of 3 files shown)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

sgratch (approved):

PR #1618, opened 2022-07-21
Force tooltips on the DonutChart to a constant width

sjd78:
## Summary
By default the `@patternfly/react-charts` chart tooltips use the same text width estimation as is used in the Victory charts [1]. This estimation works ok for ASCII characters, but underestimate sizes for other non-Latin character code pages.

Instead of creating another text size estimation function that can handle non-Latin characters (e.g. see [2]), just pick a constant width that should accommodate the expected text widths of most languages we support.

[1] - https://github.com/FormidableLabs/victory/blob/main/packages/victory-core/src/victory-util/textsize.ts
[2] - https://stackoverflow.com/questions/118241/calculate-text-width-with-javascript

## Examples
How it looks now in English:
screenshot-localhost_3000-2022 07 21-13_34_40

Previously in Japanese:
screenshot-localhost_3000-2022 07 21-13_32_02

Now in Japanese:
screenshot-localhost_3000-2022 07 21-13_33_36

### DIFF
--- a/src/components/VmDetails/cards/UtilizationCard/UtilizationCharts/DonutChart.js
+++ b/src/components/VmDetails/cards/UtilizationCard/UtilizationCharts/DonutChart.js
@@ -1,6 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { ChartDonut, ChartLabel } from '@patternfly/react-charts'
+import {
+  ChartDonut,
+  ChartLabel,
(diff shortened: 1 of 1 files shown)


ovirt-infra:
All tests passed

sgratch (approved):

sgratch:
OST fails due to other issue, not related to this fix so merging

Issue #1590, opened 2022-05-16
VM Details > Nics Card - a user can't create a new NIC with an <empty> vNIC profile, unless he has permissions to at least one NIC profile within the VM's DC


sgratch:
A non admin user with `edit_vm_properties` permission for a specific VM, can't create a new vNic with an `<Empty>  vNIC profile` for that VM via the VM Details > Nics Card dialog. The 'create NIC' option is hidden for him.

The same user can create a VM with a  new vNIC using an empty vNIC profile  via the 'Create VM' wizard (but he can't create another such NIC for the same VM via the VM details - Nics Card dialog).

The create wizard and VM details dialog should be consistent and therefore if the create wizard allows empty vNic profiles then the same should be allowed via the Nics Card.
This is not a regression.

sgratch:
The fix is probably just to remove this line and adjust the render accordingly: https://github.com/oVirt/ovirt-web-ui/blob/master/src/components/VmDetails/cards/NicsCard/index.js#L106

PR #1624, opened 2022-07-28
Always allow creating nics in VM Details, Nics Card

sjd78:
Fixes: #1590

Since the list of nics will always include "<Empty>", it is
always possible to create a new nic.  Existing restrictions
have been removed.


ovirt-infra:
All tests passed

rszwajko (approved):

isaranova:
LGTM

PR #1627, opened 2022-08-23
containers: Specify COPR chroot explicitly

tinez:
A recent change to DNF COPR plugin [1] makes 'epel-9' the default chroot
used for CentOS Stream 9. Let's specify the 'centos-stream-9' chroot
name that we use in oVirt's COPR explictly.

[1] https://github.com/rpm-software-management/dnf-plugins-core/pull/459

Signed-off-by: Marcin Sobczyk <msobczyk@redhat.com>

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -30,7 +30,7 @@ for your platform to use the same packages that will be used by CI to build the 
 
 For CentOS stream 9:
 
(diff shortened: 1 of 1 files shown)


ovirt-infra:
All tests passed

sgratch (approved):

Issue #1530, opened 2021-11-08
GitHub web hooks for docker.com not working anymore


sandrobonazzola:
Description of problem:
GitHub Web hooks for docker.com are currently in failed state for a long while.

We should either fix the integration with docker or drop them

sjd78:
Required #1442

michalskrivanek:
+1 on dropping

sjd78:
Along with merging #1630, the webhooks can be removed at any time.

sjd78:
@sgratch, @michalskrivanek - With PR #1630 removing broken docker support, any webhooks on this repo configured to docker or dockerhub should be removed.

michalskrivanek:
docker hooks dropped, as well as two old jenkins hooks, hopefully no one is going to miss them...

PR #1630, opened 2022-09-15
Remove broken docker support

sjd78:
Docker support has been broken for some time.  Remove the docker related configs and drop support for web-ui use via container.

Resolves: #1442

Following up #1530, the docker webhooks on github repo setting can be removed (cc @sgratch).


ovirt-infra:
All tests passed

PR #1636, opened 2022-10-11
Force UEFI Bios if required to enable TPM

rszwajko:
In order to enable TPM, the VM needs to use UEFI firmware. Currently oVirt supports 2 types of UEFI firmwares:
1. standard - 'q35_ovmf'
2. secure - 'q35_secure_boot'

As requested in [1], we should force 'q35_ovmf' firmware type if the current firmware prevents enabling TPM:
1. in 'create VM' scenario - override the cluster defaults
2. in 'edit VM' scenario  - override previous value

Known issues:
1. if firmware is different than cluster defaults then a warning icon is displayed in the Web Admin UI -> VM details -> General tab
2. in VM Portal it's not possible to revert firmware to previous value which prevents some OS configurations i.e. Windows XP

[1] https://github.com/oVirt/ovirt-web-ui/issues/1596#issuecomment-1263431193

Reference-Url: https://github.com/oVirt/ovirt-web-ui/issues/1596

### DIFF
--- a/src/components/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/CreateVmWizard/CreateVmWizard.js
@@ -40,6 +40,7 @@ const DEFAULT_STATE = {
         threads: 1,
       },
       tpmEnabled: undefined,
+      biosType: undefined,
     },
     network: {
       nics: [],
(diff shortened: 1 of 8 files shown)


sgratch (approved):
LGTM

isaranova:
LGTM

PR #1642, opened 2022-11-30
Optimize the entity permission to permit mappings

sjd78:
Fixes: #1647 

In some cases, when a large number of entities are processed for permission to permit mapping, the code would just hang without indication of what went wrong.  At least 2 `yield` would be needed to process each entity.  Over the course of 1000+ entities, the 2000 `yield` calls seem to cause problems.

The mapping code has been optimized to allow using a lookup cached version.  Now only a single `yield` is required.  The change allows 1000+ entities to be processed successfully.

Signed-off-by: Scott J Dickerson <sdickers@redhat.com>

### DIFF
--- a/src/sagas/index.js
+++ b/src/sagas/index.js
@@ -35,9 +35,9 @@ import {
 
 import {
   callExternalAction,
+  curryEntityPermissionsToUserPermits,
   delay,
   doCheckTokenExpired,
-  entityPermissionsToUserPermits,
   mapCpuOptions,
 } from './utils'
 
@@ -88,8 +88,10 @@ const VM_FETCH_ADDITIONAL_SHALLOW = [
 export const EVERYONE_GROUP_ID = 'eee00000-0000-0000-0000-123456789eee'
 
 export function* transformAndPermitVm (vm) {
+  const _entityPermissionsToUserPermits = yield curryEntityPermissionsToUserPermits()
+
   const internalVm = Transforms.VM.toInternal({ vm })
-  internalVm.userPermits = yield entityPermissionsToUserPermits(internalVm)
+  internalVm.userPermits = _entityPermissionsToUserPermits(internalVm)
 
   internalVm.canUserChangeCd = canUserChangeCd(internalVm.userPermits)
   internalVm.canUserEditVm = canUserEditVm(internalVm.userPermits)
@@ -107,7 +109,7 @@ export function* transformAndPermitVm (vm) {
 
   // Permit disks fetched and transformed along with the VM
   for (const disk of internalVm.disks) {
-    disk.userPermits = yield entityPermissionsToUserPermits(disk)
+    disk.userPermits = _entityPermissionsToUserPermits(disk)
     disk.canUserEditDisk = canUserEditDisk(disk.userPermits)
   }
 
--- a/src/sagas/utils.js
+++ b/src/sagas/utils.js
@@ -171,7 +171,7 @@ export function* selectorUserAndRoles () {
  * to a set of permits for the app's current user.  Function `mapEntityPermits` is called
  * to do the actual mapping.
  *
- * NOTE: For effiency sake, use the cached lookup version `cacheEntityPermissionsToUserPermits`
+ * NOTE: For effiency sake, use the cached lookup version `curryEntityPermissionsToUserPermits`
  * in calling code.  This helps keep the redux-saga effect queue small and memory use efficient.
  */
 export function* entityPermissionsToUserPermits (entity) {
@@ -189,7 +189,7 @@ export function* entityPermissionsToUserPermits (entity) {
  */
 export function* curryEntityPermissionsToUserPermits () {
   const userAndRoles = yield selectorUserAndRoles()
-  return (entity, ...rest) => mapEntityPermits(entity, userAndRoles, ...rest)
+  return (entity) => mapEntityPermits(entity, userAndRoles)
 }
 
 /**
--- a/src/sagas/utils.js
+++ b/src/sagas/utils.js
@@ -167,25 +167,38 @@ export function* selectorUserAndRoles () {
 }
 
 /**
- * Convert an entity's set of permissions to a set of permits for the app's current
- * user by mapping the permissions through their assigned roles to the permits.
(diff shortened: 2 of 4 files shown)


src/sagas/utils.js
```diff
@@ -156,29 +156,42 @@ export function* delayInMsSteps (count = 20, msMultiplier = 2000) {
...
+ *
+ * NOTE 2: For effiency sake, use the cached lookup version `.cached()` in calling
+ * code.  This help keep the redux-saga effect queue small and memory use efficient.
  */
 export function* entityPermissionsToUserPermits (entity) {
+  const userAndRoles = yield selectorUserAndRoles()
+  return mapEntityPermits(entity, userAndRoles)
+}
+
+entityPermissionsToUserPermits.cached = function* () {
```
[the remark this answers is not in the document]
sjd78: > For difficult cases, I would rather vote for using generators all along the way ( check [Generators as observers](https://exploringjs.com/es6/ch_generators.html#sec_generators-as-observers))
> 

Using generators all along the way is what seems to have actually caused the problem.  Also, I don't see your example as being any more clear than my solution.

> 
> However in this case we might get away with converting `entityPermissionsToUserPermits()` to regular function i.e.
> 
> ```
> const userAndRoles = yield selectorUserAndRoles()
> for (const dataCenter of dataCentersInternal) {
>   dataCenter.userPermits = entityPermissionsToUserPermits(dataCenter, userAndRoles)
> }
> ```

I started it out this way...do the store lookups in a single yield then a normal functions.  But it looked better to encapsulate the data needed by `entityPermissionsToUserPermits` inside the caching function.  That way the use looks pretty much the same everywhere.

michalskrivanek:
hm, so did we just push the number of templates limit to 2000?

sjd78:
> hm, so did we just push the number of templates limit to 2000?

There shouldn't be a limit, just a longer wait.  Fetching the 1080 templates already takes >20 seconds.  I haven't tested the upper limit.

michalskrivanek:
> > hm, so did we just push the number of templates limit to 2000?
> 
> 
> 
> There shouldn't be a limit, just a longer wait.  Fetching the 1080 templates already takes >20 seconds.  I haven't tested the upper limit.

Ok. Either way it's an improvement so LGTM

rszwajko (approved):

sgratch (commented):
@sjd78  LGTM and seems like solving the issue. 
I also tested that fix against an engine with a triple amount of templates (~3200) and it works as expected and doesn't hang.

- Why not using the cached lookup function `curryEntityPermissionsToUserPermits` for vm's disks as well? https://github.com/oVirt/ovirt-web-ui/blob/34a013cfacdc4428d0fafda2bd88cdd2f03a039d/src/sagas/index.js#L110  Do we assume a small amount as assumed for vms? 
- Please check few inline comments as well

src/sagas/utils.js
```diff
@@ -156,29 +156,55 @@ export function* delayInMsSteps (count = 20, msMultiplier = 2000) {
...
+ * Curry `mapEntityPermits` with a single lookup of user and roles from the redux store.
+ *
+ * NOTE: This version is most useful when processing multiple entities at a time.
+ *
+ * @returns A function that will convert an entity's set of permissions to a set of
+ *          permits using a shared set of user and roles data.
+ */
+export function* curryEntityPermissionsToUserPermits () {
+  const userAndRoles = yield selectorUserAndRoles()
+  return (entity, ...rest) => mapEntityPermits(entity, userAndRoles, ...rest)
```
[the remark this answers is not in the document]
sjd78: It just makes it so you wouldn't need to touch the curry function if `entityPermissionsToUserPermits()` or `mapEntityPermits()` params change.  Not required so I'll remove it.

sjd78:
> @sjd78 LGTM and seems like solving the issue. I also tested that fix against an engine with a triple amount of templates (~3200) and it works as expected and doesn't hang.

:+1:

> 
>    * Why not using the cached lookup function `curryEntityPermissionsToUserPermits` for vm's disks as well? 
>         Do we assume a small amount as assumed for vms?

Yes, small amounts processed so I didn't touch it.  I updated the function anyway just to be sure.

> 
>    * Please check few inline comments as well

Done.

src/sagas/index.js
```diff
@@ -88,8 +88,10 @@ const VM_FETCH_ADDITIONAL_SHALLOW = [
 export const EVERYONE_GROUP_ID = 'eee00000-0000-0000-0000-123456789eee'
 
 export function* transformAndPermitVm (vm) {
+  const entityPermissionsToUserPermits = yield curryEntityPermissionsToUserPermits()
```
sgratch, line 91: just to be consistent with other appearances of this `curryEntityPermissionsToUserPermits` call:
```suggestion
  const _entityPermissionsToUserPermits = yield curryEntityPermissionsToUserPermits()
```

src/sagas/index.js
```diff
@@ -88,8 +88,10 @@ const VM_FETCH_ADDITIONAL_SHALLOW = [
...
+
   const internalVm = Transforms.VM.toInternal({ vm })
-  internalVm.userPermits = yield entityPermissionsToUserPermits(internalVm)
+  internalVm.userPermits = entityPermissionsToUserPermits(internalVm)
```
sgratch, line 94: Based on previous suggestion:

```suggestion
  internalVm.userPermits = _entityPermissionsToUserPermits(internalVm)

```

src/sagas/index.js
```diff
@@ -107,7 +109,7 @@ export function* transformAndPermitVm (vm) {
 
   // Permit disks fetched and transformed along with the VM
   for (const disk of internalVm.disks) {
-    disk.userPermits = yield entityPermissionsToUserPermits(disk)
+    disk.userPermits = entityPermissionsToUserPermits(disk)
```
sgratch, line 112: Based on previous suggestion:

```suggestion
    disk.userPermits = _entityPermissionsToUserPermits(disk)
```

sgratch (commented):
LGTM except for a nitpick fix

sgratch (approved):

PR #1645, opened 2022-11-30
github actions: update action versions to current

sjd78:
Upgrade to currently supported versions of actions used in `check.yaml`.

Ref: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

Signed-off-by: Scott J Dickerson <sdickers@redhat.com>

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -24,13 +24,13 @@ jobs:
 
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v2
+        uses: ovirt/checkout-action@main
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -24,13 +24,13 @@ jobs:
 
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
michalskrivanek, line 27: how about switching to https://github.com/oVirt/checkout-action ?

michalskrivanek (approved):

Issue #1644, opened 2022-11-30
'Created' field within snapshot details card and NotificationDrawer events timestamp include an invalid month 


sgratch:
The Month displayed as part of the snapshot `created` date field within the **vm details -> snapshot card -> snapshot details** dialog is invalid since it represents a previous month instead of the actual one. E.g. if the snapshot was created on 30/11/2022, it appears as created on 30/10/2022:

![image](https://user-images.githubusercontent.com/18169498/204864708-00b6f3a3-a8cf-43fe-968c-c3130b63c450.png)

The same goes for user messages timestamp appears within the NotificationDrawer. E.g. for an event occurred on 30/11/2022:
![image](https://user-images.githubusercontent.com/18169498/204872438-5d944f27-79a5-46d0-aa79-7329d35cd79b.png)

The reason is that the JavaScript Date getMonth() returns the month value (0 to 11) of a date, so need to add one when using: https://www.w3schools.com/jsref/jsref_getmonth.asp

PR #1646, opened 2022-11-30
Fix getFormatedDateTime() to use current month instead of previous

sgratch:
Fixes: https://github.com/oVirt/ovirt-web-ui/issues/1644
Since the JavaScript Date.getMonth() method returns the month value starting from 0 (0 to 11 range), need to add one when using.

Check for details: https://www.w3schools.com/jsref/jsref_getmonth.asp

Bug-Url: https://github.com/oVirt/ovirt-web-ui/issues/1644
Signed-off-by: Sharon Gratch <sgratch@redhat.com>


sjd78 (approved):
LGTM

Issue #1658, opened 2024-09-26
path-to-regexp outputs backtracking regular expressions


sandrobonazzola:
`path-to-regexp` used by `ovirt-web-ui` is affected by **[CVE-2024-45296](https://github.com/advisories/GHSA-9wv6-86v2-598j) path-to-regexp outputs backtracking regular expressions**

Dependabot fails to update the dependency with:

```
Dependabot cannot update path-to-regexp to a non-vulnerable version
The latest possible version that can be installed is 0.1.7 because of the following conflicting dependencies:

react-router@5.2.1 requires path-to-regexp@^1.7.0
react-router-dom@5.3.0 requires path-to-regexp@^1.7.0 via a transitive dependency on react-router@5.2.1
webpack-dev-server@4.8.1 requires path-to-regexp@0.1.7 via a transitive dependency on express@4.17.3
The earliest fixed version is 0.1.10.
```

VM Portal version number:  1.9.3

JasperB-TeamBlue:
Fixed, version for path-to-regexp is 0.1.12 since https://github.com/oVirt/ovirt-web-ui/commit/df783a5adc62ccc858be9d75fdd0996e56337e5f

PR #1670, opened 2026-04-16
core: bump react-router and react-router-dom

JasperB-TeamBlue:
* Update routing to function with react routing v5 to v6. 
* Dropped connected-react-router, react-router-config and react-router-navigation-prompt  only available for v5, updated to v6 equivalent. 
* Use proper navigation with useNavigate instead of usage of history component.
* This finishes the changes needed to solve https://github.com/oVirt/ovirt-web-ui/issues/1658


src/store.js
```diff
@@ -66,6 +55,6 @@ export default function configureStore (): Store<any, any> & { rootTask: Task<an
   return {
     ...store,
     rootTask,
-    history,
+    // history,
```
dupondje: Why keep this commented out thing? :)

src/components/VmActions/index.js
```diff
@@ -247,27 +251,38 @@ VmActions.propTypes = {
...
+    config: state.config,
+    preferredConsole: state.options.getIn(['remoteOptions', 'preferredConsole', 'content'], state.config.get('defaultUiConsole')),
+  }),
+  (dispatch, { vm, pool, navigate }) => ({
+    onShutdown: () => dispatch(Actions.shutdownVm({ vmId: vm.get('id'), force: false })),
+    onRestart: () => dispatch(Actions.restartVm({ vmId: vm.get('id'), force: false })),
+    onForceShutdown: () => dispatch(Actions.shutdownVm({ vmId: vm.get('id'), force: true })),
+    onSuspend: () => dispatch(Actions.suspendVm({ vmId: vm.get('id') })),
+    onRemove: ({ preserveDisks }) => {
+      navigate('/')
```
SiebeVE: Navigate before dispatch, not sure if that won't cause issues

dupondje:
The "Wait until VM Portal is displayed" seems to fail. Can you check @JasperB-TeamBlue ?
Perhaps the OST needs to be adjusted for this.

JasperB-TeamBlue:
@dupondje OST works now. Problem was on my end but is fixed now.

PR #1671, opened 2026-04-23
ui: add password and username to cloud init options

### LINKED ISSUE #1514: cloud-init linux set password
Description of problem [include screenshots and browser console logs]:

When creating a VM and using cloud-init there is no way to set root password or create user for linux or BSD VMs only Windows.

VM Portal version number:1.7.1

oVirt version number:4.4.8.5-1.el8

Browser and version:Google Chrome Version 93.0.4577.82 

Steps to Reproduce:
1. Create new VM
2. select linux operating system
3. select cloud-init

Actual results:
hostname and ssh key section

Expected results:
method of creating user and setting root password 
Additional info:

JasperB-TeamBlue:
- When configuring cloud init for linux, have the ability to add a username and password.
- Make the password viewable in the summary instead of being a string value of *.
- Make the password edit field have the same checkbox as in the admin panel.

This fixes https://github.com/oVirt/ovirt-web-ui/issues/1514

### DIFF
--- a/src/components/CreateVmWizard/steps/SummaryReview.js
+++ b/src/components/CreateVmWizard/steps/SummaryReview.js
@@ -1,8 +1,8 @@
-import React, { useContext } from 'react'
+import React, { useContext, useState } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { Label } from '@patternfly/react-core'
-import { InfoCircleIcon } from '@patternfly/react-icons'
+import { EyeIcon, EyeSlashIcon, InfoCircleIcon } from '@patternfly/react-icons'
 
 import { MsgContext, enumMsg, withMsg } from '_/intl'
 import { templateNameRenderer, userFormatOfBytes } from '_/helpers'
@@ -27,6 +27,30 @@ const Item = ({ id, label, children }) => (
     </div>
   </div>
 )
+
+const RevealablePassword = ({ password }) => {
+  const [revealed, setRevealed] = useState(false)
+
+  return (
+    <div style={{ display: 'flex', alignItems: 'center' }}>
+      <span>{revealed ? password : '******'}</span>
+      <button
+        onClick={() => setRevealed(!revealed)}
+        style={{
+          background: 'none',
+          border: 'none',
+        }}
+        title={revealed ? 'Hide password' : 'Show password'}
+      >
+        {revealed ? <EyeSlashIcon /> : <EyeIcon />}
+      </button>
+    </div>
+  )
+}
+
+RevealablePassword.propTypes = {
+  password: PropTypes.string.isRequired,
+}
 
 Item.propTypes = {
   id: PropTypes.string.isRequired,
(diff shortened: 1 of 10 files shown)


src/components/VmDetails/cards/DetailsCard/CloudInit/CloudInitForm.js
```diff
@@ -34,6 +38,39 @@ const CloudInitForm = ({ idPrefix, vm, onChange }) => {
...
+      >
+        <TextInput
+          id={`${idPrefix}-cloud-init-username`}
+          value={cloudInitUsername}
+          onChange={value => onChange('cloudInitUsername', value)}
+        />
+      </FormGroup>
+      <Checkbox
+        id={`${idPrefix}-cloud-init-password-configured`}
+        label='Use already configured password'
```
peter-boden: How does i18n (translation) work for this?

peter-boden (changes requested):
One minor question/remark

src/components/VmDetails/cards/DetailsCard/CloudInit/CloudInitForm.js
JasperB-TeamBlue: Forgot about localization, should be fixed now

src/components/CreateVmWizard/steps/SummaryReview.js
```diff
@@ -223,9 +255,9 @@ const ReviewAdvanced = ({ id, operatingSystems, basic }) => {
             { basic.initTimezone &&
               <Item id={`${id}-sysprep-tz`} label={msg.sysPrepTimezone()}>{ basic.initTimezone }</Item>
           }
```
dupondje: Unrelated, but that indent 😱

src/components/CreateVmWizard/steps/SummaryReview.js
```diff
@@ -223,9 +255,9 @@ const ReviewAdvanced = ({ id, operatingSystems, basic }) => {
...
-            { basic.initAdminPassword && (
-              <Item id={`${id}-sysprep-admin-pwd`} label={msg.sysPrepAdministratorPassword()}>
-                ******
+            { basic.initPassword && (
+              <Item id={`${id}-sysprep-pwd`} label={msg.sysPrepAdministratorPassword()}>
```
dupondje: The label doesn't really match the reality.

dupondje (commented):
Tested if this doesn't conflict stuff on Windows?
Cause I think you cannot pass an initUsername to a windows sysprep? 🤔

src/components/CreateVmWizard/steps/SummaryReview.js
JasperB-TeamBlue: This is the label that is shown when the password that is saved under basic.initPassword belongs to a vm with a Windows OS, if the summary is generated for a Linux OS the label is msg.password() instead. This matches how it is on the ovirt engine.

JasperB-TeamBlue:
Tested creation of a windows VM with this setup too, no issues emerge due to the input fields being seperated between what is shown when a Linux OS is selected vs a Windows OS. Only initUsername that gets passed by the creation of a Windows VM is an empty one which gets ignored by the API.

src/components/CreateVmWizard/steps/SummaryReview.js
dupondje: This needs adjustment?

src/components/CreateVmWizard/steps/SummaryReview.js
```diff
@@ -28,6 +28,30 @@ const Item = ({ id, label, children }) => (
   </div>
 )
 
+const RevealablePassword = ({ password }) => {
+  const [revealed, setRevealed] = useState(false)
+
+  return (
+    <div style={{ display: 'flex', alignItems: 'center' }}>
+      <span>{revealed ? password : '******'}</span>
+      <button
```
dupondje, line 37: This needs `type="button"` added?

dupondje (changes requested):

PR #1673, opened 2026-06-09
Lint implementation

JasperB-TeamBlue:
The feature to lint the present code was already implemented but was rarely used throughout the project.
By adding `yarn eslint` to the CI we can ensure that the code quality and styling is consistent without the need to put that responsibility with the end user.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -79,3 +79,6 @@ jobs:
         run: |
           yarn install
           yarn build
+
+      - name: Run 'yarn eslint' to check for linting errors
+        run: yarn eslint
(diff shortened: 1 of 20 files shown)


dupondje (approved):
Don't we need some configuration for eslint?
It doesn't find the error pointed out in https://github.com/oVirt/ovirt-web-ui/pull/1671/changes#r3154259333 :)

JasperB-TeamBlue:
> Don't we need some configuration for eslint? It doesn't find the error pointed out in https://github.com/oVirt/ovirt-web-ui/pull/1671/changes#r3154259333 :)

There is config present for eslint in `config/eslint.cjs` that restricts on what indents it complains on and which not. Towards the future we can look into making eslint less forgiving on indenting.

PR #1676, opened 2026-07-16
build: update build flow

dupondje:
- Update the way version tags are added during build. Now we can tag a commit in github, and the correct version string will be used for the builds.
- Run the workflow on any tag, so it builds tags in branches also.
- Drop copr
- Combined online and offline builds in 1 workflow


.github/workflows/check.yaml
```diff
@@ -1,84 +0,0 @@
...
-      - name: Checkout sources
-        uses: ovirt/checkout-action@main
-
-      - name: Run 'yarn install' and 'yarn build' (like a developer would)
-        run: |
-          yarn install
-          yarn build
-
-      - name: Run 'yarn eslint' to check for linting errors
-        run: yarn eslint
```
peter-boden, line 84: We loose eslint because we don't run it in the "new" flow?

peter-boden (changes requested):

.github/workflows/check.yaml
dupondje: good catch, fixed!

JasperB-TeamBlue (approved):
LGTM!

peter-boden (approved):

